Issue #61, opened 2014-11-19
Add dom definition for canvas


lbolla:
document.createElement('canvas') returns a type HTMLElement, which does not define methods like "getContext". There should be a HTMLCanvasElement in dom.js similarly to HTMLImageElement.

agentcooper:
+1

facebook-github-bot-2:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

samwgoldman:
`createElement("canvas")` does not return a `HTMLCanvasElement`, so this issue is resolved.

@lbolla Regarding your comment about `getElementById` returning a specific `HTMLElement` subtype—Flow does not do that, and it is unsafe to do. Flow doesn't know what HTML the code will execute against, and therefore doesn't know anything more specific.

This is a great use case for dynamic type tests.

```javascript
var canvas = document.getElementById("mycanvas");
if (canvas instanceof HTMLCanvasElement) {
  // good to go -- type of canvas is refined this block
} else {
  // error condition -- provide some useful debug info
}
```

I see this as a good thing. Flow is helping you out and directing you to add error handling to make your code safer.

PR #66, opened 2014-11-19
Add type definitions for HTMLCanvasElement.

tetsuharuohzeki:
This adds only the definitions of `HTMLCanvasElement`. To real usecases, we need to add definitions of `CanvasRenderingContext2D`.
#61


gasi:
:bow:

Issue #114, opened 2014-11-24
Allow core-lib declarations to be ignored


tcoopman:
At the moment I cannot find a way to ignore the core-lib declarations at startup.
It would be nice if we could provide our own declarations. This would allow us to fix problems with the core declarations that are out of sync with the spec.

See for example: https://github.com/facebook/flow/pull/113 or https://github.com/facebook/flow/pull/108

samwgoldman:
You can use --no-flowlib to ignore the builtin libs.

Issue #72, opened 2014-11-20
TypeScript interface support


fdecampredon:
Flow seems to have a limited interface support, no implements (and not supported in declaration files).

Would mimicking typescript interface support seems reasonable ?

laser:
For those of you who ended up here while searching for "Facebook Flow interfaces," perhaps a simple example would be good:

```
class User {
  n: number;
  constructor(n) {
    this.n = n;
  }
}
 
class CustomError {
  message: string;
  constructor(m) {
    this.message = m;
  }
}

type NodeCallback<E, V> = (err: ?E, value: ?V) => void;
 
// Defining an interface
//
type UserDAOIFace = {
  getUserById: (x: number, callback: NodeCallback<CustomError, User>) => void;
}

// Implementing an interface
// 
var userDAOImpl: UserDAOIFace = {
  getUserById(id, callback) {
    if (id === 555) {
      callback(new CustomError("User not found!"), null);
    }
    else {
      callback(null, new User(id));
    }
  }
}

// A second implementation
// 
var fakeUserDAOImpl: UserDAOIFace = {
  getUserById(id, callback) {
    callback(null, new User(123));
  }
}
```

Hope that helps, and, I hope that my example isn't too stupid.

Erin

0xR:
+1 to this. It would really help to have type interfaces for existing libraries like lodash, underscorejs, angular, ember or whatever. Right now getting started with an existing codebase mostly requires writing long interface files. Because of that adopting flowtype might be more work than it avoids by finding bugs earlier.

Of course making proper flow interfaces from .d.ts files would require some manual work because typescript doesn't support non-nullable types. Therefore a conversion to a flow interface would allow it to be updated manually to properly fit FlowType.

samwgoldman:
Merging with #7

runn1ng:
I am sorry to write it here, but I have no idea where to write it elsewhere.

It seems to me flow now has an `interface` keyword and it does *something*. I am just not sure what and how.

Is it documented anywhere?

GGAlanSmithee:
@runn1ng I don't think there is much documentation of interface declarations. See #833 for some discussion on the subject.

runn1ng:
OK, I think I slightly understand what `interface` does. I am still not sure what `mixins` does and how it relates to `interface`.

vaukalak:
@runn1ng if you do, please explain what is advantage of interfaces compared to types: https://gist.github.com/vaukalak/64be636be0b09181eb31

[...]

PR #148, opened 2014-11-29
add node.js fs module declaration

ptmt:
This is converted declaration from https://github.com/borisyankov/DefinitelyTyped/blob/master/node/node.d.ts
1). Currently there is a makefile command `make test`, but other tests could be run only manually (like `node_tests`, `auxiliary_tests` and tests for declarations). Should we consider to add new makefile target? Also, is it possible to connect `flow` repository with CI server like travic-ci / codeship? 
2). There is no tests for `lib/*.js`, so I ran it across existing codebase, though this is only part of node API and I'm pretty sure somebody have written their own declarations for node core modules. Is it possible to overwrite system declaration? How we can ensure that all contributed declaration is valid? For example, after updating from 0.1.1 to 0.1.2 I've got this error:

```
/gulp-flowtype/index.js:24:3,61:4: function call
  /private/var/folders/tp/ffwbqwn51dd1zgb9pb8j4g7w0000gn/T/flow_potomushto/flowlib_825dbe6/lib/node.js:116:24,120:31: function type

/gulp-flowtype/index.js:24:17,61:3: function
  /private/var/folders/tp/ffwbqwn51dd1zgb9pb8j4g7w0000gn/T/flow_potomushto/flowlib_825dbe6/lib/node.js:118:15,20: object type
```

That's because the signature of `child_process.exec` was:

```
declare function exec(
    command: string,
    options?: Object, // TODO: child_process$execOpts,
    callback: (error: ?Error, stdout: Buffer, stderr: Buffer) => void
  ): child_process$ChildProcess;
```

but it didn't work with `exec(command, callback)` because `flow` can't manage Maybe types this way. I've fixed it, but the general problem is here. 

Maybe somehow we could run `flow` across `node` simple tests? https://github.com/joyent/node/blob/master/test/simple/test-fs-append-file-sync.js

3). Does it make sense at all to convert and debug other modules manually or better to wait for https://github.com/facebook/flow/issues/72? 

Flow is great, we would like to use it for both server and client side, but it requires to contribute system declarations or at least overriding declaration to be independent (or `--no-lib` option https://github.com/facebook/flow/issues/114).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ _build
 .DS_STORE
 docs/_site/
 bin
+*.diff
--- a/lib/node.js
+++ b/lib/node.js
@@ -1,12 +1,12 @@
 /**
- * Copyright (c) 2014, Facebook, Inc.
- * All rights reserved.
- *
- * This source code is licensed under the BSD-style license found in the
- * LICENSE file in the "flow" directory of this source tree. An additional grant
- * of patent rights can be found in the PATENTS file in the same directory.
- *
- */
+* Copyright (c) 2014, Facebook, Inc.
+* All rights reserved.
+*
+* This source code is licensed under the BSD-style license found in the
+* LICENSE file in the "flow" directory of this source tree. An additional grant
+* of patent rights can be found in the PATENTS file in the same directory.
+*
+*/
 
 declare class Buffer {
   constructor(value: Array<number> | number | string, encoding?: string): void;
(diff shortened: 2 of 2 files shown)


thealjey:
@unknownexception 
I'm struggling with this particular feature at the moment actually.
Flow comes with type declarations for 'child_process$ChildProcess' and the methods declared directly in ChildProcess work ok.
However, ChildProcess inherits EventEmitter and as such should also have methods like "on" available as well.
In the following code:

    var flow = spawn('flow', [], {stdio: 'inherit'});
    flow.kill();
    flow = spawn('flow', [], {stdio: 'inherit'});
    flow.on('close', function (code) {});
The last line errors with "call of method on Property not found".
Would be infinitely grateful if someone helped me figure out a workaround for this.

unknownexception:
see https://github.com/facebook/flow/pull/378

PR #200, opened 2014-12-19
Partial implementation of HTTP module

fmahnke:
First pull request. Began implementing missing node.js http module.


sktguha:
Hi @fmahnke  , 
may I know what this pull request is for? , (just curious, not a collaborator )

fmahnke:
@sktguha lib/node.js is the interface file so flow can be used with node.js projects. Many node.js modules are currently missing from that file. This pull request implemented some of the interface of node's http module.

Issue #131, opened 2014-11-25
Type Inferance problems with property and operator


fdecampredon:
```
function fn1(x) { return x* 4};
fn1('hi');

function fn2(x) { return x.length * 4; }
fn2({length: 'hi'}); // no error however the situation is exactly the same that for fn1
```

PR #223, opened 2015-01-18
Removed hard-coded .length case

commonlisp:
Enable type checker to correctly type non-standard length properties.
Fixes issue #131.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1262,12 +1262,6 @@ let rec flow cx (l,u) trace =
     | (_,
        (GetT(_,"toLocaleString",_) | SetT(_,"toLocaleString",_))) -> ()
 
-    | ((ObjT _ | ArrT _), GetT(reason_op,"length",tout)) ->
-      unit_flow cx (NumT.why reason_op, tout)
-
-    | ((ObjT _ | ArrT _),
-       (SetT(_,"length",_) | MethodT(_,"length",_,_,_,_))) -> ()
-
     | ((ObjT _ | ArrT _), GetT(reason_op,"constructor",tout)) ->
       unit_flow cx (AnyT.why reason_op, tout)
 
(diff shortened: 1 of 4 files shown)


tests/misc/misc.exp
```diff
@@ -11,4 +11,8 @@ E.js:8:6,10: boolean
 This type is incompatible with
   E.js:4:14,19: number
 
-Found 3 errors
+F.js:2:14,17: string
```
gabelevi: Hmmm...[I don't see an F.js](https://github.com/facebook/flow/tree/master/tests/misc). Did you add it and forget to add it to the commit?

commonlisp: Oops. Yes, F.js was missed in the commit. I corrected that now. Thanks!

src/typing/flow_js.ml
```diff
@@ -1262,9 +1262,6 @@ let rec flow cx (l,u) trace =
     | (_,
        (GetT(_,"toLocaleString",_) | SetT(_,"toLocaleString",_))) -> ()
 
-    | ((ObjT _ | ArrT _), GetT(reason_op,"length",tout)) ->
-      unit_flow cx (NumT.why reason_op, tout)
-
     | ((ObjT _ | ArrT _),
```
gabelevi: Can you also delete this case? It was allowing you to set `.length` to anything and call `.length()` with anything. For example this is clearly wrong

    var a = { length: "duck" };
    a.length = 123;
    a.length();

and I think you should get two errors if you delete this case.

Also, sorry for not giving this feedback last time. If you're busy just let me know. I'll merge and update it myself :)

commonlisp: Very good point! I have just pushed a squashed commit with that case removed.

gabelevi:
Awesome! Thanks for fixing this issue!

PR #246, opened 2015-02-06
Fixes #195

### LINKED ISSUE #195: Is there a reason why catch on Promise is commented out?
see https://github.com/facebook/flow/blob/master/lib/core.js#L299

tcoopman:


gabelevi:
Thanks for the fix! I had an internal diff up to fix this and then forgot to bug people to accept it. Sorry about that!

PR #247, opened 2015-02-06
Resubmit #113 this fixes #155

### LINKED ISSUE #155: Pull request #113 has been overwritten
Pull request #113 has been overwritten by commit https://github.com/facebook/flow/commit/63f049dfafe4a53fa1d4875150c88bc7d7efab3f

Can you reapply or do I need to submit a new pull request?

tcoopman:
According to https://developer.mozilla.org/en-US/docs/Web/API/Blob.Blob type and endings are optional arguments.
(http://dev.w3.org/2006/webapi/FileAPI/#dfn-BlobPropertyBag also says type is optional, but has no endings argument at all)

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -11,8 +11,8 @@
 
 declare class Blob {
     constructor(blobParts?: Array<any>, options?: {
-        type: string;
-        endings: string;
+        type?: string;
+        endings?: string;
     }): void;
     type: string;
     size: number;


gabelevi:
Great, thanks for resubmitting! Sorry about clobbering the change before...we've got a much better process now for merging PRs internally plus we're going to start publishing commit history for each release, so hopefully this won't happen in the future.

Though I suppose optional object parameters weren't really working until this v0.2.0 :)

PR #265, opened 2015-02-19
flow status --json should exit 2 on errors

### LINKED ISSUE #261: Running flow check with --json flag returns an error code 2
Hi,

When I run flow check with the json command option I get a status code of 2
`$ flow check --json`
`...`
`$ echo $?`
`2`

When I run the exact same command with status, it returns 0.
`$ flow status --json`
`...`
`$ echo $?`
`0`

I'm running flow using nodes `execFile`, the error code 2 is giving me a nice `[Command failed: ]` error.

Using flow v0.2.0.

Cheers,
Charlie

gabelevi:
Fixes #261

### DIFF
--- a/src/commands/statusCommands.ml
+++ b/src/commands/statusCommands.ml
@@ -138,9 +138,9 @@ module Impl = struct
       then Errors_js.print_errorl args.output_json e stdout
       else (
         let show_all = !(option_values.CommandUtils.show_all_errors) in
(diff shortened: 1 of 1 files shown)


gabelevi:
I should mention I checked the vim plugin, the emacs plugin, and a Facebook integration to make sure that the new return code wouldn't break them.

PR #269, opened 2015-02-21
Add RSS feed of blog posts

simonwelsh:


gabelevi:
Sweet, thanks for setting this up for us!

PR #292, opened 2015-02-27
Types for React synthetic events

jamesgpearce:
c.f. http://facebook.github.io/react/docs/events.html and react/lib/Synthetic*Event.js

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -163,3 +163,78 @@ declare class SyntheticEvent {
     timeStamp: number;
     type: string;
 }
+
+declare class SyntheticClipboardEvent extends SyntheticEvent {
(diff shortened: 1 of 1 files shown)


gabelevi:
Looks good to me, thanks! A bunch of these are `any` when they probably don't have to be, but that can be fixed in a later PR.

gabelevi:
For the record, there was a spurious Travis CI failure for this PR (looks like network issues). I restarted the job and it passed.

jamesgpearce:
Yeah, a few are DOM classes that we didn't seem to have in dom.js (and probably should). Wanted to get started here rather than disappear down that particular rabbit hole. Thanks!

PR #295, opened 2015-02-28
Add one of proptype

samwgoldman:
I noticed when digging around the source that a number of PropTypes are not supported, and thought this would be a good place to start contributing.

I tried to follow the style of the existing code, but I would appreciate any formatting critiques.

I'm not super happy with the way I'm creating the string literal EnumT, as I copied it from `convert`. Should this code be factored out into a utility, or is there a better way to go about this?

Hope this helps!

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -3072,14 +3072,14 @@ and mk_proptype cx = Ast.Expression.(function
         elements = es
       })];
     } ->
-      let rec string_literals es res = match (es, res) with
+      let rec string_literals lits es = match (es) with
         | Some (Expression (loc, Literal { Ast.Literal.
             value = Ast.Literal.String lit
-          })) :: tl,
-          Some (lits) -> string_literals tl (Some (lit :: lits))
-        | [], _ -> res
+          })) :: tl ->
+            string_literals (lit :: lits) tl
+        | [] -> Some lits
         | _  -> None in
-      (match string_literals es (Some []) with
+      (match string_literals [] es with
         | Some lits ->
             let reason = mk_reason "oneOf" vloc in
             mk_enum_type cx reason lits
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -421,11 +421,8 @@ let rec convert cx map = Ast.Type.(function
       ArrT (r, t, [])
 
   | loc, StringLiteral { StringLiteral.value; _ }  ->
-    let reason = mk_reason "string literal type" loc in
-      EnumT
-        (reason,
-         Flow_js.mk_object_with_map_proto cx reason
-           (SMap.singleton value AnyT.t) (MixedT reason))
+      let reason = mk_reason "string literal type" loc in
+      mk_enum_type cx reason [value]
 
   (* TODO *)
   | loc, Generic { Generic.id = Generic.Identifier.Qualified (_,
@@ -708,6 +705,12 @@ and mk_type_annotation cx reason = mk_type_annotation_ cx SMap.empty reason
 and mk_type_annotation_ cx map reason = function
   | None -> mk_type_ cx map reason None
   | Some (loc, typeAnnotation) -> mk_type_ cx map reason (Some typeAnnotation)
+
+and mk_enum_type cx reason keys =
+  let map = List.fold_left (fun map key ->
+    SMap.add key AnyT.t map
+  ) SMap.empty keys in
+  EnumT (reason, Flow_js.mk_object_with_map_proto cx reason map (MixedT reason))
 
 (************)
 (* Visitors *)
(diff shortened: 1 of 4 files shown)


src/typing/type_inference_js.ml
```diff
@@ -3065,6 +3065,32 @@ and mk_proptype cx = Ast.Expression.(function
...
+            let reason = mk_reason "string literal type" loc in
+            let string_literal = EnumT
+              (reason,
+               Flow_js.mk_object_with_map_proto cx reason
+                 (SMap.singleton value AnyT.t) (MixedT reason)) in
+            mk_one_of tl (UnionT (r, string_literal :: ts))
+        | [], _ -> t
+        | _  -> AnyT.at vloc in
+      let reason = mk_reason "oneOf" vloc in
+      mk_one_of es (UnionT (reason, []))
```
bhosmer: Hey @samwgoldman - first of all, very cool that you're digging into this, thanks! One edit here would be to have the helper function simply build the list of types to plug into the union, rather than unwrapping and rewrapping a union at each step. Then here you could build a UnionT with the finished list. (That would also tighten the pattern match in the helper, since you wouldn't need a wildcard at the end for non-UnionT arguments.)

samwgoldman: Good point. I didn't see it when I wrote this, but it seems obvious now.

samwgoldman: Seems like your comment disappeared. I had composed this:

Err, actually, the wildcard is to bail, not for non-UnionT arguments, but for non-literal strings.

I could have the recursive function return a `t list option`, to indicate failure, but then I'd just be un-re-wrapping the option.

I could separate the "is everything a string literal" check from the "let's create a type" part, but I'm not sure that would be more clear.

I'm still learning OCaml, so if there's another preferable way way (more procedural?), I'm happy to change this.

src/typing/type_inference_js.ml
```diff
@@ -3065,6 +3065,32 @@ and mk_proptype cx = Ast.Expression.(function
...
+          UnionT (r, ts) ->
+            let reason = mk_reason "string literal type" loc in
+            let string_literal = EnumT
+              (reason,
+               Flow_js.mk_object_with_map_proto cx reason
+                 (SMap.singleton value AnyT.t) (MixedT reason)) in
+            mk_one_of tl (UnionT (r, string_literal :: ts))
+        | [], _ -> t
+        | _  -> AnyT.at vloc in
+      let reason = mk_reason "oneOf" vloc in
```
bhosmer: Hey Sam - I think what you want to do instead here is have the helper build a map of elements to AnyTs, and then  build a final EnumT with the result. That way you'll get a single EnumT<{"foo":any; "bar": any; ...}> rather than a union of singleton EnumTs.

samwgoldman: That does make more sense.

samwgoldman:
@bhosmer Hope the latest commit addresses your comments. Constructing the oneOf EnumT still involves a de-re-structuring of an option type, but that seemed to me the cleanest way to bail out if we encounter something other than a string literal. I imagine some procedural code might allocate less, but this hardly seems like a hot spot.

samwgoldman:
Found an issue while I was trying to separate the tests from the existing spread test. If the oneOf proptype isn't required, flow doesn't warn about invalid values. The required/optional/default gymnastics are a little over my head at the moment, so any guidance here is appreciated.

bhosmer:
Hi Sam - sorry for the delay. Thanks for your update! Couple items of feedback:

* Re packing and unpacking the union: with a quick simplification to `string_literals`, you can avoid doing this. Basically, if you match only on the node list rather than the (node list, result) pair, then you don't need to build the option type until the end. Check it out:
```
     (* let rec string_literals lits es = match es with *)
     (* ...or equivalently: *)
     let rec string_literals lits = function
        | Some (Expression (loc, Literal { Ast.Literal.
            value = Ast.Literal.String lit
          })) :: tl ->
            string_literals (lit :: lits) tl
        | [] -> Some lits
        | _ -> None in
      (match string_literals [] es with
      ...
```

* Re the enforcement failure if .isRequired isn't specified. This is definitely a bug, but fixing it is beyond the scope of this PR - there are deeper things in play, and the problem is almost certainly not confined to oneOf. (In fact, it would be great if you could post an issue for this. :) For our purposes here, I would add isRequired to the test case, plus a comment noting why it's there.

* Aside: zooming out a bit, it's worth mentioning that the silent failure strategy that Flow takes here -silently ignoring propTypes specs if they're incorrect in any way, rather than giving errors - is suboptimal, and a usability TODO. I don't know if it strikes your fancy, but this sort of work (for instance here, an error on non-string-literals, or if the argument itself isn't an array literal) is more than fair game for future contributions, if you're interested. Not the most glamorous stuff but great for usability. :)

Anyway, if you could avoid the union repack as described above and add isRequired to your new test, I think we're good to go. And thanks again! Really awesome to have you contributing.

Basil

samwgoldman:
Pending the test run, I think I addressed your comments. Thanks for your patience and excellent feedback with my contribution.

I am definitely interested in spending more time with this proptypes code: adding warnings for proptypes that are unchecked and adding checks to cover more proptypes are both on my list of things to look into.

bhosmer:
Done - thanks again Sam!

PR #299, opened 2015-03-05
Declare prompt function (fixes #204)

### LINKED ISSUE #204: prompt() not declared
HTML5 function prompt() is not declared, and gives an error when used. See https://developer.mozilla.org/en-US/docs/Web/API/Window.prompt

neilparikh:
Fix for issue #204


gabelevi:
Sweet, thanks!

PR #313, opened 2015-03-09
Move methods and attributes to appropriate parent classes, and add declarations to describe the CSSOM to provide stronger typing for `HTMLElement.style` attribute.

redneckbeard:
This pull brings flow's definition of `HTMLElement` into line with the W3C HTML5 Recommendation and introduces most of the classes described by the CSS Object Model.

### DIFF
--- a/lib/cssom.js
+++ b/lib/cssom.js
@@ -1,0 +1,58 @@
+/**
+ * Copyright (c) 2014, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the "flow" directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+
+declare class CSSStyleSheet {
+  cssRules: CSSRuleList;
+  ownerRule: ?CSSRule;
+  deleteRule(index: number): void;
+  insertRule(rule: string, index: number): void;
+}
+
+declare class CSSRule {
+  cssText: string;
+  parentRule: ?CSSRule;
+  parentStyleSheet: ?CSSStyleSheet;
+  type: number;
+  static STYLE_RULE: number;
+  static MEDIA_RULE: number;
+  static FONT_FACE_RULE: number;
+  static PAGE_RULE: number;
+  static IMPORT_RULE: number;
+  static CHARSET_RULE: number;
+  static UNKNOWN_RULE: number;
+  static KEYFRAMES_RULE: number;
+  static KEYFRAME_RULE: number;
+  static NAMESPACE_RULE: number;
+  static COUNTER_STYLE_RULE: number;
+  static SUPPORTS_RULE: number;
+  static DOCUMENT_RULE: number;
+  static FONT_FEATURE_VALUES_RULE: number;
+  static VIEWPORT_RULE: number;
+  static REGION_STYLE_RULE: number;
+}
+
+declare class CSSRuleList {
+  length: number;
+  item(index: number): ?CSSRule;
+}
+
+declare class CSSStyleDeclaration {
+  cssText: string;
+  length: number;
+  parentRule: CSSRule;
+  cssFloat: string;
+  item(index: number): string;
+  getPropertyValue(property:string): string;
+  getPropertyPriority(property: string): string;
+  setProperty(property: string, value: ?string, priority: ?string): void;
+  setPropertyValue(property: string, value: string): void;
+  setPropertyPriority(property: string, priority: string): void;
+  removeProperty(property: string): string;
+}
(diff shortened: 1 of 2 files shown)


samwgoldman:
I glanced over this and it seems to check out based on the web api interfaces [documented over at MDN](https://developer.mozilla.org/en-US/docs/Web/API).

Wish we could just generate these files.

redneckbeard:
I'd be more than willing to work on a tool that parses Web IDL files and outputs Flow class declarations. I do think, however, that some degree of human intervention will continue to be necessary -- WebKit and Mozilla both have IDL files available, but they contain vendor-specific extensions that would somehow have to be merged (or pruned, depending on how Flow owners wanted to handle that), and the spec has some interesting omissions, such as the [HTMLElement IDL](http://www.w3.org/TR/html5/dom.html#htmlelement) not including `id`, `class`, or `style`, which are elsewhere named as [global attributes](http://www.w3.org/TR/html5/dom.html#global-attributes). But maybe automation could get us 3/4 of the way there.

lib/cssom.js
```diff
@@ -0,0 +1,51 @@
...
+  type: number;
+  static STYLE_RULE: number;
+  static MEDIA_RULE: number;
+  static FONT_FACE_RULE: number;
+  static PAGE_RULE: number;
+  static IMPORT_RULE: number;
+  static CHARSET_RULE: number;
+  static UNKNOWN_RULE: number;
+  static KEYFRAMES_RULE: number;
+  static KEYFRAME_RULE: number;
```
gabelevi: How did you decide which rules to include? I see a bunch more here: https://developer.mozilla.org/en-US/docs/Web/API/CSSRule

lib/dom.js
```diff
@@ -321,19 +339,31 @@ declare class Element extends Node {
     removeAttributeNS(namespaceURI: string, localName: string): void;
     querySelector(selector: string): HTMLElement;
     querySelectorAll(selector: string): NodeList<HTMLElement>;
+    getBoundingClientRect(): ClientRect;
+    insertAdjacentHTML(position: string, text: string): void;
+    blur(): void;
+    click(): void;
+    focus(): void;
```
gabelevi: I think these functions are defined on `HTMLElement` rather than `Element` (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement)

gabelevi:
Some inline comments but this looks good to me for the most part

lib/dom.js
redneckbeard: You are right, although the latest version of CSSOM has `getBoundingClientRect` on `Element`: http://dev.w3.org/csswg/cssom-view/#the-getclientrects()-and-getboundingclientrect()-methods I will move the other four. Should I move `getBoundingClientRect` as well?

redneckbeard:
@gabelevi Thanks for having a look at this. I've made all the recommended changes.

lib/dom.js
```diff
@@ -327,13 +345,25 @@ declare class HTMLElement extends Element {
...
+    accessKey: string;
+    accessKeyLabel: string;
+    dataset: {[key:string]: string};
+    contentEditable: string;
+    isContentEditable: boolean;
+    hidden: boolean;
+    tabIndex: number;
+    spellcheck: boolean;
+    getBoundingClientRect(): ClientRect;
+    insertAdjacentHTML(position: string, text: string): void;
```
gabelevi: [Looks like this belongs on `Element`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML)

lib/cssom.js
```diff
@@ -0,0 +1,58 @@
...
+  parentRule: ?CSSRule;
+  parentStyleSheet: ?CSSStyleSheet;
+  type: number;
+  static STYLE_RULE: number;
+  static MEDIA_RULE: number;
+  static FONT_FACE_RULE: number;
+  static PAGE_RULE: number;
+  static IMPORT_RULE: number;
+  static CHARSET_RULE: number;
+  static UNKNOWN_RULE: number;
```
gabelevi: Oh weird...I wonder why it's missing from the [WebIDL declaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSRule)

redneckbeard: Three potential sources of truth: the WedIDL on that page, the table of props underneath or the [CSSOM Working Draft](), none of which fully agree. The table seemed broadest -- I thought perhaps it'd be worth capturing potentially deprecated attributes, since they still might be referenced in code targeting specific browsers. Will remove `UNKNOWN_RULE` if you so recommend.

gabelevi: Ha, that's kind of frustrating. I'm totally fine with the way this is now.

lib/dom.js
redneckbeard: Thanks. I've moved `insertAdjacentHTML` to `Element`.

gabelevi:
Alright, looks good to me! Thanks for bearing with me!

PR #315, opened 2015-03-10
Implement more PropTypes

samwgoldman:
Implemented `func`, `object`, and `objectOf`. Added tests for the existing `arrayOf` proptype. "Fixed" issue flowing React props to generated proptypes (see  ba95d17).

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1549,11 +1549,9 @@ let rec flow cx (l,u) trace =
          unaliased), propertywise subtyping is sound *)
       let lit =
         let desc = (desc_of_reason reason1) in
-        let is_prefix a b =
-          let a_len = String.length a in
-          let b_len = String.length b in
-          let rec loop i = i = b_len || a.[i] = b.[i] && loop (i + 1) in
-          a_len >= b_len && loop 0 in
+        let is_prefix a b = String.(
+          let n = length b in
+          length a >= n && sub a 0 n = b) in
         desc = "object literal" || is_prefix desc "props of React element" in
       iter_props_ cx flds2
         (fun s -> fun t2 ->
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1547,7 +1547,14 @@ let rec flow cx (l,u) trace =
       ->
       (* if inflowing type is literal (thus guaranteed to be
          unaliased), propertywise subtyping is sound *)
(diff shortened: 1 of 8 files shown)


src/typing/flow_js.ml
```diff
@@ -1547,7 +1547,14 @@ let rec flow cx (l,u) trace =
       ->
       (* if inflowing type is literal (thus guaranteed to be
          unaliased), propertywise subtyping is sound *)
-      let lit = (desc_of_reason reason1) = "object literal" in
+      let lit =
+        let desc = (desc_of_reason reason1) in
+        let is_prefix a b =
```
bhosmer: Here's a simpler is_prefix:

    let is_prefix a b = String.(
      let n = length b in 
      length a >= n && sub a 0 n = b)

samwgoldman: :+1: I'll use that instead.

bhosmer:
Awesome... hopefully the march of progress will sweep away the reason hacking in short order. :) Meanwhile, if you could substitute the simpler is_prefix (or an equivalent) I'll merge this baby in. Thanks again!

samwgoldman:
Should be good to go. Only a few simple proptypes left.

tests/new_react/new_react.exp
```diff
@@ -47,10 +47,6 @@ new_react.js:4:12,33: string
 This type is incompatible with
   new_react.js:18:15,20: number
 
-new_react.js:4:12,33: string
```
bhosmer: Any idea why this goes away?

bhosmer:
Hi Sam - everything we talked about looks great. The error that gets dropped is a puzzle though... apologies for not noting it earlier. Any idea what's going on there?

tests/new_react/new_react.exp
samwgoldman: Yeah, I mentioned this in the commit message for this change. Since we used to unify props, proptypes are checked in both directions. Elsewhere in this expectation the same error exists with the line numbers reversed.

samwgoldman:
@bhosmer, yup, it's invariance vs subtyping. The error message used to fail "both ways" and now just fails one way. Intermediate states of this PR (before the proptype literal reason hacks) had a lot more of those "duplicated" errors. I think it's nicer now, and shouldn't entail a loss of type checking.

bhosmer:
Ah, cool - missed the inverted one later in the error dump. Thanks!

PR #318, opened 2015-03-12
Match string variables with literal information in oneOf proptype

samwgoldman:
This could be way off-base, but it seems like we'll often have literal information with a StrT. This makes oneOf a lot more powerful/useful.


samwgoldman:
This change isn't as foolproof as I'd hoped.

```javascript
var s = "hello";
var f = () => React.createClass({ propTypes: { s: React.PropTypes.oneOf([s]) } });
s = "oops";
f() // runtime check and static type disagree
```

PR #321, opened 2015-03-14
Support for React.cloneElement in 0.13

jamesgpearce:
As per http://facebook.github.io/react/blog/2015/03/03/react-v0.13-rc2.html#react.cloneelement


gabelevi:
Looks good to me! Thanks!

PR #322, opened 2015-03-15
Add type signature for node "url", "querystring" and "path" modules

zertosh:
_I'm just experimenting with Flow - so if this PR sucks, it's ok to just say so_

Things I'm not totally clear on:

* The `path` module exports `posix` and `win32` - they have the same interface as `path`. How do I express that? See https://github.com/joyent/node/blob/a995a6a/lib/path.js#L620-L626

* `querystring` has undocumented aliases for `stringify` and `parse` (`encode` and `decode` respectively). Should they be included?

* `url#parse` can return a `query` that can be `null | string | Object`, but I couldn't get that to work in any form.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -468,7 +468,25 @@ declare module "punycode" {
 }
 
 declare module "querystring" {
-  // TODO
+  declare function stringify(
+    obj: Object,
+    separator: ?string,
+    equal: ?string,
+    options?: {
+      encodeURIComponent?: (str: string) => string;
+    }
+  ): string;
+  declare function parse(
+    str: string,
+    separator: ?string,
+    equal: ?string,
+    options?: {
+      decodeURIComponent?: (str: string) => string;
+      maxKeys?: number;
+    }
+  ): any;
+  declare function escape(str: string): string;
+  declare function unescape(str: string, decodeSpaces?: boolean): string;
 }
 
 declare module "readline" {
--- a/lib/node.js
+++ b/lib/node.js
@@ -501,7 +501,38 @@ declare module "tls" {
 }
 
 declare module "url" {
-  // TODO
+  declare function parse(
+    urlStr: string,
+    parseQueryString: ?boolean,
+    slashesDenoteHost?: boolean
+  ): {
+    protocol: ?string;
+    slashes: ?boolean;
+    auth: ?string;
+    host: ?string;
+    port: ?string;
+    hostname: ?string;
+    hash: ?string;
+    search: ?string;
+    query: ?any; // null | string | Object
+    pathname: ?string;
+    path: ?string;
+    href: string;
+  };
+  declare function format(urlObj: {
+    href?: string;
+    protocol?: string;
+    slashes?: boolean;
+    auth?: string;
+    hostname?: string;
+    port?: string | number;
+    host?: string;
+    pathname?: string;
+    search?: string;
+    query?: Object;
+    hash?: string;
+  }): string;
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -468,7 +493,25 @@ declare module "punycode" {
...
+    obj: Object,
+    separator?: string,
+    equal?: string,
+    options?: {
+      encodeURIComponent?: (str: string) => string;
+    }
+  ): string;
+  declare function parse(
+    str: string,
+    separator?: string,
```
jeffmo: I think `separator` and `equal` probably want to be nullable (rather than optional).

The docs show examples where you might want to skip over one or both of these args to get to the options object -- so you can do that by specifying `null` in the params list.

lib/node.js
```diff
@@ -468,7 +493,25 @@ declare module "punycode" {
 }
 
 declare module "querystring" {
-  // TODO
+  declare function stringify(
+    obj: Object,
+    separator?: string,
```
jeffmo: (same here WRT nullable vs optional)

lib/node.js
```diff
@@ -468,7 +493,25 @@ declare module "punycode" {
...
+    str: string,
+    separator?: string,
+    equal?: string,
+    options?: {
+      decodeURIComponent?: (str: string) => string;
+      maxKeys?: number;
+    }
+  ): any;
+  declare function escape(str: string): string;
+  declare function unescape(str: string, decodeSpaces?: boolean): string;
```
jeffmo: I'm not seeing the `decodeSpaces` param listed in the node API docs:

https://nodejs.org/api/querystring.html#querystring_querystring_unescape

lib/node.js
```diff
@@ -501,7 +544,38 @@ declare module "tls" {
 }
 
 declare module "url" {
-  // TODO
+  declare function parse(
+    urlStr: string,
+    parseQueryString?: boolean,
```
jeffmo: nullable vs optional here?

lib/node.js
```diff
@@ -501,7 +544,38 @@ declare module "tls" {
...
+    hostname: ?string;
+    hash: ?string;
+    search: ?string;
+    query: ?any; // null | string | Object
+    pathname: string;
+    path: string;
+    href: string;
+  };
+  declare function format(urlObj: {
+    href?: string;
```
jeffmo: Rather than having to type these out twice, you might consider a type alias similar to what we've done for other APIs:

https://github.com/facebook/flow/blob/master/lib/node.js#L65

lib/node.js
zertosh: It's just poorly documented :disappointed: 

https://github.com/joyent/node/blob/v0.12.2/lib/querystring.js#L108
https://github.com/joyent/node/blob/v0.10.38/lib/querystring.js#L107
https://github.com/iojs/io.js/blob/v1.6.3/lib/querystring.js#L80

jeffmo: Ahh ok, good to know! Fine by me to document it here then :)

lib/node.js
zertosh: Done!

lib/node.js
zertosh: They look similar but they're subtly different. (1) `format` doesn't take a `path` prop, (2) and`format`'s  `query` can only be an ?Object. (3) Unlike `parse`'s `query` that can be `null | string | Object`.

I've fixed the rest of the nullable vs optional.

jeffmo:
Thanks a ton for fixing this up!

PR #333, opened 2015-03-19
Partial fs module implementation

gyzerok:
Partial fs module implementation

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -425,34 +425,10 @@ declare module "fs" {
   declare function statSync(path: string): Stats;
   declare function unlink(path: string, callback: (err: ?Error) => void): void;
   declare function unlinkSync(path: string): void;
-  declare function write(
-    fd: number,
-    buffer: Buffer,
-    offset: number,
-    length: number,
-    position?: mixed,
-    callback: (err: ?Error, write: number, str: string) => void
-  ): void;
-  declare function write(
-    fd: number,
-    data: mixed,
-    position?: mixed,
-    encoding?: string,
-    callback: (err: ?Error, write: number, str: string) => void
-  ): void;
-  declare function writeSync(
-    fd: number,
-    buffer: Buffer,
-    offset: number,
-    length: number,
-    position?: number
-  ): number;
-  declare function writeSync(
-    fd: number,
-    data: mixed,
-    position?: number,
-    encoding?: string
-  ): number;
+  declare var write: (fd: number, buffer: Buffer, offset: number, length: number, position?: mixed, callback: (err: ?Error, write: number, str: string) => void): void
+                   | (fd: number, data: mixed, position?: mixed, encoding?: string, callback: (err: ?Error, write: number, str: string) => void): void;
+  declare var writeSync: (fd: number, buffer: Buffer, offset: number, length: number, position?: number): number
+                       | (fd: number, data: mixed, position?: mixed, encoding?: string): number;
   declare function read(
     fd: number,
     buffer: Buffer,
--- a/lib/node.js
+++ b/lib/node.js
@@ -422,7 +422,7 @@ declare module "fs" {
   declare function rename(oldPath: string, newPath: string, callback: (err: ?Error) => void): void;
   declare function renameSync(oldPath: string, newPath: string): void;
   declare function stat(path: string, callback: (err: ?Error, stats: Stats) => void): void;
-  declare function stat(path: string): Stats;
+  declare function statSync(path: string): Stats;
   declare function unlink(path: string, callback: (err: ?Error) => void): void;
   declare function unlinkSync(path: string): void;
   declare function write(
--- a/lib/node.js
+++ b/lib/node.js
@@ -430,7 +430,8 @@ declare module "fs" {
     buffer: Buffer,
     offset: number,
     length: number,
-    position?: callback: (err: ?Error, write: number, str: string) => void
+    position?: mixed,
+    callback: (err: ?Error, write: number, str: string) => void
   ): void;
   declare function write(
     fd: number,
--- a/lib/node.js
+++ b/lib/node.js
@@ -416,14 +416,34 @@ declare module "events" {
 
 declare module "fs" {
   declare class Stats {
-    // TODO: fill this
+    dev: number;
+    ino: number;
+    mode: number;
+    nlink: number;
+    uid: number;
+    gid: number;
+    rdev: number;
+    size: number;
+    blksize: number;
+    blocks: number;
+    atime: Date;
+    mtime: Date;
+    ctime: Date;
+
+    isFile(): boolean;
+    isDirectory(): boolean;
+    isBlockDevice(): boolean;
+    isCharacterDevice(): boolean;
+    isSymbolicLink(): boolean;
+    isFIFO(): boolean;
+    isSocket(): boolean;
   }
 
-  declare function rename(oldPath: string, newPath: string, callback: (err: ?Error) => void): void;
+  declare function rename(oldPath: string, newPath: string, callback?: (err: ?Error) => void): void;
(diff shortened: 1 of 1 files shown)


gyzerok:
Dont understand the reason of failing. Any suggestions?

lib/node.js
```diff
@@ -415,7 +415,76 @@ declare module "events" {
...
+    position?: number,
+    encoding?: string
+  ): number;
+  declare function read(
+    fd: number,
+    buffer: Buffer,
+    offset: number,
+    length: number,
+    position: ?number,
+    callback: (err: ?Error, bytesRead: number, buffer: Buffer)
```
samwgoldman: This callback needs a return type

samwgoldman:
I don't think you can define `write` and `writeSync` twice with different signatures like this.

lib/node.js
```diff
@@ -415,7 +415,76 @@ declare module "events" {
...
 declare module "fs" {
-  // TODO
+  declare class Stats {
+    // TODO: fill this
+  }
+
+  declare function rename(oldPath: string, newPath: string, callback: (err: ?Error) => void): void;
+  declare function renameSync(oldPath: string, newPath: string): void;
+  declare function stat(path: string, callback: (err: ?Error, stats: Stats) => void): void;
+  declare function stat(path: string): Stats;
```
samwgoldman: `statSync`?

samwgoldman:
Ugly as it is, this seems to work for `write` and `writeSync`:

```
  declare var write: (fd: number, buffer: Buffer, offset: number, length: number, position?: mixed, callback: (err: ?Error, write: number, str: string) => void) => void
                   | (fd: number, data: mixed, position?: mixed, encoding?: string, callback: (err: ?Error, write: number, str: string) => void) => void;
  declare var writeSync: (fd: number, buffer: Buffer, offset: number, length: number, position?: number) => number
                       | (fd: number, data: mixed, position?: number, encoding?: string) => number;
```

lib/node.js
gyzerok: Ok, sorry. I'll fix it.

lib/node.js
gyzerok: Sure, I'll add this.

gyzerok:
@samwgoldman 
Thank you for help, Sam. I guess it now can be merged.

samwgoldman:
Nice! Looks good to me.

lib/node.js
```diff
@@ -415,7 +415,54 @@ declare module "events" {
...
+    // TODO: fill this
+  }
+
+  declare function rename(oldPath: string, newPath: string, callback: (err: ?Error) => void): void;
+  declare function renameSync(oldPath: string, newPath: string): void;
+  declare function stat(path: string, callback: (err: ?Error, stats: Stats) => void): void;
+  declare function statSync(path: string): Stats;
+  declare function unlink(path: string, callback: (err: ?Error) => void): void;
+  declare function unlinkSync(path: string): void;
+  declare var write: (fd: number, buffer: Buffer, offset: number, length: number, position?: mixed, callback: (err: ?Error, write: number, str: string) => void) => void
```
gabelevi: Optional params currently only make sense as the last params of a function. I know what you're trying to say, but I don't think it's going to work like this

```JavaScript
function foo(x: number, y?: number, z?: number) {}
foo(1); // This is fine

function bar(x?: number, y: number, z: number) {}
foo(2,3); // This is an error
```

You could in theory do

```JavaScript
(fd: number, buffer: Buffer, offset: number, length: number, position?: mixed, callback: (err: ?Error, write: number, str: string) => void) => void |
(fd: number, buffer: Buffer, offset: number, length: number, callback: (err: ?Error, write: number, str: string) => void) => void
```

To be completely honest, I think you're supposed to be able to do

```JavaScript
declare function write(fd: number, buffer: Buffer, offset: number, length: number, position: mixed, callback: (err: ?Error, write: number, str: string) => void): void;
declare function write(fd: number, buffer: Buffer, offset: number, length: number, callback: (err: ?Error, write: number, str: string) => void): void;
...etc
```

gabelevi:
Other than the optional param in the middle of the params, this looks pretty good to me. I noticed that the [.d.ts file](https://github.com/borisyankov/DefinitelyTyped/blob/master/node/node.d.ts) makes the callback for the async function optional, so that might be worth picking up too, if you like.

samwgoldman:
It's definitely possible the error I saw (and fixed locally) regarding overridden `write` was because of the penultimate optional params. Sounds like making the callback optional as well is the path forward, and it's worth checking if the override syntax works instead of the confusing union or function types.

Aside, by "doesn't make sense" does that mean that optional middle parameters are not possible to encode, or that the logic for them hasn't been implemented yet?

gabelevi:
> Aside, by "doesn't make sense" does that mean that optional middle parameters are not possible to encode, or that the logic for them hasn't been implemented yet?

So if you only allow optional parameters to be the last arguments to a function (before the rest param), then you can figure out whether or not they're being passed in based on the number of parameters passed to the function

```JavaScript
function foo(x, y, z?) {}
foo(1,2,3); // This is passing in z since it has 3 arguments
foo(1,2); // This is not passing in z since it has <3 arguments
```

If you allow optional parameters in the middle of the parameter list, then if you have 2 or more optional parameters it is ambiguous whether they are being passed in.

```JavaScript
function bar(a, b?, c?, d) {}
bar(1,2,3); // Did I pass in b or c?
```

Yes, you can sometimes use types to disambiguate. Yes, if you only have one optional param then it's not ambiguous. But in the general case they are ambiguous.

I think default parameter values have the same problem and I think the proposal only lets you have default values for function parameters at the end of a parameter list.

gyzerok:
@gabelevi suppose we can make callback optional. The dark side of it would be impossibility to notify the developer about missing callback mistake.

Before I switch here to ```declare var write``` there was an error? Is it because of optional parameters issue?
Do you think the following polymorphic declaration is correct and would work?
```javascript
declare function write(fd: number, buffer: Buffer, offset: number, length: number, position?: mixed, callback?: (err: ?Error, write: number, str: string) => void): void;
declare function write(fd: number, data: mixed, position?: mixed, encoding?: string, callback?: (err: ?Error, write: number, str: string) => void) => void;
...etc
```

gyzerok:
I suppose its ok to waste a few time with me since i'm going to contribute more into node interface. It would be great to get the right understanding of the tool.

gabelevi:
@gyzerok I think these needs to be changed so that optional parameters are always at the end of parameter lists. Optional parameters in the middle of parameter lists won't work as expected. Here's an example:

```Bash
$ echo "function foo(x?: number, y: string) {} foo(1,'hi');" | flow check-contents
No errors!

$ echo "function foo(x?: number, y: string) {} foo('hi');" | flow check-contents

-:1:40,48: function call
  -:1:40,42: function

-:1:44,47: string
  -:1:18,23: number

Found 2 errors
```

I'm not sure why the `declare function write` isn't working for you. I'm not 100% sure how that stuff works (CC @avikchaudhuri) but if the `declare var write` solution works for you then that's fine. I think you just need to add more cases so that there are no optional params in the middle of a parameter list.

gyzerok:
@gabelevi I really dont know why I cant use ```declare function ...``` polymorphic declarations. I suppose its an issue, which have to be placed. The following polymorphic declarations works.
```javascript
declare function createDiffieHellman(prime_length: number): crypto$DiffieHellman;
declare function createDiffieHellman(prime: number, encoding?: string): crypto$DiffieHellman;
```
I guess the reason of failure are different types of the arguments in my case.

gyzerok:
@gabelevi as you can see for now i've made all the callbacks optional.

gabelevi:
Looks good to me! Sorry we've been so slow to respond...we've been traveling a bunch and not handling the jet lag terribly well :)

PR #339, opened 2015-03-24
Increase base font-weight

zpao:
300 was really hard to read, especially combined with `-webkit-font-smoothing: antialiased`


gabelevi:
Sure...website looks better to me in chrome with this change.

PR #340, opened 2015-03-26
Add ES6 String methods to core.js.

bolinfest:


lib/core.js
```diff
@@ -157,7 +157,7 @@ declare class String {
     charAt(pos: number): string;
     charCodeAt(index: number): number;
     concat(...strings: Array<string>): string;
-    contains(substr: string): boolean;
+    contains(substr: string): boolean; // TODO: Remove in favor of includes()?
```
gabelevi: We'll probably want to keep supporting this for legacy code.

gabelevi:
All the changes look good to me. Would you mind fixing the tests? The test examples record the expected errors, and unfortunately the expected errors occasionally include library file line numbers. When you change library files, the expected errors line numbers are no longer correct.

You can probably do something like

```Bash
find tests/ -name "*.diff" | sed "s/\(.*\)diff/mv \1out \1exp/" | sh
```

to fix. When a test fails it should have a `.diff` and a `.out` file in the failed test directory, so replacing the `.exp` file with the `.out` file should suffice.

gabelevi:
I'm going to close this out for now. Feel free to reopen if you have time!

PR #348, opened 2015-03-28
Implement PropTypes.oneOfType

samwgoldman:
Each expression in the array becomes a type in a union. It's possible (likely even) that we will produce a union containing an AnyT, but I don't believe that will be a problem. If the array is empty, we will create an "empty union" which should not admit any value. If the array contains a spread, we fall back to AnyT.


gabelevi:
Looks great! Thanks @samwgoldman!

PR #350, opened 2015-03-29
Try to fix #336

### LINKED ISSUE #336: Array indexing into a function should be an error
I mistyped the `()` to call a function as `[]`. I was surprised that Flow had no complaints!

``` javascript
function add(x, y) {
  return x + y;
}
add[1, 2]
```

I suppose I could do `add[2] = 123` somewhere else, but this wasn't an exported function, so flow could have known that this was off the table. Is `fn[key]` a construct that's often used?

Eyyub:
This is a try to a little workaround to prevent array indexing on function(func[e], func[e] = x). .

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -2470,7 +2470,7 @@ let rec flow cx (l,u) trace =
     (* functions statics *)
     (*********************)
 
-    | (FunT (_,static,_,_), _) when object_like_op u ->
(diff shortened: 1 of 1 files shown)


gabelevi:
Thanks for taking a swing at this! This would definitely fix #336 by banning accessing computed properties of a function, but I'm not convinced that is the right solution. I've commented on #336, so let's discuss it there.

Eyyub:
Warning about dead code is better than this workaround.

PR #356, opened 2015-03-31
Add React.version to lib/react.js

tsing:


samwgoldman:
The test failure here is just because the line numbers changed in a `.exp` file (not a regression caused by this change). @tsing if you update the expectation file this will have less friction to be merged.

tsing:
@samwgoldman Thank you. now the test passed :)

gabelevi:
Looks good to me! Thanks!

And thanks @samwgoldman for the help!

PR #365, opened 2015-04-06
Allow type alias declaration in module.

Eyyub:
This pull request add type alias declaration in module. (exactly the same thing as type alias outside a module definition)

### DIFF
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -1621,6 +1621,25 @@ end = struct
             extends;
           })) in
 
+      (* This is more or less identical to `type_alias` *)
+      let declare_type_alias env start_loc =
+	Expect.token env T_TYPE;
+	Eat.push_lex_mode env TYPE_LEX;
+	let id = Parse.identifier env in
+	let typeParameters = Type.type_parameter_declaration env in
+	Expect.token env T_ASSIGN;
+	let right = Type._type env in
+	let end_loc = match Peek.semicolon_loc env with
+	  | None -> fst right
+	  | Some end_loc -> end_loc in
+	Eat.semicolon env;
+	Eat.pop_lex_mode env;
+        Loc.btwn start_loc end_loc, Statement.(TypeAlias TypeAlias.({
+          id;
+          typeParameters;
+          right;
+        })) in
+
       let declare_function env start_loc =
         Expect.token env T_FUNCTION;
         let id = Parse.identifier env in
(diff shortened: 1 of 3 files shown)


gabelevi:
Hi @Eyyub, why do you want to be able to declare type aliases inside `declare module` declarations? What kind of code are you trying to write and how is this helping you?

Eyyub:
Hi, actually we are able to declare a type alias outside module declaration but not inside even if it has the same effect.  So,  IMHO I think it's good to allow type alias declaration in module, more for  the design than a usecase. It might also prevent a user to think about why he is not able to declare a type alias in a module declaration.

I hope you will understand my cheap english,  thanks :)

gabelevi:
I understand perfectly! Thanks for the explanation! My main hesitation here is that this works in the same way that declaring a type alias outside `declare module` does. If we were able to put it inside the `declare module`, then I think it would make sense for it to only be available inside the module.

@jeffmo, what do you think? This could go well with your `import type` feature...you could declare in a lib file what type aliases a module exports with `declare type MyType = ...`.

Eyyub:
You're welcome, actually this pull request as you said allow declare type in module as if the declaration was outside of module declaration, and I also think it would be better to restrict type alias availability like this ! I did not know it was the purpose of flow, it's great :+1:

jeffmo:
Awesome. I'd been thinking about something like this for a while, so glad to see you're looking at it too.

Now that we have `export type` and `import type` in modules, I think it would be preferable for a `declare type` inside a module declaration to actually export the type alias from the module (rather than be global).

Here's the bit of code that does this for regular modules if you want to take it as a guide for implementing `declare type`:

https://github.com/facebook/flow/blob/master/src/typing/type_inference_js.ml#L2191

Eyyub:
Mmh, ok, just to be sure, we are talking about this
```javascript
declare module Foo {                                                                                    
    declare type bar = string;                                                                          
    declare function foobar(s : bar) : void;                                                            
}  
/* ... */
var Foo = require('Foo');                                                                               
                                                                                                        
function baz(s : Foo.bar) {                                                                             
    Foo.foobar(s);                                                                                      
}                                                                                                       
                                                                                                        
function barbaz(s : bar)  { // being rejected   ?                                                                         
                                                                                                        
} 
```

jeffmo:
Almost. The following is more what I meant:

```javascript
declare module Foo {
  declare type bar = string;
  declare function foobar(s : bar) : void;
}
/* ... */
var Foo = require('Foo');
import type {bar} from 'Foo';

// Error because 'Foo' is a runtime object, types don't exist at runtime, so we
// don't try to converge the two
// (it would be impossible to transform type-only object properties in some cases)
function baz(s : Foo.bar) { ... }

// This works because of the `import type` above
function barbaz(s : bar) { ... }
```

(This will match with the way we support `import type`/`export type` currently)

jeffmo:
PS: If you think it would be helpful to chat this through on IRC more synchronously, I'm happy to talk it through more (I'm usually on there on weekdays most of the day).

Eyyub:
Thanks, indeed it will be simpler to discuss about that on IRC ! (my nickname is Eyyub)

Eyyub:
See #569.

PR #370, opened 2015-04-10
Dom canvas context

e98cuenc:
Complete canvas and HTMLVideoElement API.
Add missing CSS properties to CSSStyleDeclaration, and remove a function that actually doesn’t exist (in browser nor in the standard).


lib/cssom.js
```diff
@@ -58,5 +333,4 @@ declare class CSSStyleDeclaration {
   removeProperty(property: string): string;
   setProperty(property: string, value: ?string, priority: ?string): void;
   setPropertyPriority(property: string, priority: string): void;
-  setPropertyValue(property: string, value: string): void;
```
gabelevi: Hmm...looking into it I see it in the (working draft)[http://dev.w3.org/csswg/cssom/#the-cssstyledeclaration-interface]. I don't feel too strongly, but I don't particularly mind leaving in future features.

e98cuenc: I don't have a strong opinion either. I tried to use that function and it didn't work on any browser, checked the standard and it was not there, so I assumed it was a copy & paste error.

Let me know if you want it back and I will update the PR.

gabelevi:
Nah, I'm fine taking it out...if anyone actually needs it they can add it back in later.

Thanks for taking the time to add all the declarations!

PR #378, opened 2015-04-12
Add missing stream/child_process declarations

### LINKED ISSUE #129: Flow fails to find nodeJS core module, child_process
/* @flow */

    var cp = require('child_process');

produces error:

    foo.js:3:10,33: child_process 
    Required module not found

Using nodeJS v0.10.32 on OS X together with flow v0.1.1

josephsavona:
Adds/updates declarations for lib/node.js `Readable` and `Writable` stream classes and `child_process` type. The added test is based on the nodejs.org documentation examples.

Resolve #129.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -100,10 +100,10 @@ type child_process$spawnOpts = {
   uid: number;
 }
 
-type child_process$ChildProcess = {
-  stdin: stream$Stream;
-  stdout: stream$Stream;
-  stderr: stream$Stream;
+declare class child_process$ChildProcess extends events$EventEmitter {
+  stdin: stream$Writable;
+  stdout: stream$Readable;
+  stderr: stream$Readable;
   pid: number;
   connected: boolean;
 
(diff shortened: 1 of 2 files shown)


lib/node.js
```diff
@@ -524,16 +524,26 @@ declare module "readline" {
...
+  pipe(dest: any, options?: any): void;
 }
 
 declare class stream$Readable extends stream$Stream {
   setEncoding(): void;
+  isPaused(): boolean;
+  pause(): void;
+  read(size?: number): ?(string | Buffer);
+  resume(): void;
+  unpipe(dest: any): void;
```
gabelevi: Looks like the argument here is optional, right? https://nodejs.org/api/stream.html#stream_readable_unpipe_destination

lib/node.js
```diff
@@ -524,16 +524,26 @@ declare module "readline" {
...
+  unshift(chunk: any): void;
+  wrap(oldReadable: any): stream$Readable;
 }
 
 declare class stream$Writable extends stream$Stream {
   close(): void;
+  end(): void;
```
gabelevi: I think `end()` has 3 optional arguments

https://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback

lib/node.js
```diff
@@ -524,16 +524,26 @@ declare module "readline" {
...
   finish(): void;
+  setDefaultEncoding(): void;
```
gabelevi: This takes one argument

https://nodejs.org/api/stream.html#stream_writable_setdefaultencoding_encoding

lib/node.js
```diff
@@ -524,16 +524,26 @@ declare module "readline" {
...
+  write(): void;
```
gabelevi: This has one required argument and 2 optional ones

https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback

gabelevi:
Awesome! This looks great! Thanks for fixing this issue! I had a few nits inline, but other than that it looks perfect :)

mroch:
what's the status on this guy?

PR #385, opened 2015-04-13
Refine switch cases

mroch:
Refines switch statements as if they were `===` binary expressions with the discriminant on the LHS and the case expression on the RHS.

This is similar to the handling for if/else if/else, except complicated by fallthroughs. `fallthrough_ctx` is a clone of the `ctx` from before the switch, which has been refined by the //not// of all of the case statements seen so far. `last_ctx` is similar, but it assumes the previous cases can all be true -- except that each time a `break` or `return` happens, `last_ctx` gets emptied out. merging `fallthrough_ctx` with `last_ctx` gives you `case_ctx`, which is correct for just that case statement.

For example:

```
function x(foo: string | number | boolean) {
  switch (typeof foo) {
  case 'string':
    // fallthrough_ctx: string | number | boolean
    // case_ctx: (string | number | boolean) & string => string
  case 'number':
    // last_ctx: string
    // fallthrough_ctx: (string | number | boolean) & !string => number | boolean
    // case_ctx: ((number | boolean) & number) | string => number | string
    return foo;
  case 'boolean':
    // last_ctx: {}
    // fallthrough_ctx: (number | boolean) & !number => boolean
    // case_ctx: boolean & boolean => boolean
    return foo ? 'foo' : 'bar';
}
```

Test Plan: unit tests

### DIFF
--- a/tests/refinements/switch.js
+++ b/tests/refinements/switch.js
@@ -1,0 +1,55 @@
+/* @flow */
+
+function foo(text: string | number): string {
+  switch (typeof text) {
+　　case 'string':
+　　　return text;
+    case 'number':
+      return text; // error, should return string
+　　default:
+　　　return 'wat';
+　}
+}
+
+function bar(text: string | number): string {
+  switch (typeof text) {
+    case 'string':
+      return text[0];
+  　default:
+      return (text++) + '';
+　}
+}
+
+function baz1(text: string | number): string {
+  switch (typeof text) {
+    case 'number':
+    case 'string':
+      return text[0]; // error, [0] on number
+  　default:
+      return 'wat';
+　}
+}
+
+function baz2(text: string | number): string {
+  switch (typeof text) {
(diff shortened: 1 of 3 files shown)


avikchaudhuri:
This is awesome!

tests/refinements/switch.js
```diff
@@ -0,0 +1,55 @@
...
+  　default:
+      return 'wat';
+　}
+}
+
+function baz2(text: string | number): string {
+  switch (typeof text) {
+    case 'string':
+    case 'number':
+      return text[0];
```
gabelevi: Can you add the `// error` comment?

gabelevi:
So cool! This looks good to me... @avikchaudhuri or @bhosmer can one of you guys sign off here too?

gabelevi:
Spoke to @avikchaudhuri and he says this looks good. Thanks for your patience!

PR #393, opened 2015-04-15
Symbol should be called statically. Fix #391

### LINKED ISSUE #391: Symbol statics
```javascript
/* @flow */
const FOO = Symbol('foo');
```

causes the error "function call Callable signature not found in statics of Symbol".

terinjokes:


terinjokes:
IMHO, the following test should cause a failure:

```javascript
var WAT = Symbol('foo', 'bar');
```

The declaration is only for 1 (optional) parameter, but two are given.

jeffmo:
Can you add a test for the double-parameter circumstance and put a `// TODO: Expected error` by it? I agree this should be an error...

PR #397, opened 2015-04-17
Facilitate debug builds

popham:
Make a bytecode version of `flow` for use by `ocamldebug`.  This `flow` build allows devs to step through Flow with `flow single`.


gabelevi:
Cool, this looks like it should work. We embed the flowlibs in the flow binary, but I vaguely remember adding support for having the flowlib archive in the same directory: https://github.com/facebook/flow/blob/31c1df2b50534a6ba387e8ca8f1dcf50b8d6f54a/src/embedded/flowlib.ml#L40

It looks like the archive is being built and put in `/bin` so you should be good to go

PR #401, opened 2015-04-20
Improve error message clarity in a few places

jeffmo:
This improves error message clarity for ES6 module imports and also general property/global variables when the name being looked up is not found.

Rather than "unknown global variable", now we say "Could not resolve name: %s".
Rather than "Required module not found", now we say "Required module not found: %s".

ES6 module errors are also more fine-grained now (rather than just using a summary of the import string syntax as the reason for pretty much everything).

See updated tests for example output

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1201,10 +1201,12 @@ let rec flow cx (l,u) trace =
       (* Set the 'default' property on the new object to point at the exports *)
       unit_flow cx (ns_obj, SetT(reason, "default", exported_t));
 
-      let ns_obj_sealed = mk_tvar_where cx reason (fun t ->
-        unit_flow cx (ns_obj, ObjSealT(reason, t))
-      ) in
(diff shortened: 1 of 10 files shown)


gabelevi:
This is fantastic. The error messages are way more clear. Thanks for fixing this!

m1sta:
Does this change also replace things like...

```
hello.js:5:5,19: string
  hello.js:3:10,15: number
```

...with something like...

```
hello.js:5:5,19: function 'foo' called with parameter 'x' as a string however
  hello.js:3:10,15: the definition of function 'foo' infers that 'x' be a number
```

jeffmo:
@m1sta : Ohhh what a wonder world it'd be if our error messages were that beautiful :) Definitely a target to strive for, but for now no. This fix mostly just targets error messages for `import`s of things that weren't exported on the target module.

I think @bhosmer has been thinking about our error messaging "framework" a bit -- but that's a bit of a more general problem to solve.

PR #404, opened 2015-04-22
[WIP] Let/const

samwgoldman:
I've just started digging in my heels on this one. I'm keeping this in an open PR to get feedback and ensure I don't step on anyone's toes.

TODO:

- [x] represent different scope types
- [ ] rename "block" to "scope" (@jeffmo you mentioned this, seems like a good idea)
- [x] create block-level scopes
- [x] hoist vars, classes, functions
- [x] allow defining let/const identifiers
- [ ] test for-loop statement and iteration scope
- [ ] strict mode vs. not
- [ ] should functions be block level???
- [ ] TDZ - I don't think we can statically ensure define before use, but best effort
- [ ] error on const reassignment
- [ ] let block
- [ ] let expressions

Overview:

In order to support let/const, we need to know what kind of scope each
block is.

Currently, the only blocks we have are what I am calling "Hoist" blocks.
Whenever an entry which should be hoisted is added to the env, it should
be added to the closest "Hoist" block.

Let/const identifiers should be added to the nearest "Block" block.

Currently we only create blocks corresponding to hoist points.
Subsequent commits will need to create a great many more blocks to
support block-level scope.


jeffmo:
cc @avikchaudhuri as well to follow along

samwgoldman:
The test failure in that last commit is throwing me for a loop. I'll look again tomorrow, but if anyone has some insight, I'd appreciate it.

samwgoldman:
OK! Figured that last error out. Some preliminary feedback on this approach would be really helpful.

Also, if anyone has a real-life project with let/const, and is able to try out this branch on their project, I'd be happy to look into any issues.

jeffmo:
Haven't gotten to the meatier portion of the PR yet, but wanted to send what I have so far before I run to lunch. Will come back to finish out the comments later today/tonight

src/typing/constraint_js.ml
```diff
@@ -507,7 +507,16 @@ type block_entry = {
   def_loc: Spider_monkey_ast.Loc.t option;
   for_type: bool;
 }
-type block = block_entry SMap.t ref
+
+type block_scope =
```
jeffmo: Let's call this "scope_kind" (i.e. odd that the name seems to imply one of the variants but not the other -- in fact the issue really is that "block" means two different things now, but more on that in the next comment...)

src/typing/constraint_js.ml
```diff
@@ -507,7 +507,16 @@ type block_entry = {
...
+  | Hoist (* var, class, functions hoisted up to this point *)
+  | Block (* let, const *)
+
+type block = {
```
jeffmo: Now that 'block' is term that refers to a kind of scope (which isn't the same as what this name refers to as a "block"), it's probably time to rename this structure to 'scope'

src/typing/constraint_js.mli
```diff
@@ -266,7 +266,16 @@ type block_entry = {
   def_loc: Spider_monkey_ast.Loc.t option;
   for_type: bool;
 }
-type block = block_entry SMap.t ref
+
+type block_scope =
+  | Hoist (* var, class, functions hoisted up to this point *)
```
jeffmo: Because variant names quickly lose the (often useful) context of the name of the type they're a part of, it's probably going to be more clear in callsites if you name these variants of the form `| HoistedScope | BlockScope`.

The "scope" becomes redundant here at the definition, but it becomes incredibly useful at callsites

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
     | block::blocks ->
-        (match SMap.get x !block with
+        let entries = block.entries in
+        (match SMap.get x !entries with
         | Some entry when (not entry.for_type || for_type) -> block
         | _ -> loop blocks)
   in
   loop (!env)
 
 let get_block x block =
```
jeffmo: Probably time to rename this to `get_from_scope`
(not your doing, but while you're in here let's clean things up for clarity a bit)

Maybe these terminology renames would make more sense as a precursory pull request to simplify the real work going on in this one? As mentioned above, though, I think they become even more necessary now that 'block' has quite a different meaning after this work.

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
-  SMap.find_unsafe x !block
+  let entries = block.entries in
+  SMap.find_unsafe x !entries
 
 let set_block x entry block =
```
jeffmo: `set_in_scope`

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
-  block := !block |> SMap.add x entry
+  let entries = block.entries in
+  entries := !entries |> SMap.add x entry
 
 let exists_block x block =
```
jeffmo: `exists_in_scope`

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
-  SMap.get x !block
+  let entries = block.entries in
+  SMap.get x !entries
+
+let set_env x entry hoist =
```
jeffmo: `set_in_env` and maybe use the `scope_kind` variant here (rather than a bool) so that it's more immediately clear what callsites are doing when they call this function (rather than passing `true` or `false` which could mean anything until you go and find and read the function definition).

i.e. `set_in_env "myVar" var_entry HoistedScope` or

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
+  let rec loop = function
+    | [] -> set_block x entry global_block
+    | block::blocks ->
+        if (not hoist || (hoist && block.scope = Hoist)) then
+          set_block x entry block
+        else
+          loop blocks in
+  loop (!env)
+
+let exists_env x hoist =
```
jeffmo: `exists_in_env x scope_kind`

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
+        if (not hoist || (hoist && block.scope = Hoist)) then
+          exists_block x block
+        else
+          loop blocks in
+  loop (!env)
 
 let update_block ?(for_type=false) x (specific_t, general_t) block =
-  let values = !block in
-  let new_entry = match SMap.get x values with
+  let entries = block.entries in
```
jeffmo: Love how this reads now -- so much more clear! ("entries" in a "scope" :ok_hand:)

src/typing/env_js.ml
```diff
@@ -116,9 +146,12 @@ let pop_env () =
   env := List.tl !env
 
 let flat_env () =
```
jeffmo: `flatten_env`

src/typing/env_js.ml
```diff
@@ -116,9 +146,12 @@ let pop_env () =
...
-  List.fold_left (fun acc x ->
-    SMap.union acc !x
-  ) !global_block !env
+  let rec loop acc env =
+    match env with
```
jeffmo: nit: You can shorten 149-150 to

```
let rec loop acc env = function
  | { entries; _ }::bs -> ...
  | [] -> ...
```

src/typing/env_js.ml
```diff
@@ -127,10 +160,9 @@ let switch_env block =
 let peek_env () =
   List.hd !env
 
-let init_env cx x shape =
-  let block = peek_env() in
-  match exists_block x block with
-  | None -> set_block x shape block
+let init_env cx x shape hoist =
```
jeffmo: s/`hoist`/`scope_kind`/

src/typing/type_inference_js.ml
```diff
@@ -4674,6 +4680,7 @@ and mk_body id cx param_types_map param_types_loc ret body this super =
           (create_env_entry ret ret None)
   )
   in
+  let block = { scope = Hoist; entries } in
```
jeffmo: The spec states that class bodies have a normal "lexical environment" (aka this should probably be `BlockScope` rather than `HoistedScope`)

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-runtime-semantics-classdefinitionevaluation

samwgoldman:
I completely agree about the renames, and it's encouraging that your suggestions are so in line with what I was thinking. The only reason I hesitated to rename now is that I didn't want to cause merge difficulties down the road, but I think a bit of upfront pain is better than wonky names. And hey, it's not my problem ;).

src/typing/type_inference_js.ml
```diff
@@ -1135,7 +1133,14 @@ and statement cx = Ast.Statement.(
   | (loc, Empty) -> ()
 
   | (loc, Block { Block.body }) ->
-      toplevels cx body
+      let entries = ref SMap.empty in
+      let block = { scope = Block; entries } in
+      Env_js.push_env block;
+
+      List.iter (statement_decl cx) body;
```
samwgoldman: Wan to call out this bit specifically. Is it wrong to jump back into the `statement_decl` codepath from the `statement` codepath like this? If this is wrong, then that somewhat invalidates my approach here.

jeffmo:
Heh. Well if it's not much harder for you, making the renames in a separate PR (that this one could be considered "stacked" on top of) might make my life a little easier :)

We can just make sure they get merged in proper order

src/typing/type_inference_js.ml
jeffmo: I think this is ok -- it's essentially what functions currently do (except way over in the `mk_body` function)

samwgoldman:
Yeah, I think it makes sense to do a separate PR that renames `block` to `scope`, and some of the other changes you identified, so that change can be rolled in sooner. I'll go ahead and do that before moving on with this let/const stuff.

src/typing/env_js.ml
samwgoldman: This is the only change I'm not sure I agree with. `set_in_env` and `set_in_scope` actually mutate the relevant refs, but `flat_env` returns a new value and leaves the env and global scope unchanged. `flattened_env` might be preferable to `flat_env`, but `flatten_env` makes it seem like this method would change env when defined adjacent to other mutating methods.

jeffmo: :+1:

mroch:
not sure where you guys ended up on the naming of the two different types of scopes (`HoistedScope` and `BlockScope`?) but the spec seems to refer to them as `LexicallyScoped` and `VarScoped` so perhaps we could go with `LexicalScope` and `VarScope`?

samwgoldman:
I can get behind that naming, for sure.

jeffmo:
I've always found "lexically scoped" pretty confusing (both are "lexical"). I could live with it if someone feels strongly, but I'd be even happier with "VarScope" and "LetScope"

mroch:
yeah I find it non-optimal too, but since we're following a spec I thought it might be easier if they matched so that you don't have to wonder about/remember the translation between the code and the spec. if we name one `VarScope` then `LexicalScope` vs `LetScope` probably doesn't matter as much because I can remember it as the "other" one.

jeffmo:
Ok, will leave it up to you @samwgoldman -- I can live with either I guess

samwgoldman:
We're talking about [these definitions](http://people.mozilla.org/~jorendorff/es6-draft.html#table-23), correct? It makes sense to me that `VarScope` is-a `LexicalScope`, and I don't see that as a problem. Not the most confusing subtyping relationship in flow. :)

Since it makes sense to me, and I like the idea of naming things close to the spec, I'll try it out and we can see how it feels.

mroch:
I was actually looking at these:

http://people.mozilla.org/~jorendorff/es6-draft.html#sec-block-static-semantics-lexicallyscopeddeclarations
http://people.mozilla.org/~jorendorff/es6-draft.html#sec-block-static-semantics-varscopeddeclarations

(and similar sections all over section 13)

src/typing/type_inference_js.ml
samwgoldman: To be clear, an example for the described behavior would be as follows?

```javascript
while (true) {
  class Foo { };
  break;
}
var fail = new Foo(); // error
```

samwgoldman:
Closing in favor of rebased #431

PR #417, opened 2015-04-25
Clean up some docs on the github.io site

jeffmo:
I just had some downtime, so I ran through a few of the docs on the site to clarify and bring things up to date a bit. I'm going to try and run through more of the docs later, but this is what I came up with in the last half hour.


docs/02-new-project.md
```diff
@@ -73,14 +73,14 @@ It is possible to indicate on the command line that you would like Flow to check
 $> flow check --all
 ```
 
-However, use this command cautiously, particularly if you are running Flow against a large existing project or one with many large third party frameworks or test suites. The chances are that the checker will find a *lot* of errors, and it can be quite demoralizing!
+However, use this command cautiously; Particularly if you are running Flow against a large existing project or a project with many large third party frameworks or test suites. The chances are that the checker will find a *lot* of errors or hazards, and it can be a little overwhelming!
```
samwgoldman: the ", particularly" part was more correct, because the clause starting with "particularly" is not independent.

gabelevi: Also you don't capitalize after semicolons, do you? For shame :stuck_out_tongue_winking_eye: 

Also I like how verbose the `:stuck_out_tongue_winking_eye:` emoji is! Hooray for descriptive naming!

jeffmo: Pff, duh. Everyone knows that semicolons are useless in every language.

(will revert this)

PR #421, opened 2015-04-26
Ugly printing instead of `<abstr>` under ocamldebug (and toplevel?)

popham:
I've been getting my bearings within the codebase by walking around in `ocamldebug`.  Unfortunately, the debugger obscures most of the data behind `Map` interfaces (`<abstr>` renderings).  I've integrated [Deriving](https://github.com/ocsigen/deriving) into the build process and marked a whole bunch of types `with show`.  Comments in `Makefile` clarify the workflow.  I believe that the changes would help others get started.

I'm sure that I have violated many necessary conditions with the changes (for instance, I figure that touching the `hack` directory may be out of bounds).  If y'all are interested in supporting the workflow, I'd be willing to adapt to a few constraints.


gabelevi:
Sorry for the slow reply! I haven't seen Deriving before, but if you can make ocamldebug useable then this is super valuable stuff! You're right about the hack stuff being out-of-bounds (it lives [here](https://github.com/facebook/hhvm/tree/master/hphp/hack) and is currently just vendored in). I've pinged the rest of the team to see if anyone has any valuable advice :)

gabelevi:
Sorry about the silence on this. I think the consensus among the team was that we have pretty printers for most things before and we'd rather not use a preprocessor at this point to keep things simple. That being said, it would be useful having these pretty printers installed for ocamldebug.

popham:
No problem.  It helped me get my bearings, but looks like maintenance overhead.  I've since discovered the CLI's `--traces` option.

PR #425, opened 2015-04-27
http server declaration

gyzerok:
Added missing http server and createServer declarations


lib/node.js
```diff
@@ -551,6 +551,15 @@ declare module "fs" {
 }
 
 declare module "http" {
+  declare class Server extends events$EventEmitter {
+    listen(port: number, hostname?: string, backlog: number, callback?: Function): Server;
```
gabelevi: I think backlog should be optional too?

lib/node.js
```diff
@@ -551,6 +551,15 @@ declare module "fs" {
...
+    listen(path: string, callback?: Function): Server;
+    listen(handle: Object, callback?: Function): Server;
+    close(callback?: Function): Server;
+    address(): { port: number; fmaily: string; address: string; };
```
gabelevi: I don't see this in [http.Server](https://nodejs.org/api/http.html#http_class_http_server) but I do see this in [net.Server](https://nodejs.org/api/net.html#net_server_address). Am I looking at the wrong docs?

gabelevi:
I have some small nits but this generally looks good to me!

gyzerok:
@gabelevi I guess here is a bug in node documentation since I do know that http.Server has address method. Check TypeScript [declarations](https://github.com/borisyankov/DefinitelyTyped/blob/master/node/node.d.ts#L347).

PR #427, opened 2015-04-27
Add "Immediate" global timer funcs to Node lib

tgroshon:
- add setImmediate function declaration
 - add clearImmediate function declaration

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -758,3 +758,6 @@ declare module "assert" {
   declare function doesNotThrow(block: Function, message?: string): void;
   declare function ifError(value: any): void;
 }
+
+declare function setImmediate(callback: any, ...args: Array<any>): any;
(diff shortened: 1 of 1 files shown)


tgroshon:
Done!

jeffmo:
Nice! We can type the `callback` params as `((...args: Array<any>) => void|string` to be more specific, right?

tgroshon:
@jeffmo Sure!  But how about just `((...args: Array<any>) => void)` because I do not think setImmediate takes a string.

https://github.com/joyent/node/blob/master/lib/timers.js
line 397

jeffmo:
Ah, yea that's great! (I guess I got it mixed up with the weird API that browsers have for setTimeout)

tgroshon:
This should be ready to go!

jeffmo:
Thanks a ton, this is great

PR #431, opened 2015-04-29
Let/const

samwgoldman:
Rebased #404 on top of the `block`→`scope` rename.

- [x] represent different scope types
- [x] create lexical scopes
- [x] allow defining let/const identifiers
- [x] test for-loop let-bound variables
- [x] test for-in-loop let-bound variables
- [x] test scoping behavior of functions and classes
- [x] verify scoping behaviors for interfaces, declarations, import/export stuff
- [x] verify scoping on refinements
- [x] TDZ - I don't think we can statically ensure define before use, but best effort
- [x] error on const reassignment
- [x] for-of support
- [x] destructive ops (e.g., `i++`)
- [ ] fix const destructuring
- [ ] ensure TDZ errors out in `typeof` expressions

Left for future work
* very sophisticated TDZ
* let blocks
* let expressions
* let/const in lib declarations


src/typing/constraint_js.ml
```diff
@@ -507,7 +507,16 @@ type scope_entry = {
   def_loc: Spider_monkey_ast.Loc.t option;
   for_type: bool;
 }
-type scope = scope_entry SMap.t ref
+
+type scope_kind =
+  | VarScope (* var, class, functions hoisted up to this point *)
```
jeffmo: FYI: Classes are actually let-scoped

src/typing/constraint_js.mli
```diff
@@ -266,7 +266,16 @@ type scope_entry = {
   def_loc: Spider_monkey_ast.Loc.t option;
   for_type: bool;
 }
-type scope = scope_entry SMap.t ref
+
+type scope_kind =
+  | VarScope (* var, class, functions hoisted up to this point *)
```
jeffmo: (same)

src/typing/env_js.ml
```diff
@@ -75,33 +80,58 @@ let find_env ?(for_type=false) cx x reason =
...
 let exists_in_scope x scope =
-  SMap.get x !scope
+  let entries = scope.entries in
+  SMap.get x !entries
+
+let set_in_env x entry scope_kind =
+  let rec loop = function
+    | [] -> set_in_scope x entry global_scope
+    | scope::scopes ->
+        if (scope_kind = LexicalScope || scope.kind = scope_kind) then
```
jeffmo: Won't this cause `var`s to get set in the first lexical scope they hit?

Thinking about this a bit, I actually think we need to introduce one `LexicalScope` directly under each `VarScope` when we create the scopes. Consider this test case:

```
let thing = 42;

function foo() {
  let thing = 42; // This gets scoped to the function env
}
```

samwgoldman: No, if the entry is var-scoped, then `scope_kind` will be `VarScope`, so the condition will fail.

Also, this same condition ensures that let-scoped variables will be installed into any kind of scope. (That's what the first part is all about, the second part ensures VarScope entries is installed only into VarScope scopes.)

You'll find that both of those test cases work today with the existing changes.

samwgoldman:
To handle some TDZ cases and const reassignment, need to track more information that I do now about the scope entries. I have an idea for how I can proceed with that, but if anyone has a preference for how I should proceed, I appreciate all input.

samwgoldman:
Regarding TDZ, I'm not sure it's possible to (in general) prevent errors like this:

```javascript
function bar() {
  console.log(foo);
}

bar();

let foo = "foo";
```

because the following *is* valid:

```javascript
function bar() {
  console.log(foo);
}

let foo = "foo";

bar();
```

Thoughts?

src/typing/env_js.ml
jeffmo: Ah I think you're right -- I was getting `scope_kind` and `scope.kind` reversed in my head. Cool

jeffmo: Mind adding some [regression] tests covering those scenarios at some point?

samwgoldman: Will do. I'm confident existing tests indirectly cover this stuff, but it doesn't hurt to be explicit.

jeffmo:
Right, some TDZ errors for `let` are definitely going to be harder (and sometimes impossible) to identify statically. We should try do best-effort here, but maybe that kind of granularity could be a added in more detail a follow-up PR?

For now, same-env TDZ errors seems like a good start

src/typing/type_inference_js.ml
```diff
@@ -870,7 +865,7 @@ and statement_decl cx = Ast.Statement.(
       let _, { Ast.Identifier.name; _ } = id in
       let r = mk_reason (spf "type %s" name) loc in
       let tvar = Flow_js.mk_tvar cx r in
-      Env_js.init_env cx name (create_env_entry ~for_type:true tvar tvar (Some loc))
+      Env_js.init_env cx name (create_env_entry ~for_type:true tvar tvar (Some loc)) VarScope (* TODO: hoist behavior *)
```
samwgoldman: @jeffmo What do you think the scoping behavior for type aliases should be?

src/typing/type_inference_js.ml
```diff
@@ -970,7 +965,7 @@ and statement_decl cx = Ast.Statement.(
       let _, { Ast.Identifier.name; _ } = id in
       let r = mk_reason (spf "class %s" name) loc in
       let tvar = Flow_js.mk_tvar cx r in
-      Env_js.init_env cx name (create_env_entry tvar tvar (Some loc))
+      Env_js.init_env cx name (create_env_entry tvar tvar (Some loc)) VarScope (* TODO: hoist behavior *)
```
samwgoldman: @jeffmo I'm fairly sure that declared classes should have lexical scoping rules, but what about interfaces? There's no guidance from ES6, but I think the spirit of the decision to make classes lexically scope also argues for the same behavior in interfaces.

src/typing/type_inference_js.ml
```diff
@@ -982,7 +977,7 @@ and statement_decl cx = Ast.Statement.(
       let r = mk_reason (spf "module %s" name) loc in
       let t = Flow_js.mk_tvar cx r in
       Hashtbl.replace cx.type_table loc t;
-      Env_js.init_env cx (spf "$module__%s" name) (create_env_entry t t (Some loc))
+      Env_js.init_env cx (spf "$module__%s" name) (create_env_entry t t (Some loc)) VarScope (* TODO: hoist behavior *)
```
samwgoldman: @jeffmo Modules is an interesting case. Does it even make sense to argue non-global modules? Does this argue for a GlobalScope alternative to the scope_kind type?

src/typing/type_inference_js.ml
jeffmo: Definitely var-scoped -- type aliases need mutual recursion

src/typing/type_inference_js.ml
jeffmo: I'd agree and say that interface definitions should be hoisted for the same reason type aliases should (mutual recursion is important there).

History lesson: The reason classes aren't hoisted is because of hazards with extends clauses. For example, this would be a problem if class decls were hoisted:

```
var _baseClass = SomeBaseClass;
class Foo extends _baseClass {}
```

src/typing/type_inference_js.ml
jeffmo: Hmm interesting. Yea, a (mostly one-off for this scenario) `GlobalScope` variant makes sense to me, too.

This is more an implementation detail of the way Flow deals with module declarations, but I agree that it doesn't make sense to declare modules in any scope other than global

samwgoldman: I tried sticking module entries in Env_js.global_scope, but it doesn't work. I get `Required module not found` errors. Var scope seems to work fine, and the syntax prevents declaring modules in a nested scope, so I think it should be OK to leave declared modules as `VarScope`.

src/typing/env_js.ml
```diff
@@ -327,7 +401,7 @@ let string_of_env cx ctx =
 let install_refinement cx x xtypes =
   if exists_in_scope x (peek_env ()) = None then
     let t = SMap.find_unsafe x xtypes in
-    init_env cx x (create_env_entry t t None)
+    init_env cx x (create_env_entry t t None) VarScope (* TODO: verify correct hoist behavior *)
```
samwgoldman: I'm pretty sure that the scope of the refinement depends on the type of entry corresponding to x. Furthermore, code that relies on `exists_in_scope`/`peek_env` instead of `exists_in_env`/`scope_kind` is generally suspect.

Unfortunately, we don't generally know the scope kind of the entry here, and I'm not even sure we can thread it through. The refinement logic is still a bit opaque to me, but I'll try to come up with some test cases to stress this code. Help here would be appreciated.

samwgoldman: OK, so I think I have two tests that show the refinement scope needs to depend on the scope of the thing being refined.

Here is why just `VarScope` doesn't work:
```
function scope() {
  {
    let x: {p:?string} = {p:"xxx"};

    if (x.p == null) {
      return;
    }
  }
  x.p // should fail, but $REFI x.p is hoisted to here, so there is no warning
}
```

Here is why just `LexicalScope` doesn't work:
```
function scope() {
  var x: {p:?string} = {p:"xxx"};
  {
    if (x.p == null) {
      return;
    }
  }
  var y: string = x.p // should pass, but refinement doesn't live long enough
}
```

@gabelevi, you helped me understand this refinement stuff in the first place (thanks!). Do you see any problem with my reasoning here? It seems like the scope_kind of the refinement needs to depend on the scope_kind of the thing being refined (`x` in this case).

If that's true, we currently don't store that information beyond `init_env`, so I believe I'd need to add scope_kind to the scope_entry struct. (If true, this isn't so bad, because I think we need to add this kind of information to support TDZ use cases anyway.)

samwgoldman: I've implemented this in 4659a41ed64ba840609a5c0a032612a2a90224cd. Mostly sure it's correct :)

samwgoldman:
This is ready for review.

jeffmo:
Woo! I'm so excited to read through this :)

I'm going to be fairly busy with conference stuff next week, so I will do my best to find some time to read through it (in case someone doesn't beat me to it).

samwgoldman:
So I realized that I didn't catch mutating ops like `i++` and `i += 1`. From what I can tell, there's not a great place to fit that in except in the `assignment_lhs` function's `Ast.Expression.Identifier` branch. Even then, we'd need to call out to some currently non-existant `Env_js` function that would just check if the binding were constant, which seems gross.

I think I'm probably not handling constant bindings correctly in general, so before I want to hold off before making the change I described above.

src/typing/constraint_js.ml
```diff
@@ -529,21 +529,34 @@ let copy_node node = match node with
     Root { rank; constraints = Unresolved (copy_bounds bounds) }
   | _ -> node
 
+type scope_kind =
```
jeffmo: Idea: If we called this 'binding_type' instead of 'scope_kind' we could probably get away with creating something like a `ConstantBinding` (and even a `TypeBinding`) -- which would let us get rid of the `for_type` and `constant` flags in all of the utils functions + `scope_entry` record.

This seems like more of a nice-to-have, though, so unless you're just feeling eager it's probably fine to do such a consolidation sometime after this goes in.

src/typing/constraint_js.ml
```diff
@@ -529,21 +529,34 @@ let copy_node node = match node with
...
+  | VarScope (* var, functions hoisted up to this point *)
+  | LexicalScope (* let, const, classes *)
+
 type scope_entry = {
   specific: Type.t;
   general: Type.t;
   def_loc: Spider_monkey_ast.Loc.t option;
   for_type: bool;
+  constant: bool;
+  scope_kind: scope_kind;
```
bhosmer: Reading top-down, so apologies if the motivation becomes clear later, but: wouldn't the scope_kind for an entry be completely determined by the scope the entry was in?

src/typing/env_js.ml
```diff
@@ -29,6 +29,10 @@ let refinement_key names =
 let is_refinement name =
   (String.length name) >= 5 && (String.sub name 0 5) = "$REFI"
 
+let refined_object_from_key name =
+  let i = String.index name '.' in
+  String.sub name 6 (i - 6)
```
bhosmer: Definitely need some length checks and so on here - Ocaml is pretty throw-happy on out of range errors and such.

Per another discussion we had recently, I'd move all the key-wrangling stuff into Type_inference_js.Refinement. (Note that I haven't gotten to where this is actually used, may have more useful stuff to say then.)

src/typing/env_js.ml
```diff
@@ -44,7 +48,10 @@ let swap_changeset f =
   changeset := newset;
   oldset
 
-let global_scope: scope = ref SMap.empty
```
bhosmer: Fair warning, the env/global_scope situation was tweaked recently (actually, it's the stuff @mroch did that makes it possible to refine globals). Shouldn't be too hard to merge, but nontrivial.

src/typing/constraint_js.ml
samwgoldman: Let-declarations in a var-scope behave differently than in a let-scope.

```
function x {
  let x = "x";
  let x = "y"; // ok
}
```

```
{
  let x = "x";
  let x = "y"; // error
}
```

That's why I store the scope_kind on both, but I think @jeffmo made a good point above. By using the same data type for both scopes and entries, I'm actually conflating two different concepts. I think it would be more useful to have a `binding_type` as Jeff suggested, which would come with other niceties like removing `for_type` and `constant` optional parameters, and just modeling the domain better.

src/typing/env_js.ml
samwgoldman: This is only used in `install_refinement`, which is called from `Env_js` where we've already verified `is_refinement name`. I agree this is ugly, though. I would like to encapsulate this more, but I feel like adding a dependency from `Env_js` to `Type_inference_js` is in the wrong direction here.

src/typing/env_js.ml
samwgoldman: Err, I already merged that locally. Thought I pushed, but apparently I didn't. I'll push that up now.

src/typing/constraint_js.ml
bhosmer: Ah ok - the comments on VarScope and LexicalScope led me to believe that an entry belong exclusively to one or the other. Maybe worth clarifying those? ...but separately yeah, definitely +1 on Jeff's suggestion.

src/typing/env_js.ml
bhosmer: Yeah, true on the dependency. You could always just wrap this stuff up in its own module here - I probably should've done that in the first place.

bhosmer:
Hey Sam, I'm going to defer a more thorough review until you've done the scope-vs-entry split, if you don't mind. But do LMK if there's anything in the current diff that you'd really benefit from getting more eyeballs on before then.

samwgoldman:
@bhosmer Sounds good. If you have any opinion on how to handle destructive ops, per [my comment above](https://github.com/facebook/flow/pull/431#issuecomment-107004462), that would be helpful, and is orthogonal to the scope/entry split. My main question there: is it reasonable to add a function to `Env_js` that just queries whether a binding is constant?

jeffmo:
@bhosmer should feel free to chime in if he feels differently here -- but if you were to switch from tracking scope types to tracking binding types, then it seems mostly reasonable to me to return that tag as part of a tuple from all `get_var`/`var_ref`/etc util functions for callers to make use of them at will.

This seems more efficient, as well, since we'll know the binding type at the time of looking up the var (rather than having to find the var type -- then seek again to find the kind of the var type in a separate lookup). For callers that don't care, there's always our trusty `fst` and `snd` ocaml util functions for throwing away the other part of a 2-tuple.

bhosmer:
Re is-constant queries etc. Pretty much what @jeffmo said, I think. Probably the API I'd expect to see is something like (names out of a hat):
* get_entry returns basically the whole record, binding type included, location etc etc
* get_var get_const etc would be convenience functions that filter for the desired binding type - could return the whole record or omit stuff that's redundant under the call
* if it makes client code nicer, could always add (const_exists name) => get_const name != None, etc. They don't add complexity, so it's just a matter of whether they pay their way in client cleanup.

bhosmer:
...zooming out: the right way to think of the API is that we started with a stack of single-keyed maps (on name), and have slowly been accreting ad-hoc ways of discriminating on a second key (~for_type etc.). Now's the time to move to a fully double-keyed setup, with part of the API servicing (name, *) kind of queries. 

As a matter of design, the key thing is to make sure that the second key (BindingType) is either a clean, disjoint partition, or if it isn't, that the overlaps are cleanly defined. E.g. there are places in the current code where for_type=false acts on both type and non-type entries, but for-type true restricts. That's the kind of thing that needs to be called out clearly in the BindingType definition, if only in comments.

samwgoldman:
Thanks a lot for that feedback. Returning the entire scope_entry, vs a (specific_t, bindind_type) tuple seems about equivalent to me. All existing callers will need to destructure the return value. I think the tuple version provides more guidance about how to use the return value, whereas the entry is more general purpose.

Ultimately, your insights have given me exactly the frame of reference I need to make progress here. Thanks again and stay tuned.

bhosmer:
Cool! On the edge of my seat :D

src/typing/constraint_js.ml
samwgoldman: It turns out the behavior of let-declarations in var-scope I showed above is from an older version of the spec. Newer versions have [simpler semantics](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-function-definitions-static-semantics-early-errors)

    It is a Syntax Error if the LexicallyDeclaredNames of FunctionStatementList contains any duplicate entries.

Doesn't change the fact that I should clarify the existing comments, add more comments, and work on Jeff's suggestion!

src/typing/env_js.ml
samwgoldman: In the end, adding a module here seemed like overkill. Instead I was able to combine the `is_refinement` and `refined_object_from_key` functions into a single function, which I also renamed `heap_refinement`.

samwgoldman:
It wasn't too hard to remove `for_type` from the `scope_entry` record, but it's proving incredibly hard to remove it from the parameter list of `Env_js` functions. There's just a lot of code that depends on the ability to easily flip that switch.

* `DeclareClass` vs `InterfaceDeclaration` install a `VarBinding`[1] and `TypeBinding` respectively via `init_env` in `toplevels`. Then in `statement` we need to `set_var` with the correct lookup. Is the trick here to separate out `set_var` from `set_type`? Basil, is that what you mean by a “fully double-keyed setup”?

* The story is similar for `ExportDeclaration`. `top levels` installs a `TypeBinding` when the declared export is a `TypeAlias`, or else the appropriate value binding. `statement` then needs to `set_var` with the correct lookup again.

* Indeed, we see the same pattern repeated with `ImportDeclaration`. We install a `TypeBinding` when `importKind` is `ImportType` or `ImportTypeof`, otherwise a `VarBinding`. `statement`, again, needs to `set_var` with the correct lookup rule.

* For a change of pace, there is `Type_inference_js.convert_qualification`, which I am pretty confident never actually needs to invoke `get_var` with `for_type:true`. From the code, it appears that maybe an early module implementation required that, but it’s no longer the case. I think this one is solved.

* Lastly, there’s `Type_inference_js.identifier` which is used both to resolve unqualified, non-special-cased type annotations and also to look up `Ast.Expression.Identifier` names. This seems to pretty strongly suggest a `get_type`/`get_var` split, but currently `identifier` has behaviors that apply to both, like a call to `Type_inference_hooks_js`. If I broke this method in two, I’d just be introducing duplication here.

The relationship between `toplevels` and `statement_decl` is not surprising, of course. Maybe the duplication there is a signal for a further refactoring opportunity that was obscured by the convenience of switching behaviors with the `for_type` parameter. I think this could easily be true of `identifier` as well.

Ultimately, I could use some more input on what kind of “double-keyed” API we should have for Env_js. I tried to outline the uses of the current API where the type/value divide comes into play.

I’m writing this out because the further I dig in here, I feel increasingly out of my depth. I think that changing the `Env_js` API to remove `for_type` could provide some serious benefits, but as I pull the thread, I can’t help but feel like that change needs to be done on top of some preliminary refactoring that is out of my reach.

Phew! If you’ve made it this far, thanks for sticking in there. Hope I got that out of my head in one piece. Let me know if I was unclear. :)

1. `DeclareClass` should be a `LexicalBinding` but it breaks too many existing declarations

bhosmer:
Hey Sam – thanks a bunch for this detailed progress report. This is just a heads-up that it may be as late as tomorrow AM before I get a chance to dig in at any depth - but having scanned your note, I agree that we should try to find a sweet spot – land your work in a relatively clean way, but without blocking on an exhaustive API revamp. More soon. And thanks again!

Basil

bhosmer:
Hi Sam - after a bit more in-depth review, I agree, you're at a disadvantage because you're building on top of infrastructure that's already a bit overburdened and fuzzy at the edges. Doesn't seem worth it to block your stuff on removing for_type, or deeper cleanup.

That said, we're adding a good bit of complexity here, so I'd like to make things as tight as possible where it's relatively cheap to do so. Will follow with inline notes to that end.

A logistical note: I'm in the midst of some stuff (async/await) that requires some extra structure in the same places as let/const (in particular, scopes need to carry attributes, and API needs to query those). I'll try to harmonize those changes with what you've got going here in a way that minimizes rebase pain, but just in case, apologies in advance. :)

samwgoldman:
Cool. Don't hold back on exposing complexity here, though. If there's a better way to approach this, I'd happily scrap this and start over. Please be ruthless!

samwgoldman:
If you think it makes sense to revisit this after async/await, that's cool too. I'm going to be busy next week preparing for my wedding, so there's plenty of time to get this right.

bhosmer:
Whoa - congrats! 

Okay, sounds good. I should be able to add the stuff I need for async in a way that will make your life a little easier, among other things. Let's resume once you're an honest man. :)

bhosmer:
Hi Sam - I know you're on and off currently, but a quick status update: some of the refactoring we talked about went in as part of async/await, and I'll shortly put up one more diff with some cleanup around for_type. At that point I'd be happy to go over the changes and how best to layer your let/const work on top (as you'll see, I tried to aim the refactored stuff roughly towards a meeting point with yours :).

samwgoldman:
Thanks, Basil! Still getting used to the real world, but I have had the opportunity to review your work with async/await—cool stuff. I'll be hanging around IRC this week and next, and would really appreciate any guidance you can offer. Looking forward to seeing the next batch of changes!

nmn:
Any update on this pull request yet? Really waiting for this!

samwgoldman:
@bhosmer I've been thinking a bit about the structure here, and I had a thought: Currently, the type inference code goes through 2 phases: `statement_decl` and `toplevels`. `statement_decl` parses through the AST and set's up the env structures (and detect some errors). `toplevels` actually does the inferencing, querying from the env and creating flows.

It's not quite a clean 2-phase system though. Sometimes `toplevels`-callees call back to `statement_decl`. Actually, this happens whenever we introduce a new scope, AFAICT, and adding block scopes only makes this issue more pronounced.

What if we introduced a new intermediate representation between the AST and the actually inferencing work in `toplevels`? I feel like this could actually _replace_ the env structure, so it could definitely clean up the wonky query API there. So phase 1 would turn the AST into IR, and phase 2 would consume the IR to create flows.

Big effort, for sure, but seems like it could lead to a couple big wins:
* performance - no back-and-forth between phases, which gets worse w/ let/const
* clarity - inferencer would consume data more tailored to its needs
* clarity - IR could consolidate some of the mutable state that makes the code harder to understand

Thoughts?

nmn:
While work on Let/Const continues, is there a way to tell flow to treat Let and Const as simple vars and continue to work.

Many of us have been using Babel for our code, and use Let and Const instead of var everywhere. The fact that const isn't re-assigned is checked by Babel.

But flow completely fails on such code. Having it treat Let and Const the same as var for now would be a great temporary fix. I tried to make a change in the source itself, but I don't know enough OCaml. :(

winkler1:
Voting for @nmn's suggestion. I'm on an es6 codebase, want to get Flow going. Choices are to change lets/const -> var (and incur the political capital of selling that to team), or not using Flow. Would much rather use Flow sooner than later. An interim flag to treat let/const as "var" would be ideal.

samwgoldman:
@nmn @winkler1 Treating `let` and `const` as `var` isn't a great option because it isn't safe. It's easy to create an example that has runtime errors if you treat all variable declarations as `var` declarations, but flow would happily accept:

```javascript
function bad1() {
  if (true) {
    let foo = "local var";
  }
  return foo; // out of scope, but flow doesn't know
}
```

So, unfortunately, you'll have to sit tight a bit longer while we get this right. Sorry for the delay!

pluma:
Flow might not catch it but eslint (with babel-eslint) sure does. Flow is primarily a type checker, so I'd think most people interested in reporting like that also use a linter, if only as a backup.

+1 for doing the wrong thing before doing the right thing.

nmn:
Doing the wrong thing in the mean time was only a suggestion. I'm not sure how much sense it makes from a time commitment point of view.

But while I understand the fact that it would not catch errors, it's still good to be able to at least flow. As of now, it's more of a all or nothing approach. The Let/Const pull request has been around for a while, and I've been watching it patiently for a while. If it's close to being merged than this whole discussion is pointless anyway!

samwgoldman:
If you are truly intent on treating let/const as var, it's not hard to make that happen in a fork. Just change the `Const` and `Let` branches in [the `variable_declaration` function](https://github.com/facebook/flow/blob/f3695ca93da0686b79b4d93e6ac01f18998c32f9/src/typing/type_inference_js.ml#L992) to be the same as the `Var` branch. I don't think such a change will be merged upstream, but it will be very simple to keep that change up-to-date with master.

Another possibility: You might be able to run babel with a limited number of transforms using the "whitelist" option. If you just whitelisted the "es6.blockScoping" transform, flow could process the post-transformed source, and you wouldn't run into the pitfalls of using flow on fully transformed source files.

Lastly, this feature branch is actually feature complete—but it's pending some internal refactoring. I haven't bothered to keep this up-to-date with master, so it's a bit behind on other features. If you want to go ahead and use this branch, please do let me know if you run into any issues.

nmn:
@samwgoldman thanks for pointing me to the right place. I am finally able to use flow on my code!

bsr203:
+1 Thank for working on supporting let/const and hope it gets merged soon

fredericksilva:
:+1:

nmn:
@bsr203 @fredericksilva If you are using babel to transpile your code, you may be happy enough with my fork of flow: github.com/nmn/flow
I've just made a couple of small edits that makes flow treat let/const the same way as var. It won't catch scoping mistakes, and may still complain about scoping incorrectly, but in the predominant case it will be just fine. Plus you can actually start using flow now if let/const is a blocker for you.

bsr203:
thanks @nmn your fork let me use `flow` now !

bhosmer:
Quick interim update - unf. delay will last a little longer, it's due to some refactoring I need to land in front of Sam's awesome work. I'm back on the grid end of this week, will update status here.

fredericksilva:
Thanks @nmn,  let's see how your work flow. :facepunch:

midinastasurazz:
@bhosmer Any updates on this?

matklad:
Is const + destructing in scope of this PR? At the moment, code like `const [x] = [92]` fails with "Assignment to constant variable `x`"

samwgoldman:
@matklad Thanks for the report! Looks like I mishandled destructuring, so I'm throwing an unnecessary error there. I will add a test case for that once the refactoring lands and I pick this up again.

mroch:
i'm not keeping up here, so apologies if this is already handled, but this is seriously wtf so hopefully we can detect it properly: http://es-discourse.com/t/why-typeof-is-no-longer-safe/15

samwgoldman:
@mroch We do handle [a number of TDZ errors](https://github.com/samwgoldman/flow/blob/9e6c28d93f0af22e92729d62441e2e06716db5c0/tests/let_const/common.js#L35), although it's worth adding an explicit test case for `typeof`, in case that code goes through a different code path. [Here's the code that checks](https://github.com/samwgoldman/flow/blob/9e6c28d93f0af22e92729d62441e2e06716db5c0/src/typing/env_js.ml#L269), although it probably (hopefully) won't survive the upcoming refactor.

samwgoldman:
All right. Looks like the refactoring landed (@bhosmer, let me know if there's more in the pipeline), so now I'm the one blocking this.

bhosmer:
@samwgoldman nothing imminent - the one nontrivial remaining loose end is far less important than your stuff and can wait. Definitely happy to walk through the new stuff to whatever extent you want BTW.

meagar:
+1

samwgoldman:
superseded by #784

PR #441, opened 2015-05-08
Promises should have optional parameters

runn1ng:


runn1ng:
(signed the CLA now)

runn1ng:
I am now not sure if it's correct.

The parameter should be optional only when R is void. However, I am not sure how to write that

samwgoldman:
This feels wrong to me. It would be sad to lose the type checking that promise() is invoked with a value. Making this optional means that Promise<T> becomes Promise<T?> always.

runn1ng:
@samwgoldman : yeah, but the current state doesn't allow you to resolve an empty promise

So you cannot define a promise that's not returning anything, unless you write `resolve(undefined)`. 

And you can have a promise that's not returning anything (Promise<void>), and it makes sense to have it.

And, by the way, in the current code, you *already* can do `Promise.resolve()` (line 357).

I am not sure how to do this ellegantly.

runn1ng:
In other words, because of non-option, I had to change several

     resolve()

in my codebase to

    resolve(undefined)

even when I return `Promise<void>`.

runn1ng:
(but it broke unit tests, haha.)

samwgoldman:
Well, for what it's worth, [TypeScript's definition for Promises](https://github.com/Microsoft/TypeScript/blob/cc078653be56d38d84ffd4010e2d208eb3596ba0/bin/lib.es6.d.ts#L4807) does indicate that the argument to `resolve` is optional.

samwgoldman:
I think this PR might be unnecessary after some recent changes to how `void` parameters are handled.

Here's an example:

```javascript
/* @flow */

function test(x: void) {
  return x;
}

// Even though test is declared to accept a parameter, this is OK
test()
```

`Promise<void>`'s `resolve` callback should have the same behavior.

runn1ng:
That's great. I will try it on master branch and close this later.

samwgoldman:
@runn1ng I just tried this out and I was able to run the following code without issue. Please reopen if you think there is still work to be done here:

```javascript
function f(b) {
  return new Promise((resolve, reject) => {
    if (b) {
      resolve()
    } else {
      reject()
    }
  });
}
```

runn1ng:
yep, works, great! Your work on this project is awesome, it is more mature every day!

Issue #444, opened 2015-05-09
Support ES5 getters/setters


runn1ng:
Simple example:

    class C {
      x: string;
      constructor(x: string) { this.x = x; }
      get fancyx():string {
        return "Fancy " + this.x;
      }
    }
    
    var o = new C("text");
    console.log(o.fancyx);

Expected: everything typechecks as correct

What I got:

    property fancyx
    Property not found in C

samwgoldman:
[Indeed.](https://github.com/facebook/flow/blob/6e7371a3e6e6e57d3762f2a77735d89f6623552c/src/typing/type_inference_js.ml#L2336) To me, it seems like this could potentially be tagged easyfix in case someone from the community wants to take a stab.

samwgoldman:
Actually, it looks like the code that should handle get/set for classes is [here](https://github.com/facebook/flow/blob/6e7371a3e6e6e57d3762f2a77735d89f6623552c/src/typing/type_inference_js.ml#L4412), and we should probably be spitting out warnings like we do with `object_` in the code linked above.

marijnh:
Note that getters and setters are in ES5, not 6, and already widely used.

avikchaudhuri:
We could support them, but they would not play nicely with object property refinements. Currently when o.x is tested != null, under some fairly restrictive but common cases this information is preserved so that you can pass o.x to places that expect non-null values. But with getters, o.x could have arbitrary side effects, so it would be unsound to do the refinements. Probably fine to be unsound here given that in practice o.x should not write the x property, or we can try to somehow enforce that getters don't set, but the general problem of tracking side effects is hard and is on the longer term wishlist.

marijnh:
Non-sound support would definitely preferable to no support at all. (And yes, in practice I haven't seen many getters with problematic side effects. They might populate a cache, but that will only affect the typing of the code that calls the getter in rare situations.)

samwgoldman:
Avik makes a really excellent point about refinements, but I agree that an unsound implementation (caveat implementor) is desirable vs. blocking support behind more refinement improvements and purity analysis.

Any objections to moving forward with an unsound-with-respect-to-refinements implementation of get/set properties?

Chris-Andrews:
I'd like to chime in to support the unsound-with-respect-to-refinements implementation.

gabelevi:
[...]

PR #461, opened 2015-05-21
Warn on unsupported class members

samwgoldman:
Before this change, flow would silently ignore these properties and
instead users would see "Property not found" errors when they went to
use the desired properties.

I intentionally didn't include a fallthrough case, because the compiler
will warn if this pattern ever becomes inexhaustive in the future.
However, that's currently just a warning. Can we make it an error?

Also, because of the way the AST is defined, some of these cases are
actually syntax errors, like computed- and literal-key method bodies. I
would be happy to also make the changes to the AST to restrict the
representation to exclude impossible cases, if that seems important.

See #444

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -4422,8 +4422,38 @@ and mk_signature cx reason_c c_type_params_map body = Ast.Statement.Class.(
           SMap.add name t fields,
           methods
 
-    (* skip unexpected members *)
-    | _ -> sfields, smethods, fields, methods
+    (* get/set *)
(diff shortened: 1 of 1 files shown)


mroch:
> I intentionally didn't include a fallthrough case, because the compiler
will warn if this pattern ever becomes inexhaustive in the future.
However, that's currently just a warning. Can we make it an error?

fd23321455b24a75f3ee00a9e50e246882dc295c :)

samwgoldman:
@mroch is on :fire:!

gabelevi:
Awesome, thanks @samwgoldman!

PR #470, opened 2015-05-27
Frozen object unification exception

samwgoldman:
Because flow tries to unify these two types, users see errors going both ways. Unification isn't necessary here, because we know the right-hand side is literal.

Adding the known reason for a frozen object literal takes us down to a more reasonable, singular message.

I can keep adding to this list all day. What's the real fix?


samwgoldman:
The test failures here are unrelated to the changes I made. I suspect the changes @mroch made will fix the build here, too, but only repo owners can re-run the build on Travis.

gabelevi:
In the future, we probably will want to make

```JavaScript
var o: { x: number } = Object.freeze({ x: 123 });
```

an error. I suppose frozen ObjT should not flow to a non-frozen ObjT. We'd also probably need someway to express a frozen object type. But yeah, that was already broken and can be fixed in the future.

jeffmo:
@gabelevi: If we made that an error, how would you annotate the type of that object? Maybe we need a `$Frozen<{x: number>` I guess?

PR #480, opened 2015-05-30
Clarify assignment_lhs

samwgoldman:
Maybe this is obvious to others, but I was confused that it seemed flow
would allow arbitrary expressions as the LHS of an assignment.

I was confused by this code when I went to add a check for mutating
operations to const-declared variables. Eventually I realized that the
parser prevents arbitrary expressions from reaching this code path, but
this code would have saved me some time.

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -3279,8 +3279,11 @@ and assignment_lhs cx = Ast.Pattern.(function
   | loc, Identifier i ->
       expression cx (loc, Ast.Expression.Identifier i)
 
-  | loc, Expression e ->
-      expression cx e
+  | _, Expression ((_, Ast.Expression.Member _) as m) ->
+      expression cx m
+
+  (* parser will error before we get here *)
+  | _ -> assert false
 )
 
 and assignment cx loc = Ast.Expression.(function


mroch:
can you point to where in the parser it catches this?

mroch:
is_assignable_lhs?

samwgoldman:
Yeah, that's what I found.

mroch:
i think you're right because this is `assignment_lhs`. I wonder why `Pattern.Identifier` is a thing instead of using `Pattern.Expression` with an `Expression.Identifier` inside it, like we do for `Member`.

anyway, also found this which allows it: https://github.com/facebook/flow/blob/master/src/parser/parser_flow.ml#L3273

but that's only a problem if assignment_lhs isn't always guarded by is_assignable_lhs

samwgoldman:
Weird. I'm having a hard time imagining `Pattern.pattern` would be valid for an arbitrary expr. I'm not an encyclopedia of JS syntax (far from it), but it seems like this is an AST representation issue, like you pointed out. If so, the better fix is to make the AST more precise, and I'd be happy to take a stab at it.

gabelevi:
> I wonder why `Pattern.Identifier` is a thing instead of using `Pattern.Expression` with an `Expression.Identifier` inside it, like we do for `Member`.

It comes from the [SpiderMonkey spec](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Parser_API). Long story short, the spec says that 4 things are `Pattern`'s:

1. `interface Identifier <: Node, Expression, Pattern`
2. `interface ObjectPattern <: Pattern`
3. `interface ArrayPattern <: Pattern`
4. `interface Expression <: Node, Pattern`

So yeah, the spec is somewhat redundant. I think I kept it in the AST because it's a little less verbose to match on `Pattern.Identifier id` than to match on `Pattern.Expression (Expression.Identifier id)`

When I wrote the parser, I was trying to build an OCaml representation of the SpiderMonkey AST (now what they're calling estree). It is not a perfect AST and there are a lot of valid ASTs according to the spec that are clearly invalid programs according to the ES6 spec. So yeah, you end up with some dumb things like saying that any expression can be a pattern, when clearly some expressions are not valid patterns.

As for this PR I'm totally fine adding comments and asserts to make these assumptions clear and to assert that things have already been caught by the parser.

samwgoldman:
Thanks for clarifying, Gabe. I think it is wise to keep the AST aligned with SpiderMonkey as much as we can, so I don't see a need to modify the AST in flow. As far as the changes here, I'm only luke warm about them. If there is any reason not to pull this in, feel free to close.

PR #485, opened 2015-06-01
Export default class

samwgoldman:
This change implements the class expression case for the type inferencer, which is very similar to the existing code to handle class declarations. I refactored the AST to enable more sharing of this code. See individual commits for details.

Notably, I didn't touch the `dts` subdirectory, because I am not sure what it is for.

### DIFF
--- a/src/parser/spider_monkey_ast.ml
+++ b/src/parser/spider_monkey_ast.ml
@@ -396,49 +396,6 @@ and Statement : sig
       body: Statement.t;
     }
   end
-  module Class : sig
-    module Method : sig
-      type t = Loc.t * t'
-      and t' = {
-        kind: Expression.Object.Property.kind;
-        key: Expression.Object.Property.key;
-        value: Loc.t * Expression.Function.t;
-        static: bool;
-      }
-    end
-    module Property : sig
-      type t = Loc.t * t'
-      and t' = {
-        key: Expression.Object.Property.key;
-        typeAnnotation: Type.annotation;
-        static: bool;
-      }
-    end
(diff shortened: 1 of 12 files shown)


src/parser/spider_monkey_ast.ml
```diff
@@ -850,7 +797,7 @@ and Expression : sig
     | TemplateLiteral of TemplateLiteral.t
     | TaggedTemplate of TaggedTemplate.t
     | JSXElement of JSX.element
-    | Class of Expression.Class.t
+    | Class of Class.t
```
jeffmo: woohoo! thanks for cleaning this up

src/typing/type_inference_js.ml
```diff
@@ -3073,12 +3067,20 @@ and expression_ cx loc e = Ast.Expression.(match e with
   | JSXElement e ->
       jsx cx e
 
+  | Class ({ Ast.Class.id } as c) ->
+      let (nloc, name) = (
```
jeffmo: Maybe pull this out into a `get_class_name` func (to consolidate with logic up on line 1853)

jeffmo:
This is awesome! Sorry it took me a while to review. A few minor nits, but if you don't have time to get to them today I'll just merge it in and do tweak them myself

samwgoldman:
Thanks for the review, Jeff. I agree re: extracting `get_class_name`. I'm pretty tied up all week, so if you can take it from here, I'd appreciate it.

jeffmo:
I'm on it! Thanks again for this

PR #486, opened 2015-06-02
Add methods and variables exported by the zlib module built in to node.

smockle:
Test file:
```javascript
#!/usr/bin/env node
/* @flow */
var zlib = require('zlib');
console.log(zlib.createGunzip());
```

`flow` without this PR: `index.js:4:13,31: call of method createGunzip Could not resolve name Found 1 error`

`flow` with this PR: `No errors!`

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -753,6 +753,28 @@ declare class stream$Writable extends stream$Stream {
   ): boolean;
 }
 
+declare class stream$Duplex extends stream$Stream {
+  setEncoding(): void;
+  isPaused(): boolean;
+  pause(): void;
+  read(size?: number): ?(string | Buffer);
+  resume(): void;
+  unpipe(dest?: any): void;
+  unshift(chunk: any): void;
+  wrap(oldReadable: any): stream$Readable;
+  end(
+    chunk?: string | Buffer,
+    encodingOrCallback?: string | (data: any) => void,
+    callback?: (data: any) => void
+  ): void;
+  setDefaultEncoding(encoding: string): boolean;
+  write(
+    chunk?: string | Buffer,
+    encodingOrCallback?: string | (data: any) => void,
+    callback?: (data: any) => void
+  ): boolean;
+}
+
 declare module "stream" {
   declare class Stream extends stream$Stream {}
 }
@@ -887,7 +909,7 @@ declare module "zlib" {
     Z_BUF_ERROR: number,
     Z_VERSION_ERROR: number
   };
-  declare class Zlib {
+  declare class Zlib extends stream$Duplex {
     // TODO
   }
   declare class Deflate extends Zlib {}
--- a/lib/node.js
+++ b/lib/node.js
@@ -844,15 +844,26 @@ declare module "vm" {
   // TODO
 }
 
-type zlib$ZlibOptions = {
-  flush: ?number;
-  chunkSize: ?number;
-  windowBits: ?number;
-  level: ?number;
-  memLevel: ?number;
-  strategy: ?number;
-  dictionary: ?Buffer;
+type zlib$options = {
+  flush?: number;
+  chunkSize?: number;
+  windowBits?: number;
+  level?: number;
+  memLevel?: number;
+  strategy?: number;
+  dictionary?: Buffer;
 };
+
+type zlib$asyncFn = (
+  buffer: string | Buffer,
+  options?: zlib$options
+) => void;
+
+type zlib$syncFn = (
+  buffer: string | Buffer,
+  options?: zlib$options,
+  callback?: ((error: ?Error, result: any) => void)
+) => void;
 
 declare module "zlib" {
   declare var Z_NO_FLUSH: number;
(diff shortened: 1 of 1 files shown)


samwgoldman:
From what I can glean in the docs. The `zlib` module doesn't actually export the `Zlib` class. You should be able to define the class non-exported and still inherit from the exported classes.

I think it would be more useful, especially combined with auto-suggest, to have the option object types spelled out. It would also be nice to have the callback function types be more specific.

samwgoldman:
I also think the full `options` parameter name is nicer than `o` or `opts`.

smockle:
@samwgoldman I couldn’t tell whether `Zlib` was exported. There is a `binding.Zlib` function [(here)](https://github.com/nodejs/node/blob/10e31ba56c676bdcad39ccad22ea9117733b8eb5/lib/zlib.js#L343) and each of `binding`’s keys are added to `exports` [(here)](https://github.com/nodejs/node/blob/10e31ba56c676bdcad39ccad22ea9117733b8eb5/lib/zlib.js#L29-L38). Should I remove [these lines](https://github.com/smockle/flow/blob/8c24f29c5d3b2b734cb2d3543ac19ad05dc143ba/lib/node.js#L872-L874)?

I prefer `options` too, but I thought I needed to match the parameter name used in node/zlib.js—the parameter names in declare statements are arbitrary?

smockle:
Re: spelling out the option object types, I didn't know all available options, so I used `any`. Closer inspection of [node/zlib.js](https://github.com/nodejs/node/blob/10e31ba56c676bdcad39ccad22ea9117733b8eb5/lib/zlib.js#L287-L341) shows `flush`, `chunkSize`, `windowBits`, `level`, `memLevel`, `strategy` and `dictionary`. Did I miss any?

samwgoldman:
I was just going off the published documentation, which states that the [zlib class is not exported](https://nodejs.org/api/zlib.html#zlib_class_zlib_zlib).

And while `Zlib` is exported, it doesn't seem to be usable:

```
> var zlib = require("zlib")
undefined
> zlib.Zlib
[Function: Zlib]
> new zlib.Zlib()
TypeError: Bad argument
```

I am able to construct the subclasses, so I'm pretty sure we should exclude `Zlib` here. I don't think you should delete those lines, but you should be able to make a class declaration that is not exported (no declare keyword). I haven't tried that myself, so I could be wrong—let me know.

Regarding options, the documentation [specifies the general shape](https://nodejs.org/api/zlib.html#zlib_options), but also notes that some options are ignored by some modes. I'd be fine with using a single object type for them all, but it might be nice to sort this out as well.

Regarding parameter names, for positional parameters, there is no need to match the source code. Instead of going with the code, I would go with the language used in the docs, which is `options`.

samwgoldman:
I also noticed the constants are all numeric. Specifying `number` instead of `any` normally wouldn't offer any benefit, but error constants are always negative, so adding a more specific annotation there could also be useful.

smockle:
@samwgoldman Sounds good, I’ll fix each of those issues. Thanks again for the review! Should I squash the commits when I’m done making the changes?

samwgoldman:
I won't be approving or merging what you do here—I'm an enthusiastic
contributor like yourself! I wouldn't worry about squashing, personally.

smockle:
The `Zlib` class should extend `stream$Duplex`, which needs to be declared. I’ll commit this later

lib/node.js
```diff
@@ -822,8 +822,102 @@ declare module "vm" {
...
 
+type zlib$ZlibOptions = {
+  flush: ?number;
+  chunkSize: ?number;
+  windowBits: ?number;
+  level: ?number;
+  memLevel: ?number;
+  strategy: ?number;
+  dictionary: ?Buffer;
+};
```
samwgoldman: I think it's more accurate to make the keys optional than the values maybe types. I.e., `{ foo?: T }` instead of `{ foo: ?T }`.

lib/node.js
```diff
@@ -822,8 +822,102 @@ declare module "vm" {
...
+  declare function deflateRaw(buffer: string | Buffer, options: zlib$ZlibOptions, callback?: ((error: ?Error, ...args: Array<any>) => void)): void;
+  declare function deflateRawSync(buffer: string | Buffer, options: zlib$ZlibOptions): void;
+  declare function unzip(buffer: string | Buffer, options: zlib$ZlibOptions, callback?: ((error: ?Error, ...args: Array<any>) => void)): void;
+  declare function unzipSync(buffer: string | Buffer, options: zlib$ZlibOptions): void;
+  declare function inflate(buffer: string | Buffer, options: zlib$ZlibOptions, callback?: ((error: ?Error, ...args: Array<any>) => void)): void;
+  declare function inflateSync(buffer: string | Buffer, options: zlib$ZlibOptions): void;
+  declare function gunzip(buffer: string | Buffer, options: zlib$ZlibOptions, callback?: ((error: ?Error, ...args: Array<any>) => void)): void;
+  declare function gunzipSync(buffer: string | Buffer, options: zlib$ZlibOptions): void;
+  declare function inflateRaw(buffer: string | Buffer, options: zlib$ZlibOptions, callback?: ((error: ?Error, ...args: Array<any>) => void)): void;
+  declare function inflateRawSync(buffer: string | Buffer, options: zlib$ZlibOptions): void;
```
samwgoldman: You could avoid the duplication in the above by defining `zlib$asyncFn` and `zlib$syncFn` types and declare these as vars with those function type—which would be equivalent to `declare function`, iiuc.

The option argument to these functions is optional in the docs, but not market optional here.

I think you can be even more specific with the callback function types, too. `(error: ?Error, result: any) => void` should do it.

mroch:
nice, thanks for this @smockle and thanks for reviewing, @samwgoldman.

> Should I squash the commits when I’m done making the changes?

nope, no need. instead of directly merging pull requests, we squash the PR and commit it internally, and then sync it back to github with your name on it.

smockle:
@samwgoldman I think I addressed your line comments in my latest commit. Please let me know if you see anything else!

lib/node.js
```diff
@@ -822,8 +844,113 @@ declare module "vm" {
...
+  buffer: string | Buffer,
+  options?: zlib$options
+) => void;
+
+type zlib$syncFn = (
+  buffer: string | Buffer,
+  options?: zlib$options,
+  callback?: ((error: ?Error, result: any) => void)
+) => void;
+
```
samwgoldman: I think you got the async/sync reversed. And shouldn't sync return something?

smockle: Oops, that’s embarrassing. Thanks for catching that! :)

mroch:
this looks good to me. anything else, @samwgoldman?

samwgoldman:
I'm a bit unsure about the `stream$Duplex` definition, but I think we can sort out any issues there if/when we run into them. :+1: Nice work, Clay.

smockle:
Thanks for the review!

@samwgoldman I’d love to improve stream$Duplex—either before or after this is merged—if you have any suggestions. The Zlib docs say “Each class…is a readable/writable Stream.” [[link]](https://nodejs.org/api/zlib.html#zlib_zlib) The Stream docs say “A 'duplex' stream is one that is both Readable and Writable.” [[link]](https://nodejs.org/api/stream.html#stream_class_stream_duplex_1) The Stream docs go on to explain how to properly inherit both stream$Readable and stream$Writable, but I wasn't sure how to best express that with Flow’s class and type syntax.

samwgoldman:
In addition to the docs and the source, I think it's good to refer to TypeScript's node.d.ts definitions. They won't map 1-1, but I imagine those definitions have been used in conjunction with those APIs in real code. I still think we can move forward with these changes as-is.

smockle:
@samwgoldman Great idea! Here are the TypeScript definitions for Duplex streams: https://github.com/Microsoft/TypeScript/blob/6d7045ec5abfa3dff4419d9d166b546ef73ce8e2/src/server/node.d.ts#L347-L360. A quick search of the TypeScript repo in GitHub didn’t turn up any Zlib-related definitions—do you know where those are stored?

samwgoldman:
Looks like DefinitelyTyped has some more fully baked definitions:
https://github.com/borisyankov/DefinitelyTyped/blob/master/node/node.d.ts#L521

PR #489, opened 2015-06-02
Stop claims of file re-checks by server for ignored files.

garylocke:
While running the server with `flow server`, it would report re-checks for files that match expressions in the [ignore] section of .flowconfig. This change should prevent that from occurring.


gabelevi:
Thanks for taking the time to clean this up!

PR #494, opened 2015-06-04
Type check React.Components when the type of defaultProps is unspecified

samwgoldman:
Using ObjTestT like this is a bit imprecise, but in the case of react
props, if the defaultProps isn't an object type, we have other problems.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -58,7 +58,7 @@ declare class ReactComponent<D, P, S> {
  * existential type (*). The * can be thought of as an "auto" instruction to the
  * typechecker, telling it to fill in the type from context.
  */
-type ReactClass<D, P, S> = _ReactClass<D, P, S, *>;
+type ReactClass<D, P, S> = _ReactClass<$ObjTest<{},D>, P, S, *>;
 type _ReactClass<D, P, S, C: ReactComponent<D, P, S>> = Class<C>;
 
 /**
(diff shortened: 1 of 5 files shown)


lib/react.js
```diff
@@ -58,7 +58,7 @@ declare class ReactComponent<D, P, S> {
  * existential type (*). The * can be thought of as an "auto" instruction to the
  * typechecker, telling it to fill in the type from context.
  */
-type ReactClass<D, P, S> = _ReactClass<D, P, S, *>;
+type ReactClass<D, P, S> = _ReactClass<$ObjTest<{},D>, P, S, *>;
```
avikchaudhuri: Clever idea. But I'm not convinced this is the best way to solve this problem. See below.

src/typing/flow_js.ml
```diff
@@ -1616,6 +1616,11 @@ let rec __flow cx (l, u) trace =
       then rec_flow cx trace (l, u)
       else rec_flow cx trace (default, u)
 
+    | (ObjTestT(reason, default, u), _) ->
```
avikchaudhuri: What is the intended behavior here? I bet object_like u is failing because u is type variable, which it will almost always be, even when defaultProps is provided (at least until the concrete object type provided as defaultProps gets unified with the type variable). Note also that l is the entire ObjTestT type here, is that what you intended in the calls below?

samwgoldman: To be honest, I didn't have a plan here. I wrote a too small test case, saw "type is incompatible" and (poorly—typos abound) reversed the above rule without considering what the upper and lower bounds would actually be.

Is there any way to "defer" this kind of check until the upper bound is concrete?

samwgoldman:
Doesn't work. I should add a test that exercises the expected defaultProps behavior, which wouldn't have passed with this code.

PR #503, opened 2015-06-10
Correcting removeAllListeners

runn1ng:
removeAllListeners gets one event, not array of events

See

https://github.com/joyent/node/blob/ef4344311e19a4f73c031508252b21712b22fe8a/lib/events.js#L258-L297


gabelevi:
Great, thanks!

PR #519, opened 2015-06-14
reduceRight's initialValue should be optional

acdlite:
Small change to fix what I believe is a bug: the second parameter, `initialValue`, of `Array.prototype.reduceRight` should be optional, just like `reduce`.


gabelevi:
Cool, thanks!

PR #527, opened 2015-06-17
Revamp "Base Types" docs page

jeffmo:
* Switch to a list + summary format with links to more detailed descriptions for each type below.
* Adds some missing built-ins
* Adds examples for each type


jeffmo:
Since I don't know of a good way to preview this, here's a screen-shot/pdf-print of the page running locally:

flow _ named types-1

jeffmo:
I should mention and reference core.js for "all the other Lang built-ins" somewhere too, I guess

samwgoldman:
Nice! Seems like the caveats entry for mixed can be removed, since it has it's own section now.

jeffmo:
Yea, I started with that but decided to leave it in caveats since it seemed like a good "overall" caveat to be aware of (i.e. the first time you scan this doc as learning material -- before you're using it as a reference).

Actually, I suppose it wouldn't hurt to also put it in the details section for `mixed` too.

PR #531, opened 2015-06-17
Updating EventEmitter.on()

runn1ng:
on() certainly doesn't get only one argument, and doesn't need to return void

( it could be probably written as `(...args: any) => any`, but that's probably just `Function` anyway )

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -429,7 +429,7 @@ declare class events$EventEmitter {
   addListener(event: string, listener: Function): events$EventEmitter;
   emit(event: string, ...args:Array<any>): boolean;
   listeners(event: string): Array<Function>;
-  on(event: string, listener: (data: any) => void): events$EventEmitter;
+  on(event: string, listener: Function): events$EventEmitter;
   once(event: string, listener: Function): events$EventEmitter;
   removeAllListeners(event?: string): events$EventEmitter;
   removeListener(event: string, listener: Function): events$EventEmitter;


runn1ng:
Also, once() and on() should have the same declaration, I guess :)

gabelevi:
The rules for callbacks are a little different. If a function `foo` takes a callback and always calls it with a single number and expects a string, the correct declaration is

```JavaScript
function foo(callback: (x: number) => string) { ... }
```

You can still do `foo(() => 'hi')`, since it's fine to pass a function more arguments than it is expected. The callback's declaration should express how `foo` will use that callback.

I'm having trouble finding a good spec for `on()`, but it did seem like it's callback will be passed some argument: https://nodejs.org/api/events.html#events_emitter_on_event_listener.

Whether the callback should return void is a good question. It is a common coding mistake to return a value and think it means something when the returned value is being ignored. However it's also a common mistake to pass more arguments to a function than it takes, and JavaScript is cool with that. So maybe in the spirit of easily using array functions we should ignore the return types. This has come up before for [List.prototype.forEach()](https://github.com/facebook/flow/blob/90eb4245c95a1a7c7a59c92cb799e21839cba911/lib/core.js#L334)

mroch:
passing a callback that expects fewer arguments is fine, but passing one that expects more arguments is not. 

the callback will receive whatever arguments are passed to `emit`:

```
emitter.on('foo', (a, b, c, d) => console.log(a + b + c + d));
emitter.emit('foo', 1, 2, 3, 4);
```

I don't think we can accurately type that so `(...args: any)` (which doesn't work right now) or `Function` seem more correct than expecting a single param.

runn1ng:
As @mroch said. Emitter listener takes whatever is passed to the emit function as arguments.

It would be nice to somehow typecheck that - somehow declare what gets passed to what emitter - but since the event itself can be an arbitrary string, I don't think it's possible (unless emitters somehow become more "first class").

runn1ng:
The return value should be void though, that's a good point

Issue #341, opened 2015-03-26
Confusing error message with mistyped function arguments on intersection types


samwgoldman:
In the following code, `c.a()` is called with the wrong argument (nothing instead of string).

``` javascript
/* @flow */

type A = { a(x: string): string; }
type B = { b(): void; }
type C = A & B

function test(c: C) {
  return c.a()
}
```

I get this error:

```
test2.js|8 col 10 error|  call of method a
|| Method cannot be called on
test2.js|5 col 10 error|  intersection type
|| 
```

I expect this error:

```
test2.js|8 col 10 error|  call of method a
|| Too few arguments (expected default/rest parameters in function)
test2.js|3 col 12 error|  function type
|| 
```

threehams:
Are there any alternatives to intersection types when an object contains all its own properties, plus additional ones? I've been avoiding using intersection types entirely because of the error messaging in this issue, to the point of type duplication.

samwgoldman:
Intersection errors have all the necessary information now. Calling this resolved.

```
Error: test.js:8
  8:   return c.a()
              ^^^^^ call of method `a`. Method cannot be called on any member of intersection type
  8:   return c.a()
              ^ intersection
  Member 1:
    5: type C = A & B
                ^ A
  Error:
    8:   return c.a()
                ^^^^^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
    3: type A = { a(x: string): string; }
                       ^^^^^^ string
  Member 2:
    5: type C = A & B
                    ^ B
  Error:
    8:   return c.a()
                  ^ property `a`. Property not found in
    5: type C = A & B
                    ^ object type
```

PR #534, opened 2015-06-18
ES6 Generators

### LINKED ISSUE #474: ES6 Generators
I'd like to reach a consensus on how we should implement type checking for generators. Parsing is already implemented, which makes this a good candidate for a next ES6 feature to tackle. I am particularly excited for this work as a prelude to async/await eventually.

The behavior from the spec force us to balance between ergonomics in types vs ergonomics in code, particularly with regard to completion values. For example:

```javascript
function *gen() {
  yield 1;
}
var res = gen();
res.next(); // { value: 1, done: false }
res.next(); // { value: undefined, done: true }
```

Should we infer `Iterator<?number>` as the return value of `gen()`? The existing `lib` definitions have a nice definition:

```javascript
interface IteratorResult<T> {
    done: boolean;
    value?: T;
}

interface Iterator<T> {
    next(): IteratorResult<T>;
}
```

However, what do we do about return values?

```javascript
function *gen() {
  yield 1;
  return "end";
}
var res = gen();
res.next(); // { value: 1, done: false }
res.next(); // { value: "end", done: true }
res.next(); // { value: undefined, done: true }
```

What is the return type of gen()? `Iterator<number|string>`? TypeScript has decided that generator functions can't have return expressions. That seems like a reasonable constraint to me, but I can also live with the union type.

Beyond the manual iterator interface through `next()`, generators produce a value that is also an Iterable.

```javascript
var [...all] = gen(); // all is [1]
```

`for-of` loops have similar semantics, where only the yielded values are ever bound to the loop variable.

```javascript
 // prints "1"
for (let i of gen()) {
  console.log(i);
}
```

Whatever solution we come to, let's make sure that for cases that only look at yielded values, Flow can infer the most precise type possible. In the above cases, `all` and `i` should be `number[]` and `number`, respectively.

If we do decide to allow return expressions in generator functions, the Iterable interpretation of the return value should only consider the yielded values:

```javascript
function *gen() {
  yield 1;
  return "end";
}
```

The above function could be `Iterator<number|string>`, but should be `Iterable<number>` only.

If a generator without an explicit type signature has consecutive `yield` statements with conflicting types, I assume we should keep widening the types to fit all yielded values.

```javascript
// Should we infer `Iterator<number|string>`?
function *gen() {
  yield 1;
  yield "a";
}
```

What about recursive/infinite generators? Is it actually possible to type them?

```javascript
// Should be `Iterator<number>`, but can we figure it out?
function *gen() {
  yield 1;
  yield *gen()
}
```

Now that I've mentioned `yield*` I should note that there is an interaction with return values here:

```javascript
// `yield*` returns the generator return value; can we make this type check?
function *foo() {
  yield 1;
  return "foo";
}

function *bar() {
  let foo: string = yield *foo();
  return foo;
}
```

I haven't even touched exceptions, but this is getting pretty long. So far this is just a brain dump of what I'm currently thinking. Ideally, I'd like this discussion to generate test cases that we can use to drive out the level of support we'd like to have.

Specific test cases that I'd like to see:
- [x] `yield` return values (sending values via `next`, including)
- [x] sending values via `next` "through" another generator that uses `yield*`
- [x] `yield*` return values
- [x] `yield*` with plain old iterables, too
- [x] inferring generator types containing `yield*` expressions
- [x] type-safe use of generators as iterables
- [ ] yield only inside generators (not nested callbacks)
- [ ] abnormal exit

---

### Closing note

Readers, please refrain from otherwise contentless "+1" comments, which would make the discussion I'd like to have here difficult to follow. Your enthusiasm is appreciated, and you can show it by subscribing to this topic! Let's keep comments substantive here. Thanks!

samwgoldman:
### Behavior

Generators are iterable. Each value yielded from the generator function
is iterated.

Generators are iterators. Values can be passed in through the `next`
method. The iterator result values will be either a) the yielded values,
b) the value returned from the generator (the completion value) or c)
`undefined`, if `next` is called after completion.

Generators can delegate to other iterables, including other generators.
Values passed to `next` on the delegator generator will be passed
through to the delegatee. Similarly, values yielded by the delegatee
will be returned along the iterable/iterator interfaces.

The "delegate yield" expression, `yield *`, returns the completion value
of the delegatee.

### Implementation

While normal functions' return type is the type of the `return`
statement, generators are different. The return type is always a
Generator, where the second type argument is the type of the `return`
statement.

We use the same process to infer this type--an internal tvar named
"return"--but then override the return type when creating the function
type, if the function is a generator.

In order to track the `next` and `yield` types, we introduce new
internal tvars, which accumulate lower bounds according to their use,
and upper bounds from explicit annotation of the Generator type.

### Caveats

The generator interface is highly dynamic, and not particularly amenable
to static typing.

The type of Generator<Y,R,N> enforces that all next values and all
yielded values be described by a single type, but it's quite normal to
use values of different types. Consider the following example:

```javascript
function *userGen() {
  var name = yield("What is your name?");
  var age = yield("How old are you?");
  return { name, age }
}

var gen = userGen();
var name, age, user;
if (name = prompt(gen.next().value)) {
  if (age = prompt(gen.next(name).value)) {
    user = gen.next(parseInt(age, 10)).value;
    // user is { name: string|number, age: string|number }
    // but we "want" { name: string, age: number }
  }
}
```

To escape this pitfall, you can use `Generator<any,any,any>` or write
the necessary dynamic type tests. Ideally, Flow would enforce that you
first pass a `string` to `next`, then a `number`, but that isn't
possible in general, and in specific instances is still very hard, so
I'm leaving it for future work (if ever).

Another difficulty is the fact that the argument to `next` is optional.
This is because, in order to "start" the generator, you need to call
`next` once. Put differently, the first argument to `next` is not
returned by the first `yield` expression; the second argument to `next`
is. Consider the following example:

```javascript
function *bmiCalc() {
  var height: number = yield("What is your height, in meters?");
  var weight: number = yield("What is your weight, in kilograms?");
  return weight / (height * height);
}

// The first argument to next is always `void`. The value of this
// expression is the string "What is your height..."
bmiCalc.next();

// This call to `next` expects the value for `height`, above, but
// because the type of `next` marks its argument optional, we allow
// this. :(
bmiCalc.next();
```

In this implementation, I wanted to make things strict, so the return
type of a yield expression is always `Y|void`, and the generator needs
to do a dynamic type test in order to get at the expected `Y`.

The above caveats, and the strict interpretation of this implementation,
has the consequence that most, if not all, generator code "in the wild"
will result in type errors from Flow. We might consider sacrificing
correctness here for practical purposes, but I think it will be easier
to go from strict to less strict than the other way around.


lib/core.js
```diff
@@ -300,6 +300,15 @@ interface Iterable<T> {
     @@iterator(): Iterator<T>;
 }
 
+/* Generators */
+interface Generator<Y,R,N> {
+    @@iterator(): Iterator<Y>;
+    next(value?: N): IteratorResult<R|Y>;
```
samwgoldman: If I change the union here from `R|Y` to `Y|R`, the test outcomes change. Tests that currently exhibit expected failures start to succeed unexpectedly. It almost seems like the speculative match failure in the first branch isn't working—instead of falling through to the other case, we see an error right away.

lib/core.js
```diff
@@ -300,6 +300,15 @@ interface Iterable<T> {
...
+}
+
+// we use this signature when typing yield* expressions
+declare var $yieldDelegate: (<R,N>(g: Generator<*,R,N>, n:N) => R) & ((i: Iterable<*>, n:void) => void);
```
samwgoldman: The [test for delegate yields to iterables](https://github.com/facebook/flow/pull/534/files#diff-235c03962cd0739e93a3d5d899a2a634R96) fails, and I think this intersection type is to blame. The unexpected error states that Array<number> doesn't have a `next` method, which is a part of the Generator interface. My expectation was that flow would fail through and use the second case when that happens, instead of spitting out an error.

src/typing/type_inference_js.ml
```diff
@@ -797,6 +796,21 @@ let rec convert cx map = Ast.Type.(function
...
+        let return = Flow_js.mk_tvar cx (prefix_reason "return of " reason) in
+        let next = Flow_js.mk_tvar cx (prefix_reason "next of " reason) in
+        let gt = Flow_js.get_builtin_typeapp
+          cx
+          reason
+          "Generator"
+          [yield; return; next] in
+        Flow_js.flow cx (ret, gt);
+        gt
+      else ret in
```
samwgoldman: Above is where the "hijacking" which I mentioned in the PR description takes place.

tests/generators/generators.js
```diff
@@ -0,0 +1,107 @@
...
+for (var x of delegate_yield_generator()) {
+  (x : number) // error: string ~> number
+}
+
+function *delegate_return_generator() {
+  function *inner() {
+    return "";
+  }
+
+  var x: number = yield *inner() // TODO: error: string ~> number
```
samwgoldman: Here I expect the expression `yield *inner()` to be of type `string`, because `inner` should be inferred as `Generator<*,string,*>`, and the [handler for yield delegate expressions](https://github.com/facebook/flow/pull/534/files#diff-c600001e37d25814dbc857496fecd26dR3301) should resolve to the `string` by the [first case in the intersection type of the `$yieldDelegate` type](https://github.com/facebook/flow/pull/534/files#diff-f265d6a6b6b64d4d096fddcf7d620a76R310).

tests/generators/generators.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// only generators can make use of a value passed to next
+function *delegate_next_iterable(xs: Array<number>) {
+  yield *xs;
+}
+delegate_next_iterable([]).next(""); // TODO: error: string ~> void
```
samwgoldman: Here flow gives an unexpected error, instead of the expected error:

```
generators.js:94:3,11: yield* delegate
Error:
[LIB] core.js:306:5,40: property next
Property not found in
generators.js:94:10,11: array type
```

It seems like the [first case of the `$yieldDelegate` intersection](https://github.com/facebook/flow/pull/534/files#diff-f265d6a6b6b64d4d096fddcf7d620a76R310) is failing instead of failing over to the second case, which should pass. It's also very likely I don't understand the speculative match failure stuff.

tests/generators/generators.js
```diff
@@ -0,0 +1,107 @@
...
+
+function *delegate_yield_iterable(xs: Array<number>) {
+  yield *xs;
+}
+for (var x of delegate_yield_iterable([])) {
+  (x : string) // error: number ~> string
+}
+
+function *delegate_return_iterable(xs: Array<number>) {
+  var x: number = yield *xs // TODO: error: void ~> number
```
samwgoldman: Here, because `xs` is not a Generator, but is an Iterable, I expect the return type of `yield *xs` to be inferred as `void`, because of the second case in the [definition of `$yieldDelegate`](https://github.com/facebook/flow/pull/534/files#diff-f265d6a6b6b64d4d096fddcf7d620a76R310). Instead, flow doesn't detect any error.

lib/core.js
gabelevi: I half wonder if you're hitting the horrible Cache. Could you rebase past 7a4a6294b32d285c0b496570b911a07b00d31e44 and try again?

samwgoldman: !!!

That did it! I also get a much nicer error message for one of the unexpected failures.

tests/generators/generators.js
samwgoldman: After merging in the commit that does away with the cache, I now get this error, which seems reasonable. I haven't looked into it yet, but it's likely my intersection type is shenanigans.

```
generators.js:94:3,11: yield* delegate
Function cannot be called on
[LIB] core.js:317:30,102: intersection type
```

samwgoldman: OK, so it turns out that this error message is actually caused by the intended failure, but the message is incredibly unhelpful because I was "too clever" in how I implemented the type checking here.

For reference, the `$yieldDelegate` declaration is:

```
declare var $yieldDelegate: (<R,N>(g: Generator<*,R,N>, n:N) => R) & ((i: Iterable<*>, n:void) => void);
```

Then the code flows the `yield*` expression's argument, and the parent generator's `next` tvar as arguments, and returns the return value. I'm using this construction because I don't know any other way to handle both Generators and Iterables like this. I picked up the trick from @bhosmer and @nmote's async/await changes.

So in this case, since the `yield*` expression's argument is an Iterable, we can pretend the declaration is the simpler `(i: Iterable<*>, n:void) => void`, into which we flow `Array<number>` (from `xs`) and `string` (from the `""` argument to `next()`), respectively.

Since the argument to the parameter `n` is non-`void`, flow complains, like I wanted it to. However, the error message blames the entire intersection type, not the specific flow `string ~> void`.

I actually already have a (possibly) related issue open for this at #341.

Any idea how I can either a) rewrite the type checking to avoid the `$yieldDelegate` function declaration trick of b) improve the messaging around intersection types to provide a more helpful error message?

samwgoldman:
OK, so thanks to @gabelevi's insight, I was able to resolve one of the more confusing problems here. There are 2 outstanding issues here:

1. [ ] `yield*` expression return values are not captured correctly—can't make them fail as expected.
2. [ ] super confusing error messages around passing mistyped arguments to `next` "through" a generator in a delegate yield scenario.

gabelevi:
Really glad that the [termination strategies commit](https://github.com/facebook/flow/commit/7a4a6294b32d285c0b496570b911a07b00d31e44) helped! Go @avikchaudhuri!

We were speaking a little on irc about `declare function` and how to deal with async and generators. A little background. I believe non-`declare` async functions and generator functions should annotated like this:

```JavaScript
async function foo(): Promise<number> { return 42; }
function *bar(): Generator<number, bool, string> {
  var str = yield 42;
  return str === "hello";
}
```

The type of `foo` is `() => Promise<number>`. The `async`-ness of `foo` is an implementation detail and not included in its type. Someone using `foo` doesn't care if `foo` is an async function or just a function that returns a `Promise`. Sure, we check that an async function has a `Promise` return type, but then we can forget whether `foo` is `async`. This is similar to how we check the body of a function, but how the body itself doesn't show up in the function's type.

Same thing for generators. The type of `bar` is `() => Generator<number, bool, string>`. The generator-ness of `bar` is an implementation detail and not included in the type. A caller of `bar` doesn't care if `bar` is a generator or just a function that returns a generator.

`declare` declarations are really just declaring the interfaces/APIs/types/whatever. It's not necessary for them to declare `async`-ness or generator-ness. It's sufficient for them to just express the type of the function. So if we were to declare `foo` and `bar` we would just write

```JavaScript
declare function foo(): Promise<number>;
declare function bar(): Generator<number, bool, string>;
```

samwgoldman:
@gabelevi Thank you for clarifying. I'm glad you're thinking about these things, because now that you've put it so plainly, it's obvious. I will revert the [part of this patch](https://github.com/samwgoldman/flow/commit/cce04ece9de0ef9ecac1a55f00041996e88203da) that adds parser support for *function syntax in declaration files—the `convert` part was [nonsense](https://github.com/samwgoldman/flow/commit/6ffacce00764d13086083720bd068c33f383a4ed) anyway.

tests/generators/generators.js
gabelevi: If I understand correctly, you would like to raise an error when someone does something like

```JavaScript
function *foo() { yield *[1]; }
foo().next('bar');
```

even though this isn't a runtime error. If I'm reading the spec correctly, this will basically call 

```JavaScript
[1][Symbol.iterator]().next('bar')
```

and the next function [expects zero arguments](https://github.com/facebook/flow/blob/3f6411d58d6a666dba5fdc3bb4065cf72a44982c/lib/core.js#L303).

In Flow so far, it's not an error to call a function with more arguments than it expects, mainly because people do this in JS so bloody often. So you probably could just ignore next altogether if you're dealing with an `Iterable` and avoid the problem completely.

I think the error will never be as clear as you like if you're using an intersection type. If none of the intersected types match, it's really hard to figure out the right error message. We could explain that each type doesn't fit or we can explain that the whole intersection type doesn't fit.

I think what you want for a good error message (if you decide to ban passing things to next when you delegate to iterables) is a custom use type that behaves like an intersection, except that it makes it's decision exclusively based on `yield *`'s argument. You would flow the type of the expression to your custom use type, into which you'd pack all the info you might possibly need. When that lower bound finally resolves to something concrete, you would check if it's a `Generator` (and if so flow your next & return types around) or if it's an `Iterable` (and make sure next is void) or if it's something else (and blow up).

Now that I think about it, I don't think the error you are seeking is a good thing. I can construct an `Iterable` object who's `Iterator`'s `next()` method takes an argument. Here is an example:

```JavaScript
var MyIterable = {
  [Symbol.iterator]: function *() { return yield 4; }
}
```

I tested out the behavior in [Babel](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&playground=true&code=function%20*simpleGen(x)%20%7B%20return%20yield%20x%3B%20%7D%0A%0Avar%20myThing%20%3D%20%7B%0A%20%20%5BSymbol.iterator%5D%3A%20()%20%3D%3E%20simpleGen(1)%0A%7D%3B%0A%0Afunction%20*foo()%20%7B%0A%20%20console.log(%22delegate%20returned%22%2C%20yield%20*myThing)%3B%0A%7D%0A%0Avar%20g1%20%3D%20foo()%3B%0Aconsole.log(g1.next('firstNext'))%0A%0A%2F%2F%20Honestly%20I'm%20surprised%20this%20is%20what%20is%20returned%0Aconsole.log(g1.next('secondNext'))%0A%0A%0Afunction%20*bar()%20%7B%0A%20%20console.log(%22delegate%20returned%22%2C%20yield%20*simpleGen(2))%3B%0A%7D%0A%0Avar%20g2%20%3D%20bar()%3B%0Aconsole.log(g2.next('firstNext'))%0A%0A%2F%2F%20Again%2C%20I'm%20surprised%20this%20is%20what%20is%20returned%0Aconsole.log(g2.next('secondNext'))) and also Chrome so it is possible to pass a `next` through to an `Iterable`'s `Iterator`. This stuff is pretty complicated :P

gabelevi:
I should also mention how impressed I am with this PR. This is tricky stuff and you're doing an awesome job with it. Thanks a ton for working on this!

tests/generators/generators.js
samwgoldman: I see what you mean about `Iterator`'s `next` taking no arguments. I agree that passing too many arguments should be OK.

I need to think on this a bit. In particular, I'm curious about generators like this:

```
function *hmm(xx: Array<number>): Generator<number, bool, *> {
  var str = yield 42;
  yield *xs;
  return str === "hello";
}
```

What is the type of `*` here? Is it `string` or `string | void`? If we choose `string`, then I would need to always call `next` with an argument, when that only matters once.

...continuing to type as I think, I'm realizing that the first call to `next` is always going to have a void argument, and indeed the argument to next is optional already...

OK, I think I'm convinced. Still going to think this over, but right now I think they way forward is to mark this test as OK and try to get the error to disappear.

Separately, thanks for the hint on use types for this. Sounds like a fun challenge.

lib/core.js
```diff
@@ -307,6 +307,15 @@ interface Iterable<T> {
     @@iterator(): Iterator<T>;
 }
 
+/* Generators */
+interface Generator<Y,R,N> {
+    @@iterator(): Iterator<Y>;
+    next(value?: N): IteratorResult<Y|R>;
+}
+
```
gabelevi: Super rough idea, but I wonder if we could express this as something like

```JavaScript
interface $Iterator<Y, R, N> {
  next(value?: N): IteratorResult<Y>;
}
type Iterator<T> = $Iterator<T, *, *>

interface $Iterable<Y, R, N> {
  @@iterator(): $Iterator<Y, R, N>;
}
type Iterable<T> = $Iterable<T, *, *>;

// Generator<Y, R, N> should be an $Iterable<Y, R, N>
// Generator<Y, R, N> should be an Iterable<Y>
// Generator<Y, R, N> should be an $Iterator<Y, R, N>
// Generator<Y, R, N> should be an Iterator<Y>
interface Generator<Y, R, N> {
    @@iterator(): Iterator<Y>;
    next(value?: N): IteratorResult<Y|R>;
}

declare function $yieldDelegate<R, N>(g: $Iterable<*, R, N>, n:N): R
```

If we do this, I think we could deal with something like

```JavaScript
yield *({ [Symbol.iterator]: function *() { return yield 4; } })
```

ignoring the fact that Flow doesn't have support for `Symbol.iterator`, of course :P

tests/generators/generators.js
gabelevi: > What is the type of `*` here?

It seems to me that the only constraint here is that the result of `yield 42` is strictly compared to a string. Following the spec, the `yield *xx` will flow `N` along to `ArrayIterator.prototype.next()` which ignores it and imposes no restrictions.

>  I'm realizing that the first call to next is always going to have a void argument, and indeed the argument to next is optional already...

Oh yeah...duh...that totally is why my babel example behaves the way it does. Gabe is dumb :P

lib/core.js
samwgoldman: Sorry, I'm a bit unclear about what that buys us. Could you extend your example with a test case showing an expected type error?

samwgoldman:
So I thought of another situation with typing Generators that is basically a Huge Bummer.

In the original issue I showed this example, showing how a common usage of generators is not terribly amenable to type checking:

```javascript
function *userGen() {
  var name = yield("What is your name?");
  var age = yield("How old are you?");
  return { name, age }
}

var gen = userGen();
var name, age, user;
if (name = prompt(gen.next().value)) {
  if (age = prompt(gen.next(name).value)) {
    user = gen.next(parseInt(age, 10)).value;
    // what is the type of user?
    // ideally { name: string; age: number }
    // but we can't match up `next` calls to `yield` calls
    // so we get { name: string|number, age: string|number }
  }
}
```

There's a related issue as well, caused by the fact that the argument to `next` is optional.

```javascript
function *bmiCalc() {
  var height: number = yield("What is your height, in meters?");
  var weight: number = yield("What is your weight, in kilograms?");
  return weight / (height * height);
}
bmiCalc.next(); // initial value to next is always blank
bmiCalc.next(); // this should be an error, but we allow it
```

So, unsound. Bummer. We have a few options:

* Leave it as is—super unsatisfying because of soundness issues
* Force yield return values to always be `N | void`—bad ergonomics, but also similar the union building we do for next arguments already
* Properly model generators: allow next/yield pairs to have different types—not possible in general, afaict, and way beyond my ability to implement

lib/core.js
gabelevi: So here's what I'm trying to do. If you `yield *` to a generator, your PR accurately tries to match the delegating generator's `R` and `N` to the delegated generator's `R` and `N`. This is right and correct and awesome.

If you're `yield *`'ing to something that is NOT a generator but IS an `Iterable`, then you're not doing anything with the delegating generator's `R` and `N`. This is fine if the `Iterable` is something like an array, since `R` is `void` and `N` is `void`. However, I can conceive of a custom `Iterable` who's `Iterator` is something fancy, like a generator.

So given the example above

    yield *({ [Symbol.iterator]: function *() { return yield 4; } })

If you call `next('batman')`, this `yield *` expression should evaluate to `batman`. (The example in babel: https://goo.gl/1Xq5VV)

I think it's possible to capture this idea, by creating `$Iterator` and `$Iterable` which keep track of the `N` and `R` types too. I'm not possible it will work, but it would be cool!

lib/core.js
```diff
@@ -307,6 +307,15 @@ interface Iterable<T> {
...
+// we use this signature when typing yield* expressions
+declare var $yieldDelegate: (<R,N>(g: Generator<*,R,N>, n:N) => R) & ((i: Iterable<*>, n:mixed) => void);
```
gabelevi: The `void` return type might also be wrong. As in my example above, if the `Iterable` has a generator for an `Iterable`, that generator might return and that returned value will be used here.

gabelevi:
I wouldn't get too bummed. There are a ton of common practices in dynamic programming languages that aren't amenable to static typing. When one encounters these there are a couple options.

1) Fight for soundness at all costs. Soundness is great, but this will cause users pain and they'll be unhappy.
2) Allow the users to do clowny things at the cost of soundness. Users will be happy until they shoot themselves in the foot and blame us for not warning them.

There isn't a right answer. We just need to balance how hard we're making the users' lives versus how likely they are to shoot themselves in the foot.

> Force yield return values to always be `N | void`

There's precedence for going either way. We make users check for `undefined` in optional function parameters, but we ignore the fact that indexing into an `Array<number>` might return `undefined`. I'm slightly more tempted to make the return type `N | void` since you're right, the user already has to deal with the union building. If they need to inspect the union before they use it, they might as well check for `undefined` too.

In any case, we have a somewhat iterative approach to Flow. It's not terrible to make a choice now and change it later if we come to regret it. It's always easier to start stricter and relax the rules later than to go from relaxed to strict (cause then you need to go fix your code). So that's my 2 cents...take it or leave it!

> Properly model generators: allow next/yield pairs to have different types—not possible in general, afaict, and way beyond my ability to implement

It does seem really bloody hard in general, especially with tricky control flows in the generator or in the calling code. I definitely think we shouldn't try this at the moment.

nmn:
Any update on this?

samwgoldman:
@nmn This is pretty close to being done, but I'm having trouble triggering an expected error, which confounded me for a while. To be honest, I haven't spent much time on it in a while because I've been busy with my day job. Definitely want to dig into this soon, though.

nmn:
@samwgoldman I'm sorry to bug you about this. Love all the work you're putting in and I'm using my fork to use let/const as var anyway.

Anyway, I've been learning a lot about the AST recently. Though I'm still completely illiterate in ocaml.
I hope to learn some ocaml soon, so I can contribute instead of just complaining.

If it's something that can be explained, I'm very interested in knowing what the missing error is?

samwgoldman:
@nmn If you read through the activity on this PR, you'll find [this comment](https://github.com/facebook/flow/pull/534#discussion_r32694874) which attempts to explain it.

Also, @gabelevi made a good point about iterables with generators as iterators, which I need to handle, although that should be easier.

Looking forward to seeing your contributions! Learning OCaml isn't too too hard. :)

nmn:
@samwgoldman oh i did read that note, but I wasn't sure you were blocked by the same thing. The whole AST, compiling stuff is new for me. Currently practicing my skills with eslint (and now Babel) plugins.

Reading ocaml tutorials on the side. That whole `+.` to add floats is quite funky, but I seem to get most of it.

samwgoldman:
@gabelevi OK, I went with your suggestion to include `<Y,R,N>` type params into the `Iterable` and `Iterator` interface, `$`-prefixed to keep the convenient definitions we have now.

I also rebased this and re-wrote the description on this issue to reflect the current state of things. This is ready for review!

nmn:
👏👏👏👏
So excited about this.

(please now let/const is the only pending ES6 feature, right?)

AprilArcus:
I can't wait to see if this is practical to use with Koa.

samwgoldman:
I brought up in IRC that it might be worth exploring "soundiness" w.r.t. generators. I think the highest-reward trade-off we could make is the return type from `yield` expressions. Right now we wrap with `OptionalT` because the call site is optional, but unwrapping the VoidT is crazy annoying.

tests/generators/generators.js
```diff
@@ -0,0 +1,106 @@
...
+for (var x of delegate_yield_generator()) {
+  (x : number) // error: string ~> number
+}
+
+function *delegate_return_generator() {
+  function *inner() {
+    return "";
+  }
+
+  var x: number = yield *inner()
```
gabelevi: Can you comment that this is an error?

lib/core.js
```diff
@@ -295,18 +295,26 @@ declare class JSON {
 }
 
 /* Iterators */
-interface IteratorResult<T> {
+interface IteratorResult<Y,R> {
```
gabelevi: Why did you decide to parameterize `IteratorResult` on two things rather than just one? They seem equivalent to me...I'm just curious

lib/core.js
```diff
@@ -295,18 +295,26 @@ declare class JSON {
...
     done: boolean;
-    value?: T;
+    value?: Y|R;
```
gabelevi: So for `Iterator<string>`, you'll have `value?: string | void`. It's a little weird that there's a `void` due to being optional and a `void` due to there being no return, but I think it's fine. 

I actually wonder if maybe we can make value not be optional, since the union'd `void` captures the behavior perfectly. Like if iterate over a `Generator<string, number, void>`, will the `IteratorResult` ever be omitted? Looking at [the mozilla docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) it seems like it shouldn't be `undefined` if there is a return

gabelevi:
It's always easier to go from stricter->less strict than the other way around. I think you're making the right call with the `OptionalT`, and if we change our minds later we can go the less sound route.

But yeah, this PR is looking great! I think we can get this merged for the next deploy!

lib/core.js
samwgoldman: The value is optional—it becomes undefined upon completion of the iterator. i.e., when `done=false`, `value=undefined`.

lib/core.js
samwgoldman: Just a bit of future proofing. Once #577 is fixed, we can use a better type for this that takes advantage of this parameterization:

`type IteratorResult = { done: false, value: Y } | { done: true, value: ?R }`

(Annoyingly, the `value` member is `R` the *first time* `done` switches to `true`, but if you call `next` again, it becomes `undefined`. >:(.

samwgoldman: That said, #577 hasn't changed yet, and interfaces as type unions isn't possible either, so I'll commit this change and you can take it at your discretion.

samwgoldman:
@gabelevi OK. Unless I'm mistaken regarding the optional-ness of `IteratorResult`'s `value`, the changes should cover your comments. Merged into latest master as well.

lib/core.js
```diff
@@ -300,13 +300,21 @@ interface IteratorResult<T> {
...
-    @@iterator(): Iterator<T>;
+interface $Iterator<Y,R,N> {
+    @@iterator(): $Iterator<Y,R,N>;
+    next(value?: N): IteratorResult<Y|R>;
 }
+type Iterator<T> = $Iterator<T,void,void>;
 
-interface Iterable<T> {
-    @@iterator(): Iterator<T>;
+interface $Iterable<Y,R,N> {
```
jeffmo: Can we name these type vars more explicitly? They can show up in error messages, and sometimes without context it makes the error more confusing/unclear.

samwgoldman: Do you have an opinion on the more explicit names? `Yield`, `Return`, and `Next`?

samwgoldman: I have no objection. I can add a commit here, but it seems like an easy enough change to make in your internal merge, too, if you want to make the change yourself. (I'm not sure what the internal merge looks like, of course, so this could be wrong-headed of me.)

jeffmo: Yea, `Yield`/`Return`/`Next` seem good. I'll pass it along to @gabelevi who's submitted the internal merge request

src/typing/type_inference_js.ml
```diff
@@ -5084,6 +5113,8 @@ and mk_class_elements cx instance_info static_info body = Ast.Class.(
       let this, super, method_sigs, getter_sigs, setter_sigs =
         if static then static_info else instance_info
       in
+      let yield = MixedT (mk_reason "no yield" loc) in
```
jeffmo: Is this considering generator methods? It looks like you're not pivoting on the `generator` param of the Function expression above...

(Also mind adding a test or two for class methods in any case?)

samwgoldman: I... actually didn't realize generator methods were possible. I'll add a suite of tests and add the probably missing implementation stuff. This PR should be blocked until then. Thanks for catching this.

samwgoldman:
@gabelevi @jeffmo OK, so I think I've responded to all feedback here, now. There's some commentary in the individual commits.

The changes that enable class method generators also fixed a stupid bug with generators causing unexpected BoundT errors. However, I'm not able to get simple generic generator definitions to type check. For example:

```javascript
function *generic_yield<Y>(y: Y): Generator<Y,void,void> {
  yield y;
}
```
generators.js:108:25,25: Y
generators.js:108:1,110:1: some incompatible instantiation of Y
```

I don't see anything wrong with the above definition, though. At the moment I'm pretty stuck on that, and I'm not sure banging on it more is going to lead to a lot of insight. Either of you able to show that a) the error is correct or b) what's wrong with the implementation?

jeffmo:
I know several people who've been asking for this for a long time now... You're going to be a hero!

We're going to do a deploy today or tomorrow and wanted to be able to get this in. We can just circle back on the generics stuff later.

gabelevi:
Woo generators are in! Nice job @samwgoldman!

@jeffmo ++ on the generics stuff. You're right, that example shouldn't error, but what you have is great and I really want it in this deploy.

Again, thanks so much for building this. It's awesome, you're awesome, and I'm excited to get it out!

dchambers:
@samwgoldman / @gabelevi: is there any documentation explaining the work that was done for ES6 generators as I've been unable to find anything by skimming the Flow docs? For example, this snippet was shown in the preceding comments:

~~~es6
function *generic_yield<Y>(y: Y): Generator<Y,void,void> {
  yield y;
}
~~~

yet I get errors when I try to use `Generator`. Additionally, I'd like to know how to insist that the type of an object after yielding won't be `undefined`, since for my generator it never will be.

samwgoldman:
@dchambers Right now the only docs are [on the blog](http://flowtype.org/blog/2015/11/09/Generators.html), but hopefully that should cover everything you want to know. Let me know if it doesn't.

dchambers:
Excellent documentation @samwgoldman. First class stuff!

In case it helps anyone else I had to mark `Generator` as being a global in my `.eslint` config to prevent linting errors, for example:

~~~
"globals": {
  "Generator"
},
~~~

but everything else worked exactly as described in the documentation.

samwgoldman:
Excellent! That's really nice to hear. Also, thanks for reminding me that I'll need to update that post soon, since @mroch landed some refinement improvements that will make using generators a lot more convenient.

Issue #188, opened 2014-12-12
No entry for `new Date(...)` in lib/core.js


bolinfest:
Reading through https://github.com/facebook/flow/blob/master/lib/core.js, there is no `constructor` entry in `declare class Date {`.

josephsavona:
@avikchaudhuri I confirmed that `var date: Date = new Date()` passes. Looks like this issue can be closed?

bolinfest:
@josephsavona I do not believe that is a sufficient test. `Date` has a very complicated constructor: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date.

josephsavona:
@bolinfest Flow accepts all of the below cases. Do you have an example that is valid JS but that Flow rejects?

```
var date1: Date = new Date();
var date2: Date = new Date('2015/01/01');
var date3: Date = new Date(2015, 01, 01);
var date4: Date = new Date(1);
```

bolinfest:
@josephsavona I don't know if that proves anything. I think Flow ignores extra arguments, which is why it doesn't flag anything you are seeing. The real test is whether it rejects invalid arguments. For example, if I write:

    new Date({}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {});

I don't see any errors, but I should.

knsn1994:
@bolinfest you are right. Flow ignores extra arguments. So we cannot really Flag 
    new Date({}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {});

jjergus:
We can flag new Date({}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}, {}); because {} doesn't match the required type of the arguments (string or int for the first argument, int for others). #539 takes care of this.

Ignoring extra arguments is probably OK because JavaScript ignores them as well. Extra arguments make your code ugly, but won't cause actual bugs.

One thing we might want to flag though are invalid argument combinations like new Date(string, int). This is not solved by #539 because Flow doesn't supported overloaded constructor definitions yet.

samwgoldman:
This can be closed, now that #539 is merged.

samwgoldman:
`Date` constructor is was added to flowlib in 0.13.

PR #539, opened 2015-06-18
Date constructor definition (fixed tests)

jjergus:
Pull request #537, now with fixed tests (changed line numbers in core.js).

Partially solves #188

Includes updated test case.

It looks like overloaded constructors are not supported, so this doesn't detect all invalid constructor calls (see examples in the test case).

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -207,6 +207,11 @@ declare class RegExp {
 }
 
 declare class Date {
+    // new Date();
+    // new Date(timestamp);
+    // new Date(dateString);
+    // new Date(year, month[, day[, hour[, minute[, second[, millisecond]]]]]);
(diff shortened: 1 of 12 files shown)


knsn1994:
Nice.

lib/core.js
```diff
@@ -207,6 +207,11 @@ declare class RegExp {
 }
 
 declare class Date {
+    // new Date();
```
bhosmer: Hi Jan - if you could add a comment here noting that we want to use constructor overloads once they become available, that would be great.

bhosmer:
One comment request, but looks good - thanks a bunch! Will merge as soon as you update. :)

jjergus:
Done.

bhosmer:
Sweet! Will merge once Travis build is done. Thanks again!

bhosmer:
Dunno what's going on with Travis... merging

PR #540, opened 2015-06-18
Remove duplication between function/method type constructors

samwgoldman:
Just a small refactoring. I think this formulation makes the differences
more immediately obvious, while highlighting the similarities as well.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -44,14 +44,6 @@ let dummy_prototype =
 let dummy_this =
   MixedT (reason_of_string "global object")
 
-let mk_methodtype this tins ?params_names tout = {
-  this_t = this;
-  params_tlist = tins;
-  params_names;
-  return_t = tout;
-  closure_t = 0
-}
-
 let mk_methodtype2 this tins ?params_names tout j = {
   this_t = this;
   params_tlist = tins;
@@ -60,21 +52,12 @@ let mk_methodtype2 this tins ?params_names tout j = {
   closure_t = j
 }
 
-let mk_functiontype tins ?params_names tout = {
-  this_t = dummy_this;
-  params_tlist = tins;
-  params_names;
-  return_t = tout;
-  closure_t = 0
-}
+let mk_methodtype this tins ?params_names tout =
+  mk_methodtype2 this tins ?params_names tout 0
 
-let mk_functiontype2 tins ?params_names tout j = {
-  this_t = dummy_this;
-  params_tlist = tins;
-  params_names;
-  return_t = tout;
-  closure_t = j
-}
+let mk_functiontype2 = mk_methodtype2 dummy_this
+
+let mk_functiontype = mk_methodtype dummy_this
 
 (* An object type has two flags, sealed and exact. A sealed object type cannot
    be extended. An exact object type accurately describes objects without


bhosmer:
Hi Sam! Just saw this work come over the transom and made some comments on the internal version, @jeffmo suggested I put them up here as an FYI as well:

I don't think these names are quite right. Functions may be identical to methods with a dummy `this` in the current implementation, but that's not what they actually are abstractly, so I'd rather not use names that imply it.

Also +1 to @mroch's comment. [He was venting about the `foo`/`foo2` pattern that's still in here from the early days - BH] Whenever we touch these (or `f`, `f_` pairs), it's worth (a) figuring out the actual difference between them and (b) renaming them to reflect it.

Some suggested names in inline comments...
Inline Comments
flow/src/typing/flow_js.ml
47
```
mk_function_or_method_type_with_closureid = { ... }

let mk_methodtype_with_closureid = mk_function_or_method_type_with_closureid
```
58
```
mk_functiontype_with_closureid
```

...this isn't a call to action, Jeff will take whatever of these suggestions he feels good about and fold them in on our side - just wanted to proselytize. :) ...but of course it goes without saying that I'm totally keying off of the impulse to factor that led you to put up this PR in the first place, so thanks again!

samwgoldman:
Based on that, I think it makes sense to close this, because the proposed refactoring is actually a step in the **wrong** direction. Glad I could help bring this issue to a head. :stuck_out_tongue: 

The comment above this section of the code ("function types and object types, which unfortunately have many fields, not all of which are meaningful in all contexts") and your comment, above ("that's not what they actually are abstractly") are a bit vague—is the issue that we don't quite know what the constraints are, or that we do know the constraints, but not the representation?

For my own curiosity, if someone could explicate the constraints surrounding a correct implementation of these types, I would certainly appreciate it.

Thanks for sharing the internal discussion with me!

bhosmer:
I don't know that it's a step in the wrong direction at all - it's just that there are two levels of meaning here. Your refactoring makes perfect sense at the level of implementation: as of now, functions and methods really are modeled as variations on a theme, the variations being the meaning of `this`. 

OTOH I just wanted the names to reflect the distinction between the two concepts at the level of the language's abstractions: we don't actually want to lock into an abstract definition of `function` that's simply a synonym of `method with dummy this`. 

WRT the other comment you mention, I'm going to have to defer a deeper discussion, but the problem is a general one of data design: a big record is used to represent several different concepts, and not everybody uses all the fields all the time. The tension lies in the fact that tightening the representation, for instance with inheritance of by introducing tagged unions either at the top level or in subsections, adds complexity. It basically boils down to a fight between tightness/correctness by construction/architectural heaviness on one side, and simple data design along with looseness, asserts, private agreements on the other. No perfect answer, but the general movement is from the latter towards the former, as design stabilizes and problem boundaries start to settle down.

An example of this shows up in the upcoming `Env_js` stuff, which BTW the second update I mentioned is in process and should land soon, at which point I can go over it with you from the perspective of Let/Const.

samwgoldman:
I suppose what I was asking was where does the notion that function = method w/ dummy this actually break down? Do we lose out on correctness in the type checker? This seems like a pretty subtle area of the language, so I probably just need to read up on it. No need to spend a bunch of time explaining it to me.

Moving from simple data with private agreement to heavier, correct-by-construction implementations seems like a good path forward. Even in the short time I have been following along, I have witnessed this trend.

Continuing to look forward to the Env_js update. :smiley:

PR #541, opened 2015-06-18
Allow `undefined` to be passed to optional params

### LINKED ISSUE #437: Inconsistent behavior when passing `undefined` to a function
@gabelevi points out that [this doesn't type check][1]:

```javascript
function foo(x?: number) {}

function bar() {
    foo(undefined);
}
```

It seems like it should.

There is an observable distinction between `foo(undefined)` and `foo()` in that `arguments.length` will have a different value. So there is an argument that the current behavior isn't totally wrong. But distinguishing between an explicit and implicit `undefined` is generally considered bad API design.

From IRC:

danvk: flow-help: I’m puzzled by the error in this gist: https://gist.github.com/danvk/894d978ee9229400b72c
danvk: I’m passing a value that’s Object|undefined to a constructor that expects an optional Object parameter
danvk: which seems like it should be OK
glevi: Like this is an error: http://tryflow.org/#90ca4cbbaf98fe3d5c14eb8a8e2c53e1
glevi: But this is not: http://tryflow.org/#434a04ebff6649df1f66f222e453c75d
glevi: danvk: Sure...we seem to be somewhat inconsistent with what we let flow into an optional param...this doesn't work either: http://tryflow.org/#c788cea5d233f6449ee51b6d75297dbf

[1]: http://tryflow.org/#c788cea5d233f6449ee51b6d75297dbf

samwgoldman:
There is already a flow for `VoidT ~> OptionalT`, but in the case of
function parameters, `multiflow` actually unwraps the `OptionalT`, so by
the argument flows into the parameter, flow isn't able to detect this
case.

I fixed this by adding a redundant case to `multiflow` instead of a
larger overhaul that would use the existing flow.

Note that functions with default argument values use the same mechanism
as optional parameters. The included test ensures that the default
expression type still flows correctly.

Fixes #437

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -4238,29 +4238,19 @@ and multiflow_partial cx trace ?strict = function
   *)
   | (_,[]) -> []
 
-  | ([],(OptionalT t)::touts) -> (OptionalT t)::touts
-
-  | ([RestT tin],(OptionalT tout)::touts) ->
-    rec_flow cx trace (tin, tout);
-    multiflow_partial cx trace ?strict ([RestT tin],touts)
-
-  | ((OptionalT tin)::tins,[RestT tout]) ->
-    rec_flow cx trace (tin,tout);
-    multiflow_partial cx trace ?strict (tins,[RestT tout])
-
   | ([RestT tin],[RestT tout]) ->
     rec_flow cx trace (tin,tout);
     []
 
-  | (tins,[RestT tout]) ->
-    tins |> List.iter (fun tin ->
-      rec_flow cx trace (tin,tout);
-    );
-    [RestT tout]
+  | ([RestT tin],tout::touts) ->
+    rec_flow cx trace (tin,tout);
+    multiflow_partial cx trace ?strict ([RestT tin], touts)
 
-  | (tin::tins,(OptionalT tout)::touts) ->
-    rec_flow cx trace (tin,(OptionalT tout));
-    multiflow_partial cx trace ?strict (tins,touts)
+  | (tin::tins,[RestT tout]) ->
+    rec_flow cx trace (tin,tout);
+    multiflow_partial cx trace ?strict (tins, [RestT tout])
+
+  | ([],[RestT _]) -> []
 
   | ([],ts) ->
     (match strict with
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -4248,6 +4248,9 @@ and multiflow_partial cx trace ?strict = function
     rec_flow cx trace (tin,tout);
     multiflow_partial cx trace ?strict (tins,[RestT tout])
 
+  | ((VoidT _)::tins,(OptionalT _)::touts) ->
+    multiflow_partial cx trace ?strict (tins,touts)
+
   | ((OptionalT tin)::tins,(OptionalT tout)::touts)
   | (tin::tins,(OptionalT tout)::touts) ->
     rec_flow cx trace (tin,tout);
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -4248,14 +4248,6 @@ and multiflow_partial cx trace ?strict = function
     rec_flow cx trace (tin,tout);
     multiflow_partial cx trace ?strict (tins,[RestT tout])
 
-  | ((VoidT _)::tins,(OptionalT _)::touts) ->
-    multiflow_partial cx trace ?strict (tins,touts)
-
-  | ((OptionalT tin)::tins,(OptionalT tout)::touts)
-  | (tin::tins,(OptionalT tout)::touts) ->
-    rec_flow cx trace (tin,tout);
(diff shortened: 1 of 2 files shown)


samwgoldman:
There's another case that this doesn't fix:

```javascript
function foo(x?: number) {}
function foo2(x?: number) {
  foo(x) // ok
}
function foo3(o: { x?: number }) {
  foo(o.x) // fails?
}
```

Not sure what's going on there, though.

avikchaudhuri:
@samwgoldman Cool, I recently fixed a related issue in multiflow that doesn't cover this, but FYI nevertheless for rebasing etc.: https://github.com/facebook/flow/commit/32acbe7ff087f0b0d3930d1f59ade5b05f7ed6a3

samwgoldman:
@avikchaudhuri Saw it! That's what made me remember this issue & helped me know where to look ;)

samwgoldman:
@avikchaudhuri I pushed [a commit](https://github.com/samwgoldman/flow/commit/04d964e4180d05108c7970f64f2828704b18927f) that does just what you suggested, and that works. Thinking about this more, I realized that I could actually [simplify more aggressively](https://github.com/samwgoldman/flow/commit/5bf35a934e198ff3e58325d583d656bb4c871f1c), which I committed separately.

Can you see any case where the cases I landed on are insufficient?

avikchaudhuri:
This is excellent! Let's try to prove your aggressive simplification correct as follows: take every case in the original code and see that we get the same constraints in the new code. I briefly went through that exercise and couldn't find any errors. Can you?

samwgoldman:
Thanks for pushing me to be rigorous about this change. I have attempted to do so case-by-case, below. I think I identified a small difference regarding VoidT lower bounds, which I bolded for easy identification, but maybe I am mistaken—still wrapping my head around this stuff.

---

Here is a list of the previous cases in this function.

    | (_,[]) -> []

Still there. Still first.

    | ([],(OptionalT t)::touts) -> (OptionalT t)::touts

Will be handled by the `([],ts)` case, which will delegate to existing normal
flows when `strict = Some reason`. Before this was a noop, and now we will run
through `(VoidT, OptionalT)` flows, which are also noops, so OK.

    | ([RestT tin],(OptionalT tout)::touts) ->
      rec_flow cx trace (tin, tout);
      multiflow_partial cx trace ?strict ([RestT tin],touts)

Will be handled by the `([RestT tin],tout::touts)` case, which will use the flow
for `(tin,(OptionalT tout))`, which is similar to existing behavior except for
unwrapping the `OptionalT`. The existing `(VoidT, OptionalT)` and
`(t1, OptionalT(t2))` flows make this equivalent.

    | ((OptionalT tin)::tins,[RestT tout]) ->
      rec_flow cx trace (tin,tout);
      multiflow_partial cx trace ?strict (tins,[RestT tout])

Will be handled by the `(tin::tins,[RestT tout])` case, which will use the flow
for `((OptionalT tin),tout)`, which is similar to existing behavior except for
unwrapping the `OptionalT`. The existing flow for `(OptionalT(t), _)` will pick
this up, and accumulate two lower bounds: `VoidT` and `t`. I believe this is
actually **different behavior** (before we didn't create a `VoidT` lower bound), but
I think that it is correct.

    | ((OptionalT tin)::tins,(OptionalT tout)::touts)
    | (tin::tins,(OptionalT tout)::touts) ->
      rec_flow cx trace (tin,tout);
      multiflow_partial cx trace ?strict (tins,touts)

Mostly the same as the two above cases. The  **different behavior** is the lower
bound `OptionalT` now picks up a lower bound `VoidT`, where it did not before.

    | ([RestT tin],[RestT tout]) ->
      rec_flow cx trace (tin,tout);
      []

Same case, just moved around.

    | (tins,[RestT tout]) ->
      tins |> List.iter (fun tin ->
        rec_flow cx trace (tin,tout);
      );
      [RestT tout]

This same case exists, but instead of iterating through all `tins`, the new code
only handles one and recurses on the rest. This rule is replaced by two rules,
one for non-nil tins, one for nil.

    | ([],ts) ->
      (match strict with
      | Some reason_op ->
          let reason = replace_reason
            "undefined (too few arguments, expected default/rest parameters)"
            reason_op
          in
          ts |> List.iter (fun t -> rec_flow cx trace (VoidT reason, t))
      | None -> ()
      );
      ts

Same rule exists unchanged, but it is used in more cases (described above) and
has the correct semantics for those cases.

    | (tin::tins,tout::touts) ->
      rec_flow cx trace (tin,tout);
      multiflow_partial cx trace ?strict (tins,touts)

Base case. Unchanged, but picks up more cases (described above) and has the
correct semantics for those cases.

samwgoldman:
Merged with recent master to get over merge conflicts. Only difference is I return the `[RestT tout]` in the `([], [RestT tout])` case, because your intermediate commit did, even though the return value is just ignored.

samwgoldman:
@avikchaudhuri is that the kind of analysis you were looking for?

avikchaudhuri:
@samwgoldman yeah this looks great. And it fixes a couple of cases, too!

Issue #499, opened 2015-06-05
Allow checking of .es6 files


mikz:
Currently flow supports just checking `*.js` and `*.jsx` files.

Consider supporting `*.es6` or just making it a configurable option.

https://github.com/facebook/flow/blob/8a34b0743850ea12a322c175112e18cc285f3da8/src/common/files_js.ml#L16-L19

mroch:
good idea. I vote for making it configurable

mroch:
Fixed by #562

PR #553, opened 2015-06-22
Allow checking of .es6 files

Eyyub:
This pull request allows checking of .es6 files with `flow {start|single|server} --check-es6-files`.
See #499

### DIFF
--- a/src/commands/serverCommands.ml
+++ b/src/commands/serverCommands.ml
@@ -51,6 +51,8 @@ module OptionParser(Config : CONFIG) = struct
         ~doc:"Do not include embedded declarations"
     |> flag "--log-file" string
         ~doc:"Path to log file (default: /tmp/flow/<escaped root path>.log)"
+    |> flag "--check-es6-files" no_arg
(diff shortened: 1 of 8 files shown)


aikeru:
Is ".es6" an extension we want to stick with? ES7 is going to come a lot quicker than ES6 did.
Isn't there already an option to target arbitrary file names/paths with flow?

Eyyub:
Maybe, not to my knowledge.
Adding an option to target arbitrary file extension is easy, it's more a design issue than anything else.

jeffmo:
Thanks for sending this over!

Rather than trying to predict all the possible extensions, it would be better if we created a .flowconfig option that allows one to specify a custom extension to expect (where the default is the existing `[".js";".jsx"]`

Check out https://github.com/facebook/flow/blob/master/src/common/flowConfig.ml to see how the config file is parsed/passed around

Eyyub:
Oh I see, it's a much better idea, I will work on tonight !

PR #556, opened 2015-06-23
Don't create default constructor for subclasses

### LINKED ISSUE #214: Inherited constructor's arguments are not type-checked
```js
/* @flow */

class A {
  constructor(name: string) {
    this.name = name;
  }
};

class B extends A {
}

new B(1);
```

Expected:
```
class.js:12:7,7: number
  class.js:4:21,26: string
```

Actual:
```
No errors!
```

samwgoldman:
This behavior prevented classes from correctly inheriting the
constructor function from their superclass.

Fixes #214


samwgoldman:
Build failed, but current `master` is failing in the same way.

jeffmo:
Whoa, how did we miss this one? Awesome!

Technically a default constructor is made of the form `constructor(...args) { super(...args); }`...

The only ways I can imagine this approximation becoming observable is if someone does an equality test on `this.constructor` or if we start tracking the literal values for `constructor.{name,length}` at some point.

I can't think of anything right now, though, so at the very least would you mind adding a comment over this code explaining the approximation in case it surfaces at some point?

avikchaudhuri:
Wow, I never thought this was intended behavior. In Java e.g. you'd get a type error if you extended a class with non-trivial constructor with a class without a constructor. Anyway, very cool!

samwgoldman:
@jeffmo Interesting point about constructor inheritance being an approximation. Does Babel have [a related issue](http://bit.ly/1K8y5VP), then?

jeffmo:
Yes, I believe it does:

https://people.mozilla.org/~jorendorff/es6-draft.html#sec-runtime-semantics-classdefinitionevaluation

```
10. If constructor is empty, then,
  a. If ClassHeritage(opt) is present, then
    i. Let constructor be the result of parsing the String 
       "constructor(... args){ super (...args);}" using the syntactic grammar 
       with the goal symbol MethodDefinition.
  b. Else,
    i. Let constructor be the result of parsing the String "constructor( ){ }" using 
       the syntactic grammar with the goal symbol MethodDefinition.
```

sebmck:
@jeffmo Yup, [it does](http://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=class%20Foo%20%7B%0A%20%20constructor()%20%7B%0A%20%20%20%20console.log(%22foobar%22)%3B%0A%20%20%7D%0A%7D%0A%0Aclass%20Bar%20extends%20Foo%20%7B%0A%20%20%0A%7D%0A%0Anew%20Bar%3B&playground=false)?

sebmck:
Oh wait, misunderstood about the equality stuff. Will investigate.

jeffmo:
@sebmck: It properly falls through to executing the right logic, but given your example code:

`Foo.constructor` should not `===` `Bar.constructor` (but it seems to do this per @samwgoldman 's example)

sebmck:
Wait, `Foo.constructor === B.constructor` is correct. Their constructors are `Function`. Output from the latest io.js:

```
> class Foo {}
[Function: Foo]
> class Bar extends Foo {}
[Function: Bar]
> Bar.constructor === Foo.constructor
true
> Bar.constructor
[Function: Function]
> Foo.constructor
[Function: Function]
```

jeffmo:
Aha! It's not that `Foo.constructor !== Bar.constructor`, but that `Foo.prototype.constructor !== Bar.prototype.constructor` (which Babel seems to handle properly)

samwgoldman:
Ah, that makes sense. Regardless, the only time flow cares about equality is in comparison with `null`, `undefined`, and `typeof` expressions, so the semantic distinction between "materialized" default constructors and inherited ones doesn't factor into what we are trying to analyze—at least, not yet.

Issue #209, opened 2015-01-06
Make the flow extensions configurable


lfrancke:
Currently the Flow server only checks files with `js` and `jsx` file endings: https://github.com/facebook/flow/blob/master/src/common/files_js.ml#L16-L19

It'd be great if these extensions were configurable. One potential use-case would be to check `ts` files.

lfrancke:
I brought this up on IRC and @gabelevi agreed that this might be a useful feature.

For my use case just being able to configure the extensions would be enough: I used Flow for the first time to day and have never used TypeScript but from reading around and looking at the two the syntax seems to be the same/very similar. I use IntelliJ IDEA as my idea and it supports TypeScript for `ts` files. If I write JavaScript with type annotations for flow it complains about just about everything. So what I'd like to do is to use `ts` files, let Flow check them and then process from there... 

I agree that this is not the most useful feature but as tooling support is lacking this would be one workaround to make it a bit easier.

avikchaudhuri:
OK, makes sense. But supporting a .ts extension does give the false impression that any .ts file is a syntactically valid Flow file, or worse, if it is syntactically valid it is also semantically valid. Also, I'm not sure whether IntelliJ will do TypeScript transforms automatically for .ts files, but they're not going to work with Flow files in general. Which is why I'm hesitating. Otherwise, making the supported extensions configurable is not a problem. @gabelevi: any further thoughts given these considerations?

lfrancke:
I totally agree: I wouldn't want flow to give false impressions. Making the extensions configurable doesn't need to mention `ts` at all though. It can be just another config option. People could have all kinds of reasons why they don't use `js` as the extension for their JavaScript files.

This is a very low priority wish and a dirty workaround and I won't be offended if you decide not to implement it. I can always try to compile my own flow version if needed.

despairblue:
I use the extension to tell 6to5 which files to transpile and which not to.
By using `require('6to5/register')({extensions: ['.es6']});` it will only transpile files with the es6 extension instead of all javascript files in my project.

That would be a use case for configurable extensions. At the moment flow just ignores my es6 files.

sporto:
I'm using .es6 as well and transpiling them with babel via webpack. +1 for a configuration for custom extensions.

Niggler:
[...]

PR #562, opened 2015-06-24
Add support for configurable choice of file extensions in .flowconfig

Eyyub:
This pull request provides support for configurable choice of file extensions in .flowconfig.
By default, as before, .js and .jsx are valid file extensions.

If user wants to allow some file extensions he has to add it in .flowconfig, for instance :
```
[flow-extensions]
.js
.jsx
.es5
.es6
.es7
```

Note : I'm not sure the regexp I used is good

hey @jeffmo


samwgoldman:
#209, #499

Eyyub:
After discussion with jeffmo, the file extensions config is now part of options section : 
```
[options]
module.file_ext=.js
module.file_ext=.jsx
module.file_ext=.es5
module.file_ext=.es6
module.file_ext=.es7
module.file_ext=.foobar
```
Thoughts ? (about the syntax and the regex)

aikeru:
I think I have a better idea for the es* file extension.
Instead of specifying a version of es (is: es6) just use simply .es
The purpose of a different extension than .js could simply be to allow the source and output files to coexist and to denote that "this is a transpired source file".
When are you realistically going to be using multiple es versions in one project or within flow itself? 

However I think the best idea overall is not to specify extension at all even in configuration.
Instead, use file glob syntax that catches the files you want.

aikeru:
Sorry for errors. Im on my mobile.

Eyyub:
Mmh there is already `[include]` which has this purpose.
`module.file_ext` is just here to specify which file extension is enabled or not.

src/common/flowConfig.ml
```diff
@@ -118,7 +120,7 @@ let empty_config root = {
   include_stems = [];
   include_map = PathMap.empty;
   libs = [];
-  options = (default_options root);
+  options = (default_options root);
```
jeffmo: extraneous whitespace?

src/common/flowConfig.ml
```diff
@@ -380,6 +382,8 @@ module OptionsParser = struct
     options
 end
 
+let file_extension = Str.regexp "^\\(\\.[^ \t\\.]+\\)$"
```
jeffmo: If I'm reading this correctly, this excludes something like ".es6.js" (two dots) -- is there a reason you restricted this?

src/common/flowConfig.ml
```diff
@@ -61,6 +62,7 @@ let default_module_system = Node
 let default_options root = {
   moduleSystem = default_module_system;
   module_name_mappers = [];
+  module_file_exts = [];
```
jeffmo: This should probably just be `[".js"; ".jsx"]`

src/common/flowConfig.ml
```diff
@@ -516,4 +541,4 @@ let get root =
 let get_unsafe () =
   match !cache with
   | Some config -> config
-  | none -> failwith "No config loaded"
+  | None -> failwith "No config loaded"
```
jeffmo: Nice catch!

src/common/files_js.ml
```diff
@@ -13,10 +13,9 @@
 open Utils
 open Modes_js
 
-let flow_extensions = [
-    ".js"  ;      (* Standard JavaScript files *)
-    ".jsx" ;      (* JavaScript files with JSX *)
-  ]
+let get_flow_extensions root =
```
jeffmo: It's a shame that we have to thread `root` through these just for purposes of some internal implementation logic.

It seems `files_js.ml` has an `init` method already, though, for initializing stuff like storing the `config`.Can we just pass the `options` object in through there and store it at the toplevel?

(similar to how things work over in [`module_js.ml`](https://github.com/facebook/flow/blob/master/src/typing/module_js.ml#L472) )

src/typing/constraint_js.ml
```diff
@@ -2167,6 +2167,40 @@ let reasons_of_trace ?(level=0) ?(tab=2) trace =
     else []
   in f level trace
 
+
```
jeffmo: Hmm, this seems like git/rebase noise?

src/typing/module_js.ml
```diff
@@ -263,37 +263,48 @@ module Node = struct
       match get_key "main" tokens with
       | None -> None
       | Some file ->
-          let path = Files_js.normalize_path dir file in
-          if path_is_file path
-          then Some path
-          else seq
-            (fun () ->
-              seqf
+	 let opts = match !flow_options with
```
jeffmo: Seems prime for pulling this out into a local `get_options` function or something (since you're repeating this below as well)

jeffmo:
This is great. Just a few comments and then I think we'll be good to go!

jeffmo:
Oh also: Can you add some tests for this as well?

src/typing/module_js.ml
Eyyub: Great, actually I wanted to do that, but I was not sure if the "guidelines" allowed this.

src/common/flowConfig.ml
Eyyub: `module_file_exts` is an empty list by default, it allows the user to choose if `.js` or `.jsx` will be checked or not, if not `module.file_ext` are specified in `[options]` then `module_file_exts = [".js"; ".jsx"]`.
But if you want to force the user to has its `.js` and `.jsx` to be checked, I can do it.
What are you thoughs ?

jeffmo:
@aikeru: Using file glob syntax is probably a good thing, too. For now this seems strictly minimal and still supportive of a more flexible glob syntax in the future; So I think this is a good first step in the right direction.

I suspect that this will cover most use cases, though (having an explicit list like this worked pretty well for Jest configs, also).

src/common/flowConfig.ml
Eyyub: No there is absolutely no reason, it was just to prototype :)

src/common/flowConfig.ml
```diff
@@ -443,6 +457,10 @@ let parse_options config lines =
     |> List.map (fun (ln, line) -> ln, String.trim line)
     |> List.filter (fun (ln, s) -> s <> "") in
   let options = OptionsParser.parse config options_parser lines in
+  if options.module_file_exts = [] then
+    let options = { options with module_file_exts = [".js"; ".jsx"] } in
```
jeffmo: It seems we're referring to this list below as well, so lets just define this default-extensions list in one place somewhere (maybe at the toplevel?).

src/common/flowConfig.ml
jeffmo: Ahh, this is a good point. We want users to start from an empty list if they specify *something*, but get the default if they specify *nothing*.

The more I think about this, the more it makes me believe we should move the "default" value into something like a `set_default` function in the `OptionsParser.option_entry` record so that it can be called when any option is *not* encountered.

Anyway, don't worry too much about doing this now -- it's more cleanup work than pertinent to this PR, so I don't think it's worth shaving that yak here and now.

src/common/files_js.ml
Eyyub: Agreed, so we do not need a `get_options` function in `modules_js.ml` if we store `options` at `Modules_js`, right ?

jeffmo: The only reason a `get_options` function is useful is just because it has the boilerplate for unpacking the optional ref (and giving a consistent assert error if somehow some call path attempts to access options before `init` was called).

src/common/flowConfig.ml
Eyyub: Good idea !!

src/typing/constraint_js.ml
jeffmo: Are you sure this PR is properly rebased on top of master? These changes shouldn't be showing up here since they happened in another commit...

Eyyub: Nop, indeed it's not, I messed up, I have to take time to fix it because I fear to break something :/

Eyyub:
Oh god :/

jeffmo:
Hehe, git is tricky...

Ok, what I would do here is this:
First, make a backup copy of of your repo directory just in case :p

Then, in the branch that you have these changes in, do:

```
$ git pull --rebase upstream master
$ git push --force origin config-flow-extensions
```

Hopefully this will clear out all of the noise here...

gabelevi:
To be clear, this PR is waiting on recovering from the git drama, right?

Eyyub:
You're right ! I just finished my school year so I will work on it soon.
If you have some ideas about how to fix it I will take it :)

Eyyub:
Yay, the first time ! I feared to have broken something but it seems all is good.

gasi:
:+1: for making file extensions configurable, e.g. `.es` :smile:

samwgoldman:
Is this PR blocked by anything? Seems like a good feature to me.

jeffmo:
Not anymore -- looks like I missed the comment about the commits/rebase stuff being fixed...so I'll take a look/import it tomorrow or Monday when I get back to a comp!

davedx:
Would love to see this merged. It's a blocker for our project which has hundreds of ".es" files :)

Eyyub:
@davedx now it's merged, you can use Flow :)

davedx:
Thank you! :) :+1:

jeffmo:
woo!

Also: Huge thanks to @Eyyub for doing this!

Eyyub:
With pleasure, and thanks for your advices !

mroch:
Closes #209

PR #569, opened 2015-06-26
Declare type proposal

Eyyub:
*do not merge*

This pull request adds `declare` type to `declare module`.

```typescript
/**                                                                                                     
 * @flow                                                                                                
 */                                                                                                                                                                                              
declare module ModuleAliasFoo {                                                                         
    declare type baz = number;                                                                          
    declare type toz = string;                                                                          
    declare function foo(bar : baz) : toz;                                                
} 
```

```javascript
/**                                                                                                     
 * @flow                                                                                                
 */                                                                                                     
                                                                                                        
///////////////////////////////////////                                                                 
// == Import Declared Type Alias  == //                                                                 
///////////////////////////////////////                                                                 
import type {baz} from "ModuleAliasFoo";                                                                
import {foo} from "ModuleAliasFoo";                                                                     
var keh: baz = 42;                                                                                       
var k2: baz = "shab"; // Error: string to int                                                           
var k3: toz = foo(keh); // Error: unknown identifier toz                                                 
                                                                                                        
import type {toz} from "ModuleAliasFoo";                                                                
var k4: toz = foo(keh); 
```

Any thoughts ?


src/parser/parser_flow.ml
```diff
@@ -2852,6 +2852,26 @@ end = struct
...
+	let id = Parse.identifier env in
+	let typeParameters = Type.type_parameter_declaration env in
+	Expect.token env T_ASSIGN;
+	let right = Type._type env in
+	let end_loc = match Peek.semicolon_loc env with
+	  | None -> fst right
+	  | Some end_loc -> end_loc in
+	Eat.semicolon env;
+	Eat.pop_lex_mode env;
+	Loc.btwn start_loc end_loc, Statement.(TypeAlias TypeAlias.({
```
jeffmo: This should generate a `DeclareTypeAlias` node instead of a `TypeAlias` node, right?

src/typing/type_inference_js.ml
```diff
@@ -2123,20 +2123,29 @@ and statement cx = Ast.Statement.(
...
       | None ->
-          let map = SMap.map (
-            fun { specific = export; _ } -> export
-          ) module_scope.entries in
-          Flow_js.mk_object_with_map_proto cx reason map (MixedT reason)
-    ) in
+	 let for_types, nonfor_types =
+	   SMap.partition (fun _ { for_type; _ } -> for_type) module_scope.entries
+	 in
+         let map = SMap.map (
```
jeffmo: nit: can you push this `let` back one char to match the indentation level of the rest of the case logic?

jeffmo:
Can you add some tests here as well?

src/parser/parser_flow.ml
Eyyub: Oups my bad :s

src/typing/type_inference_js.ml
Eyyub: Np, it's weird because on my emacs it's well intented

tests/declare_type/.testconfig
```diff
@@ -0,0 +1 @@
+cmd: check --all --show-all-errors --strip-root --lib lib
```
jeffmo: Actually for tests like this one you don't really need a `.testconfig` file like this (the test runner assumes the absence of one means that you want to test `flow check` on the contents of the dir and compare to the `.exp` file in the dir).

For specifying `--lib lib`, you can just do this in the adjacent `.flowconfig` file:

```
[libs]
lib
```

jeffmo:
I think this is ready to go -- mind just making that one final tweak to eliminate .testconfig? Then we can merge this!

Eyyub:
It's done, sorry I'm not used to flow options already !

gabelevi:
This is awesome! Thanks a ton for building this!

PR #571, opened 2015-06-26
Add support for scoped type variables

samwgoldman:
OK, so all of my pull requests are actually requests for help in disguise.

This one is not quite done. Explained below:

---

Before this commit, type params were passed around in a map in `convert`
and related functions, but were quickly "forgotten."

This commit adds a ton of plumbing to pass around this map, accumulating
entries as scopes pile on.

I also changed the name of this variable from `map` to
`type_params_map`, for readability and because the scope of this
variable is super broad now, so it should have a specific name.

The plumbing is pretty gnarly, and hopefully this diff will cause
someone to speak out and offer a better way.

I attempted managing this data in the context, which is already passed
around, but dismissed that because it was extremely error prone because
we need to rewind the modifications as we pop stacks.

There are a few changes that I should call out specifically, because the
diff noise will make this annoying to review. (I will add inline
comments on GitHub as well.)

* In mk_signature, method signatures seemed to depend on *only* the
  f_type_params_map, but the changed implementation uses the fully
  accumulated map, i.e., `union f_type_params_map c_type_params_map`.

* In function_decl, I map Flow_js.subt over the type params map, just as
  mk_class_elements does, which is necessary (otherwise we pass BoundT
  into Flow_js.flow).

* I moved the code that merges the old map with the new map into
  mk_type_param_declaration, because that should happen everywhere, I'm,
  like, 80% sure.

* I didn't change this, but convert has a check for existentials that
  changes behavior if the type params map is empty. I don't understand
  what the significance of that is, but since that map accumulates more
  now, is this check still sufficient?

Lastly, and most importantly, type checking doesn't actually work
correctly after these changes. For example:

```javascript
var constant = function<U>(u: U) {
  return function(): U {
    return u;
  }
}
var test: number = constant(0)("")
```

leads to the type error:

```
test.js:6:20,34: U
test.js:6:11,16: number
```

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -577,7 +577,7 @@ let are_getters_and_setters_enabled () = FlowConfig.(
 (**********************************)
 
 (* converter *)
-let rec convert cx map = Ast.Type.(function
+let rec convert cx type_params_map = Ast.Type.(function
 
   | loc, Any -> AnyT.at loc
 
@@ -589,10 +589,10 @@ let rec convert cx map = Ast.Type.(function
 
   | loc, Boolean -> BoolT.at loc
 
-  | loc, Nullable t -> MaybeT (convert cx map t)
+  | loc, Nullable t -> MaybeT (convert cx type_params_map t)
 
   | loc, Union ts ->
-      let ts = List.map (convert cx map) ts in
+      let ts = List.map (convert cx type_params_map) ts in
       (* "Flatten" out any unions in this union, like
        * var a: number | (string | bool) *)
       let ts = List.map (function
@@ -602,7 +602,7 @@ let rec convert cx map = Ast.Type.(function
       UnionT (mk_reason "union type" loc, ts)
 
   | loc, Intersection ts ->
-      let ts = List.map (convert cx map) ts in
+      let ts = List.map (convert cx type_params_map) ts in
       IntersectionT (mk_reason "intersection type" loc, ts)
 
   | loc, Typeof x ->
@@ -618,7 +618,7 @@ let rec convert cx map = Ast.Type.(function
         error_type cx loc "Unexpected typeof expression")
 
   | loc, Tuple ts ->
-      let elts = List.map (convert cx map) ts in
+      let elts = List.map (convert cx type_params_map) ts in
       let reason = mk_reason "tuple type" loc in
       let element_reason = mk_reason "tuple element" loc in
       let tx =
@@ -629,7 +629,7 @@ let rec convert cx map = Ast.Type.(function
 
   | loc, Array t ->
       let r = mk_reason "array type" loc in
-      let t = convert cx map t in
+      let t = convert cx type_params_map t in
       ArrT (r, t, [])
 
   | loc, StringLiteral { StringLiteral.value; _ }  ->
@@ -657,7 +657,7 @@ let rec convert cx map = Ast.Type.(function
     let typeParameters = extract_type_param_instantiations typeParameters in
     let params = if typeParameters = []
       then None else Some typeParameters in
-    mk_nominal_type cx reason map (t, params)
+    mk_nominal_type cx reason type_params_map (t, params)
 
   (* type applications: name < params > *)
   | loc, Generic {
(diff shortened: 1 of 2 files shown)


src/typing/type_inference_js.ml
```diff
@@ -4644,15 +4644,14 @@ and mk_signature cx reason_c c_type_params_map superClass body = Ast.Class.(
...
+      (* SG: Do we really need the separate type_params_map here? *)
+      let typeparams, type_params_map =
+        mk_type_param_declarations cx type_params_map typeParameters in
 
-      let params_ret = mk_params_ret cx map
+      let params_ret = mk_params_ret cx type_params_map
         (params, defaults, rest) (body_loc body, returnType) in
       let reason_m = mk_reason (spf "method %s" name) loc in
-      let method_sig = reason_m, typeparams, f_type_params_map, params_ret in
+      let method_sig = reason_m, typeparams, type_params_map, params_ret in
```
samwgoldman: Before, `method_sig` would only have a map with the type params in `typeparams`, but after this change it will have all the type params accumulated up to this point. Did I inadvertently break something with this change?

src/typing/type_inference_js.ml
```diff
@@ -5031,11 +5033,13 @@ and function_decl id cx (reason:reason) ~async
   let save_return_exn = Abnormal.swap Abnormal.Return false in
   let save_throw_exn = Abnormal.swap Abnormal.Throw false in
   Flow_js.generate_tests cx reason typeparams (fun map_ ->
+    let type_params_map =
+      type_params_map |> SMap.map (Flow_js.subst cx map_) in
```
samwgoldman: The type params map is passed down into `mk_body` and may feed into flows, so we need to eliminate the `BoundT`s.

src/typing/type_inference_js.ml
```diff
@@ -5219,10 +5223,10 @@ and mk_type_params cx ?(map=SMap.empty) (types: Ast.Identifier.t list) =
   let typeparams, smap =
     List.fold_left mk_type_param ([], SMap.empty) types
   in
-  List.rev typeparams, smap
+  List.rev typeparams, SMap.union smap type_params_map
```
samwgoldman: Always return the old type params with the new ones mixed in. Before, some call sites would do this, but this seems cleaner, assuming it's sound.

src/typing/type_inference_js.ml
```diff
@@ -876,7 +876,8 @@ let rec convert cx map = Ast.Type.(function
     (* Do not evaluate existential type variables when map is non-empty. This
        ensures that existential type variables under a polymorphic type remain
        unevaluated until the polymorphic type is applied. *)
-    let force = SMap.is_empty map in
+    (* SG: Does this logic still make sense with scoped type variables? *)
+    let force = SMap.is_empty type_params_map in
```
samwgoldman: Not sure if this check will still serve the intended purpose with scoped type vars.

samwgoldman:
I was able to get a type error similar to the one in my commit message on current master, with none of these changes:

```javascript
function identity<U>(x: U) {
  return x;
}
var test: number = identity(0)
```
```
generics.js:2:10,10: U
generics.js:4:11,16: number
```

samwgoldman:
A different example of scoped type variables *does* work correctly.

```javascript
type Merge<T> = (a: T, b: T) => T;

// hypothetical immutable map
declare class Map<K,V> {
  (): Map<K,V>;
  insertWith(fn: Merge<V>, k: K, v: V): Map<K,V>;
}

declare function foldr<A,B>(fn: (a: A, b: B) => B, b: B, as: A[]): B;

function insertMany<K,V>(merge: Merge<V>, vs: [K,V][], m: Map<K,V>): Map<K,V> {
  function f([k,v]: [K,V], m: Map<K,V>): Map<K,V> {
    return m.insertWith(merge, k, v)
  }
  return foldr(f, m, vs)
}
```

samwgoldman:
(If you're interested in this patch, I'm happy to resolve merge conflicts. Would prefer to time it to minimize conflicts with whatever is inflight internally, of course.)

bhosmer:
Hi Sam - if you want to resolve conflicts, we can work through any misbehavior you're still seeing and get it landed. (Aside: we should be able to replace the explicitly passed maps with scope push/pops, but no reason to block this work in the meantime.)

samwgoldman:
@bhosmer OK, rebased against master and added a couple tests. :sunglasses:

samwgoldman:
The only misbehavior I'm seeing is actually unrelated to these changes, so I filed a separate issue (#656).

samwgoldman:
This fell out of sync, but it was trivial to get things back in order. Ping?

PR #575, opened 2015-06-26
Making storage indexable

runn1ng:
Note: this is not at all tested, I just copy-pasted the indexing line

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1129,6 +1129,7 @@ declare class Storage {
     clear(): void;
     removeItem(key: string): void;
     key(index: number): ?string;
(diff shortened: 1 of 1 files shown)


mroch:
this is allowed by this line of [§4.1 of the Web Storage spec](https://w3c.github.io/webstorage/#the-storage-interface), right?

> The supported property names on a Storage object are the keys of each key/value pair currently present in the list associated with the object.

runn1ng:
Yes, that's at least my reading.

Actually every object is indexable by string, at least if I understand JS right :) but yeah it's specifically written in the storage spec

PR #583, opened 2015-06-28
Fix error handling for nested unions and intersections

### LINKED ISSUE #582: Union of unions doesn't work as I expected
I am trying to create a union of union types, and it doesn't work as I expect it to.

```JavaScript
/* @flow */

type Foo = 'foo' | 'stuff'
type Bar = 'bar'

type Status = Foo | Bar

type Thingy = {
  id: string;
  status: Status
}

function doStuff(thing: Thingy): Thingy {
  thing.status = 'bar'
  return thing
}
```

Running `flow` on this gives the following error:

```bash
/Users/jergason/code/flow-test/durp.js:14:18,22: string
/Users/jergason/code/flow-test/durp.js:3:12,26: union type
```

I expected to be able to create a union out of two union types, but it appears the `Status` type doesn't pick up values from the `Bar` type. If I change the code to `thing.status = 'foo'` or `thing.status = 'stuff'` it works. If I change the types to:

```JavaScript
type Foo = 'foo'
type Bar = 'bar' | 'baz'
```

then I can assign `thing.status = 'bar'` or `thing.status = 'baz'` without problem. It looks like it is only happening when the first type in the union is also a union.

This is on flow 0.12 on Mac OS X.

samwgoldman:
Union and intersection errors are handled using a speculative match
process, where an error on one side of the match raises an exception,
which is caught by the speculative match code, which then tries the
other side of the match.

If, however, there are nested speculative matches in a given flow, the
innermost one would raise an error too early. This can happen if a type
alias is a union containing union types.

To fix, instead of using a boolean to track whether to log the error or
raise, we use an integer representing the "depth" of the speculative
match.

Before, there was some code to "flatten" unions of unions during the
convertion of annotations to types, but this doesn't work for the case
where the unions are themselves type aliases.

Unfortunately, this change affects error messages around union types,
and in my opinion, the new messages are worse and less helpful than they
used to be. Any ideas how to make this change without making error
messages worse?

Fixes #582

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -98,14 +98,14 @@ let mk_objecttype ?(flags=default_flags) dict map proto = {
 
 (**************************************************************)
 
-(* for now, we do speculative matching by setting this global.
-   it's set to true when trying the arms of intersection or
+(* For now, we do speculative matching by incrementing this global.
+   It's incremented when trying the arms of intersection or
    union types.
-   when it's true, the error-logging function throws instead
+   When it's non-zero, the error-logging function throws instead
    of logging, and the speculative harness catches.
    TODO
  *)
-let throw_on_error = ref false
+let throw_on_error = ref 0
 
 (* we keep a stack of reasons representing the operations
    taking place when flows are performed. the top op reason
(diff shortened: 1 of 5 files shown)


samwgoldman:
I have a (pretty gross rn) change building on top of this that might help improve the error message problem. It keeps track of the error message with the greatest depth, and prints that instead of the latest one in the event of an error during speculative match. Early testing is promising, but it changes a lot of existing error messages, so I need to comb through and do a sanity check.

samwgoldman:
Even I don't like this solution anymore. This is definitely a wart, though.

zgao:
@samwgoldman As this issue is blocking for a very wide range of things, I think it would be good to add this as an "experimental" option. Thoughts?

PR #597, opened 2015-07-02
Match returns null with no matches

runn1ng:
"foobar".match(/Foo/) returns null.

See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match)

> Returns
>
> array
> 
> An Array containing the matched results or null if there were no matches.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -196,7 +196,7 @@ declare class String {
     indexOf(searchString: string, position?: number): number;
     lastIndexOf(searchString: string, position?: number): number;
     localeCompare(that: string): number;
-    match(regexp: string | RegExp): Array<string>;
+    match(regexp: string | RegExp): ?Array<string>;
     startsWith(searchString: string, position?: number): boolean;
     endsWith(searchString: string, position?: number): boolean;
     repeat(count: number): string;
--- a/tests/overload/overload.exp
+++ b/tests/overload/overload.exp
@@ -7,6 +7,18 @@ overload.js:1:9,14: number
 
 overload.js:1:18,28: call of method `match`
 Error:
+overload.js:1:18,31: access of computed property/element
+Computed property/element cannot be accessed on possibly null value
+overload.js:1:18,28: null
+
+overload.js:1:18,28: call of method `match`
+Error:
+overload.js:1:18,31: access of computed property/element
+Computed property/element cannot be accessed on possibly undefined value
+overload.js:1:18,28: undefined
+
+overload.js:1:18,28: call of method `match`
+Error:
 overload.js:1:27,27: number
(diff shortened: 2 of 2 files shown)


runn1ng:
...well, all the tests failed, because `match()` result can now be null. I don't feel qualified enough to fix them.

samwgoldman:
@runn1ng Those errors aren't too bad! I think you can do it.

If you aren't familiar with the test set up, it's not too hard to explain.

If there is a directory under `test`, like `test/example`, then there will be a file called `test/example/example.exp` that includes the expected output from `flow check run against the `.js` files in `test/example`.

The test runner will run `flow check` and redirect the output to `test/example/example.out` and generate a diff between the `.exp` and the `.out` file into `test/example/example.diff`. If the diff is non-empty, the test "failed."

It is possible (and quite easy) to make a change that creates a diff, but does not represent an actual test failure. In this case, you can use patch to update the `.exp` file, like so: `patch test/example/example.ext test/example/example.diff`.

OK, for these failures, here's what I would do:

1. For each error, I would use the filename/lineno to locate the code in the test file.
2. If it looks like the return type of the match is part of the assertion, I would change the test to first do a dynamic type test for null, then proceed with the normal check. (Otherwise, just patch and you're done.)
3. This will probably change some line numbers, so run again, and if only line numbers are different, re-patch.

Unfortunately, some of the tests are not as clear as they could be about what they mean/expect, so if you are uncertain about that, feel free to paste it here and I (or someone else) will help you get unstuck!

runn1ng:
Haha, all right, let's try this.

gabelevi:
Looks like the last failing test is just because it assumes that `match()` returns an array. If you update the [test](https://github.com/facebook/flow/blob/de990dea17873393d1d91ad10255d6dc0c638a27/tests/overload/overload.js#L1) you should be set

samwgoldman:
@gabelevi Do you think we can take this? I know you can make the necessary changes during your import workflow. (I'd do it myself, but I'm on the outside looking in here.)

runn1ng:
Sorry that I left this PR rot, I don't really have time to look at it :/

runn1ng:
Well I fixed it by modifying the expected values, not the test itself. (Plus I rebased the branch on master, but that was without any problems)

Is that OK?

gabelevi:
Sweet, thanks for updating this PR! Looks good to me now!

gabelevi:
@facebook-github-bot shipit

gabelevi:
Hmm, that didn't seem to work...let's try one more time...

gabelevi:
@facebook-github-bot shipit

Issue #525, opened 2015-06-16
ES6 import of default and properties doesn't work


AgentME:
These two things work:

    import _ from 'lodash';
    import {map, filter} from 'lodash';

but flow fails to parse this which I'm pretty sure is valid:

    import _, {map, filter} from 'lodash';

jeffmo:
For reference, can you post the interface definition you're using for lodash?

AgentME:
The error happens without any interface definition files. Here's the error for reference:

```
.../foo.js:7:12,14: identifier map
Could not resolve name
```

(Not directly related, but I can't get any sort of interface definitions for external modules to work anyway: https://github.com/facebook/flow/issues/526)

avikchaudhuri:
@AgentME see my comment on that task, maybe you're setting up your libs wrong? Please report back here if fixing the setup causes this issue to go away, thanks!

AgentME:
I just found that Flow fails to parse the following too:

```javascript
/* @flow */
import 'babel/polyfill'; // just execute it without importing anything from it
```

jeffmo:
@AgentME: Nice catch! Can you open another issue for that so we can track it there? This should mostly just be parser work

AgentME:
This issue is a parse issue too (same sort of error; it had nothing to do with the definition files), though I've opened that one as https://github.com/facebook/flow/issues/574.

nmn:
+1

Same error for me. Flow fails to work in this case:
```
import _, {map, filter} from 'lodash';
```

jeffmo:
BTW I totally glosses over the fact that this was a default+named import in one single statement. Apologies for not seeing that sooner.

As @samwgoldman mentioned on the PR itself, #600 should fix both of these issues

jeffmo:
https://github.com/facebook/flow/pull/600 just landed which should fix this (and will be in the upcoming release)

Issue #574, opened 2015-06-26
ES6 import without default or members doesn't work


AgentME:
Flow fails to parse this:

```javascript
/* @flow */
import 'babel/polyfill'; // just execute it without importing anything from it
```

jeffmo:
https://github.com/facebook/flow/pull/600 just landed which should fix this (and will be in the upcoming release)

PR #600, opened 2015-07-04
Allow simultaneous named and default imports

jeffmo:
Previously Flow considered `import Foo from ...` and `import Foo, {Bar, Baz} from ...` to be mutually exclusive -- though they should not be (per spec).

This fixes Flow to properly handle `import Foo, {Bar, Baz} from ...`


src/typing/type_inference_js.ml
```diff
@@ -1220,46 +1220,53 @@ and statement_decl cx = Ast.Statement.(
...
-            ) in
-            List.iter init_specifier named_specifiers
-          | Some(ImportDeclaration.NameSpace(_, (loc, local_ident))) ->
-            let local_name = local_ident.Ast.Identifier.name in
-            let reason =
-              mk_reason (spf "%s * as %s" import_str local_name) loc
+          | None ->
+            failwith (
+              "Parser error: Non-default imports must have a set of " ^
+              "named import specifiers!"
```
samwgoldman: Isn't it possible to import for side effects only, without specifying any bindings, default or non? See #574 for an issue, and [the spec at 15.2.2; `import ModuleSpecifier`](http://www.ecma-international.org/ecma-262/6.0/#sec-imports).

samwgoldman:
Cool! This is a simple enough diff, and clearly does what it says on the tin. For tracking purposes, this fixes #525.

src/typing/type_inference_js.ml
jeffmo: Oh yea, forgot about that issue. Will fix now

nmn:
Just want to report in saying that I merged this temporarily on my local repo and it works as intended for me. Will run the full test suite and report back.

PR #616, opened 2015-07-08
[easy] Minor edits to declarations

fabiopelosin:
- Added stylesheet related declarations to `Document`
- Made optional the title in `createHTMLDocument` function

### DIFF
--- a/lib/cssom.js
+++ b/lib/cssom.js
@@ -11,10 +11,25 @@
 declare class StyleSheet {
   disabled: boolean;
   href: string;
+  media: MediaList;
   ownerNode: Node;
   parentStyleSheet: ?StyleSheet;
   title: string;
   type: string;
+}
+
+declare class StyleSheetList {
+  length: number;
+  [index: number]: StyleSheet;
+}
+
+declare class MediaList {
+  mediaText: string;
+  length: number;
+  item(index: number): ?string;
+  [index: number]: ?string;
+  deleteMedium(oldMedium: string): void;
+  appendMedium(newMedium: string): void;
 }
 
 declare class CSSStyleSheet extends StyleSheet {
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -202,6 +202,7 @@ declare class Document extends Node {
     documentElement: HTMLElement;
     implementation: DOMImplementation;
     scripts: HTMLCollection;
+    styleSheets: StyleSheetList;
     title: string;
     embeds: HTMLCollection;
     forms: HTMLCollection;
@@ -235,6 +236,7 @@ declare class Document extends Node {
     createElement(tagName: 'media'): HTMLMediaElement;
     createElement(tagName: 'script'): HTMLScriptElement;
     createElement(tagName: 'select'): HTMLSelectElement;
+    createElement(tagName: 'style'): HTMLStyleElement;
     createElement(tagName: 'textarea'): HTMLTextAreaElement;
     createElement(tagName: 'video'): HTMLVideoElement;
     createElement(tagName: string): HTMLElement;
@@ -1046,7 +1048,7 @@ declare class DOMImplementation {
     hasFeature(feature: string, version?: string): boolean;
 
     // non-standard
-    createHTMLDocument(title: string): Document;
+    createHTMLDocument(title?: string): Document;
 }
 
 declare class DocumentType extends Node {


lib/dom.js
```diff
@@ -202,6 +202,7 @@ declare class Document extends Node {
     documentElement: HTMLElement;
     implementation: DOMImplementation;
     scripts: HTMLCollection;
+    styleSheets: HTMLCollection;
```
mroch: looks like it's actually a [StyleSheetList](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheetList). would you mind creating a type for that?

lib/cssom.js
```diff
@@ -11,12 +11,22 @@
...
   type: string;
 }
 
+declare class StyleSheetList {
+    length: number;
+    [index: number]: StyleSheet;
+}
+
+declare class MediaList {
+
```
fabiopelosin: Unfortunately not yet documented on MDN: https://developer.mozilla.org/en-US/docs/Web/API/MediaList

lib/dom.js
fabiopelosin: Good call… done (I've amended the commit)

lib/cssom.js
mroch: wow, i've never seen MDN missing anything!

looks like the spec is here: http://www.w3.org/TR/DOM-Level-2-Style/stylesheets.html#StyleSheets-MediaList

you don't have to keep going down a deeper and deeper rabbit hole, though. if you want to leave it empty for now that's fine and we can come back to it later.

lib/dom.js
mroch: i don't think this change came through

lib/dom.js
```diff
@@ -54,6 +54,10 @@ declare class FileList {
 
 /* DOM */
 
+declare class DOMString extends string {
+
+}
+
```
fabiopelosin: Not sure wether the following is supported:

```
declare class DOMString: string
```

If it is, it might be more appropriare given that it is a direct mapping.

lib/dom.js
fabiopelosin: Ouch… fixed

lib/cssom.js
```diff
@@ -11,12 +11,26 @@
...
   type: string;
 }
 
+declare class StyleSheetList {
+  length: number;
+  [index: number]: StyleSheet;
+}
+
+declare class MediaList {
+  mediaText: DOMString;
```
mroch: let's just use `string` instead of `DOMString`. we do that in a lot of other places so I think it's how we're handling it (before my time, though)

fabiopelosin: I've changed it.

lib/cssom.js
```diff
@@ -11,12 +11,26 @@
...
+  length: number;
+  item(index: number): DOMString;
```
mroch: if `index` doesn't exist, this returns `null` so this should be `string | null` (not `?string` since I don't think there's any way for it to return undefined)

also, let's add an `[index: number]: string` here too, I think array lookups are supported:

pick any page with CSS (say, this one) and try this in the console:
```
document.querySelector("link[rel=stylesheet]").media = "screen"; document.styleSheets[0].media[0]
```

also supported by [cssom](http://dev.w3.org/csswg/cssom/#the-medialist-interface) and this kinda ambiguous line from [DOM](http://www.w3.org/TR/DOM-Level-2-Style/stylesheets.html#StyleSheets-MediaList) -- "The items in the MediaList are accessible via an integral index, starting from 0."

mroch:
you're fast :) updating faster than I can write comments...

lib/cssom.js
```diff
@@ -11,12 +11,26 @@
...
 
+declare class StyleSheetList {
+  length: number;
+  [index: number]: StyleSheet;
+}
+
+declare class MediaList {
+  mediaText: string;
+  length: number;
+  item(index: number): string;
```
mroch: if `index` doesn't exist, this returns `null` so this should be `string | null` (not `?string` since I don't think there's any way for it to return undefined)

also, let's add an `[index: number]: string` here too, I think array lookups are supported:

pick any page with CSS (say, this one) and try this in the console:
```
document.querySelector("link[rel=stylesheet]").media = "screen"; document.styleSheets[0].media[0]
```

also supported by [cssom](http://dev.w3.org/csswg/cssom/#the-medialist-interface) and this kinda ambiguous line from [DOM](http://www.w3.org/TR/DOM-Level-2-Style/stylesheets.html#StyleSheets-MediaList) -- "The items in the MediaList are accessible via an integral index, starting from 0."

fabiopelosin: I've also accounted for the fact that the braces return undefined if the index is not found.

```javascript
document.querySelector("link[rel=stylesheet]").media = "screen";
document.styleSheets[0].media.item(1)
// null
document.styleSheets[0].media[1]
//  undefined
```

fabiopelosin: ... The tests failed so I changed `string | null` to `?string`

mroch:
Nice, thanks! We have an odd internal review process so I'll get this merged hopefully later today.

fabiopelosin:
Awesome :smile: Thanks for the guidance!

PR #634, opened 2015-07-15
Second index can be number

runn1ng:
I am not sure if this will not break things!

Specifically, I am not sure if this will work with mapFn with only one parameter now. It should though.


mroch:
thanks!

I changed it to not be optional. `Array.from` always passes the index, so it will never be undefined, and the callback, if it takes the 2nd param, shouldn't have to worry about it being undefined. flow allows a function to be called with extra params for exactly this use case.

runn1ng:
Thanks for accepting, one less annoying "non-error" I will see in the flow output :)

PR #636, opened 2015-07-16
Add ES2015 String method includes

rjn945:
Per the specification at http://www.ecma-international.org/ecma-262/6.0/#sec-string.prototype.includes

I'm not certain whether the first argument should accept a RegExp. The specification says: 

> "Throwing an exception if the first argument is a RegExp is specified in order to allow future editions to define extensions that allow such argument values."

So using a RegExp won't work today, but allowing that argument type is forward-compatible, so I've included it.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -170,7 +170,7 @@ declare class String {
     charCodeAt(index: number): number;
     concat(...strings: Array<string>): string;
     contains(substr: string): boolean;
-    includes(searchString: string | RegExp, position?: number): boolean;
+    includes(searchString: string, position?: number): boolean;
     indexOf(searchString: string, position?: number): number;
     lastIndexOf(searchString: string, position?: number): number;
     localeCompare(that: string): number;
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks for the contribution!

The purpose of the type checker is to prevent runtime errors. Since a regexp argument currently throws, I think we should catch that use case as an static error, as well.

rjn945:
@samwgoldman Okay, I've removed the RegExp from the accepted types and I've signed the CLA.

samwgoldman:
@rjn945 Cool! One annoying thing about our test setup: adding a line to a flowlib file changes the expected output files, which include line numbers. This change does cause a bunch of false negatives in our tests.

Do you think you could make the necessary changes to bring this PR into a passing state? In order to fix the tests, you need to change the `test/xxx/xxx.exp` with the updated line numbers. The easiest way to do this is using the `diff` command, like so: `diff test/xxx/xxx.{exp,diff}`, but you can also change them by hand.

Thanks again!

PR #648, opened 2015-07-20
No underscore munge CLI and config options

popham:
Adds a `--no-underscore-munge` flag to the CLI.  Adds a `munge_underscores` configuration option.


jeffmo:
Awesome!

Honestly, though, I'd rather we make this opt-in than opt-out. This is a non-standard FBism with very little chances of becoming standard as-is (it's far more likely to standardize with both different syntax and possibly even different semantics).

Mind switching this to `--munge-underscore-members` and `munge_underscore=false`?

popham:
Sure.  I'm already sick of typing `--whatever`.  I trust that you're familiar enough with the ecosystem to understand the implications with respect to anybody's build process.

jeffmo:
Indeed it may be a breaking change for anyone who was previously relying on this -- but most of our releases incur new (usually previously uncovered) errors anyway, so we consider them breaking changes for this purpose.

For those who hit this particular change and want to retain the previous behavior, adding `munge_underscore=true` to their .flowconfig in the process of upgrading should be pretty easy.

gabelevi:
Looks like a test needs to be updated, but otherwise this looks good to me! @jeffmo convinced me that munging should be off by default. It makes sense to me because

A) Babel has it off by default
B) JSTransform has it on by default but shouldn't
C) In the future, JS engines will support ES6 classes and won't munge by default.

src/commands/singleCommand.ml
```diff
@@ -66,6 +68,9 @@ let main all weak debug verbose verbose_indent json profile quiet module_
   let config_root = CommandUtils.guess_root (Some(root)) in
```
gabelevi: Ahhh....I think the assert you're hitting is due to this. One of the things that `Path.make` does is `realpath`'s the path. So if you're using symlinks and stuff, maybe this `config_root` is different than the `opt_root` below? Then you're hitting an assert when the config things you're using two different roots?

That being said, I can't seem to repro myself :/

popham: I clobbered all of my symlinks and replaced them with deep copies, and the assert persists.  My environment is a Chromebook with Ubuntu running in a chroot--I'm betting that my environment's the problem.  I'm going to quit chasing this one and hope for the best.

tests/config_module_no_underscore_munge/.flowconfig
```diff
@@ -0,0 +1,8 @@
+[ignore]
```
jeffmo: Should probably rename this test dir to `config_munge_underscores` now

tests/config_module_no_underscore_munge/chain.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+}
+A._sProperty = 48;
+
+class B extends A {
+  _property1: string;
+  static _sProperty: string;
+
+  constructor() {
+    //super();
```
jeffmo: Why commented? It's actually a bug that Flow allows this (which should hopefully be fixed soon...but might as well anticipate the fix here)

gabelevi:
Alright! This is looking good! Merging!

PR #657, opened 2015-07-23
Support `new someInstance.constructor(...)`

popham:
ES6 class instances have property `constructor` non-writable.  This PR maps `constructor` lookups to the class so that static access and construction can follow.  Currently, `new someInstance.constructor(...)` fails.

I think that ES5 allows users to clobber `constructor`--if so, then that's a potential problem area.  Also, I don't know how well the tests cover the interaction between ES5 newable functions and ES6 classes (and ES6 newable functions).  Let me know if I need more tests.

I came upon this while trying:

```js
function newC<T: K | L>(i: T): T {
  if (i instanceof K) {
    return new K(5); // see tests
  } else {
    return new L("some string"); //see tests
  }
}
```

This fails (as it should) with an obtuse error message that comes from the `bot` case of `generate_tests`.


jeffmo:
I was just poking through the code to suggest that you implement this inside type_inference_js.ml instead when I realized that code already seems to do what I was going to suggest...

So i tried it out and this seems to work for me now:

https://gist.github.com/jeffmo/bb24d24aef322fbbc610

Maybe this was fixed in the last 2 weeks sometime?
(I'll wait for you to confirm it's working before we close out)

popham:
Just barely different: https://gist.github.com/popham/0e833b94da820f07ad42 (the `new` misses the ctor in yours).

Confirmed this works.  Thanks.

PR #663, opened 2015-07-24
Document .flowconfig [options]

mroch:


docs/02-advanced-configuration.md
```diff
@@ -42,7 +42,45 @@ specified [declarations](http://flowtype.org/docs/declarations.html) when type c
...
+  ```
+  
+  will match a comment like this:
+  
+  ```
+  // $FlowFixMe: suppressing this error until we can refactor
+  var x : string = 123;
+  ```
+  
+  and suppress the error.
```
jeffmo: It's worth mentioning in here somewhere that it will also cause an error if there is not actually a Flow error on the following line

PR #670, opened 2015-07-28
Colorize test output

mroch:
screen shot 2015-07-28 at 3 03 31 pm


samwgoldman:
Oooo... pretty. :heart_eyes:

Issue #478, opened 2015-05-29
Interface declarations and other exotic syntax


motiz88:
Could someone comment on the current status and maturity of the following annotation language constructs, which don't seem to be documented?
1. `interface`, `implements`, and interface (single|multiple) inheritance - As in [lib/core.js](https://github.com/facebook/flow/blob/master/lib/core.js#L288) and [elsewhere](https://github.com/facebook/flow/search?utf8=%E2%9C%93&q=interface).
2. [Indexed properties?](https://github.com/facebook/flow/blob/master/lib/dom.js#L52) (e.g. `[index: number]: File`)
3. [(Static) Function call operator overloading?](https://github.com/facebook/flow/blob/master/lib/core.js#L161) (e.g. `static (...values:Array<any>): Array<any>`)

I am working on a WebIDL to Flow converter motivated by #477 (with some relevant discussion in #165 and #140). The basics work, but for a robust conversion I could really use some pointers as to the semantics of the above (and whether it's wise to rely on them at all).

For example, am I correct in assuming that an `interface`, like `type` and **unlike** `class`, does not result in its identifier added as a property on the global scope? This interpretation is analogous to the effect of [`[NoInterfaceObject]`](https://heycam.github.io/webidl/#NoInterfaceObject) on a WebIDL interface.

samwgoldman:
Cool! I'm excited to see the result of this work.

You are correct that interfaces are "compiled away" and don't result in any runtime bindings on their own.

`implements` is not implemented currently (no pun intended). I'm not sure if anyone is working on it. Can you share some IDL that you would like to translate using it? Maybe we can find alternatives, or possibly try to get it working.

Indexed properties might have some issues surrounding them, but I think it's safe to use them assuming they work properly. Any bugs there we should fix, but I don't think it should block what you're working on.

What do you mean exactly by operator overloading as it relates to static function calls? Overloading in flow doesn't quite work as you might expect coming from TypeScript, but I don't think anyone has spent a lot of time working out the details there. Again, an example of what you might like to generate given IDL would help here.

Overall, I'd recommend that you not let the maturity of features in flow block you from getting this important work done. If anything, your efforts will inspire us to get these things in working order! Thanks!

motiz88:
Thanks for the encouragement, it means a lot :)

---

Re: Interfaces being "compiled away" - so is this a bug?
```javascript
interface I {}
type T = {};
new I();	// <--- No error!
new T();	// Identifier T: could not resolve name
```

---

[...]

PR #672, opened 2015-07-30
Stupid simple interfaces

popham:
This is a stub demonstrating an extension of Flow's object type notation to class types.  Basically object types, but with statics allowed, e.g.

```
type BI = class {
  bMethod1(): number;
  static bsMethod1(): number;
};
```

These are cute in that intersection provides extension semantics.

Is this worth the time to polish, or will a completed implementation get rejected in favor of some other syntax and/or semantics?

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -2470,7 +2470,7 @@ let rec __flow cx (l, u) trace =
          _;
        }))
       ->
-      structural_subtype cx trace l (reason1, super, fields_tmap, methods_tmap)
+      structural_subtype cx trace l (super, fields_tmap, methods_tmap)
 
     (********************************************************)
     (* runtime types derive static types through annotation *)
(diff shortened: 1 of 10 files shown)


src/typing/flow_js.ml
```diff
@@ -1481,13 +1481,15 @@ let rec __flow cx (l, u) trace =
     (* The sink component of an annotation constrains values flowing
        into the annotated site. *)
 
-    | _, AnnotT (sink_t, _) ->
+    | l, AnnotT (sink_t, _)
+    | ShiftT(l), ShiftT(AnnotT (sink_t, _)) ->
```
popham: This and a few other sites have been defering to the generic `ShiftT~>ShiftT => unify` case.  My operating assumption is that Flow has been using the `Node.utils` model of inheritence (no prototype chaining on the classes themselves, just the instances), so everything tends to work (as long as you don't try using `Function` methods of a class inside the body of a function, for instance).  If I'm trampling on some important invariants here, I'd appreciate some feedback.

avikchaudhuri:
@popham I'm working on adding proper interface support to Flow.

popham:
@avikchaudhuri Yeah, I remember from https://github.com/facebook/flow/issues/478#issuecomment-113314880.  That's why I asked if a completed implementation would be rejected.  I assume by your response that yes, a completed implementation will be rejected.

Is the advantage of "proper" interfaces over "stupid simple" ones in error message sanity?  I've fallen in love with this syntax and the idea of intersections yielding interface refinements.  I feel it fits very naturally into the syntax of Flow, and I'm a little weary of using the `interface` word throughout my code.

The only challenge I expected was rebinding type parameters across an intersection, e.g. `type A<T> = B<T> & class C {...}`.  Is there any chance of someday desugaring this syntax (I imagine with `type A<T> = B<T> & interface C {...}`, `type A<T> = B & C<T>`, etc.)?  It seems like the natural analogue to class expressions versus class declarations.

PR #674, opened 2015-07-30
Type definitions for screen and navigator

nmn:
Type definitions for screen and navigator objects. Properties added are a union of the properties from different browsers. Properties that are not found in all browsers has a `?` at the end. This will essentially make you do existential check before using browser-specific APIs.

browsers tested: the latest versions of Chrome, Safari, Firefox

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -9,9 +9,72 @@
  */
 /* BOM */
 
-declare var screen: any;
+declare class Screen {
+    availHeight: number;
+    availLeft: number;
+    availop: number;
+    availWidth: number;
+    colorDepth: number;
+    height: number;
+    left: number;
+    mozOrientation?: string;
+    onmozorientationchange?: any;
+    orientation?: {
+        angle: number;
+        onchange: any;
+        type: string;
+    };
+    pixelDepth: number;
+    top: number;
+    width: number;
+    mozLockOrientation?: Function;
+    mozUnlockOrientation?: Function;
+}
+
+declare var screen: Screen;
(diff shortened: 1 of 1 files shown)


nmn:
*Question*;
instead of using class definitions, perhaps it makes sense to define the types as Objects?
`declare var navigator: {...}`

This way, we can add `[key: string]: any` as one of the props, to account for props we may not have covered? That is a good way to provide autocomplete with little error checking.

The stricter types defined now, may run into limitations when someone uses a future property that is not covered, but it is much more likely to catch stupid spelling errors.

gabelevi:
I think you're right to model them as class instances, since it looks like you should be able to do things like `screen instanceof Screen`

nmn:
can you use `instanceof` keyword in flow? I haven't seen it in documentation, and I'm working on documentation updates, so would love to know.

gabelevi:
Sure! It's a useful way of checking at runtime what something is! Here's an example: http://tryflow.org/#82c78ec0219f4234b91aa05cc2a3d4bd

In general, Flow does try to support all of JavaScript. Sure, we might not have gotten around to some of the less common or harder-to-type parts, but we're trying!

nmn:
Oh, I was wondering if `instanceof` was something flow used in annotations. Like how it uses `typeof` in annotations.

But this example helps a lot. I see how `if(a instance b)` can help flow refine the type of a. Thanks, this really clears up the value of using classes to define types.

gabelevi:
Alright, it's merged! Thanks for the contribution!

PR #675, opened 2015-07-30
add innerText to Node, with a ?

nmn:
innerText may be non-standard, but it is supported by all non-Firefox browsers and so it's relatively well supported.
It also has some unique benefits over textContent.

I've added it to the Node Class with a `?` so you can't use it without checking for it first.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -11,7 +11,7 @@
 
 declare class Blob {
     constructor(blobParts?: Array<any>, options?: {
-        type?: string;b
+        type?: string;
         endings?: string;
     }): void;
     type: string;
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -11,7 +11,7 @@
 
 declare class Blob {
     constructor(blobParts?: Array<any>, options?: {
-        type?: string;
+        type?: string;b
         endings?: string;
     }): void;
     type: string;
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -364,6 +366,7 @@ declare class Element extends Node {
     hasAttribute(name: string): boolean;
     hasAttributeNS(namespaceURI: string, localName: string): boolean;
     insertAdjacentHTML(position: string, text: string): void;
+    insertAdjacentElement?: (position: ('beforebgin' | 'afterbegin' | 'beforeend' | 'afterend'), node: Node) => void;
```
mroch: `beforebegin`

lib/dom.js
```diff
@@ -344,8 +344,10 @@ declare class Element extends Node {
     clientWidth: number;
     id: string;
     innerHTML: string;
+    innerText?: string;
```
mroch: seems reasonable to me, since it has the `?`... thoughts, @jeffmo?

jeffmo: Agreed, seems legit. I like that it'll cause you to check for it first.

lib/dom.js
nmn: I fixed this in a new commit.

PR #686, opened 2015-08-01
Flat module system (.flowconfig root~baseUrl)

popham:
This is a trivial module system intended for use with ES6 modules.  Dump all of your libraries in a directory with a .flowconfig, and you can import with absolute paths.  There's no configuration and/or magic to facilitate single word imports, e.g. `import "someLib";` will not look for `someLib/someMagicFile`, but it would find `someLib.js` in the .flowconfig root.

This admits paths aliasing by symlinks within the .flowconfig root.

### DIFF
--- a/src/commands/serverCommands.ml
+++ b/src/commands/serverCommands.ml
@@ -116,7 +116,8 @@ module OptionParser(Config : CONFIG) = struct
     let flowconfig = FlowConfig.get root in
     let opt_module = FlowConfig.(match flowconfig.options.moduleSystem with
     | Node -> "node"
-    | Haste -> "haste") in
+    | Haste -> "haste"
+    | Flat -> "flat") in
     let opt_libs = FlowConfig.(match lib with
     | None -> flowconfig.libs
     | Some libs ->
--- a/src/common/flowConfig.ml
+++ b/src/common/flowConfig.ml
@@ -15,7 +15,7 @@ module Json = Hh_json
 
 let version = "0.16.0"
 
-type moduleSystem = Node | Haste
+type moduleSystem = Node | Haste | Flat
 
 type experimental_feature_mode =
   | EXPERIMENTAL_IGNORE
@@ -118,6 +118,7 @@ end = struct
     in let module_system = function
       | Node -> "node"
       | Haste -> "haste"
+      | Flat -> "flat"
 
     in fun o config ->
       let options = config.options in
@@ -459,7 +460,7 @@ let options_parser = OptionsParser.configure [
 
   ("module.system", OptionsParser.({
     flags = [];
-    _parser = enum ["node", Node; "haste", Haste] (fun opts (_, moduleSystem) ->
+    _parser = enum ["node", Node; "haste", Haste; "flat", Flat] (fun opts (_, moduleSystem) ->
       {opts with moduleSystem}
     );
   }));
--- a/src/common/flowConfig.mli
+++ b/src/common/flowConfig.mli
@@ -7,7 +7,7 @@
  * of patent rights can be found in the PATENTS file in the same directory.
  *
  *)
-type moduleSystem = Node | Haste
+type moduleSystem = Node | Haste | Flat
 
 type experimental_feature_mode =
   | EXPERIMENTAL_IGNORE
--- a/src/typing/module_js.ml
+++ b/src/typing/module_js.ml
@@ -242,8 +242,22 @@ and file_exists path =
     in SSet.mem (Filename.basename path) files
   )
 
+(!!)
 let resolve_symlinks path =
   Path.to_string (Path.make path)
+(!!)
+let relative r =
+  Str.string_match Files_js.dir_sep r 0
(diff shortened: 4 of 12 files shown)


gabelevi:
I'm a bit of a module system n00b. With what would you use this flat module system? The ES2015 spec leaves the resolution algorithm unspecified, so this must be following the module resolution system for something, right?

popham:
Absolute paths are treated as based at the .flowconfig parent directory.  Since that directory cannot contain two packages with identical names, resolution is unambiguous.

The AMD spec has a [baseUrl](http://requirejs.org/docs/api.html#config-baseUrl) that works similarly to this.  I would like the Flat module system to evolve with ES6 modules, so I haven't actually used the term `baseUrl` anywhere.
* I expect a `baseUrl` analogue in ES6, so eventually there would be a configuration option to point somewhere other than the .flowconfig root.
* I'm a little worried about `.js`, etc. suffixes on imported modules because I've seen an MDN article with imports that include the `.js`, but since AMD and CommonJS exclude the suffixes, I expect no suffixes on ES6 imports.

I'm getting ready to publish a library updated to ES6.  For `npm` publication I was thinking of adding a `package.json` entry like `flowRoot: "someDirectoryContainingFlowAnnotatedVersion"`.  An `npm`ish tool could leverage the `npm` infrastructure to symlink the `someDirec...` to a project root, e.g.

Initial state:
* /project/.flowconfig
* /project/package.json

`npmish install --save myLib`

Post-install state:
* /project/.flowconfig
* /project/package.json
* /project/node_modules/myLib/package.json (includes `flowRoot: "flow"`)
* /project/node_modules/myLib/flow/...
* /project/node_modules/myLib/...
* /project/myLib (symlink to `node_modules/myLib/flow/`)

Similarly, a `bower`ish tool could leverage github infrastructure, but I don't know how the `bower` people feel about ad hoc entries in their config file.

jeffmo:
In isolation this seems like a fine resolution algorithm, but in terms of building a tool like what you describe and using this in reality: I think you'll run into some tricky challenges there.

The first that I can think of is that you'll have to do more than just symlink into a `node_modules` dir: You'll also need to re-write the internal require/import path strings in those files (and each of their recursive dependencies) to map the node resolution algorithm back to this one for the code to work at runtime. After dealing with each of the corner cases of the Node algorithm (and there certainly are plenty :p ) the tool itself will probably just end up as a re-write of the "Node" system that's built into flow now -- except it'll exist inside a pre-processor tool.

Anyway, if you can work through the challenges of making this work in practice and provide a good set of tooling/docs for others who might want to use it then I don't have an issue with taking this; But I'd rather we have a good understanding of the practicalities before merging upstream and maintaining this.

PS: You won't hurt our feelings if you want to have your fork track upstream while you experiment with this :)

PPS: If you have any ideas of how we might abstract the resolution system better to make it easier to implement custom resolvers and experiment like this without a fork, I'm all ears!

popham:
> The first that I can think of is that you'll have to do more than just symlink into a node_modules dir: You'll also need to re-write the internal require/import path strings in those files (and each of their recursive dependencies) to map the node resolution algorithm back to this one for the code to work at runtime. After dealing with each of the corner cases of the Node algorithm (and there certainly are plenty :p ) the tool itself will probably just end up as a re-write of the "Node" system that's built into flow now -- except it'll exist inside a pre-processor tool.

I'm thinking that the Flat schematic is no smarter than AMD for version resolution.  The `flowRoot` for a package provides a parallel implementation to the Node implementation that uses ES6 importing and Flow annotations, but no `require`ing; the whole of npmjs is incompatible since no package has a `flowRoot`.  I'm no `bower` expert, but my impression is that they suggest a package version that is compatible with all package users if such a version exists; this is how my `npmish` would need to cope with conflicts.  `bower`'s behavior is probably a good analogue to what I had in mind (for instance, I do know that `bower` provides a flat directory of dependencies instead of the nested dependencies of Node).  This is probably incompatible with Flow's Browserify bent.  And anything evocative of "npm" would be a terrible name, since `require` is out of bounds.

This PR is just a branch that I'm developing in parallel with the needs of an ES6 project that I'm trying to integrate with Flow (I've had trouble under the current state of Flow making this work prettily).  The package management aspect is speculation on how the project layout could be integrated with existing infrastructure, but I'm content with manually symlinking into the root directory in the short-term.

Hopefully it evolves to couple nicely with a package manager.  It looks like `bower` exposes an API that I may be able to exploit.

I take by the P.S. that I'm okay manners-wise leaving this open?

On the P.P.S., a quick search for `ocaml dynamic linkage` yields [Dynlink](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Dynlink.html).  I've got a mountain of stuff to cover before this blocks, though (not to mention a few external blocks :().

jeffmo:
Let's close this out for now so we can keep our PR list down to things we're actively pursuing.

If you manage to get some good mileage and community support from this, it should be easy to open and start up the discussion again.

PR #697, opened 2015-08-05
Update "Advanced configuration" docs

kylehg:
Attempting to write a `.flowconfig` based on existing docs will give an
"Unable to parse line" error. The updated docs reflect the formatting permitted
by `flowConfig.ml:384`


gabelevi:
Thanks for fixing this up!

PR #700, opened 2015-08-05
Create binding for named class expressions

### LINKED ISSUE #699: class name visibility in class expressions
I believe Flow (version 0.13.1) incorrectly handles visibility of class name in class expression. 

The following code fails with `error identifier Foo Could not resolve name`

```javascript
var Bar = class Foo {
    static factory(): Foo {
        return new Foo()
    }
};
```

It seems that the name `Foo` should be visible inside the class body (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/class).

If I use class statement instead of class expression, everything is Ok:

```javascript
class Foo {
    static factory(): Foo {
        return new Foo()
    }
}
```

samwgoldman:
If a class expression specifies a name, that name serves as a binding to
the class itself within the expression only.

I initially attempted to do this by converting the `Ast.Class` into a
`Ast.Statement.ClassDeclaration`, then running that through
`statement_decl` and `toplevels`, but in order to return the type from
the `expression` function, I needed to read the entry from the
environment, which doesn't work if the class expr is anonymous.

So, instead, I copied in the relevant parts of `statement_decl` and
`toplevels` into this part of the code. This kind of manual env
management inside of `expression` isn't unprecedented, but I'm open to
suggestions to improve this.

Fixes #699

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -3436,7 +3436,6 @@ and expression_ ~is_cond cx loc e = Ast.Expression.(match e with
           Env_js.push_env cx scope;
           let class_t = mk_class cx loc reason c in
           Env_js.pop_env ();
-          Hashtbl.replace cx.type_table loc class_t;
           Flow_js.flow cx (class_t, tvar);
           class_t;
       | None -> mk_class cx loc reason c)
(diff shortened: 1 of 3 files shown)


gabelevi:
Sweet, this looks good to me! Nice improvement!

PR #705, opened 2015-08-06
fix typo in 01-typeof.md

matklad:


gabelevi:
Whoops! Thanks for fixing!

PR #711, opened 2015-08-10
Improve docs for dynamic type tests

samwgoldman:
* code examples galore
* where type tests can be used
* (very nearly) exhaustive list of supported tests
* caveat section about havocked refinements

### DIFF
--- a/docs/01-dynamic-type-tests.md
+++ b/docs/01-dynamic-type-tests.md
@@ -10,34 +10,255 @@ next: primitives.html
 In addition to `null` checks, there are several other dynamic type tests
 (*predicates*) on local variables that Flow recognizes and uses to refine
 types. Refining a type with a predicate means narrowing the original type with
-the type satisfied by values satisfying the predicate. Some of these
-predicates are:
+the type satisfied by values satisfying the predicate.
 
-- `typeof x == 'string'`, `typeof x == 'number'`, and so on: narrows the type
-  of `x` with `string`, `number`, and so on;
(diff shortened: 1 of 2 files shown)


docs/01-dynamic-type-tests.md
```diff
@@ -10,34 +10,255 @@ next: primitives.html
...
+    return y; // error: Flow doesn't know that
+  } else {
+    return "default";
+  }
+}
+
+function bar(x: ?string): string {
+  function baz() { /* this doesn't mutate x */ }
+  if (x) {
+    bar();
```
mroch: should be `baz`

docs/01-dynamic-type-tests.md
```diff
@@ -10,34 +10,255 @@ next: primitives.html
...
+  }
+}
+
+function baz(x: boolean): string {
+  if (x) {
+    return "true";
+  } else {
+    return "false";
+  }
+}
```
samwgoldman: The above examples (foo, bar, and baz) don't actually make good use of the refinement. Flow does actually "know" that x is not empty, or non-zero, or true, but I had a lot of trouble coming up with examples that actually used that knowledge in a useful way.

mroch: knowing a boolean is truthy is useful for nested logical operators like `x && "truthy" || "falsy"`. without knowing that, it'd think it could return `boolean | string` (false || "falsy", which is clearly just "falsy" to a human). but I agree, it's hard to find a user-level example without getting into internals that should appear to Just Work.

Issue #584, opened 2015-06-29
Fix paths on http://facebook.github.io/flow/


mroch:
We should use relative paths to static resources so that it works from http://flowtype.org/ and also http://facebook.github.io/flow/. The latter URL works over HTTPS, so it'd be nice to get it working.

mroch:
links between pages should be relative too

mroch:
that site is no longer being updated

PR #713, opened 2015-08-10
Fix broken link

samwgoldman:
My bad :neutral_face:

### DIFF
--- a/docs/01-dynamic-type-tests.md
+++ b/docs/01-dynamic-type-tests.md
@@ -44,7 +44,7 @@ function bar(x: ?string): string {
 }
 {% endhighlight %}
 
-Read more about <a href="docs/nullable-types.html">Maybe Types</a>.
+Read more about <a href="/docs/nullable-types.html">Maybe Types</a>.
 
 ### Truthiness and Existance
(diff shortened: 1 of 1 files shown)


docs/01-dynamic-type-tests.md
```diff
@@ -44,7 +44,7 @@ function bar(x: ?string): string {
 }
 {% endhighlight %}
 
-Read more about <a href="docs/nullable-types.html">Maybe Types</a>.
+Read more about <a href="/docs/nullable-types.html">Maybe Types</a>.
```
mroch: let's do `../docs/nullable-types.html` so that it also works on facebook.github.io/flow/

samwgoldman: That site is pretty broken right now. There are a bunch of other absolute hrefs on this site already. Can we merge this as-is and I'll got back in and fix all the hrefs in a separate PR?

mroch: (this'll matter more after #584)

mroch:
yup, i'll just change #584 to include that too

PR #715, opened 2015-08-12
Upgrade year

rafaelstz:


rafaelstz:
Thank's! Ready :+1: :smile:

mroch:
thanks! I checked with our legal team, and can we do "2014-present" instead?

rafaelstz:
That's?

mroch:
that'll work, thanks!

PR #717, opened 2015-08-13
Fix some DOM definition issues

AgentME:
I'm loving flow, but there's some issues with the DOM definitions that have been driving me crazy!

* document.location.origin was missing.
* All MutationObserverInit properties were marked as required.
* Node.insertBefore didn't allow null for second argument.
* HTMLCollection was declared as two different classes.
* Node.remove was missing. (It conflicted with HTMLSelectElement.remove, so I commented that one out. The remove method can always be called with extra arguments.)

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -90,6 +90,7 @@ declare var history: History;
 
 declare class Location {
     hash: string;
+    origin: string;
     protocol: string;
     search: string;
     href: string;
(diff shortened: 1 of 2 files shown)


lib/bom.js
```diff
@@ -130,13 +131,13 @@ declare class MutationRecord {
...
     constructor(callback: (arr: Array<MutationRecord>, observer: MutationObserver)=>any): void;
     observe(target: Node, options: {
-        childList: boolean;
-        attributes: boolean;
-        characterData: boolean;
-        subtree: boolean;
-        attributeOldValue: boolean;
-        characterDataOldValue: boolean;
-        attributeFilter: Array<string>;
+        childList?: ?boolean;
```
mroch: let's make all of these optional but not nullable (`childList?: boolean`). even though the API might not care, i don't think we need to be that loose.

lib/dom.js
```diff
@@ -937,7 +931,7 @@ declare class HTMLSelectElement extends HTMLElement {
   value: string;
 
   add(element: HTMLElement, before?: HTMLElement): void;
-  remove(index: number): void;
+  //remove(index: number): void;
```
mroch: add this back but make index optional (`index?: number`), per https://html.spec.whatwg.org/multipage/forms.html#dom-select-remove

what a crazy API. >:(

lib/dom.js
```diff
@@ -646,13 +647,6 @@ declare class HTMLCanvasElement extends HTMLElement {
     toBlob(callback: (v: File) => void, type?: string, ...args: any): void;
 }
 
-declare class HTMLCollection {
```
mroch: hah, nice find

PR #734, opened 2015-08-21
Allow property-wise subtyping for destructured bindings

### LINKED ISSUE #729: Property-wise subtyping is safe in destructuring assignments
```javascript
/* @flow */

declare var obj: { foo: string };
var { foo } : { foo: string | number } = obj;
```

Actual output:
```
test.js|4 col 31 error|  number
|| This type is incompatible with
test.js|3 col 25 error|  string
|| 
```

Expected: No errors

samwgoldman:
Generally, property-wise subtyping is not safe because of aliasing.

```javascript
type X = { foo: string, bar: number };
type Y = { foo: string, bar?: number };
var x: X = { foo: "foo", bar: 0 };
var y: Y = x; // Unsound, even though X < Y
y.bar = undefined; // Here's why: (x : X) no longer holds
```

However, with destructured bindings, you don't end up with a reference
to the object on the RHS, so this kind of error isn't possible.

The reason-mangling is a bit unfortunate here. `mk_type_annotation` does
take a reason parameter, but discards it when there is a type
annotation and uses the reason generated from `convert` instead. For
objects, we end up with reason description of `"object type"`.

This commit adds a prefix to the generated reason, which is not too
intrusive, so we have `"destructured object type"` instead. This makes
it possible to identify the destructured binding in order to whitelist
the pattern for subtyping.

Fixes #729


samwgoldman:
Let me know if I should rebase this change—it should apply trivially, in spite of conflicts.

bhosmer:
Unfortunately the destructuring only strips the top-level reference. Consider:

    var x: {p: {q: number}} = {p: {q: 0}};
    var {p}: {p: {q: number|string}} = x;
    p.q = "hey";

samwgoldman:
Ack. You're right. We need const types (not bindings)!

bhosmer:
Yup, exactly! Cried these same tears recently over read-only params, actually. Const types are the shining path. :)

PR #735, opened 2015-08-21
Fix #730 + typo Uninitialized rather than Unintialized

### LINKED ISSUE #730: Destructuring assignments should always have an initializer.
```js
/* @flow */
var {foo};
var {bar} : {bar : number};
```

Flow do not throw an error here, it seems it should.

I think it comes from [here](https://github.com/facebook/flow/blob/master/src/typing/type_inference_js.ml#L2801).

If it's not urgent, I would be happy to fix it soon.

Eyyub:
I have done the same changes in `parser_dts` because it seems +/- sync with `parser_flow` but I'm not sure at 100%.

Also in `destructuring.js` I added init for 2 decl. (line 10 & 11) to keep the wanted behavior (ie. `unsupported destructuring`).

### DIFF
--- a/src/dts/parser_dts.ml
+++ b/src/dts/parser_dts.ml
@@ -1087,7 +1087,11 @@ end = struct
         then begin
           Expect.token env T_ASSIGN;
           Some (Parse.assignment env)
-        end else None in
+        end else Ast.Pattern.(
+          match id with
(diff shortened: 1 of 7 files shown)


gabelevi:
Sorry for the delay! This looks great, thanks for fixing this up! Merging...

Issue #125, opened 2014-11-25
unreachable code for function declarations


timruffles:
Function declarations are not flow friendly. Since they're hoisted to the top of the scope people use them to keep helpers separate from what the function is 'doing', with access to its params e.g:

```javascript
/* @flow */

function foo(a) {

  return helper() + 1;

  function helper() {
    return a * a;
  }
}
```

gives

```sh
$ flow check
./test.js:5:3,22:
unreachable code
```

tommilimmot:
Would be nice to have this supported!

XrXr:
+1

runn1ng:
+1

ghengeveld:
+1

davidmason:
I am happy to have a try at implementing this in my spare time.

@avikchaudhuri Can you point me to the right area in the code that I should be looking if I want to make this sort of change?

samwgoldman:
@davidmason I am pretty sure you want to look at [the code that processes statements in the type inference code](https://github.com/facebook/flow/blob/151759d6057e9a08dd9dd4efc6fdd034dda2ce33/src/typing/type_inference_js.ml#L1289).

For a bit of background, check out the [Abnormal module](https://github.com/facebook/flow/blob/151759d6057e9a08dd9dd4efc6fdd034dda2ce33/src/typing/type_inference_js.ml#L29).

A heavy handed approach would be to actually iterate through the remaining statements, and only complain about unreachable code if there are any non-`VariableDeclaration` statements.

davidmason:
@samwgoldman thanks, I'll take a look at those.

>  (like `VariableDeclaration` and `FunctionDeclaration`—there might be more that hoists).

I'll also try to check whether a `VariableDeclaration` is part of an assignment. The variable declaration is hoisted, but the assignment part of the statement will still never run so should still be flagged as unreachable code.

avikchaudhuri:
Yeah this strategy should work.

samwgoldman:
@davidmason Are you working on this?

davidmason:
@samwgoldman yes, had to study up a bit on OCaml. I made some changes, I just need to get the build working to check if it works.

I'm on Fedora 20. I did:

```
yum install ocaml elfutils-libelf-devel
```

```
[]$ ocamlc -v
The OCaml compiler, version 4.00.1
Standard library directory: /usr/lib64/ocaml

[]$ yum info ocaml
Installed Packages
Name        : ocaml
Arch        : x86_64
[...]

PR #737, opened 2015-08-22
Do not report function declarations as unreachable.

davidmason:
See #125 

This should fix the warning for hoisted functions. I'm having some trouble with types for also skipping a warning on variable declarations that have no assignment, so I will deal with that separately.

This is my first ever OCaml, and my first ML in a few years, so please let me know how I could have written it better.

I haven't added any tests yet, I've just been manually running it against a file like this:

```
/* @flow */

function foo(x, y) {
  "use strict";
  return bar(x) + baz(y);

  // should not generate warning
  function bar (ecks) {
    return x + ecks;
  }

  // should generate warning
  var baz = function (why) {
    return y + why;
  }

  // should not generate warning
  var z;

  // should generate warning
  var w = 'w';
}

foo(1, 2);
```

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -1292,15 +1292,13 @@ and toplevels cx stmts =
       if uc < List.length stmts
       then (
         let msg = "unreachable code" in
-        let rec drop n lst =
-          match (n, lst) with
+        let rec drop n lst = match (n, lst) with
           | (_, []) -> []
           | (0, l) -> l
-          | (x, h :: t) -> drop (x-1) t
+          | (x, _ :: t) -> drop (pred x) t
         in
         let trailing = drop uc stmts in
-        trailing |> List.iter Ast.Statement.(fun stmt ->
-          match stmt with
+        trailing |> List.iter Ast.Statement.(function
           (* function declarations are hoisted, so not unreachable *)
           | (_, FunctionDeclaration _ ) -> ()
           (* TODO also skip variable declarations if they do not have
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -1302,15 +1302,7 @@ and toplevels cx stmts =
         trailing |> List.iter Ast.Statement.(fun stmt ->
           match stmt with
           (* function declarations are hoisted, so not unreachable *)
-          | (_, FunctionDeclaration {
-                FunctionDeclaration.id;
-                params; defaults; rest;
-                body;
-                returnType;
-                typeParameters;
-                async;
-                _
-              } ) -> ()
+          | (_, FunctionDeclaration _ ) -> ()
           (* TODO also skip variable declarations if they do not have
                   any assignments *)
           | (loc, _) -> Flow_js.add_warning cx [mk_reason "" loc, msg]
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -1292,8 +1292,29 @@ and toplevels cx stmts =
       if uc < List.length stmts
       then (
         let msg = "unreachable code" in
-        let (loc, _) = List.nth stmts uc in
-        Flow_js.add_warning cx [mk_reason "" loc, msg]
+        let rec drop n lst =
+          match (n, lst) with
+          | (_, []) -> []
+          | (0, l) -> l
+          | (x, h :: t) -> drop (x-1) t
+        in
+        let trailing = drop uc stmts in
+        trailing |> List.iter Ast.Statement.(fun stmt ->
+          match stmt with
+          (* function declarations are hoisted, so not unreachable *)
+          | (_, FunctionDeclaration {
+                FunctionDeclaration.id;
+                params; defaults; rest;
+                body;
+                returnType;
+                typeParameters;
+                async;
+                _
+              } ) -> ()
+          (* TODO also skip variable declarations if they do not have
+                  any assignments *)
+          | (loc, _) -> Flow_js.add_warning cx [mk_reason "" loc, msg]
+        )
       );
       Abnormal.raise_exn exn
     )
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -1292,6 +1292,7 @@ and toplevels cx stmts =
       if uc < List.length stmts
       then (
         let msg = "unreachable code" in
+        let warn_unreachable loc = Flow_js.add_warning cx [mk_reason "" loc, msg] in
         let rec drop n lst = match (n, lst) with
           | (_, []) -> []
           | (0, l) -> l
@@ -1304,11 +1305,12 @@ and toplevels cx stmts =
           (* variable declarations are hoisted, but associated assignments are not,
              so skip variable declarations with no assignments.
              Note: this does not seem like a practice anyone would use *)
-          | (_, VariableDeclaration d ) when VariableDeclaration.(d.declarations |>
-              List.for_all Declarator.(function
-              | (_, { init = None }) -> true
-              | _ -> false)) -> ()
-          | (loc, _) -> Flow_js.add_warning cx [mk_reason "" loc, msg]
+          | (_, VariableDeclaration d) -> VariableDeclaration.(d.declarations |>
+              List.iter Declarator.(function
+              | (_, { init = Some (loc, _) } ) -> warn_unreachable loc
+              | _ -> ()
+            ))
+          | (loc, _) -> warn_unreachable loc
         )
       );
       Abnormal.raise_exn exn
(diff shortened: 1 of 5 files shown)


src/typing/type_inference_js.ml
```diff
@@ -1292,8 +1292,29 @@ and toplevels cx stmts =
...
+          match (n, lst) with
+          | (_, []) -> []
+          | (0, l) -> l
+          | (x, h :: t) -> drop (x-1) t
+        in
+        let trailing = drop uc stmts in
+        trailing |> List.iter Ast.Statement.(fun stmt ->
+          match stmt with
+          (* function declarations are hoisted, so not unreachable *)
+          | (_, FunctionDeclaration {
```
Eyyub: Hi, as it seems you just want to know if the statement is a `FunctionDeclaration` and do not want to do something with(correct me if I'm wrong), you might just want to write ` (_, FunctionDeclaration _) -> ()` instead of deconstruct the record `FunctionDeclaration.t`.

Hope you enjoyed OCaml :) !

samwgoldman:
This is a great start! As far as getting your test case into the automated test suite, there are a few steps but it's not so bad:

1. Create a directory, something like `tests/unreachable`
2. Put your test file in there, `tests/unreachable/unreachable.js` is fine
3. Add a `.flowconfig`, which can be empty, to that directory
4. Add a `unreachable.exp` file, which contains the expected output
5. Run `./runtests.sh bin/flow unreachable` (or just `make test`) and see the test failure

A test failure means that the actual output is different from the expected output, in the `unreachable.exp` file. In this case, the expected output is empty, and the actual output may be expected. If so, you can change the expectation to make your test pass: `mv tests/unreachable/unreachable.{out,exp}`.

src/typing/type_inference_js.ml
```diff
@@ -1292,8 +1292,29 @@ and toplevels cx stmts =
...
+                params; defaults; rest;
+                body;
+                returnType;
+                typeParameters;
+                async;
+                _
+              } ) -> ()
+          (* TODO also skip variable declarations if they do not have
+                  any assignments *)
+          | (loc, _) -> Flow_js.add_warning cx [mk_reason "" loc, msg]
```
samwgoldman: The case for variable declarations is slightly trickier, because a single `VariableDeclaration` statement can have multiple actual declarations. You want to figure out if there is a `VariableDeclaration` where any of the `declarations` have a non-None `init` member. You could use a pattern guard for this.

src/typing/type_inference_js.ml
davidmason: @Eyyub thanks, I had played around trying to figure out how to do that, but didn't quite hit the right syntax, so I copied this pattern from elsewhere.

Can you point me at some reading on pattern matching with modules/records? I still want to match a `VariableDeclaration` with a guard that checks that for all its `declarations`, `init` is `None`, but I think I need to learn a bit more to get that working without type errors.

Eyyub: With pleasure, [here](https://realworldocaml.org/v1/en/html/records.html#idm181618235488) there is an explaination about how correctly pattern match on record which is part of a module. (the syntax maybe seems ugly but there is a proposal for adding local open in pattern matching to do `Foo.{bar; toz}` instead of `{Foo.bar; toz}`, hope it will be accepted)

[ocaml.org](https://ocaml.org/learn/tutorials/) and [Real World OCaml](https://realworldocaml.org/v1/en/html/index.html) are really great places to learn OCaml, but maybe you already known these websites.

To do what you want with the `when` guard, I suggest you to use [List.for_all](http://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html#VALfor_all) with `declarations` and a predicate function.

I hope you will understand my english, don't hesite to ask me something (and come to #flowtype channel on freenode if you want).

src/typing/type_inference_js.ml
```diff
@@ -1292,8 +1292,21 @@ and toplevels cx stmts =
...
-        let (loc, _) = List.nth stmts uc in
-        Flow_js.add_warning cx [mk_reason "" loc, msg]
+        let rec drop n lst =
+          match (n, lst) with
+          | (_, []) -> []
+          | (0, l) -> l
+          | (x, h :: t) -> drop (x-1) t
+        in
+        let trailing = drop uc stmts in
+        trailing |> List.iter Ast.Statement.(fun stmt ->
```
mroch: code golf: you can do 

```
trailing |> List.iter Ast.Statement.(function
| ...
```

instead of the extra `match` line

(could do something similar with `drop` by passing `(n, lst)` instead of `n lst`)

not something to change, just thought I'd mention it since it's an interesting ocamlism

Eyyub: :+1: 

While we're at telling "good" OCaml guideline, in your third pattern branch in `drop`, you don't use `h` so you can shadow it by replacing it with `_` like `(x, _ :: t)`.

Also most OCamllers love to use `pred x` instead of `x - 1` (same thing with `succ x` and `x + 1`).

mroch: > Also most OCamllers love to use pred x instead of x - 1 (same thing with succ x and x + 1).

TIL

Eyyub: Sorry, I do not find the meaning of this acronym on internet :d

mroch: heh, sorry. "today i learned", from https://www.reddit.com/r/todayilearned/

Eyyub: Thx !

src/typing/type_inference_js.ml
davidmason: @samwgoldman Yeah, I tried to write something like:

```
| (_, VariableDeclaration { declarations }) where declarations |> for_all (fun decl ->
    match decl.init with
    | None -> true
    | _ -> false
) -> ()
```

but I had some trouble getting the syntax right and dealing with the module imports and such.

src/typing/type_inference_js.ml
davidmason: Awesome, this is all good stuff. It is so satisfying to get rid of unneeded syntax.

davidmason: > (could do something similar with drop by passing (n, lst) instead of n lst)

It seems nicer to keep it in curried form, but if there are curry/uncurry functions I could have the best of both worlds. I'll have a look into that now.

src/typing/type_inference_js.ml
samwgoldman: I'm pretty new to OCaml myself, but here's what I came up with:
```
| (_, VariableDeclaration { VariableDeclaration.declarations })
  when declarations |> List.for_all (function
    | (_, { VariableDeclaration.Declarator.init; _ }) -> Option.is_none init
  ) -> ()
```

However, I think it would be even nicer if we pointed to the `loc` of the init itself, so maybe something more like this:

```
(* variable declarations are hoisted, but initializers are not *)
| (_, VariableDeclaration { VariableDeclaration.declarations }) ->
    declarations |> List.iter VariableDeclaration.Declarator.(function
      | (_, { init = Some (init_loc, _); _ }) ->
        Flow_js.add_warning cx [mk_reason "" init_loc, msg]
      | _ -> ()
    )
```

davidmason: I got it working, with slightly different code - should have checked back here first.

> I think it would be even nicer if we pointed to the loc of the init itself

I agree, I'll add that now.

davidmason: @samwgoldman done, mine looks slightly different, but I think it will execute the same - I just had the modules attached to slightly different scopes. Happy to change it if a different way is more idiomatic/performant/clear.

samwgoldman: Looks great!

samwgoldman:
Nice work, here!

Looks like using the init locs changed an expected error in the abnormal test, but the actual output is now expected after that change.

Also, I didn't realize we had a file for unreachable expectations in the abnormal tests. You could move the tests into that file instead of having a separate unreachable test, or alternatively kill that file in the abnormal tests, or just change the expectation and leave things how they are.

davidmason:
@samwgoldman ah, that was confusing me - didn't notice it was in abnormal tests. I'll sort it out. Should have run all tests before committing, instead of just the new one.

davidmason:
@samwgoldman is there anything else to do with this before it can proceed?

samwgoldman:
@davidmason This looks good to me! I'm just a co-contributor, so we'll have to wait for someone at FB to review. This can take a bit of time, in my experience, so we might have to be a bit patient. :) Thanks for taking this on!

mroch:
> This can take a bit of time

challenge-accepted-barney-5619

samwgoldman:
I should employ reverse psychology in my own PRs... :stuck_out_tongue_winking_eye:

PR #741, opened 2015-08-23
Adds merlin file configuration

Eyyub:
Now, you will not have merlin saying there is 274 errors because of no `.merlin` :D !


frantic:
+1

PR #742, opened 2015-08-24
React.cloneElement attribute param type should be $Shape<P>

ianobermiller:
Since you will already have a valid element (flow will complain earlier) `cloneElement` should only take `$Shape<Props>`. You only pass the props you want to change to `cloneElement`.


gabelevi:
I think you're absolutely correct! Thanks for your help! Merging...

PR #747, opened 2015-08-26
Fix build caused by warn 50.

Eyyub:
The build is currently broken because of `warn 50`.

`warn 50` was introduced in ocaml 4.02.2 with the PR#149, documentation comments are since attached to parsetree, so they should be on expected places.
Flow use a lot of documentation comments but it seems it's more for having a different color in text editors ?

The travis build for 4.01.0 seems to be also broken but it's weird because `warn 50` comes from 4.02.2 :/

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,9 +1,9 @@
-<**/*.ml*>: ocaml, warn_A, warn(-4-6-29-35-44-48), warn_error_A
-<hack/utils/*.ml*>: warn(-3-27)
-<hack/heap/*.ml*>: warn(-27-34)
-<src/commands/*.ml*>: warn(-27-32-34-45)
-<src/common/*.ml*>: warn(-27-32-34-45)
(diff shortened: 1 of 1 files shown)


samwgoldman:
@bhosmer @mroch bae1b75b30a3ac5532a2b2846353407523366b3f is trying to fix this problem, but @Eyyub already fixed it!

mroch:
also weird because travis is supposed to be running with 4.02.1, not 4.02.2.

mroch:
@Eyyub, thanks for jumping on this so quickly and sorry for missing the PR (and for breaking it in the first place in d1f696005f49f138ef5aee5417304bef860a7d5a). As I understand `_tags`, it applies all of the matching rows for each file, so the `-50` only needs to be added to the wildcard at the top. Alright if I close this?

Eyyub:
Of course !

PR #750, opened 2015-08-26
Add support for non function event listeners

Gozala:
For more details see:
1. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
2. https://developer.mozilla.org/en-US/docs/Web/API/EventListener


gabelevi:
Sweet, makes sense to me! Thanks for the PR! Merging...

PR #756, opened 2015-08-28
Build hack/utils/handle_stubs.o

mroch:
See related hack diff: https://github.com/facebook/hhvm/commit/7115d5c37d4cf7fedc747f193fddc33927505fcb

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,8 @@ NATIVE_OBJECT_FILES=\
   hack/utils/nproc.o\
   hack/hhi/hhi_elf.o\
   hack/utils/get_build_id.gen.o\
-  hack/utils/get_build_id.o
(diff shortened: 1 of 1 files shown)


samwgoldman:
The build also hangs on autocomplete for me locally.

mroch:
appears to be all server commands, autocomplete is just the first test to run.

it looks like the server creates the socket and the client (at least tries to) writes the build ID to it, but the server stays blocked on the Unix.select in `serverFunctors.ml`

mroch:
blames to https://github.com/facebook/flow/commit/af9a76888eec8262f250f68d8ae535769d51128e, but i'm not sure which part

mroch:
#758 fixes the hanging

PR #758, opened 2015-08-28
Call Daemon.check_entry_point ASAP

mroch:
`Daemon.spawn` sets an environment variable and then calls `flow`, expecting to hit a call to `Daemon.check_entry_point ()` which will hijack execution to call into the "entry point" specified in the environment variable.

We weren't calling `Daemon.check_entry_point ()`. Hack calls it in `serverMain.ml`, but since they have separate client and server binaries, `serverMain.ml` gets hit. our server codepath doesn't get hit unless you call `flow server`; everything else is a client command that execs `flow server`, so we have to do this right upfront.

### DIFF
--- a/src/flow.ml
+++ b/src/flow.ml
@@ -56,6 +56,8 @@ end = struct
   let commands = ShellCommand.command :: commands
 
   let main () =
+    Daemon.check_entry_point (); (* this call might not return *)
+
     Sys.set_signal Sys.sigpipe Sys.Signal_ignore;
     let default_command = DefaultCommand.command in
     let argv = Array.to_list Sys.argv in


hack/utils/daemon.ml
```diff
@@ -173,10 +173,8 @@ let spawn
           fn)  in
   let out_fd =
     Unix.openfile out_path [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC] 0o666 in
-  let pid =
-    Unix.create_process
-      Sys.executable_name [|Sys.executable_name|]
-      null_fd out_fd out_fd in
+  let exe = Sys_utils.executable_path () in
```
mroch: this is so we use an absolute path to the same binary, instead of relying on PATH. sometimes `Sys.executable_name` is not correct, like if you do `PATH='~/flow/bin/' flow` (where the path is in single quotes so that the shell doesn't expand it)

PR #760, opened 2015-08-28
 Restore state on slave after the fork/exec.

OCamlPro-Henry:
Following change in Hack that use 'Daemon.spawn' to create slave, the
'global references' have to be copied "manually" on slave. This is done
in `server/serverWorker.ml`.

This is tied with D45423.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -80,10 +80,11 @@
  * 'caml_' */
 #define CAML_NAME_SPACE
 #include <caml/mlvalues.h>
-#include <caml/unixsupport.h>
 #include <caml/memory.h>
 #include <caml/alloc.h>
 #include <caml/fail.h>
+
+#include "handle.h"
 
 #include <assert.h>
 
@@ -104,6 +105,43 @@
 #include <sys/time.h>
 #include <sys/types.h>
 #include <unistd.h>
+
+/****************************************************************************
+ * Quoting the linux manpage: memfd_create() creates an anonymous file
+ * and returns a file descriptor that refers to it. The file behaves
+ * like a regular file, and so can be modified, truncated,
+ * memory-mapped, and so on. However, unlike a regular file, it lives
+ * in RAM and has a volatile backing storage. Once all references to
+ * the file are dropped, it is automatically released. Anonymous
+ * memory is used for all backing pages of the file. Therefore, files
+ * created by memfd_create() have the same semantics as other
+ * anonymous memory allocations such as those allocated using mmap(2)
+ * with the MAP_ANONYMOUS flag. The memfd_create() system call first
+ * appeared in Linux 3.17.
+ ****************************************************************************/
+#ifdef HAVE_MEMFD_CREATE
+#include <sys/memfd.h>
+#else
+#if HAVE_DECL___NR_MEMFD_CREATE
+#define HAVE_MEMFD_CREATE 1
+#include <asm/unistd.h>
+static int memfd_create(const char *name, unsigned int flags) {
+  return syscall(__NR_memfd_create, name, flags);
+}
+#endif
+#endif
+
+/****************************************************************************
+ * When used with open(2), O_TMPFILE allows to create an unnamed
+ * temporary file.  The pathname argument specifies a directory; an
+ * unnamed inode will be created in that directory's filesystem.
+ * Anything written to the resulting file will be lost when the last
+ * file descriptor is closed, unless the file is given a name.
+ ****************************************************************************/
+#ifdef __O_TMPFILE
+#define O_TMPFILE __O_TMPFILE
+#endif
+
 #endif
 
 // The following 'typedef' won't be required anymore
@@ -118,8 +156,7 @@ typedef unsigned __int64 uint64_t;
 #endif
 
 #ifdef _WIN32
-static int win32_getpagesize(void)
-{
+static int win32_getpagesize(void) {
   SYSTEM_INFO siSysInfo;
   GetSystemInfo(&siSysInfo);
   return siSysInfo.dwPageSize;
@@ -170,7 +207,7 @@ static size_t heap_size;
 
 /* Fix the location of our shared memory so we can save and restore the
  * hashtable easily */
-#define SHARED_MEM_INIT 0x500000000000ll
+#define SHARED_MEM_INIT ((char *)0x500000000000ll)
 
 /* As a sanity check when loading from a file */
 static uint64_t MAGIC_CONSTANT = 0xfacefacefaceb000ll;
@@ -254,6 +291,14 @@ value hh_hash_slots() {
   CAMLreturn(Val_long(HASHTBL_SIZE));
 }
 
+#ifdef _WIN32
+
+struct timeval log_duration(const char *prefix, struct timeval start_t) {
+   return start_t; // TODO
+}
+
+#else
+
 struct timeval log_duration(const char *prefix, struct timeval start_t) {
   struct timeval end_t;
   gettimeofday(&end_t, NULL);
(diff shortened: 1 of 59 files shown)


OCamlPro-Henry:
At the moment, I have no idea what might cause the sigbus on Travis. The tests run fine here and all usual suspects seem correctly protected with error messages.

int3:
@gabelevi, can you look into this sigbus?

gabelevi:
Is this missing hack/server/serverWorker.ml? I see it in https://reviews.facebook.net/D45423 but it's not in this PR, so Hack fails to build when I import this PR.

gabelevi:
Hmm...so running the internal tests on linux I get a ton of failures too: https://gist.github.com/gabelevi/1c206fe3637f311a362f

I also tried cloning this PR and running the tests on OSX and this is what I get from `make test`: https://gist.github.com/a5382ccd2ae4d19d62ab

So yeah, something is up with the tests. When you say they're running fine for you, are you testing with Windows?

lefessan:
@gabelevi : I think the last commits (except the one that I added that will be reverted later during rebase) should fix yours errors on Linux and OSX, but there is still a problem on Travis, that we were not able yet to replay on our computers. We will review the PR completely, to see what might be wrong, probably in the new shared-memory part.

gabelevi:
Tested on Linux and OSX and the build and tests are working. The Travis CI issue seems to be due to Travis testing in a container. I tried disabling the containers (https://github.com/gabelevi/flow/commit/5ded38a003758838b69ee0b53257a7660b492f15) and the tests [starting passing again](https://travis-ci.org/gabelevi/flow/builds/79944477).

I tried naively running Flow using docker on Ubuntu 12.04, but `make test` still passed. So I looked into reproing what Travis is doing. I noticed in the Travis build log system-info section it said

> Cookbooks Version
a68419e https://github.com/travis-ci/travis-cookbooks/tree/a68419e

So I went to that link, followed the instructions and managed to reproduce it. Here's my commands, more or less

```Bash
git clone git@github.com:travis-ci/travis-cookbooks.git
cd travis-cookbooks
git checkout a68419e
# Downloaded most recent version of vagrant from https://www.vagrantup.com/downloads.html
vagrant plugin install deep_merge
vagrant up # should use precise64 by default. I used VirtualBox
vagrant ssh
sudo add-apt-repository ppa:avsm/ppa
sudo apt-get update
sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install ocaml ocaml-native-compilers camlp4-extra opam libelf-dev git
git clone https://github.com/OCamlPro/flow.git
cd flow
export OCAML_VERSION=4.02.1
export OPAM_VERSION=1.2.0
export TRAVIS_OS_NAME="linux"
bash -ex .travis-ci.sh # This gave me a random error, but it set up opam for me. Probably could have set it up by hand instead
make test # And see the failures
```

Hope this helps! And I hope there isn't something intrinsic to containers that makes them not work with librt or something like that!

gabelevi:
Bah, in my attempt at reproing I forgot to checkout the correct branch. If I do `git checkout origin/ocp-windows-sharing && make test` then the tests pass.

So maybe it repros when you use OpenVZ instead of VirtualBox? Or maybe there's some configuration that leads to this? It's going to be really annoying to debug through Travis CI, since you don't have access to sudo :/

hack/server/serverWorker.ml
```diff
@@ -0,0 +1,20 @@
+(**
+ * Copyright (c) 2014, Facebook, Inc.
```
int3: 2015

hack/utils/relative_path.ml
```diff
@@ -18,6 +18,12 @@ type prefix =
 let root = ref None
 let hhi = ref None
 
+type slave_state = string option * string option
+let save () = !root, !hhi
+let restore (saved_root, saved_hhi) =
+  root := saved_root;
+  hhi := saved_hhi
```
int3: the hhi path shouldn't be part of Relative_path... can we define this in ServerWorker instead?

OCamlPro-Bozman:
I rebase and edit the code. I am still investigating the PR, specially `hh_shared.c` and trying to understand what can cause the failures with Travis.

lefessan:
This bug is hard to understand. It happens while scanning "hashtbl", but in 19b3e2c, it was possible to scan completely hashtbl just before failing at scanning it again. I think we should avoid shm_open on Travis, and keep the former mechanism (based on fork) on Linux, instead of spawning (fork+exec) the workers. That way, a simple mmap is enough to share the memory segment between the server and the workers. The other mechanism, based on spawning, should be kept only for Windows, where it seems to work well.

gabelevi:
Ok, I tried again.

1. I patched this PR
2. I fixed the build (small things...internal build, rebased, `s/SharedMem.(init default_config)/let _handle = SharedMem.init_default ()/`
3. For good measure I also patched [the corresponding Hack PR](https://reviews.facebook.net/D45423) but there were no important changes

As of my testing

1. Flow builds! Yay!
2. Flow tests pass! Yay!
3. All the Hack projects build! Yay!
4. Hack tests massively fail (failed 1487 out of 2033 tests) :sob:

I see this message quite often

  Unix.Unix_error(Unix.ESPIPE, "fallocate", "")

plus the tests pass on master, so it's definitely this diff. That said, I'm not sure the best way to share the test output with you. I'm pinging the hhvm oncalls but haven't heard back.

gabelevi:
Also @int3 had some feedback, which I'll reproduce here:

> I have yet to properly review worker.ml, but we can have more comments on it? I'd like an overview / explanation of what it's doing and why the extra complexity is necessary for Windows compatibility

hack/heap/hh_shared.c
```diff
@@ -104,6 +104,24 @@
...
 #include <unistd.h>
+
+#ifdef HAVE_MEMFD_CREATE
+#include <sys/memfd.h>
+#else
+#if HAVE_DECL___NR_MEMFD_CREATE
+#define HAVE_MEMFD_CREATE 1
+#include <asm/unistd.h>
+static int memfd_create(const char *name, unsigned int flags)
+{
```
gabelevi: @int3 

> Nit: { should be on the same line as the rest of the declaration

> Also, can we have a comment here about what memfd_create does? As far as I understand it's the same as an atomic shm_open + mmap + unlink, right?

hack/heap/hh_shared.c
```diff
@@ -264,31 +290,180 @@ struct timeval log_duration(const char *prefix, struct timeval start_t) {
...
+  if (memfd < 0) {
+#if defined(__APPLE__)
+    char memname[255];
+    sprintf(memname, "/fb_heap.%d", getpid());
+    memfd = shm_open(memname, O_CREAT | O_RDWR, 0666);
+    if (memfd < 0) {
+        uerror("shm_open", Nothing);
+    }
+#elif defined(O_TMPFILE)
+    memfd = open(shm_dir, O_TMPFILE|O_RDWR|O_CLOEXEC|O_EXCL, 0666);
```
gabelevi: @int3 

> Would also like a comment about O_TMPFILE versus shm_open

> Also, if O_TMPFILE is not defined and we're not building on an Apple, it looks like this function will never succeed. Why not use shm_open on Linux as well?

hack/heap/hh_shared.c
```diff
@@ -264,31 +290,180 @@ struct timeval log_duration(const char *prefix, struct timeval start_t) {
...
+/* in a mmaped segment that begin at `start` */
+/*********************************************/
 
 #ifdef _WIN32
-  if (!VirtualAlloc(mem,
-                    global_size_b + page_size +
-                      2 * DEP_SIZE_B + HASHTBL_SIZE_B,
-                    MEM_COMMIT, PAGE_READWRITE)) {
+
+static void memfd_reserve(char * mem, size_t sz) {
```
gabelevi: @int3 

> what's the purpose of this function? Is it just so we raise clearer errors upon running out of memory?

hack/procs/multiWorker.ml
```diff
@@ -38,57 +38,52 @@ let single_threaded_call_dynamic job merge neutral next =
   done;
   !acc
 
+let multi_threaded_call_dynamic
```
gabelevi: @int3 

> hmm so why was this refactoring necessary? Just so we keep around the handles?

hack/procs/multiWorker.ml
```diff
@@ -38,57 +38,52 @@ let single_threaded_call_dynamic job merge neutral next =
...
+            (* ... send a job to the worker.*)
+            let handle =
+              Worker.call worker
+                (fun xl -> job neutral xl)
+                bucket in
+            dispatch workers (handle :: handles) acc
+  and collect workers handles acc =
+    let { Worker.readys; waiters } = Worker.select handles in
+    let workers = List.map ~f:Worker.get_worker readys in
+    (* Collect there results. *)
```
gabelevi: @int3 

> s/there/the/

hack/procs/worker.ml
```diff
@@ -24,91 +22,60 @@ external hh_worker_init: unit -> unit = "hh_worker_init"
...
-  let oc = Unix.out_channel_of_descr descr_oc in
-  let input() = Marshal.from_channel ic in
-  let output data = Marshal.to_channel oc data [Marshal.Closures]; flush oc in
-  { pipe_descr_in = descr_ic;
-    pipe_fin = input;
-    pipe_descr_out = descr_oc;
-    pipe_fout = output;
-  }
+type request =
+  | Request: (('a -> unit) -> unit) -> request
```
gabelevi: @int3 

> nit: can we use of instead of :?

OCamlPro-Henry:
> 2. I fixed the build (small things...internal build, rebased, `s/SharedMem.(init default_config)/let _handle = SharedMem.init_default ()/`

Is that fixes in Hack or in Flow ? Because for Flow, I currently use `let _handle = SharedMem.(init default_config FlowConfig.default_shm_dir)`. The `SharedMem.init_default` will use the Hack default directory for shared memory, which might be changed by setting the `HACK_SHMDIR` environment variable. While `FlowConfig.default_shm_dir` might be changed with `FLOW_SHMDIR`.

> 1. Flow builds! Yay!
> 2. Flow tests pass! Yay!
> 3. All the Hack projects build! Yay!

Yay!

> 4. Hack tests massively fail (failed 1487 out of 2033 tests) :sob:
>
> I see this message quite often
>
>   Unix.Unix_error(Unix.ESPIPE, "fallocate", "")

This is what I called, with too much enthusiasm, "a better error message than SIGBUS"! It still needs to be more explicit. Actually, it means: "Out of memory in the shared heap."

That's the inconvenient of using `shm_open` (or an equivalent): the shared heap is bound to the available "disk space" in `/dev/shm` (which is usually a `tmpfs`, so not really a hard drive). That's usually not a problem on a desktop computer where `/dev/shm` is huge. It mights be a problem on Travis or others CIs. For instance, on Travis, `/dev/shm` is only 64MB.

That's why I used `O_TMPFILE` or `mkstemp` instead of `shm_open`. This allows to move the shared heap from `/dev/shm` to another path, e.g. `/tmp`. Currently, this could be achieved by setting `HACK_SHMDIR` or `FLOW_SHMDIR`.

All of this burden is not required on recent Linux, where `memfd_create` can be used and which is not bound to the limits of of `/dev/shm`.

OCamlPro-Henry:
On Fri, Nov 06, 2015 at 02:25:54PM -0800, Gabe Levi wrote:
> Also @int3 had some feedback, which I'll reproduce here:
> 
> > I have yet to properly review worker.ml, but we can have more comments on it? I'd like an overview / explanation of what it's doing and why the extra complexity is necessary for Windows compatibility

Ok, I will add some documentation in `worker.ml`.

One of the motivation to the "extra complexity" is this comment:

```
(* As explained in the header, we wrap every function with Obj.magic, because
 * the type-checker does not allow us to have a polymorphic worker.
 *)
```

And all the `Worker.get_result proc (Obj.magic 0)`.

At first, I felt challenged by the remarks but restrained myself. Later, while trying to adapt it to Windows, I get lost. Totally. This was like programming without typechecker :-) Too many `Segfault` for trivial mistakes. At the end, I felt safer to partially rewrite and to minimize the `magic`.

hack/heap/hh_shared.c
OCamlPro-Henry: On Fri, Nov 06, 2015 at 02:26:43PM -0800, Gabe Levi wrote:
> > @@ -104,6 +104,24 @@
> >  #include <sys/time.h>
> >  #include <sys/types.h>
> >  #include <unistd.h>
> > +
> > +#ifdef HAVE_MEMFD_CREATE
> > +#include <sys/memfd.h>
> > +#else
> > +#if HAVE_DECL___NR_MEMFD_CREATE
> > +#define HAVE_MEMFD_CREATE 1
> > +#include <asm/unistd.h>
> > +static int memfd_create(const char *name, unsigned int flags)
> > +{
> 
> @int3 
> 
> > Nit: { should be on the same line as the rest of the declaration
> 
> > Also, can we have a comment here about what memfd_create does? As far as I understand it's the same as an atomic shm_open + mmap + unlink, right?

Ok, I will add a comment.

That's more like `shm_open + unlink` without the inconvenient of being limited by the size of `/dev/shm`. The file descriptor is not bound to a mount point.

The `mmap` has still to be made.

hack/heap/hh_shared.c
OCamlPro-Henry: > > Would also like a comment about O_TMPFILE versus shm_open

Ok.

> > Also, if O_TMPFILE is not defined and we're not building on an Apple, it looks like this function will never succeed. Why not use shm_open on Linux as well?

There is always the fallback case (on Linux and Apple): `mkstemp`. It should always succeed.

I prefered `O_TMPFILE/mkstemp` over `shm_open` because it allows to move the `shared heap` from `/dev/shm` to another "disk", when `/dev/shm` is too small.

hack/heap/hh_shared.c
OCamlPro-Henry: > > what's the purpose of this function? Is it just so we raise clearer errors upon running out of memory?

Yes. Without this function, this is SIGBUS. Which might even happen on a read.

hack/procs/multiWorker.ml
OCamlPro-Henry: > hmm so why was this refactoring necessary? Just so we keep around the handles?

Yes. But I changed it more than required. I removed a very small code
duplication (the exception handler for `Exit` is not required
anymore). I might simplify the "diff" and stand closer to the actual
code, if you prefer.

hack/procs/worker.ml
OCamlPro-Henry: > > +type request =
> > +  | Request: (('a -> unit) -> unit) -> request

> nit: can we use of instead of :?

Hmm. Let me thinks.

Yes, it mights be replaced by:

```
type serializer = { send : 'a. 'a -> unit }
type request =
  | Request of (serializer -> unit)
```

which is more explicit: the function expects a "serializer" as
argument. This "serializer" should be able to serialize any `'a`.

OCamlPro-Henry:
I have no idea what's wrong with Travis, the same exact commit builds perfectly on our repo: https://travis-ci.org/OCamlPro/flow/builds/91608156

OCamlPro-Henry:
Two questions:
- About commit https://github.com/OCamlPro/flow/commit/97c51d730359e3e22ac8e08771ae90c50db047d4 Is there any reason not to immediatly take the lock ?
- Is there a reason for 'ocamlbuild' not to build Flow with `-DOSS_SMALL_HH_TABLE_POWS` ?

OCamlPro-Henry:
> > 4. Hack tests massively fail (failed 1487 out of 2033 tests) :sob:
> > I see this message quite often
> >
> >   Unix.Unix_error(Unix.ESPIPE, "fallocate", "")
>
> This is what I called, with too much enthusiasm, "a better error message than SIGBUS"! It still needs to be more explicit. Actually, it means: "Out of memory in the shared heap."

Any news ? I improved the error message, it should be more explicit.

Could you confirm that your internal CI use a linux kernel < 3.17 and owns a too "small" '/dev/shm' ? In that case, does setting `HH_SHMDIR=/tmp` fix the problem ? if `/tmp` is not too small also.

If the kernel >= 3.17, it should not use `/dev/shm`. Then, is 'memfd_create` correctly detected by your internal build system ?

OCamlPro-Henry:
Any news ?

OCamlPro-Bozman:
Please check [#1184](https://github.com/facebook/flow/pull/1184).

PR #761, opened 2015-08-28
Use full path to executable in Daemon.spawn

mroch:
(**note**: this PR should be against hack directly, but my plane is landing, I don't have an hhvm checkout, and I'll probably be off the grid for the next couple days so I wanted to get it out)

`argv[0]` is not always correct enough to find the binary. One repro we've found so far is when you have an unexpanded `~` in `PATH`, e.g. `PATH='~/flow/bin/':/usr/bin flow ...` (note the single quotes). Bash does expansion to find `~/flow/bin/flow`, but when ocaml reads `PATH`, it just sees `~/flow/bin/` and doesn't do any expansion, so can't find the executable.

`Sys_utils.executable_path ()` does the expansion.

One concern is that this may not be Windows compatible. However, switching to `Daemon.spawn` caused flow to regress because of this, so we should probably fix it and then unbreak Windows if necessary.

### DIFF
--- a/hack/utils/daemon.ml
+++ b/hack/utils/daemon.ml
@@ -173,10 +173,8 @@ let spawn
           fn)  in
   let out_fd =
     Unix.openfile out_path [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_TRUNC] 0o666 in
-  let pid =
-    Unix.create_process
-      Sys.executable_name [|Sys.executable_name|]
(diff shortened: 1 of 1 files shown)


mroch:
cc @OCamlPro-Henry

mroch:
Closed by 455262b80cbfc1669b4638fd993bd829df62a25e

PR #763, opened 2015-08-31
Warn when rest parameters are annotated with a tuple type

### LINKED ISSUE #762: Tuple type-annotated rest parameters assertion failure ("Did not expect BoundT")
```javascript
/* @flow */
function oops<T>(...xs: [T]): void { }
```

```javascript
/* @flow */
type Fn = <T>(...xs: [T]) => void;
```

```javascript
/* @flow */
class A<T> { f(...xs: [T]): void {}; }
```

Output:

```
<<<<<<<<
Did not expect BoundT:
Called from file "list.ml", line 73, characters 12-15
>>>>>>>>

/path/to/test.js:
Failure("Did not expect BoundT")
```

samwgoldman:
Before this change, annotating a rest parameter with a tuple type could
lead to an ugly exception, "Did not expect BoundT"

Instead, warn the user when this pattern occurs, because a tuple type
does not represent rest parameters well. The tuple type implies a fixed
number of elements, but there can be any number of rest parameters, by
definition.

While debugging this, I also noticed an issue with bound type parameters
in another rest parameter use case: `fn<T>(...xs: T)`. Currently, this
also throws the ugly exception, but it's a bit tricker to resolve
because the error is thrown from the unification that occurs before the
type param can be substituted.

Fixes #762

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -933,7 +933,7 @@ and mk_rest cx = function
   | ArrT (_, t, []) -> RestT t
   | AnyT _ as t -> RestT t
   | ArrT (r, t, _) ->
-      let msg = "Tuple type should not be used to annotate rest parameters" in
+      let msg = "rest parameters should be an array type, got a tuple type instead" in
       Flow_js.add_warning cx [r,msg];
       RestT t
   | t ->
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -930,8 +930,12 @@ and convert_qualification ?(lookup_mode=ForType) cx reason_prefix
 )
 
 and mk_rest cx = function
-  | ArrT(_, t, []) -> RestT t
+  | ArrT (_, t, []) -> RestT t
   | AnyT _ as t -> RestT t
+  | ArrT (r, t, _) ->
+      let msg = "Tuple type should not be used to annotate rest parameters" in
+      Flow_js.add_warning cx [r,msg];
+      RestT t
   | t ->
       (* unify t with Array<e>, return (RestT e) *)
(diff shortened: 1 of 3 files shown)


samwgoldman:
Feel free to bike shed on the error message here, of course.

src/typing/type_inference_js.ml
```diff
@@ -930,8 +930,12 @@ and convert_qualification ?(lookup_mode=ForType) cx reason_prefix
 )
 
 and mk_rest cx = function
-  | ArrT(_, t, []) -> RestT t
+  | ArrT (_, t, []) -> RestT t
   | AnyT _ as t -> RestT t
+  | ArrT (r, t, _) ->
+      let msg = "Tuple type should not be used to annotate rest parameters" in
```
jeffmo: This says what can't be done, but maybe you can specify that `Array<T>` is what can?

samwgoldman: Good idea! Pushed a change.

src/typing/type_inference_js.ml
```diff
@@ -930,8 +930,12 @@ and convert_qualification ?(lookup_mode=ForType) cx reason_prefix
 )
 
 and mk_rest cx = function
-  | ArrT(_, t, []) -> RestT t
+  | ArrT (_, t, []) -> RestT t
   | AnyT _ as t -> RestT t
+  | ArrT (r, t, _) ->
+      let msg = "rest parameters should be an array type, got a tuple type instead" in
+      Flow_js.add_warning cx [r,msg];
+      RestT t
```
samwgoldman: A pattern exhibited elsewhere is to propagate `AnyT` in the event of an error (presumably, so errors don't just pile up). Would it be better to error and return `RestT AnyT` here instead of plowing ahead with the combined element type of the union?

PR #766, opened 2015-09-01
Filter instance and class types when checking non-existence

### LINKED ISSUE #757: Flow sometimes detects incorrect type in expressions involving && and ||
```
/* @flow */
class A {};
type B = {};

function foo(a: ?A, b: ?B) {
  var s: string = a && "X" || "Y"; // error
  var t: string = b && "X" || "Y"; // ok
}
```
Also, quite confusingly, the error is reported in parameter `a` of the function, instead of in the location where the parameter fails to typecheck: http://tryflow.org/#dd51333184c595d324852a62d95ed062

samwgoldman:
Values of these types will always be "truthy" in a conditional
expression.

Fixes #757

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3986,10 +3986,13 @@ and filter_not_exists t = match t with
   | StrT (r, (Literal _ | Truthy))
   | ArrT (r, _, _)
   | ObjT (r, _)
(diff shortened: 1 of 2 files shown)


mroch:
thanks, this looks great. our sync script is busted again right now, trying to get this merged though.

PR #778, opened 2015-09-03
Respect flowconfig [ignore] always when searching for node modules

samwgoldman:
Before this change, we ignored the path for a `package.json` file based
on the [ignore] section of the flowconfig, but we still tried to find an
`index.js` file regardless. Later on, Flow would complain that the file
could not be found.

Before this change, for the provided test case, users would get an
error:

```
test.js:3:17,21: index.js
Required module not found

Found 1 error
```

After this change, we succeed in using the libdef, so we might see an
expected error:

```
test.js:5:2,8: number
test.js:5:12,17: string

Found 1 error
```

cc @jeffmo

### DIFF
--- a/src/typing/module_js.ml
+++ b/src/typing/module_js.ml
@@ -250,8 +250,10 @@ module Node = struct
   let guess_exported_module file _content = file
 
   let path_if_exists path =
-    if file_exists path then Some path
-    else None
+    if not (file_exists path) ||
+      FlowConfig.(is_excluded (get_unsafe ()) path)
+    then None
+    else Some path
 
   let path_is_file path =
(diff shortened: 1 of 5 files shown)


mroch:
just to confirm, index.js and package.json are empty files, right?

gabelevi:
Awesome! This has been annoying me for ages!

samwgoldman:
@mroch Yeah, I actually didn't intend to commit the `package.json` file, and I think there are actually two code paths, both searching for `index.js`, in either case. However, both code paths go through `path_if_exists` so the coverage gap isn't so bad, IMO.

PR #783, opened 2015-09-05
Allow const let

paulshen:
See https://github.com/facebook/flow/pull/431#issuecomment-118206002


paulshen:
sorry didn't mean to send this here :)

PR #784, opened 2015-09-06
let/const

samwgoldman:
This commit allows for the creation of let/const bindings and ensures
all block-scoped bindings are installed into the appropriate scope. This
involves creating new LexScopes, as outlined in `test/bindings/scope.js`.

Before this change, the environment stack, i.e., frames, changesets, and
scopes, would change in lockstep. Now, lex scopes change separately from
frames and changesets via `push_lex` and `pop_lex`.

When adding heap refinements, we may traverse over lex scopes to find
the scope where the base object was bound.

In unusual cases, like try/catch, envs may have an uneven number of
scopes when merging. The uneven scopes will always be lex. I have
updated `merge_env` to accept this imbalance.

Note that this is not a complete implementation of let/const:
- [x] switch statements should forbid rebinding lexicals between cases
- [x] functions, amazingly, are block scoped *and* hoisted in ES2015
- [ ] numerous TDZ cases, surely, which I have not yet investigated
- [x] can't update const bindings (assignment already handled)
- [x] var bindings can not shadow a lexical binding, ever
- [x] more tests

### DIFF
--- a/src/typing/env_js.ml
+++ b/src/typing/env_js.ml
@@ -761,9 +761,11 @@ let merge_env =
         (* ...in which case we can forget them *)
         ()
       (* changeset entry exists only in lex scope *)
-      | None, None, Some e when is_lex scope2 && Entry.is_lex e ->
+      | None, Some _, Some _ when is_lex scope1 && is_lex scope2 ->
         ()
-      | None, Some e, None when is_lex scope1 && Entry.is_lex e ->
+      | None, Some _, None when is_lex scope1 ->
+        ()
+      | None, None, Some _ when is_lex scope2 ->
         ()
       (* otherwise, non-refinement uneven distributions are asserts. *)
       | orig, child1, child2 ->
--- a/src/typing/env_js.ml
+++ b/src/typing/env_js.ml
@@ -824,7 +824,7 @@ let copy_env  = Entry.(
   let rec copy_entry cx reason (env1, env2) name =
     match env1, env2 with
 
-    | scope1 :: env1_, scope2 :: env2_ -> Scope.(
+    | scope1 :: env1, scope2 :: env2 -> Scope.(
       match get_entry name scope1, get_entry name scope2 with
 
       | Some (Value _ as v1), Some (Value _ as v2) ->
@@ -842,10 +842,10 @@ let copy_env  = Entry.(
 
       | None, None ->
         (* not found, try outer scopes *)
-        copy_entry cx reason (env1_, env2_) name
+        copy_entry cx reason (env1, env2) name
 
       (* global lookups may leave new entries in env2 *)
-      | None, Some _ when env1_ = [] ->
+      | None, Some _ when env1 = [] ->
         (* ...in which case we can forget it *)
         ()
 
@@ -854,12 +854,6 @@ let copy_env  = Entry.(
         ()
       | Some e, None when is_lex scope1 && Entry.is_lex e ->
         ()
-
-      (* walk through uneven lex scopes *)
-      | Some _, None when is_lex scope2 ->
-        copy_entry cx reason (env1, env2_) name
-      | None, Some _ when is_lex scope1 ->
-        copy_entry cx reason (env1_, env2) name
 
       | entry1, entry2 ->
         (* uneven distributions *)
--- a/src/typing/env_js.ml
+++ b/src/typing/env_js.ml
@@ -62,8 +62,8 @@ open LookupMode
 (****************)
 
 (* the environment is a scope stack, frame stack and changeset stack.
-   currently these are in lockstep, though this may change depending
-   on how lexical scoping is implemented. *)
+   except for lex scopes, these are in lockstep. multiple lex scopes
+   may exist within the same frame and changeset. *)
 
 type 'a stack = 'a list ref
 
@@ -89,17 +89,32 @@ let get_scopes () =
 let iter_scopes f =
   List.iter f !scopes
 
+(* use to walk scope chain up to the topmost var scope *)
+let iter_lex_scopes f =
+  let rec loop = function
+    | [] -> assert_false "empty scope list"
+    | scope::scopes ->
+        f scope;
+        (match scope.kind with
+        | LexScope -> loop scopes
+        | _ -> ())
+  in
+  loop !scopes
+
 (* clone the current scope stack (snapshots entry maps) *)
 let clone_scopes scopes =
   List.map Scope.clone scopes
 
 (* true if current top scope is an async function *)
-(* TODO will need to walk once LexScopes appear *)
-let in_async_scope () = Scope.(
-  match (peek_scope ()).kind with
-  | VarScope { async } -> async
-  | _ -> false
-)
+let in_async_scope () =
+  let rec loop = function
+    | [] -> false
+    | scope::scopes ->
+      (match scope.kind with
+      | VarScope { async } -> async
+      | _ -> loop scopes)
+  in
+  loop !scopes
 
 (* build a map of all var entries - no refis - in the current
    scope stack.
@@ -174,6 +189,18 @@ let pop_env () =
   frames := List.tl !frames;
   changesets := List.tl !changesets
 
+(* push a lex scope but NOT a frame or changeset
+ * scope rag affects merge_env -- see comments *)
+let push_lex () =
+  let scope = Scope.fresh_lex () in
+  scopes := scope :: !scopes
+
+let pop_lex () =
+  scopes := match !scopes with
+  | { kind = LexScope; _ }::scopes -> scopes
+  | [] -> assert_false "empty scope list"
+  | _ -> assert_false "top scope is non-lex"
+
 (* initialize a new environment (once per module) *)
 let init_env cx module_scope =
   clear_env ();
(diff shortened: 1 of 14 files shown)


samwgoldman:
All right. This is about as far as I can get right now. Feedback on what's here is appreciated. I may have missed something, but TDZ *should* be the only part outstanding here.

@bhosmer, did you have an approach in mind for handling that? I noticed you want to kill two birds with one stone here, and allow declaring non-maybe vars without an initializer.

kastermester:
I have only briefly tried this out, but so far it seems to work as advertised! Great work! :) Will let you know if I run into any issues :)

src/typing/env_js.ml
```diff
@@ -40,19 +40,22 @@ open Scope
...
       not raised here.
 
-   4. ForTypeof lookups are in fact ForValue lookups, but due to the order in
+   5. ForTypeof lookups are in fact ForValue lookups, but due to the order in
       which AST traversal takes place, these lookups may legitimately violate
       rule #2, hence the need for a special mode.
  *)
 module LookupMode = struct
-  type t = ForValue | ForType | ForTypeof
+  type t = ForValue | ForUpdate | ForType | ForTypeof
```
bhosmer: Per IRC, if you take the modes path, I'd suggest

    type t = ForValue of value_op | ForType | ForTypeof 
    and value_op = Read | Write

The important thing is to keep the modes nonoverlapping, and pushing the read/write distinction down into `ForValue` will also spare you the occasional `mode = ForValueRead || mode = ForValueWrite` down below.

AgentME:
Is TDZ compliance possible to statically check for in general? Babel only supports using runtime checks because [its earlier attempt at static checking had problems](https://github.com/babel/babel/issues/527).

For what it's worth, Flow already misses TDZ-like issues with vars:

```javascript
/* @flow */
console.log(s.length);
var s = "abc";
```

```javascript
/* @flow */
f();
var s = "abc";
function f() {
  console.log(s.length);
}
```

src/typing/env_js.ml
```diff
@@ -293,49 +342,93 @@ let already_bound_error =
...
+      | LexScope, Entry.Value { Entry.kind = Entry.Const; _ }
+      | VarScope _, _ ->
+        Some scope
+      | LexScope, Entry.Type _ ->
+        find_scope scopes
+      | LexScope, Entry.Value { Entry.kind = Entry.Var; _ } ->
+        match get_entry name scope with
+        (* nested scope var shadows a lex-scoped lexical *)
+        | Some prev
+          when is_lex scope && Entry.is_lex prev ->
```
bhosmer: `is_lex scope` is unnecessary here...

src/typing/env_js.ml
```diff
@@ -293,49 +342,93 @@ let already_bound_error =
...
+            already_bound_error cx name prev reason;
+            None
+        (* keep searching for the var's scope *)
+        | _ -> find_scope scopes
```
bhosmer: This pattern will fire for the `Some _` case when the side condition is false, causing us to look past existing entries. Don't think that's right, is it?

src/typing/env_js.ml
```diff
@@ -293,49 +342,93 @@ let already_bound_error =
...
 
+let bind_entry cx name entry reason =
+  (* lex scopes can only hold let/const bindings
+   * var scopes can hold all bindings
+   * var bindings can't shadow lex bindings
+   * type entries are hoisted to var scope *)
+  let rec find_scope = function
+    | [] -> assert_false "empty scope list"
+    | scope::scopes -> Scope.(
+      match scope.kind, entry with
```
bhosmer: Clearer and tighter to do nested matches on scope kind then entry kind here: 

    match scope.kind with
    | VarScope _ -> Some scope
    | LexScope -> Entry.(
      match entry.kind with
      | Let
      | Const -> Some scope
      | Type -> find_scope scopes
      | Var -> ...
    )

Also, I haven't worked it out completely but it doesn't seem right that both this function and the caller raise errors and need to do `get_entry` calls on the found scope. Don't know if it would be better to merge the two functions, or return an optional entry from here, or something else, but I think a cleaner separation into find-scope/do-stuff is possible. LMK if I'm talking nonsense and I'll take a crack at it. :)

src/typing/env_js.ml
```diff
@@ -638,12 +748,12 @@ let merge_env =
     match env0, env1, env2 with
     | [], [], [] ->
       ()
-    | scope0 :: env0, scope1 :: env1, scope2 :: env2 -> Scope.(
+    | scope0 :: env0_, scope1 :: env1_, scope2 :: env2_ -> Scope.(
```
bhosmer: Trailing underscores?

samwgoldman: Haha. Got me. Naming things is hard!

bhosmer: :) I feel you on trying to avoid rebinding, but in this case I think it actually makes sense, given that you're peeling off the heads and (maybe) recursing straight into the tails, with nothing in between.

samwgoldman: If you squint, there are a couple cases where we need to revisit the whole list, which is why I changed the names—it's not just to avoid shadowing. In Haskell, it's fairly common to do `x:xs'` in these cases, but a better name/more intention-revealing structure is called for here.

src/typing/env_js.ml
samwgoldman: So, I added this "check as we walk the lex scopes" logic a bit hastily, and I should have refactored here. The reason the `LexScope, Entry.Value _` case even exists is to ensure that a var doesn't shadow a lex binding in an intermediate lex scope, see [the associated test case](https://github.com/facebook/flow/pull/784/files#diff-d5b8aa5064f80d1a4462d05b7f7fa872R173), for an example. This is also why I added the `is_lex` guard, above; we'll catch the non-`is_lex` case below.

I think the solution here is to actually merge the `find_scope` logic with the rest of the logic below, and eliminate the intermediate `Scope.t option` stuff. I should be able to add enough comments to make the intention clear.

src/typing/env_js.ml
```diff
@@ -661,24 +771,21 @@ let merge_env =
...
-         residual entries in child contexts, when catch blocks combine
-         conditionals and early returns.
-         So for now we need to let extra child entries pass, though we
-         should still assert on extra parent entries.
-         Once let/const/LexScope is in place, we can reimplement let_env
-         in terms of those constructs instead, and the full assert can
-         be reinstated.
-       *)
-      | None, _, _ ->
+      (* changeset entry exists only in lex scope *)
```
bhosmer: How about when both children are lex scopes with entries?

...in general this is combinatorial enough now that I'd consider splitting this match into an outer match on the `get scope0` with inner matches on the kids. NP if you'd rather leave it alone though - I can come in after yours lands if I feel strongly enough about it. It's just that we're now in that territory where we're testing a lot of fairly specific things and then falling through to an assert - basically relying on testing to make sure we've covered all the cases.

src/typing/env_js.ml
```diff
@@ -661,24 +771,21 @@ let merge_env =
...
+      | None, None, Some e when is_lex scope2 && Entry.is_lex e ->
+        ()
+      | None, Some e, None when is_lex scope1 && Entry.is_lex e ->
         ()
-      | Some _ as orig, child1, child2 ->
+      (* walk through uneven lex scopes *)
```
bhosmer: Analogously, what happens when both kids are lex scopes here?

src/typing/env_js.ml
samwgoldman: > How about when both children are lex scopes with entries?

If the changeset entry appears in both lex children, we'll prune one side, then the other, yeah? Adding a `None, Some e, Some e` case would just be optimization, as I understand it.

src/typing/env_js.ml
samwgoldman: Similarly, my reasoning was that we'd recuse on one side (peeling off the top scope from env2, say), then immediately hit the other side. Adding a `Some _, None, None` case would just be optimization.

samwgoldman: Although, if we did explicitly handle the `Some _, None, None` case, we would never need to reference the original array, and we could do away with the `envN_` bindings... that seems worthwhile.

src/typing/env_js.ml
samwgoldman: You're right that these functions should be merged. I will do that. The bit of logic that complicates this, is that we need to check as we walk the lex scopes that a var binding doesn't shadow a lexical anywhere between the declaration site and the eventual var scope.

src/typing/env_js.ml
bhosmer: Your call. Personally I'd do the pair cases explicitly rather than relying on the recursion just to convey intent, but I don't want to peck your commit to death. :) Do whatever feels right - if I see a way to matricize the patterns in here a bit more tightly that's worth the churn, I'll fly it in after your PR is merged.

samwgoldman: Peck! I have a lot to learn from you and I appreciate your sensibilities about this stuff.

src/typing/env_js.ml
bhosmer: Yeah, got that thanks to your excellent comments. :) Awesome comment work throughout, BTW - in code like this good comments are a huge asset, and you've done a great job adding (and editing!) them here.

src/typing/env_js.ml
bhosmer: :) ...so I was about to just bang out a show-don't-tell, but first let me just ask: will these lists ever actually be ragged? 
  - they should be uniform from the bottom (global) up to the topmost VarScope, since we only ever merge after traversing local control structures
  - any nonuniform LexScopes should have been popped off before the merge takes place, I think. This is from first principles, but a quick scan of type_inference_js.ml in this PR seems to confirm it...

If you can set me straight with an example that makes it ragged, please do. :) Either way, I'll come back with a scratch version of merge_entry that explains what I'm trying to get at... basically I always try to a) constrain the space of possibilities to only what should happen (for instance, if there's any possibility of raggedness, it definitely stops at the topmost VarScope) and b) partition what's left hierarchically, as soon as the flat partition starts to have that whack-a-mole feel to it.

bhosmer: Just to be clear: I'm saying I don't think the lists will ever be ragged, *if* `merge_env` is called properly, i.e. I think it's a precondition of calling `merge_env` that you've come out of any divergent lexical scopes. And I think we should capture that by asserting if we get scope lists that couldn't happen if the precondition is satisfied. (Rather than handling them.)

samwgoldman: Gotcha—I added this because I actually hit some rag during development. Will try to repro and come at you with an example, assuming the call site isn't violating that invariant...

samwgoldman: OK, so this test from `tests/try/test.js`

```
function foo() {
    var x = 0;
    var y;
    try {
        x = "";
    } catch(e) {
        x = false;
        throw -1;
    } finally {
        y = {};
    }
    x(); // x is neither a number nor a boolean
    y(); // y is not undefined
}
```

Causes this error without the rag cases:

```
test.js:
Failure("merge_env /Users/sam/Developer/OSS/flow/tests/try/test.js:4:5,11:5:\ntry: non-uniform distribution of entry x: Some var, Some var, None")
```

But I haven't investigated whether this is an error in how I'm handling the catch lex scope, or an expected scenario...

bhosmer: Hmmm, ok. LMK if you don't find an error and I can look upstream and see if I can figure out why it would be happening. Seems like it shouldn't be.

samwgoldman: Think I found it & it's a bug in the env copying. Going to try to write a
few tests to suss out the expected behavior, because I think there is a
correlated bug.

samwgoldman: See 2c04e7f16c0fae81706dd49d0febd6dc8bf4fa80

src/typing/env_js.ml
samwgoldman: OK, so I went down this path and things got pretty ugly. Happy to go over it in more depth, but I think the alternative approach of defining a set_var-like entrypoint for Update/PlusAssign. As far as a name for that entrypoint... any ideas? To be honest, I don't really see the problem with ForUpdate, so maybe this is a better change for you to make separately/subsequently?

src/typing/env_js.ml
samwgoldman: (FWIW, I did add the both children case, but GitHub only collapses comments when the line they are attached to changes.)

tikotzky:
:clap: :clap: :clap: :clap: :clap: 
:smile:

danawoodman:
Huzzah! Great work! :fireworks: :boom: :clap: :hand: :+1: :ok_hand: :beers:

cesarandreu:
:tada: :tada: :tada:

iamdustan:
Here’s a quick [jscodeshift](https://github.com/facebook/jscodeshift) script you can use to transform all your `var`s to `let` or `const` and most likely not having anything break:

http://felix-kling.de/esprima_ast_explorer/#/tGblyJN52T/4

Flow and unit tests caught the two issues my current project codebase had post-transform. Both are documented in the transform.

I’ll update it again #831 lands.

davis:
how do we use this?

PR #788, opened 2015-09-09
Fix definition of Array reducers when no initial value is provided

samwgoldman:
When no initial value is provided to the reducer, the first value from
the array is provided as the initial value of the accumulator.

Before this commit, we assumed that T = U, but that's not true. The
included tests did not raise any error before this change.

The error messages are _very_ unfortunate, but bad error messages
shouldn't prevent us from fixing basic correctness issues.


gabelevi:
> The error messages are very unfortunate, but bad error messages shouldn't prevent us from fixing basic correctness issues.

Agreed. Thanks for fixing...we really do need to find a better solution for errors involving intersections and unions. Merging...

Issue #786, opened 2015-09-07
Type not widened by assignment in switch statements


samwgoldman:
```javascript
var a = "";
var b = "";
declare var x: mixed;
switch (x) {
  case "foo":
    a = 0;
  default:
    b = 0;
}
(a : string); // no error, expected number ~> string
(b : string); // no error, expected number ~> string
```

@mroch Any idea what's going on in this example? I was trying to write a counter example for let/const, but wasn't able to trigger an expected error, leading me to this.

samwgoldman:
Oh, after slightly more digging, it's the presence of the `default` case that prevents widening. If I remove the `default` case, then I see the expected error for `a`.

mroch:
adding a `break` at the end of the first case also seems to make both `a` and `b` error, so this sounds like a problem with fallthrough merging.

ngavalas:
I think I've found the bug (at least from my testing) and the fix is relatively simple; I can make a pull request for it tonight.  I was wondering if there was a good way to add test cases for this so it doesn't break in the future -- the test cases I wrote to debug this obviously are intended to fail once the bug is fixed; I basically just adapted the code from the original issue and it fails once the proper types are inferred.

samwgoldman:
@ngavalas That's awesome! IMO, the best test cases are the ones that add expected error output, because then the test suite assets not only on the existence of the error, but also the quality of it—in particular the locations associated with the error. Obviously expected to fail is the best kind of expected to fail. :smiley:

ngavalas:
Ah, I see now, the output gets diffed with the .exp file in the test directory.  I'll work on making sure this change doesn't break anything else then.

ngavalas:
We can move the discussion to #792 if you'd like, but I have definitely identified the issue and *why* it happens, but I'm not sure what the most correct way to fix it is.  What I did certainly works and all of the test cases still pass (it produces the correct errors from number->string at the annotation, not at the a=0 assignment site), so I do believe it's correct, but we could also go down the route of havoc'ing the new changeset in all cases also -- I'm not sure of what the consequences of either choice are in the long run.

PR #792, opened 2015-09-11
Fix type widening for switch statements without break statements

ngavalas:
In response to #786, in switch statements without any break statements (that is, every branch falls through) where there is a default case, the type refinements from the different case branches are not preserved.

The cause for this is that after gathering type information from the branches, we wipe the changeset and have a special case for switch statements without default statements that has the side effect of adding in any of the type information we were going to wipe.  Additionally, if there was a break statement at all in the code, we havoc the variables.  This happens to leave out the case where we have a default and every case falls through.

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -1594,8 +1594,8 @@ and statement cx type_params_map = Ast.Statement.(
         merge_with_last_ctx cx reason default_ctx newset
       );
 
-      if Abnormal.swap (Abnormal.Break None) save_break_exn
-      then Env_js.havoc_vars newset;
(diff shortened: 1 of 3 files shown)


ngavalas:
As far as I can tell, the code the prevented this in the case of having no break statements and also having no default case was around type_inference_js.ml:1589, where we have:

```ocaml
 if not !default
       then (
         let default_ctx = Env_js.clone_scopes !fallthrough_ctx in
         let reason = mk_reason "default" loc in
         Env_js.update_env cx default_ctx;
         merge_with_last_ctx cx reason default_ctx newset
);
```

That last line merges back the newset changes that we just got rid of with Env_js.merge_changeset.

bhosmer:
Thanks Nick!

PR #797, opened 2015-09-14
Support `return` and `throw` generator methods

samwgoldman:
Per the spec[1], generator objects have two methods which I neglected to
implement on the first go 'round.

`return` is somewhat special, because it means the return type can be
widened outside the generator function declaration itself. The argument
passed to `return` becomes the completion value of the generator itself,
and return value is the iterator result with `done:true` and the
provided value.

`throw` is more straightforward, but led me to a (perhaps controversial)
change: I think the type of the catch parameter should be `mixed`,
instead of an open tvar. In a perfect world, Flow would know exactly
what type was thrown, but barring that, we should be pessimistic and
require the user to do a dynamic type test (`invariant()` is useful for
this).

This change is unrelated to generators, but I think the throw API
invites people to pass values into a generator, and I don't want Flow to
ignore potential runtime errors caused by that use.

1. http://www.ecma-international.org/ecma-262/6.0/#sec-generator-objects

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -330,6 +330,8 @@ type Iterable<T> = $Iterable<T,void,void>;
 interface Generator<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
+    return(value: Return): IteratorResult<Yield,Return>;
+    throw(error?: any): IteratorResult<Yield,Return>;
 }
 
 /* Maps and Sets */
--- a/lib/core.js
+++ b/lib/core.js
@@ -330,7 +330,7 @@ type Iterable<T> = $Iterable<T,void,void>;
 interface Generator<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
-    return(value: Return): IteratorResult<Yield,Return>;
+    return<R : Return>(value: R): IteratorResult<Yield,R>;
     throw(error?: any): IteratorResult<Yield,Return>;
 }
 
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -5695,11 +5695,11 @@ and mk_body id cx type_params_map ~async ~generator ?(derived_ctor=false)
       TypeAppT (promise, [VoidT.at loc])
     else if generator then
       (* The return value for generators can be provided by consumers via the
-         `throw` method, so it isn't correct to infer Return = VoidT in the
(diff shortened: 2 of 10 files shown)


lib/core.js
```diff
@@ -330,6 +330,8 @@ type Iterable<T> = $Iterable<T,void,void>;
 interface Generator<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
+    return(value: Return): IteratorResult<Yield,Return>;
```
gabelevi: Could we do something like

```JavaScript
return<R: Return>(value: R): IteratorResult<Yield,R>;
```

So this would allow us to write

```JavaScript
function foo(g: Generator<string, ?string, void>): IteratorResult<string, string> {
  return g.return('hello');
}
```

src/typing/type_inference_js.ml
```diff
@@ -5694,8 +5694,12 @@ and mk_body id cx type_params_map ~async ~generator ?(derived_ctor=false)
...
       TypeAppT (promise, [VoidT.at loc])
     else if generator then
-      let reason = mk_reason "return Generator<Yield,void,Next>" loc in
-      let ret = VoidT.at loc in
+      (* The return value for generators can be provided by consumers via the
+         `throw` method, so it isn't correct to infer Return = VoidT in the
+         absence of a return statement. Instead, return a tvar that can go on to
+         accumulate more lower bounds. *)
+      let reason = mk_reason "implicit generator return" loc in
+      let ret = Flow_js.mk_tvar cx reason in
```
gabelevi: It's ok to let it grow, but we probably should fllow `VoidT` nevertheless. So if you have

  function *g() {}
  g().return('hello'); // IteratorResult<void, string>
  g().next(); // IteratorResult<void, ?string>

gabelevi:
Actually are we overthinking it? Should we just do something like

 ```JavaScript
return(value: Return): { done: true, value: value };
```

I'm having trouble deciding why not, especially since a generator, once done, stops returning the value.

src/typing/type_inference_js.ml
samwgoldman: Yup. This is very true. Fixing this caught a bug in the tests, too.

lib/core.js
samwgoldman: Smart.

samwgoldman:
Regarding your "overthinking it" comment—maybe. I don't have the b/w to do due diligence on that, but if you do, feel free to add it in if/when you merge.

For the rest, I'm pushing up a couple commits...

gabelevi:
Sorry it took me so long to land this! It's in now, though! Thanks!

PR #798, opened 2015-09-14
setInterval requires ms argument

iamdustan:
Additionally, the return type for `setTimeout` and `setInterval` are different in node. Is there currently a clean way to support overloading the return value of setTimeout and setInterval in node?

In node, both timer methods return on object which has (amongst a few other properties) `ref` and `unref` methods. https://nodejs.org/api/timers.html#timers_settimeout_callback_delay_arg

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -548,7 +548,7 @@ declare function atob(encodedString: string): string;
 declare function clearInterval(intervalId?: number): void;
 declare function clearTimeout(timeoutId?: any): void;
 declare function setTimeout(callback: any, ms?: number, ...args: Array<any>): number;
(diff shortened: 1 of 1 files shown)


sebmck:
Actually these are optional. See the [timers spec](http://www.w3.org/TR/2011/WD-html5-20110525/timers.html#).

iamdustan:
Ah. Thanks @sebmck. So I suppose I should update the MDN docs instead :smile:: https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/setInterval

So disregarding the PR itself—is there a way to manage the `{ref, unref}` return object in node cleanly and (assuming the documentation is correct) mark the timeout values as required?

PR #837, opened 2015-09-21
Fix and Regression test for #819

### LINKED ISSUE #819: getContext("2d") should return CanvasRenderingContext2D
I'm currently using this typecast-ridden code to work with HTML `<canvas>`:

```javascript
var canvas = (this.refs.canvas.getDOMNode() : HTMLCanvasElement),
    ctx = ((canvas.getContext('2d') : any) : CanvasRenderingContext2D);
```

The `HTMLCanvasElement` cast would be hard to avoid, but the `CanvasRenderingContext2D` cast shouldn't be. The problem is [this declaration][1]:

```
type RenderingContext = CanvasRenderingContext2D | WebGLRenderingContext;
declare class HTMLCanvasElement extends HTMLElement {
    getContext(contextId: string, ...args: any): ?RenderingContext;
}
```

- If `contextId` is `"2d"`, `getContext` will return a `CanvasRenderingContext2D`.
- If it's `webgl`, `getContext` will return a `WebGLRenderingContext`.

If a string literal is passed to `getContext` (and it's hard to imagine a situation where it wouldn't be), then the return type should always be one of these, not the hard-to-work-with union type.

See also #61

[1]: https://github.com/facebook/flow/blob/25e6368959740e845cb962fe2843c503c7d6c4cc/lib/dom.js#L642

danvk:
Fixes #819 
See also #61 

This _should_ work, but I'm running into another bug. `canvastest.js` produces this error instead of the expected one:

```
canvastest.js:4:11,29: call of method getContext
Function cannot be called on
[LIB] dom.js:649:5,14: HTMLCanvasElement

Found 1 error
```

This is caused by overloading string parameters to `getContext` in `lib/dom.js`. If there's just one declaration, it works as expected.

It looks like something similar is going on for the existing `createElement` declarations. For example, this line:

```
var el = document.createElement('canvas');
```

produces this non-sense error:

```
canvastest.js:2:10,41: call of method createElement
Function cannot be called on
[LIB] dom.js:235:5,17: Document
```

@samwgoldman thoughts? Is this the same as #583?

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -677,7 +677,9 @@ type RenderingContext = CanvasRenderingContext2D | WebGLRenderingContext;
 declare class HTMLCanvasElement extends HTMLElement {
     width: number;
     height: number;
-    getContext(contextId: string, ...args: any): ?RenderingContext;
+    getContext(contextId: "2d", ...args: any): ?CanvasRenderingContext2D;
+    getContext(contextId: "webgl", ...args: any): ?WebGLRenderingContext;
+    getContext(contextId: string, ...args: any): ?RenderingContext; // fallback
     toDataURL(type?: string, ...args: any): string;
     toBlob(callback: (v: File) => void, type?: string, ...args: any): void;
 }
--- a/tests/canvas/.flowconfig
+++ b/tests/canvas/.flowconfig
@@ -1,0 +1,3 @@
+[options]
+module.system=haste
+
--- a/tests/canvas/canvas.exp
+++ b/tests/canvas/canvas.exp
@@ -1,0 +1,14 @@
+
+canvastest.js:8:3,22: call of method `moveTo`
+Error:
+canvastest.js:8:14,16: string
+This type is incompatible with
+[LIB] dom.js:660:15,20: number
+
+canvastest.js:8:3,22: call of method `moveTo`
+Error:
+canvastest.js:8:19,21: string
+This type is incompatible with
+[LIB] dom.js:660:26,31: number
+
+Found 2 errors
--- a/tests/canvas/canvastest.js
+++ b/tests/canvas/canvastest.js
@@ -1,0 +1,9 @@
+var el = document.createElement('canvas');
+var ctx: ?CanvasRenderingContext2D = el.getContext('2d');
+if (ctx == null) {
+  // error handling
+} else {
+  // ctx here is specifically a CanvasRenderingContext2D (not a WebGL context)
+  ctx.fillRect(0, 0, 200, 100);
+  ctx.moveTo('0', '1');  // error: should be numbers
+}


samwgoldman:
Smells like the nested union/intersection bug from #582. I had an inelegant fix in #583, but I think @bhosmer had a better idea for it. Basil, this issue has come up a lot (see the many linked issues in 582 and 583). Time for a fix?

danvk:
@samwgoldman @gabelevi any word on getting a fix for #582? I rebased this with the Flow 0.17.0 release and saw no changes.

As a reminder, the issue is that this line:

```
var el = document.createElement('canvas');
```

produces an error because Flow can't handle the overloads for `createElement` in `lib/dom.js`.

samwgoldman:
@danvk Unfortunately, no. If someone wanted to revive my fix from #583, it should unblock this PR. Not sure if the core team is interested in that.

danvk:
Sigh. It's really surprising that `lib/dom.js` includes untested definitions which are incompatible with Flow. (e.g. `document.createElement`.)

samwgoldman:
OK, so I'm sorry about this, but the unexpected error is actually unrelated to #582. I really should have looked into this, but I trusted my instinct, which turned out to be wrong.

To be clear, you are getting this error:

```
canvastest.js:2:10,41: call of method createElement
Function cannot be called on
[LIB] dom.js:235:5,17: Document
```

But you expected an error that more specifically explains that the argument to `moveTo` is wrong:

```
canvastest.js:10:3,22: call of method moveTo
Error:
canvastest.js:10:14,16: string
[LIB] dom.js:628:15,20: number
```

The first error is actually a result of the `string` / `number` incompatibility, but due to the advanced type inference that Flow performs, the error is attributed in a somewhat strange way.

I'll try to explain what's going on, but first let me show you how a small annotation can bring back the expected type error.

```javascript
var el = document.createElement('canvas');
var ctx: ?CanvasRenderingContext2D = el.getContext('2d');
if (ctx == null) {
  // error handling
} else {
  // ctx here is specifically a CanvasRenderingContext2D (not a WebGL context)
  ctx.fillRect(0, 0, 200, 100);
  ctx.moveTo('0', '1');  // error: should be numbers
}
```
```
canvastest.js:8:3,22: call of method `moveTo`
Error:
canvastest.js:8:14,16: string
[LIB] dom.js:660:15,20: number

canvastest.js:8:3,22: call of method `moveTo`
Error:
canvastest.js:8:19,21: string
[LIB] dom.js:660:26,31: number

Found 2 errors
```

Note that I added an annotation to the `ctx` variable declaration. So, why does adding an annotation change the error message so much? The answer lies in Flow's approach to intersection (and union) type checking.

When Flow encounters an intersection type, it goes into a special mode where type errors aren't immediately raised, but rather caught and handled. When an error is caught, Flow will try the next "branch" of the intersection type.

So, when the test calls `document.createElement('canvas')`, Flow enters this state, and dependent paths in the flow analysis do not immediately re-raise. The call to `ctx.moveTo` is ultimately dependent on the original `createElement` intersection branching. No branch of that intersection yields a type upon which `getContext` can be called to receive a type which in turns allows `moveTo` to be called with a string argument. That inconsistency "bubbles up" and we error on the intersection for `createElement` itself.

It's a fairly complicated thing to explain, so I hope that makes sense. Let me share another example:

```javascript
var el: HTMLCanvasElement = document.createElement('canvas');
var ctx = el.getContext('2d');
if (ctx == null) {
  // error handling
} else {
  // ctx here is specifically a CanvasRenderingContext2D (not a WebGL context)
  ctx.fillRect(0, 0, 200, 100);
  ctx.moveTo('0', '1');  // error: should be numbers
}
```
```
canvastest.js:2:11,29: call of method `getContext`
Function cannot be called on
canvastest.js:2:11,29: HTMLCanvasElement

Found 1 error
```

Note that here we annotate the `el` variable declaration instead of `ctx`. Notice how the error is similar to the original confusing error, but is attributed to `getContext` rather than `createElement`. The type annotation has the effect of simplifying the flow analysis, by making the upstream and downstream paths independent!

OK, so all that said, improving the error situation for intersection types and union types is a very well known issue and I don't think we should let it hold up this change. If you update your test to include an annotation on `ctx`, you can get an error that is expected. Once that change is in, I think this PR will be ready to merge in.

Again, sorry for misleading you about the root cause of this issue. If anything I've said here is unclear, please let me know so I can try to shed more light. Thanks for your contribution!

danvk:
Thanks for the explanation @samwgoldman. It makes sense why I got the error I did now, thought it is unfortunate. This sort of deferred error reporting reminds me of the infamous error messages associated with [C++'s templates SFINAE][1].

I've updated the test and this PR should be good to go.

I still don't understand why this produces an intersection type, though:

```javascript
var el = document.createElement('canvas');
```

given this declaration:

```
declare class Document extends Node {
    ...
    createElement(tagName: 'canvas'): HTMLCanvasElement;
    createElement(tagName: string): HTMLElement;
    ...
}
```

how could it result in anything _but_ an `HTMLCanvasElement`? Does Flow need a notion of the specificity of a declaration?

[1]: https://en.wikipedia.org/wiki/Substitution_failure_is_not_an_error

samwgoldman:
It's not quite precise to say that expression "produces an intersection type." Rather, That expression results in a type whose dependent paths are evaluated in the context of an intersection.

Here's a simpler example that shows the same behavior:

.flowconfig
```
[libs]
lib
```

lib/lib.js
```javascript
declare function f(x: string): number;
declare function f(x: number): string;
```

test.js
```javascript
/* @flow */

f('').length;
```

```
test.js:3
  3: f('').length;
     ^^^^^ function call. Function cannot be called on
  3: f('').length;
     ^ intersection type
```

Also interesting to note is that we get a different error message if the intersection type information is local to the file, rather than in a lib def. I haven't spent the time to investigate why this is.

When we call `f('')`, the return expression isn't an intersection type itself, but since `f` does have an intersection type, the expression is evaluated in what the code calls a "speculative match" process, described in my original comment.

I should add, this bad error message is not a necessary result of Flow's type inference. We can, and should, provide much better error messages. It's just an open issue, and kind of separate from this issue. The fix is out of my reach at the moment, but quite a few folks much more clever than I are working on Flow and might have some ideas.

danvk:
Yeah, that seems like a pretty bad error, too :scream: Hopefully we get a fix for this sort of thing at some point in the future.

In the meantime, this PR should be good to go. The build failure looks transient: https://travis-ci.org/facebook/flow/jobs/85177715. If an owner could re-run that test, it will most likely pass.

samwgoldman:
Thanks! Looks good to me :)

danvk:
Cool! What's the next step? Does a FB person need to take this over?

samwgoldman:
Yup. This is a pretty simple change (with a nice, fun, complex detour), so
hopefully you won't need to wait very long. :-)

gabelevi:
FB person to the rescue! Re-ran travis and it looks good!

@danvk thanks for the PR and @samwgoldman thanks for reviewing it!

gabelevi:
@facebook-github-bot shipit

gabelevi:
@facebook-github-bot shipit

PR #840, opened 2015-09-21
Fix incremental make on OSX when lib files change

gabelevi:
@danvk noticed that when you changed the lib files and than ran `make` (without a `make clean`) that the resulting binary would still use the old libraries. I looked into it and found an issue with incremental make on OSX.

Long story short, we want ocamlbuild to re-link the binary when the lib files change, since on OSX we rely on ocamlbuild to embed the lib files in the binary (`objcopy` is not available). However ocamlbuild doesn't understand that it needs to relink when non-ocaml dependencies change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,8 @@ EXTRA_LIB_PATHS=
 
 OS=$(shell uname -s)
 
+FLOWLIB=bin/flowlib.tar.gz
+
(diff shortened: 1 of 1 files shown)


danvk:
LGTM

mroch:
@facebook-github-bot shipit

PR #848, opened 2015-09-23
New terminal output format

frantic:
This commit introduces a new CLI output format that is more user-friendly:
- Source code is printed next to error messages
- Parts of the line that causes error is highlighted
- Each error is clearly separated from each other

*Disclaimer: I've never used OCaml before, any code structuring and style suggestions are very welcome*

### Outstanding issues

- [x] Tests are broken because they rely on the old style of output
- [x] Not sure what the deal is with `--one-line` attribute
- [x] Need to remove old code
- [x] If the terminal has no colors (or the output is redirected to file) error sections in lines are not highlighted. I haven't gotten to it yet, but I'm imagining using `^^^^` on the next line to underline tokens in question, but only if color is not available.
- [x] Printing relative paths
- [x] What if file comes from stdin?
- [x] Warnings vs errors. Looks like flow differentiates between warnings and error. We could incorporate this information into the output if it seems useful. I've experimented with a couple of approaches but neither felt right yet:
  - highlight parts of the lines in red vs yellow (might be confusing)
  - prefix each section with ERROR or WARNING (too intense)

### Demo

Before [see full output](https://dl.dropboxusercontent.com/u/4980295/flow/old-flow.html)

screen shot 2015-09-23 at 9 39 15 am

After [see full output](https://dl.dropboxusercontent.com/u/4980295/flow/new-flow.html)
screen shot 2015-09-23 at 9 38 59 am

### DIFF
--- a/src/common/errors_js.ml
+++ b/src/common/errors_js.ml
@@ -171,7 +171,7 @@ let print_file_at_location main_file loc s = Loc.(
     | Some filename ->
       let content = Sys_utils.cat filename in
       let lines = Str.split_delim (Str.regexp "\n") content in
-      let code_line = if (List.length lines) >= l0 then List.nth lines (l0 - 1) else "" in
+      let code_line = if (List.length lines) >= l0 && (l0 > 0) then List.nth lines (l0 - 1) else "" in
       let code_line = Str.global_replace (Str.regexp_string "\n") "" code_line in
       let highlighted_line = if (l1 == l0) && (String.length code_line) > 0
         then highlight_error_in_line code_line c0 c1
--- a/src/common/errors_js.ml
+++ b/src/common/errors_js.ml
@@ -162,6 +162,16 @@ let highlight_error_in_line line c0 c1 =
     source_fragment_style suffix;
   ]
 
+(* TODO: Ideally this would also work if the file was read from stdin *)
+let read_line_in_file line filename =
+  let content = try Sys_utils.cat filename with
+    | Sys_error _ -> ""
+  in
+  let lines = Str.split_delim (Str.regexp "\n") content in
+  if (List.length lines) > line && (line >= 0)
+    then List.nth lines line
+    else ""
+
 let print_file_at_location main_file loc s = Loc.(
   let l0 = loc.start.line in
   let l1 = loc._end.line in
@@ -170,11 +180,8 @@ let print_file_at_location main_file loc s = Loc.(
   match loc.source with
     | Some LibFile filename
     | Some SourceFile filename ->
-      let content = Sys_utils.cat filename in
-      let lines = Str.split_delim (Str.regexp "\n") content in
-      let code_line = if (List.length lines) >= l0 && (l0 > 0) then List.nth lines (l0 - 1) else "" in
-      let code_line = Str.global_replace (Str.regexp_string "\n") "" code_line in
-      let highlighted_line = if (l1 == l0) && (String.length code_line) > 0
+      let code_line = read_line_in_file (l0 - 1) filename in
+      let highlighted_line = if (l1 == l0) && (String.length code_line) >= c1
         then highlight_error_in_line code_line c0 c1
         else [source_fragment_style code_line]
       in
--- a/src/common/errors_js.ml
+++ b/src/common/errors_js.ml
@@ -217,7 +217,12 @@ let append_comment blame comment =
   | BlameM(loc, s) ->
     (match comment with
     | "Error:" -> BlameM(loc, s) (* Almost everywhere we have "Error:" that hurts readability *)
-    | comment -> BlameM(loc, s ^ ". " ^ comment))
+    | comment ->
+      let combined_comment = if String.length s > 0
+        then s ^ ". " ^ comment
+        else comment
+      in
+      BlameM(loc, combined_comment))
   | CommentM(_) -> failwith "should not be comment"
 
 (* TODO: Having a bunch of List.rev's is clowny, there must be another way *)
--- a/src/common/errors_js.ml
+++ b/src/common/errors_js.ml
@@ -36,6 +36,8 @@ type flags = {
   old_output_format: bool;
 }
 
+type stdin_file = (string * string) option
+
 let default_flags = {
   color = Tty.Color_Auto;
   one_line = false;
@@ -164,17 +166,28 @@ let highlight_error_in_line line c0 c1 =
     source_fragment_style suffix;
   ]
 
-(* TODO: Ideally this would also work if the file was read from stdin *)
-let read_line_in_file line filename =
-  let content = try Sys_utils.cat filename with
+let read_file_safe filename =
+  try Sys_utils.cat filename with
     | Sys_error _ -> ""
+
+let dbg a =
+  Printf.printf "%s\n" a;
+  a
+
+let read_line_in_file line filename stdin_file =
+  let content = match stdin_file with
+    | Some (stdin_filename, content) ->
+      if stdin_filename = filename
+        then content
+        else read_file_safe filename
+    | None -> read_file_safe filename
   in
   let lines = Str.split_delim (Str.regexp "\n") content in
   if (List.length lines) > line && (line >= 0)
     then List.nth lines line
     else ""
 
-let print_file_at_location main_file loc s = Loc.(
+let print_file_at_location stdin_file main_file loc s = Loc.(
   let l0 = loc.start.line in
   let l1 = loc._end.line in
   let c0 = loc.start.column in
@@ -183,7 +196,7 @@ let print_file_at_location main_file loc s = Loc.(
     | Some LibFile filename
     | Some SourceFile filename ->
       let line_number_text = Printf.sprintf "%3d: " l0 in
-      let code_line = read_line_in_file (l0 - 1) filename in
+      let code_line = read_line_in_file (l0 - 1) filename stdin_file in
       let highlighted_line = if (l1 == l0) && (String.length code_line) >= c1
         then highlight_error_in_line code_line c0 c1
         else [source_fragment_style code_line]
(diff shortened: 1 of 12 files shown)


src/common/errors_js.ml
```diff
@@ -106,6 +201,21 @@ let file_of_error err =
...
+let remove_newlines (color, text) =
+  (color, Str.global_replace (Str.regexp "\n") "\\n" text)
+
+let print_error_color ~one_line ~color error =
+  let level, messages, trace_reasons = error in
+  let messages = append_trace_reasons messages trace_reasons in
+  let messages = merge_comments_into_blames messages in
+  let header = print_error_header (List.hd messages) in
+  let main_file = file_of_error error in
+  let formatted_messages = List.map (print_message_nice main_file) messages in
```
avikchaudhuri: I think you're relying on a wrong assumption here: you're assuming that all messages point to the same file, but that's not necessarily true. You could easily have an error span multiple files. In which case the format needs to be:
`file1`
`code line` <- message
`file2`
`code line` <- message

etc.

You could still omit printing out a file name if it matches the previous one, which will make the output look like what you have now when you're dealing with errors within the same file.

Anyway, what this means eventually is that you can't always pass `main_file` to `print_message_nice`, you need to extract the file out from every message.

ianobermiller:
`--one-line` is useful to grep through the output, say, looking only for a certain file or path. Example command looking for error lines containing "liverail":

```bash
flow --color=always --one-line | grep liverail | sed 's/\\n/\n/g'
```

This preserves the color output and unescapes the newlines.

src/common/errors_js.ml
```diff
@@ -106,6 +201,21 @@ let file_of_error err =
   let loc = loc_of_error err in
   match loc.Loc.source with Some filename -> filename | None -> ""
 
+(* TODO: Is this actually needed? Does it make sense with the new format? *)
```
avikchaudhuri: We have this so that we can easily diff errors on a large codebase when making changes to Flow's behavior.

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+
+let append_comment blame comment =
+  match blame with
+  | BlameM(loc, s) ->
+    (match comment with
+    | "Error:" -> BlameM(loc, s) (* Almost everywhere we have "Error:" that hurts readability *)
+    | comment -> BlameM(loc, s ^ ". " ^ comment))
+  | CommentM(_) -> failwith "should not be comment"
+
+(* TODO: Having a bunch of List.rev's is clowny, there must be another way *)
```
avikchaudhuri: Right, here's a suggestion that would make this more efficient. Assume that `messages` is maintained in reverse order, so that `merge_comments_into_blames` below reverses the result of the fold.

Then make the following changes:

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+let maybe_combine_message_text messages message =
+  match message with
+    | BlameM (_, _) -> messages @ [message]
```
avikchaudhuri: This becomes `message::messages` (see above).

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+    | CommentM s ->
+      match List.rev messages with
```
avikchaudhuri: This becomes `match messages with`

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+      | x :: xs -> (List.rev xs) @ [append_comment x s]
```
avikchaudhuri: This becomes `(append_comment x s)::xs`

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+    file_location_style (Printf.sprintf "%s" relfilename);
+    default_style "\n"
+  ]
+)
+
+let append_comment blame comment =
+  match blame with
+  | BlameM(loc, s) ->
+    (match comment with
+    | "Error:" -> BlameM(loc, s) (* Almost everywhere we have "Error:" that hurts readability *)
```
avikchaudhuri: This is OK for now, but we need to remember on the error producing side to reserve the string "Error:" for its current purpose. Eventually we'll make the errors more structured so that this special case handling can be checked by the compiler.

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+let load_file f =
+  let ic = open_in f in
+  let n = in_channel_length ic in
+  let s = Bytes.create n in
+  really_input ic s 0 n;
+  close_in ic;
+  (s)
+
+(* TODO: Add proper support for relative path, for example ../../.. *)
+let relative_path filename =
```
avikchaudhuri: Could you check that `src/common/utils_js.ml` or `src/common/files_js.ml` doesn't already have something you can reuse? We like to keep path manipulation code in one place.

src/common/errors_js.ml
```diff
@@ -88,11 +88,106 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
-  print_reason_color ~first:true ~one_line ~color (List.hd messages);
-  List.iter (print_reason_color ~first:false ~one_line ~color) (List.tl messages)
+let file_location_style text = (C.Underline C.Default, text)
+let default_style text = (C.Normal C.Default, text)
+let source_fragment_style text = (C.Normal C.Default, text)
+let error_fragment_style text = (C.Normal C.Red, text)
+let line_number_style text = (C.Dim C.Default, text)
+
+(* TODO: Not sure how to handle the usecase when flow reads file content from stdin *)
+let load_file f =
```
avikchaudhuri: Probably OK for now, but this is going to be heavy eventually. We already cache ASTs somewhere, maybe we can cache contents too.

There's a util called `cat` that we use elsewhere, by the way. If you can't find it in `src/`, look for it in `hack/utils`. There should also be something that splits content into lines.

ianobermiller:
Once caveat with this output over something like `flowery` is that you must have color for the intra-line marking to be useful. If you instead point to something from the line below, you don't have this issue, plus you have more horizontal space to work with before wrapping and don't have error messages right aligned far away from the code. Taking one of the examples:

```
examples/01_HelloWorld/hello.js
  7: foo("Hello, world!");                              
	   ↑ function call
  7: foo("Hello, world!");                              
	     ↑ string. This type is incompatible with
  4:   return x*10;                                     
	          ↑ number

**Comment by avikchaudhuri** [comment]
This is very cool, by the way! And the OCaml code is very clean, I can't believe you're writing OCaml for the first time! :)

**Comment by avikchaudhuri** [comment]
For the tests, I suggest you retain the old behavior and put either the new or the old under a flag. Making the new format the default is totally fine, as long as runtests.sh can be modified to pass the right flag to get old output.

We can deprecate the old output later if we want, but I think it's still going to be used.

File: src/common/errors_js.ml (hunk 1 above)
Comment (frantic) [inline src/common/errors_js.ml]: Actually I do handle that case :) Look for "See:" line here: https://dl.dropboxusercontent.com/u/4980295/flow/new-flow.html

The idea is if the blame points to a different file we are printing "See: that/other/file.js". This way visually we can pretend it's one "main" file per error, and the error trace can still span different files.

**Comment by frantic** [comment]
Awesome, thanks for the feedback! I'll do another pass soonish.

Good to know about `--one-line`!

@ianobermiller I like flowery output too. Do you guys think I should be aiming for that? Having error text side by side is nice because it's compact and IMHO slightly easier to follow, however I also see benefits of using new lines for error explanations.

What do you think about warning vs error? I like the idea of just presenting everything as an error, otherwise it's easy to end up in situation where your code is okay but flow has warnings.

**Comment by ianobermiller** [comment]
@frantic, you have a good point about compact output, flow tends to have long error messages with multiple sites as it is, and we might not want three errors to fill your terminal's entire vertical space.

**Comment by frantic** [comment]
Addressed the comments and rebased. `incremental_basic` test is failing, but so it does on master. Is it expected?

**Comment by avikchaudhuri** [comment]
Awesome work, thanks!

**Comment by frantic** [comment]
Incorporated changes from `master`. Would you like me to squash the commits into one?

**Comment by frantic** [comment]
Alternative error output layout (improves readability when colors are turned off)

screen shot 2015-09-29 at 6 45 00 pm

**Comment by avikchaudhuri** [comment]
No need to squash, I think. I can import all your stuff at once.

**Comment by ianobermiller** [comment]
Nice! I love the alternate output.

File: src/common/errors_js.ml [hunk 10]
```diff
@@ -131,13 +133,89 @@ let print_reason_color ~first ~one_line ~color (message: message) =
...
+  let prefix = String.sub line 0 c0 in
+  let fragment = String.sub line c0 (c1 - c0) in
+  let suffix = String.sub line c1 ((String.length line) - c1) in
+  [
+    source_fragment_style prefix;
+    error_fragment_style fragment;
+    source_fragment_style suffix;
+  ]
+
+(* TODO: Ideally this would also work if the file was read from stdin *)
```
gabelevi: So what happens when you run

  flow check-contents < foo.js

? If that needs love, then I think you could actually pass along the stdin contents. Take a look at [where the code for check-contents calls into this stuff](https://github.com/facebook/flow/blob/master/src/commands/checkContentsCommand.ml#L55). You could probably pass the `file` into `print_error_summary`. Then you can compare the `file`'s name to the file you're trying to read. If `file` doesn't have a name, then it will probably show up in the error as `-`.

Here are more examples of check-contents: https://gist.github.com/gabelevi/d77c00701eb7dd3429d8

frantic:
@gabelevi done. See example here: https://gist.github.com/frantic/15697bc391db33d370f1

PR #855, opened 2015-09-24
Adds support for registerElement for web components

jamesgpearce:
Just adds the registerElement function as per http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -202,6 +202,12 @@ declare class HTMLCollection {
     [index: number]: Element;
 }
 
+// from http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register
+declare class ElementRegistrationOptions {
+    prototype?: any;
+    extends?: string;
+}
+
 declare class Document extends Node {
     documentElement: HTMLElement;
     implementation: DOMImplementation;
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -202,6 +202,12 @@ declare class HTMLCollection {
     [index: number]: Element;
 }
 
+// from http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register
+declare class ElementRegistrationOptions {
```
mroch: does this need to be a class, or can it just be an object type, e.g. `type ElementRegistrationOptions = {...`? it doesn't look like you `new` it

jamesgpearce:
:v:

jamesgpearce:
wait, I messed that up completely

lib/dom.js
```diff
@@ -263,6 +269,8 @@ declare class Document extends Node {
     xmlStandalone: boolean;
     xmlVersion: string;
 
+    registerElement(type: string, options: ElementRegistrationOptions): any;
```
gabelevi: I believe the spec says that `options` should be optional.

jamesgpearce:
yes indeed, thanks

gabelevi:
Great, thanks for addressing my nit! Merging...

Issue #596, opened 2015-07-01
Calling a method on a class method should probably fail


karelbilek:
I don't think this should typecheck correctly. (It has burned me right now in real code, and I think this is exactly the types of errors Flow should catch.)

``` js
class Foo {
    bar(): Promise<number> {
        return Promise.resolve(3);
    }
}

var foo = new Foo();

foo.bar.then(i => i+1)
```

You shouldn't be able to call a non-existing methods on class methods like that.

runn1ng:
(if you didn't notice, the error is in `foo.bar.then`, it should be `foo.bar().then`)

avikchaudhuri:
You're absolutely right. But loosening type checking here was a deliberate temporary step. We get requests to go the other way lots of times as well: since under the hood functions are objects, they often hold properties. In particular, at some point we starting seeing cases where static properties in a class were accessed via `this.constructor.foo`, and since constructor is just another method in Flow, to support that pattern we decided to loosen type checking on method statics altogether. This is obviously not the end of this story, thanks for reporting and especially for pointing out that this burned you in real code! Will fix.

runn1ng:
It didn't really burn me that hard, since browser recognized that `foo.bar.then` is not a function, but I was still surprised that Flow didn't catch that

gkz:
It's an error now (unbinding of `foo.bar`)

Issue #852, opened 2015-09-24
`new someInstance.constructor(...)` regression


popham:
The following fails:

```js
class T {
  n: number;
  constructor(n: number) {
    this.n = n;
  }
}

var t = new T(1);
var tp: T = new t.constructor(19);

/*
someFile.js:2:6,11: property `n`
Property not found in
someFile.js:3:3,5:3: new object

Found 1 error
*/
```

I mentioned it back in https://github.com/facebook/flow/pull/657, and I closed it as fixed.  If I remove the `T` annotation on `tp`, then Flow yields no errors.  This may be how I fooled into believing that it was fixed.  Or maybe it was fixed and has regressed.

samwgoldman:
@popham I just tried this on master and I don't get the same error. Can you confirm this is still an issue?

popham:
It's kinda fixed by a PR.

Issue #399, opened 2015-04-19
Support for chainable APIs aka `This` type


claudio-silva:
This code:
```js
/* @flow */

class A {
  x():A {
  	return this;
  }
}

class B extends A {
  y():B {
  	return this;
  }
}

var w = new B();

w.x().y();
```

Gives this error:

```
hello.js:17:1,9: call of method y
Property not found in
  hello.js:3:7,7: A
```
How can I express the return type of `A.x()` to be the same type as the instance's type where `x()` is invoked on?
This is very important for any javascript library that implements a chainable API where elements on the chain may subclass a base class and inherit chainable methods.

samwgoldman:
I don't have a solution here, but this sounds like Objective-C's `instancetype` to me.

@claudio-silva if you were writing in a different statically typed language, how would you write this?

claudio-silva:
Hi @samwgoldman, I don't know Objective-C, but doing a quick search it seems that `instancetype` is something quite similar to this.
As for other examples of this feature on other languages:
The Typescript team is investigating this issue and brainstorming solutions for it, as you can see in [this issue on GitHub](https://github.com/Microsoft/TypeScript/issues/513).
Scala has `this.type`. See [this article](http://scalada.blogspot.pt/2008/02/thistype-for-chaining-method-calls.html).
In PHPDocumentor, you can use the `$this` keyword as a function return type. See [this reference](http://phpdoc.org/docs/latest/guides/types.html#keywords). It is a very useful feature that allows you to get correct autocompletion on IDEs (ex: PHPStorm).
PHPDocumentor also defines `self` and `static` which have other meanings, but that may potentially also be useful for use with Javascript.

claudio-silva:
@samwgoldman I see you have closed this, but no reason was presented. Is this issue solved?

samwgoldman:
@claudio-silva Sorry about that. I closed this because we released support for `this` types in 0.20.0!

claudio-silva:
Great! Thanks!
But I can't find anything about it on the Flow documentation site!
Also, on the site's Blog section (a changelog), you've skipped version 0.20 and went from 0.19 to 0.21, so there is no mention of what's new on version 0.20!!!
Where can I find more info about these new features?

samwgoldman:
[...]

PR #857, opened 2015-09-24
Support `new this.constructor(...)` and `new x.constructor(...)`

popham:
This redirects `constructor` property lookups on class instances to the class rather than its constructor function.  With the ctor handled separate from other static methods, #596 can be closed.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -33,11 +33,8 @@ open Type
    meaningful in all contexts. This part of the design should be revisited:
    perhaps the data types can be refactored to make them more specialized. *)
 
-(* Methods may use a dummy statics object type to carry properties. We do not
-   want to encourage this pattern, but we also don't want to block uses of this
-   pattern. Thus, we compromise by not tracking the property types. *)
 let dummy_static =
-  AnyObjT (reason_of_string "object type for statics")
+  MixedT (reason_of_string "empty statics object")
 
 let dummy_prototype =
   MixedT (reason_of_string "empty prototype object")
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -2511,6 +2511,9 @@ let rec __flow cx (l, u) trace =
     | InstanceT (_, _, super, _), GetPropT (_, (_, "__proto__"), t) ->
       rec_flow cx trace (super, t)
 
+    | InstanceT _ as instance, GetPropT (_, (_, "constructor"), t) ->
+      rec_flow cx trace (ClassT instance, t)
+
     | InstanceT (reason_c, static, super, instance),
       GetPropT (reason_op, (reason_prop, x), tout) ->
       Ops.push reason_op;
--- a/tests/method_properties/method_properties.exp
+++ b/tests/method_properties/method_properties.exp
@@ -1,8 +1,14 @@
 
+test.js:5:11,28: property `x`
+Error:
+test.js:5:28,28: property `x`
+Property not found in
+test.js:1:7,7: statics of C
+
 test.js:7:1,3: property `x`
 Error:
 test.js:7:3,3: property `x`
 Property not found in
 test.js:1:7,7: statics of C
 
-Found 1 error
+Found 2 errors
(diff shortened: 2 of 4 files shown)


popham:
Issues: #852, #596

popham:
Upon reflection, I see that this is related to #399. I think that the complexity comes from the decision to admit class derivation where a subclass constructor does not necessarily subtype the superclass constructor. (I'm sure that this decision was the correct one.)

gabelevi:
I think this is also a good PR. Sure, when #399 comes along we'll beef this up even more, but this is a great start.

I think I'll take this PR except for the part that updates the dummy static object. We probably should make that type `MixedT` eventually, but it's somewhat unrelated and will cause errors in existing code when we land it.

gabelevi:
Thanks for fixing this, by the way! Merging now...

popham:
You sure?  I see backwards compatibility with this as a blocker to #399.  If constructors were constrained by subtyping rules like other class methods, then I think that implementing the `this` type would be trivial.  

As this PR stands, the following demonstrates a false positive:

```js
class A {
  constructor(n: number) { console.log(n); }
}

class B extends A {
  constructor(s: string, t: string) { console.log(s+t); }
}

function k(a: A): A {
  return new a.constructor(5); // No error.  :(
}

var a: A = k(new B("a", "b")); // No error.  :)
```

I'm okay with false negatives, but false positives like the body of `k` give me pause.  Also, if it comes down to a duel between `this` construction or `this` typing on classes, I would prefer to see `this` typing on classes win (it just landed in TS: https://github.com/Microsoft/TypeScript/pull/4910).  My feeling is that `this` construction is 10% to the 90% of `this` typing on classes.

After playing with this, my instinct is

* Add a `final` qualifier for constructors or
* Add a general annotations mechanism and then qualify constructors with an `lsp_ctor`-like flag, e.g. `constructor<; lsp_ctor>...`.  

Either of these options imply that subclasses are subtypes.  The above example becomes:

```js
class A {
  constructor(n: number) { console.log(n); }
}

class B extends A {
  final constructor(s: string, t: string) { console.log(s+t); }
  // XOR
  constructor<; lsp_ctor>(s: string, t: string) { console.log(s+t); }
}

function k(a: A): A {
  return new a.constructor(5); // Error:  Constructor on A is non-final or non-LSP.
}

var a: A = k(new B("a", "b")); // No error.  :)
```

samwgoldman:
I agree that it is useful and desirable to permit subclasses to have static signatures which do not exhibit a subtype relationship with the parent class's static signature, like @popham's constructor example above. 

I also agree that, as a consequence, it is unsafe to "reach out" to the static members from an instance, which, while annotated as an `A` instance, could be an instance of `A` or any of `A`'s subtypes.

I support reverting this change, and going further and evaluating the subtyping relationship for classes and instances in general. For example, if the general principle that I outlined above is correct, that subtyping should be against the instance-level (non-static) signature only, then what does a static member of an interface declaration mean? I'd argue that we shouldn't allow that, either.

popham:
I feel that only constructors should be allowed to violate subtyping.  Even that allowance annoys me, but I know that it's a necessary one.  (My knee-jerk opinion comes from my recent experience toying with the `this` type:  allowing further violations would make this more difficult under any strategies that I've come up with.)  Personally, I reject the remainder of @samwgoldman's argument on that basis.

Speaking as objectively as I'll manage....  In ES6 classes, statics behave identically to instance methods (prototype chaining up to `Function` instead of `Object`).  I think that statics under Flow should behave the same as nonstatics since there's little/no difference.  Why create a semantic division within Flow where none was imagined in ES6?  This tact is the path of least resistance, so the onus to come up with compelling reasons otherwise lies with detractors.  (A couple of months ago, I spent a few hours researching the gotchas that plague statics in other languages.  As far as I could tell, the common knowledge that "statics suck" doesn't extrapolate to ES6 classes.)

Out of curiosity, @samwgoldman or others that share his opinion, is this a source of incompatibilities with existing libraries, a spoiler of current plans, etc.?

P.S. Thanks for backing me up on the revert @samwgoldman.  I only wish you were a little quicker on the draw so I could avoid arguing against my own PR :).

Issue #822, opened 2015-09-19
`new this(...)` fails within class static methods


popham:
I want to create a class instance within a static method of the same class, e.g.

```js
class A {
  n: number;
  constructor(n: number) {
    this.n = n;
  }
  static create(n: number): A {
    return new A(n+1); // I'd like to generalize this to `new this...`
  }
}
```

In my particular use case, I'm extending the class many times to embellish it with different non-static methods; there is no need to include additional constructor logic.  I would like to avoid creating a `static create(...)` method within each of my extending classes, but the following fails:

```js
class A {
  n: number;
  constructor(n: number) {
    this.n = n;
  }
  static create<T>(n: number): T {
    return new this(n+1);
  }
}

class B extends A {
  someMethod(a: string): void {
    console.log(a);
  }
}

var a: A = A.create(5);
var b: B = B.create(5);

/*
someFile.js:7:12,24: constructor call
Constructor cannot be called on
someFile.js:1:7,7: statics of A

Found 1 error
*/
```

My extending classes are generated code, so this isn't a big deal in my case (no maintenance overhead and gzipping should vanish the verbose code from the wire).  The error message is obviously a false negative.

samwgoldman:
@popham This is resolved by your latest contributions, right?

popham:
Ugh.  Kinda.  Closing....

Issue #611, opened 2015-07-07
Constructor cannot be called on & Property not found in


bsr203:
Hi,

Below code produces following errors. Are they bugs, or not yet supported? The code is working functionally to get a singleton instance.

```
/* @flow weak */

var _ALLOW = false;

class A {
  name: string;

  static instance() {
    _ALLOW = true;
    try {
        global.__CACHE = global.__CACHE || {};
        if (global.__CACHE[this.name] === undefined) {
          global.__CACHE[this.name] = new this();
        }
        return global.__CACHE[this.name];
    } finally {
      _ALLOW = false;
    }
  }
}
```

errors

```
Found 4 errors
~/t/flow ❯❯❯ flow                                                                                            ⏎

/Users/bsr/tmp/flow/test.js:12:28,36: property name
Property not found in
/Users/bsr/tmp/flow/test.js:5:7,7: statics of A

/Users/bsr/tmp/flow/test.js:13:26,34: property name
Property not found in
/Users/bsr/tmp/flow/test.js:5:7,7: statics of A

/Users/bsr/tmp/flow/test.js:13:39,48: constructor call
Constructor cannot be called on
/Users/bsr/tmp/flow/test.js:5:7,7: statics of A

/Users/bsr/tmp/flow/test.js:15:31,39: property name
Property not found in
/Users/bsr/tmp/flow/test.js:5:7,7: statics of A
```

flow --version                                                                                  ⏎
Flow, a static type checker for JavaScript, version 0.13.1

samwgoldman:
What does `this` refer to in a static method call? I think the `name`-related errors are real. If you declare `static name: string`, they go away.

`new this()` is interesting. I believe `this` in a static method should be the constructor function, but we don't understand that in flow. I think that part of this is a bug in flow. @avikchaudhuri, does that seem right?

bsr203:
Hi Sam, thanks for the pointers. I have seen the usage of `this` in JS, and know doesn't allow in other statically typed languages. See a qn in [SO](http://stackoverflow.com/questions/11249037/javascript-this-in-static-methods), though it is poorly worded, some answers suggest it's ok to do so.

calebmer:
[...]

PR #858, opened 2015-09-24
Support `new this(...)` within static methods

popham:
Bind `this` within class static methods to the class itself rather than just the statics.  This is necessary to allow instantiating `this`.

I would still like to admit polymorphic statics that yield derived instances, e.g.

```js
class A {
  static create<T: A>(): T {
    return new this();
  }
}

class B extends A {}

var b: B = B.create(); // NG
```

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -5361,6 +5361,7 @@ and mk_class_elements cx instance_info static_info body = Ast.Class.(
         let ret = Flow_js.subst cx map_ ret in
         (* determine if we are in a derived constructor *)
         let derived_ctor = match super with
+          | ClassT (MixedT _) -> false
           | MixedT _ -> false
           | _ -> name = "constructor"
         in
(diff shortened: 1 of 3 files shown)


popham:
Issues: #611, #822

popham:
Upon reflection, I see that this is related to #399.  I think that the complexity comes from the decision to admit class derivation where a subclass constructor does not necessarily subtype the superclass constructor.  (I'm sure that this decision was the correct one.)

gabelevi:
I think this is ok. In the future, when we build the `this` type, we can build on top of this PR :)

gabelevi:
Ok, merging! Thanks for the contribution!

samwgoldman:
Unfortunately, this exhibits a subtyping issue as well, and I think we need to revert this.

```javascript
class A {
  constructor(s: string) {}
  
  static foo() {
    return new this("");
  }
}

class B extends A {
  constructor(n: number) {
    super(n.toFixed(2));
  }
}

B.foo();
```

(Note: subclasses inherit static members)

popham:
@samwgoldman +1.  Half baked.

Issue #850, opened 2015-09-23
Support ES7 class property initializers


bolinfest:
This would make it easier to follow the new ES6+ pattern for declaring React classes as described in http://babeljs.io/blog/2015/06/07/react-on-es6-plus/. In particular, this is needed for declaring `static propTypes`.

samwgoldman:
We're lucky to have @jeffmo, who is not only a Flow maintainer, but also the champion of this proposal to TC39.

My understanding is that we're not going to move on this until the proposal has made some more progress (which is in progress, presumably as I type), but I'll defer to Jeff.

bolinfest:
True, but our collective co-workers at FB would really like to write React in the style that I linked to today :) I'm optimistic.

On the bright side, as of Flow 0.16.0, it is at least possible to use this feature now and suppress the error from Flow, so the parser knows about this syntax, it just kinda ignores it.

Our `.flowconfig`:

```ini
[options]
suppress_comment=.*\\$FlowIssue.*
```

A React component:

```js
class FileTreePanel extends React.Component {
  // $FlowIssue https://github.com/facebook/flow/issues/850
  static propTypes = {
    initialWidth: PropTypes.number,
    onChangeExperimentalHgIntegration: PropTypes.func.isRequired,
    store: PropTypes.instanceOf(FileTreeStore).isRequired,
  };

  render() {
    return (
      <div className="nuclide-file-tree-deux-container">
        <PanelComponent
          dock="left"
          initialLength={this.props.initialWidth}
          ref="panel">
          <FileTree store={this.props.store} />
        </PanelComponent>

        <div className="nuclide-file-tree-deux-footer">
          <NuclideCheckbox
            checked={atom.config.get('nuclide-file-tree-deux.enableExperimentalVcsIntegration')}
            onChange={this.props.onChangeExperimentalHgIntegration}
            >
            &nbsp;Use experimental Hg integration
          </NuclideCheckbox>
        </div>
      </div>
    );
  }

  getFileTree(): FileTree {
    return this.refs['panel'].getChildComponent();
  }

  getLength(): number {
    return this.refs['panel'].getLength();
  }
}
```

LegNeato:
Some outside Facebook would like this too :wink:

LegNeato:
Also unfortunate that the workaround requires adding an extra `;` but I'll take it!

bolinfest:
Currently, it's possible to have Flow suppress all diagnostics about decorators by specifying the following in `.flowconfig`:

[...]

PR #860, opened 2015-09-25
Point class property initializers error message to #850

LegNeato:
This might be a good habit to get into?


mroch:
yeah, seems like a good idea

mroch:
@facebook-github-bot shipit

PR #861, opened 2015-09-25
Create `esproposal.class_property_initializers =warn|ignore` config option

LegNeato:
This was discussed a bit in https://github.com/facebook/flow/issues/850

### DIFF
--- a/src/common/flowConfig.ml
+++ b/src/common/flowConfig.ml
@@ -24,6 +24,7 @@ type experimental_feature_mode =
 type options = {
   enable_unsafe_getters_and_setters: bool;
   experimental_decorators: experimental_feature_mode;
+  experimental_class_property_initializers: experimental_feature_mode;
   moduleSystem: moduleSystem;
   module_name_mappers: (Str.regexp * string) list;
   munge_underscores: bool;
@@ -84,6 +85,7 @@ let default_module_system = Node
 let default_options root = {
   enable_unsafe_getters_and_setters = false;
   experimental_decorators = EXPERIMENTAL_WARN;
+  experimental_class_property_initializers = EXPERIMENTAL_WARN;
   moduleSystem = default_module_system;
   module_name_mappers = [];
   munge_underscores = false;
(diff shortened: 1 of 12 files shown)


LegNeato:
Note I haven't written OCaml since college, not sure if this is right but the tests all pass! :crystal_ball:

jamesgpearce:
:clap:

mroch:
@jeffmo if we go forward with static initializers before instance initializers, does this name work for you? presumably we could create a `static_class_property_initializers` option to let people enable those while disabling the instance ones, if they wanted to.

LegNeato:
:point_right:

jeffmo:
Awesome! Two small requests: 

1. Borderline nit: In order to stay in sync with spec terminology, let's start calling property declarations "field declarations"?
2. There's a chance that the semantics (and possibly syntax) of instance fields might continue to iterate, so it would be good to be able to toggle them separately from static field decls. As such, can we split this option into `esproposal.class_instance_fields` and `esproposal.class_static_fields`?

After this lands I'll start building on top of it to support a third "enable" option for the configs :)

jeffmo:
(Also I think you need to rebase on top of @mroch's latest tweaks in https://github.com/facebook/flow/commit/c81e41dda64819795fab15d0cf7864356c02334a )

jeffmo:
@LegNeato:are you working on following up with the tweak? Or should I just patch+tweak for you?

LegNeato:
@jeffmo: I'm a bit sick so probably won't get to it for a couple of days while I fight this thing. Feel free to take it on! If you don't have time I'll take a look when I am feeling better.

jeffmo:
Ok cool. I have a few things to finish up first, so I'll comment here if I end up nabbing it (otherwise feel free to finish it out when you feel better)

jeffmo:
FYI: Going to nab this from you and I can just tweak it on my end. Thanks for sending this over!

jeffmo:
FYI: I broke the config out into 2 and named them `esproposal.class_instance_fields` and `esproposal.class_static_fields` so that the two portions of the proposal can run independently (I'm starting work on a third "enable" option for both of these as well).

`class_static_fields` is less controversial than the `instance` counterpart, so I wanted to make it possible to toggle either independently depending on risk tolerance for new features.

PR #865, opened 2015-09-27
Add @@iterator to NodeList and HTMLCollection

jeffutter:
Add @@iterator support to HTMLCollection and NodeList. This adds support for things like:

```
let inputs = document.querySelectorAll('input, select');
for ( let el of billingInputOrSelect ) { ... }
```

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -170,6 +170,7 @@ declare class Node extends EventTarget {
 }
 
 declare class NodeList<T> {
+    @@iterator(): Iterator<T>;
     length: number;
     item(index: number): T;
     [index: number]: T;
@@ -196,6 +197,7 @@ declare class Attr extends Node {
 }
 
 declare class HTMLCollection {
+    @@iterator(): Iterator<HTMLElement>;
     length: number;
     item(nameOrIndex?: any, optionalIndex?: any): Element;
     namedItem(name: string): Element;


gabelevi:
I was confused since these things aren't iterable on Chrome, but they are on Firefox. Found http://stackoverflow.com/questions/31283360/are-htmlcollection-and-nodelist-iterables which answers this question. Since they're supposed to be iterable I think we can merge this

lib/dom.js
```diff
@@ -170,6 +170,7 @@ declare class Node extends EventTarget {
 }
 
 declare class NodeList<T> {
```
gabelevi: Unrelated to this diff, but I suppose we can constrain `T` like

```JavaScript
declare class NodeList<T: Node> {
```

gabelevi:
Ok, I think I understand the spec now.

* [This describes what it takes for something to be iterable](http://heycam.github.io/webidl/#es-iterators)
* [NodeList is explicitly iterable<Node>](https://dom.spec.whatwg.org/#nodelist)
* [HTMLCollection is implicitly iterable<Element>](https://dom.spec.whatwg.org/#htmlcollection) since it has length and a getter.

jeffutter:
@gabelevi is this the correct solution for @@iterable then? Or do we need to change how flow interprets for of, to consider anything with length and a getter iterable?

gabelevi:
This is right! I'll get it merged today!

gabelevi:
Merged! Thanks for the contribution!

PR #866, opened 2015-09-27
Add checkValidity() function HTMLFormElement and HTMLInputElement

jeffutter:
Both of these are defined in the HTML5 spec.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -662,6 +662,7 @@ declare class HTMLFormElement extends HTMLElement {
     name: string;
     target: string;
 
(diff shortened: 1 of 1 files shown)


gabelevi:
There seem to be 8 interfaces with this method, but I'm happy to take this PR to update the most commonly used two.

jeffutter:
I've added a few more of them. There are a handful more elements (that I have never heard of) which flow doesn't yet have any interfaces for. I have left those off.

gabelevi:
Ok great! Thanks for the contribution! Merging...

PR #868, opened 2015-09-28
Use full names for type parameters

ForbesLindesay:
Although errors regarding `getDefaultProps` still appear at the call-site rather than at definition, and don't mention `getDefaultProps`, at least the error message will contain some reference to DefaultProps.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -11,22 +11,22 @@
 
 /**
  * Base class of all React classes, modeled as a polymorphic class whose type
- * parameters are D (= DefaultProps), P (= PropTypes), S (= State).
+ * parameters are DefaultProps, PropTypes, State.
  */
-declare class ReactComponent<D, P, S> {
-  props: P;
-  state: S;
+declare class ReactComponent<DefaultProps, PropTypes, State> {
+  props: PropTypes;
+  state: State;
   refs: any;
   context: any;
 
-  getInitialState(): S;
+  getInitialState(): State;
   getChildContext(): any;
 
-  setProps(props: $Shape<P>, callback?: () => void): void;
-  replaceProps(props: P, callback?: () => void): void;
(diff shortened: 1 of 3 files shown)


gabelevi:
Awesome! It looks like some of the recorded test output includes these names, so you need to re-record those tests. Here's an easy way to do it

```Bash
cd /path/to/flow/project
make
./runtests ./bin/flow
for test in `find tests/ -name "*.out" | sed "s/\(.*\)\.out$/\1/"`; do mv "$test.out" "$test.exp"; done
```

ForbesLindesay:
Was hoping to get away without having to set up the entire tool chain locally:

```
echo "const char* const BuildInfo_kRevision = \"$(git rev-parse HEAD 2>/dev/null || hg identify -i)\";" > hack/utils/get_build_id.gen.c
ocamlbuild -ocamlc "ocamlopt   -ccopt -DNO_LZ4"\
		hack/fsevents/fsevents_stubs.o hack/heap/hh_shared.o hack/hhi/hhi_elf.o hack/utils/files.o hack/utils/get_build_id.gen.o hack/utils/get_build_id.o hack/utils/handle_stubs.o hack/utils/nproc.o hack/utils/realpath.o hack/utils/sysinfo.o src/embedded/flowlib_elf.o
/bin/sh: ocamlbuild: command not found
make: *** [build-flow-native-deps] Error 127
```

ForbesLindesay:
It should have been:

```
./runtests.sh ./bin/flow
```

ForbesLindesay:
The command to do the renames results in:

```
mv: rename tests//more_react/more_react.out.out to tests//more_react/more_react.out.exp: No such file or directory
mv: rename tests//new_react/new_react.out.out to tests//new_react/new_react.out.exp: No such file or directory
```

ForbesLindesay:
I've updated the two test cases manually.

PR #870, opened 2015-09-28
Lex object keys in normal mode

### LINKED ISSUE #829: Can't use `implements` as an object key
Not sure if it's related to just a library file but I'm currently unable to use reserved words as property names. I have the following declare class:

```javascript
declare class BabelNodeClassDeclaration extends BabelNode {
  type: "ClassDeclaration";
  id: BabelNodeIdentifier;
  body: BabelNodeClassBody;
  superClass?: void | BabelNodeExpression;
  decorators: any;
  typeParameters: any;
  superTypeParameters: any;
  implements: any;
}
```

Note the `implements` property name. This causes the parser to fail with:

```
[LIB] types.js:365:3,12: Library parse error:
[LIB] types.js:365:3,12: Use of future reserved word in strict mode
```

However if I wrap `implements` in quotes like:

```javascript
"implements": any;
```

then I get a different error, this time not from the parser of:

```
[LIB] types.js:522:3,11: Library type error:
[LIB] types.js:522:3,11:
illegal name
```

Is there anyway around this? I'm unable to change the name of that property and need to annotate it.

Any help is appreciated, thanks!

samwgoldman:
Fixes #829, where @sebmck couldn't use `implements` as the key in a
declared class type. This was happening because the key was lexed in
"type mode" which spat out a T_IDENTIFIER instead of a T_IMPLEMENTS, so
later on, in `Parser_flow.identifier_or_reserved_word`, the logic to
explicitly allow reserved words didn't fire, and we fell back to
`identifier`, which complains about the future reserved word.

After this change object keys in a type context are lexed in normal
mode, so the lexer will spit out the more specific token type, and
`identifier_or_reserved_word` can do its job.

This change does allow `@@foo` in object literals and class
declarations, which is absolutely wrong. However, we should really be
using [Symbol.iterator] instead of @@iterator, so adding a bunch of
error handling in the parser (tried it, was ugly) just didn't seem worth
it. Feel free to set me straight if I'm just being lazy... :)

### DIFF
--- a/src/parser/lexer_flow.mll
+++ b/src/parser/lexer_flow.mll
@@ -639,8 +639,11 @@ rule token env = parse
                        { illegal_number env lexbuf w (T_NUMBER NORMAL) }
   | wholenumber
   | floatnumber        { env, T_NUMBER NORMAL }
+
   (* Keyword or Identifier *)
-  | word as word       {
+  (* TODO: Use [Symbol.iterator] instead of @@iterator. *)
+  | ("@@"? word) as word
+                       {
                          unicode_fix_cols lexbuf;
                          try env, Hashtbl.find keywords word
                          with Not_found -> env, T_IDENTIFIER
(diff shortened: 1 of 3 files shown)


gabelevi:
Looks good to me :)

gabelevi:
Merged! Thanks for tracking this down!

PR #872, opened 2015-09-29
Default param values can reference earlier params

### LINKED ISSUE #827: Default parameter initialisers can't refer to other parameters
```javascript
/* @flow */

function foo(a, b = a) {}
```

Incorrectly reports an error of:

```
$ flow check

/Users/sebmck/Scratch/flow-test/index.js:3:21,21: identifier a
Could not resolve name

Found 1 error
```

Default parameter initialisers should allow reference to all parameters to the left.

samwgoldman:
Summary: Defer evaluation of default expr until mk_body

This commit introduces a module, `Params`, which encapsulates lots of
information that was passed around in a tuple of interdependent values
between the various functions that build up function types.

The original motivation for this refactor was to defer the evaluation of
the default expressions. Adding yet another value to the "params_ret"
tuple was a step in the wrong direction, so I tried something else.

This also fixes a case where users could force an unexpected BoundT
error, by adding a default expr to a generically-typed param. Now the
default expr is evaluated within the generate_test callback.

This commit is larger and noisier than I wanted it to be, which is a
consequence of the added module. The module depends on some functions,
and other functions depend on the module, but all of those functions
were simultaneously declared with `and`. In order to define the module,
I needed to split the declarations, which involved moving some
declarations up. If there is a better way to organize this with less
churn, I'd be happy to learn it.

Fixes #827

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -6089,8 +6089,9 @@ and mk_body id cx type_params_map ~kind ?(derived_ctor=false)
       Flow_js.flow cx (te, t)
     ) params;
     (* add to scope *)
-    let entry = Entry.new_var t ~loc in
-    Scope.add_entry name entry function_scope);
+    let reason = mk_reason (spf "param %s" name) loc in
+    Env_js.bind_implicit_let ~state:Entry.Initialized
+      Entry.ParamBinding cx name t reason);
 
   (* early-add our own name binding for recursive calls *)
   (match id with
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -409,15 +409,6 @@ let destructuring_assignment cx t =
     Env_js.set_var cx name t reason
   )
 
-(* instantiate pattern visitor for parameters *)
-let destructuring_map cx t p =
-  let tmap, lmap = ref SMap.empty, ref SMap.empty in
-  p |> destructuring cx t (fun _ loc name t ->
-    tmap := !tmap |> SMap.add name t;
-    lmap := !lmap |> SMap.add name loc
-  );
-  !tmap, !lmap
-
 (*
  * type refinements on expressions - wraps Env_js API
  *)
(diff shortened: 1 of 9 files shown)


samwgoldman:
Let me know if I should rebase this one—because I moved a few function definitions, we need to handle it carefully.

mroch:
can't merge unless you rebase so i'd say go for it. it's hard to tell what's a move and what changed but I don't think it'll be too bad to review. the only way I can think of to reduce churn/make this easier to review would be to do all of the moving in one commit, and then make changes in a separate one (even separate PRs, so we can stamp the first one quickly and avoid future merge conflicts)

samwgoldman:
@mroch Cool, done! The only conflict was @jeffmo's backtick stuff, which I preserved through the rebase.

samwgoldman:
(This diff will, however, conflict with my other open PR, #878—can cross that bridge when/if we come to it.)

samwgoldman:
I'm happy to rebase this again, but maybe ping me when it's ready for review, so I can just rebase once.

mroch:
#878 is in... sorry about the delay.

bashful dog

samwgoldman:
@mroch Thanks :) Rebased.

samwgoldman:
:point_right: poke

jeffmo:
Haha I just went through and did a rebase of some related stuff (basically put all class+method signature info into records with nicely named/consistent fields -- but should actually go nicely with this as well).

Anyway you've had this out for a while now (sorry about that :/ ) so I'll take the burden of rebasing mine onto this. I need to run for the night, but I'm leaving the tab open so I don't forget to come back to this tomorrow morning!

samwgoldman:
Thanks, Jeff!

FWIW, my only apprehension about the changes here is that I split the definitions in a semi-arbitrary way—I moved the definitions that I needed. There might be a more sensible way.

Another small thing that I think I got wrong here: If the default expressions can also refer to the binding to the function itself (I haven't checked), that entry should be added to the scope before the params. Currently the params are added first.

jeffmo:
The parameter scope is just a lexical scope whose parent is the outer lexical scope -- so indeed it is possible to refer to the function itself in the parameter list.

Here's the relevant spec text:  http://www.ecma-international.org/ecma-262/6.0/index.html#sec-functiondeclarationinstantiation

src/typing/type_inference_js.ml
```diff
@@ -5939,38 +6076,40 @@ and mk_body id cx type_params_map ~kind ?(derived_ctor=false)
...
-      add_entry (internal_name "yield") (new_entry yield) scope;
-      add_entry (internal_name "next") (new_entry next) scope;
-      add_entry (internal_name "return") (new_entry ret) scope
-    );
-    scope
-  in
+  let function_scope = Scope.fresh ~var_scope_kind:kind () in
 
+  (* push the scope early so default exprs can reference earlier params *)
   Env_js.push_var_scope cx function_scope;
```
jeffmo: Actually, parameter lists with default-value expressions should sit in their own intermediary scope in between the outer-scope and the toplevel inner-scope of the function.

This guarantees that any function expressions created in the parameter environment can't see any of the declarations that might sit in the function body:

```javascript
var a = 'outer';
function foo(param1=(function(){ console.log(a); })) {
  var a = 'inner';
  param1();
}
foo(); // Prints 'outer'
```

src/typing/type_inference_js.ml
```diff
@@ -5939,38 +6076,40 @@ and mk_body id cx type_params_map ~kind ?(derived_ctor=false)
...
 
+  (* add param bindings *)
+  params |> Params.iter Scope.(fun (name, t, loc) ->
+    (* add default value as lower bound, if provided *)
+    Params.with_default name (fun expr ->
+      let te = expression cx type_params_map expr in
+      Flow_js.flow cx (te, t)
+    ) params;
+    (* add to scope *)
+    let entry = Entry.new_var t ~loc in
```
jeffmo: I think you're just moving this around, but params should be defined using a let-binding (let-decls inside a function that collide with parameter names is an error)

jeffmo:
Other than the scope and binding things, I think this looks pretty good!

src/typing/type_inference_js.ml
samwgoldman: IIUC, we should use `let` bindings here to prevent argument name clash, even though doing so would not prevent the static error for function body `let`s (due to separate scopes)? I'm fine w/ that, just want to make sure we don't lose track of that static error either.

src/typing/type_inference_js.ml
samwgoldman: Good catch. Adding test...

samwgoldman:
@jeffmo So I added tests for all the cases you mentioned in your review and ensured this code exhibits the correct behavior. However, I did not add a separate lexical scope for params. The reason we don't need to is because we "evaluate" the default exprs before we process the function body, so bindings established therein are not visible. This behavior is somewhat non-obvious, so normally I would add the scope just for clarity. However, by not adding a separate scope, we are actually able to catch the error where a `let` binding in the fn body shadows a param. With a separate scope, this would be trickier to catch. LMK if you disagree with the tradeoff I made there.

src/typing/type_inference_js.ml
```diff
@@ -5939,38 +6076,41 @@ and mk_body id cx type_params_map ~kind ?(derived_ctor=false)
...
 
+  (* add param bindings *)
+  params |> Params.iter Scope.(fun (name, t, loc) ->
+    (* add default value as lower bound, if provided *)
+    Params.with_default name (fun expr ->
+      let te = expression cx type_params_map expr in
+      Flow_js.flow cx (te, t)
+    ) params;
+    (* add to scope *)
+    let reason = mk_reason (spf "param %s" name) loc in
```
jeffmo: Reminder about backticks (I'll just add with the merge, though -- no worries)

jeffmo:
This is awesome, thanks! Working on importing now...

bhosmer:
Hey @samwgoldman just did the behind-the-walls review for this - really nice work pulling the param stuff together! Good example of where it's worth the diff churn, I think (and I say this as someone who'll be rebasing against it :P). Thanks once again!

PR #878, opened 2015-09-30
Generator improvements: working generics and proper variance

samwgoldman:
The yield and next tvars were created outside of the generate_tests
callback, so naturally we were creating inconsistencies between the
different bounds tests. Each test should use a fresh tvar, because we
only need to be consistent within a test.

### DIFF
--- a/tests/generators/generators.exp
+++ b/tests/generators/generators.exp
@@ -9,10 +9,6 @@ class.js:15:8,8: number
 This type is incompatible with
 class.js:15:12,17: string
 
-class.js:23:39,44: number
-This type is incompatible with
-class.js:24:12,13: string
-
 class.js:24:12,13: string
 This type is incompatible with
 class.js:23:39,44: number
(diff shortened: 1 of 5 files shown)


tests/generators/generators.exp
```diff
@@ -9,10 +9,6 @@ class.js:15:8,8: number
 This type is incompatible with
 class.js:15:12,17: string
 
-class.js:23:39,44: number
```
gabelevi: It's not clear to me why this is no longer an error

tests/generators/generators.exp
```diff
@@ -63,10 +59,6 @@ generators.js:14:6,6: number
 This type is incompatible with
 generators.js:14:10,15: string
 
-generators.js:22:46,51: number
```
gabelevi: Same here

tests/generators/generators.exp
samwgoldman: This is a result of the variance change. If you look elsewhere in this file, you'll see the same error in reverse. We no longer unify, so we only get the error "one way." I called this out in the individual commit message, which I generally pack with insights, so don't miss out!

tests/generators/generators.exp
samwgoldman: Same here

PR #883, opened 2015-09-30
Type definitions for node's net module

JohnyDays:


JohnyDays:
Question: Is there a way to event emitter listener argument values based on the event?

lib/node.js
```diff
@@ -664,9 +664,47 @@ declare module "https" {
...
+
   declare function isIP(input: string): number;
   declare function isIPv4(input: string): boolean;
   declare function isIPv6(input: string): boolean;
+
+
+  declare type connectionListener = (socket: Socket) => any;
+  declare function createServer(
+    options?: {
+      allowHalfOpen: boolean,
```
mroch: i think these keys should be optional too. `net.createServer({pauseOnConnect: false})` is valid, right?

JohnyDays: Makes perfect sense, I'll add that

mroch:
@facebook-github-bot shipit

PR #889, opened 2015-10-01
Add MouseEvent

### LINKED ISSUE #774: Please include MouseEvent in lib/dom.js
It has a bunch of important properties beyond `Event`:

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

jeffutter:
Adding definition for MouseEvent based on docs:

https://w3c.github.io/uievents/#interface-MouseEvent  and https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

fixes #774

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -88,6 +88,28 @@ declare class Event {
     BUBBLING_PHASE: number;
 }
 
+declare class UIEvent extends Event {
+    detail: number;
+}
+
+declare class MouseEvent extends UIEvent {
+    altKey: boolean;
+    button: number;
+    buttons: number;
+    clientX: number;
+    clientY: number;
+    ctrlKey: boolean;
+    metaKey: boolean;
+    movementX: number;
+    movementY: number;
+    region: number;
+    screenX: number;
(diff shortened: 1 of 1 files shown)


mroch:
@facebook-github-bot shipit

lib/dom.js
```diff
@@ -88,6 +88,28 @@ declare class Event {
...
+    altKey: boolean;
+    button: number;
+    buttons: number;
+    clientX: number;
+    clientY: number;
+    ctrlKey: boolean;
+    metaKey: boolean;
+    movementX: number;
+    movementY: number;
+    region: number;
```
ssorallen: Looks like this can be `null`:

> Returns the id of the hit region affected by the event. If no hit region is affected, null is returned.

`region: ?number;`

mroch: thanks, fixed in https://github.com/facebook/flow/commit/00f6411eb721cc46ef89bdc8cc0274e15ead82b1

ssorallen: :beers:

Issue #823, opened 2015-09-19
Re-export default or as default both fail


cesarandreu:
I'm trying to re-export the default export of a module, but it's failing.

The following syntax is valid ES6 module syntax ([source](http://exploringjs.com/es6/ch_modules.html#leanpub-auto-re-exporting)):

```js
// Re-export the Module's default
export { default } from './Module'
```

```js
// Re-export Module as the default
export { Module as default } from './Module'
```

## Reproducible example:

**A/A.js**
```js
/* @flow */
export default function a (value: number): number {
  return value
}
```

**A/index.js**
```js
/* @flow*/
export { default } from './A'
```

**B.js**

```js
/* @flow */
import a from './A'
a(123)
a('123')
```

**.flowconfig**

```
[options]
module.system=node
strip_root=true
```

**Output**

```
A/index.js:2:10,16: Unexpected token default

B.js:2:15,19: index.js
Required module not found

Found 2 errors
```

lapanoid:
@samwgoldman Why it is closed? It will be fixed in 0.17? Have exactly the same now with 0.16.

PR #896, opened 2015-10-02
Allow reserved keywords in import/export when safe to do so

### LINKED ISSUE #828: default keyword not allowed as import name from sourced module declarations
Flow throws the syntax error "Unexpected token default" on the following input source code:

```javascript
export { default as foo } from "./bar";
```

```
$ cat test.js
export { default as foo } from "./bar";
$ flow ast test.js
{"errors":[{"loc":{"source":null,"start":{"line":1,"column":9},"end":{"line":1,"column":16}},"message":"Unexpected token default"}],"tokens":[],"type":"Program","loc":{"source":null,"start":{"line":1,"column":0},"end":{"line":1,"column":39}},"range":[0,39],"body":[{"type":"ExportDeclaration","loc":{"source":null,"start":{"line":1,"column":0},"end":{"line":1,"column":39}},"range":[0,39],"default":false,"declaration":null,"specifiers":[{"type":"ExportSpecifier","loc":{"source":null,"start":{"line":1,"column":9},"end":{"line":1,"column":23}},"range":[9,23],"id":{"type":"Identifier","loc":{"source":null,"start":{"line":1,"column":9},"end":{"line":1,"column":16}},"range":[9,16],"name":"default","typeAnnotation":null,"optional":false},"name":{"type":"Identifier","loc":{"source":null,"start":{"line":1,"column":20},"end":{"line":1,"column":23}},"range":[20,23],"name":"foo","typeAnnotation":null,"optional":false}}],"source":{"type":"Literal","loc":{"source":null,"start":{"line":1,"column":31},"end":{"line":1,"column":38}},"range":[31,38],"value":"./bar","raw":"\"./bar\""},"exportKind":"value"}],"comments":[]}
```

samwgoldman:
If we are doing a delegate export, the identifiers in the export
specifiers come from and go to object keys, so they follow the same
rules as object keys. If there is no "from" in the export statement, the
identifiers come from bindings, and thus follow those rules.

Import statements can rebind identifiers that come from object keys to
valid binding identifiers. If there is an "as" in the import binding
list item, the left-hand identifier can be a reserved keyword.

In order to allow the parse with the option of raising later, I modified
the signature of `identifier_or_reserved_keyword` to return the parsed
identifier along with the error that would have been raised. I factored
out the logic to determine this error from `error_unexpected`.

Fixes #828, #823

### DIFF
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -105,7 +105,7 @@ let rec filter_duplicate_errors acc = function
     filter_duplicate_errors acc rl
 | x :: rl -> filter_duplicate_errors (x :: acc) rl
 
-let unexpected_error = function
+let get_unexpected_error = function
   | T_EOF, _ -> Error.UnexpectedEOS
   | T_NUMBER _, _ -> Error.UnexpectedNumber
   | T_JSX_TEXT _, _
@@ -124,7 +124,7 @@ let error_unexpected env =
    * unexpected token *)
   error_list env (lookahead.lex_errors);
   clear_lookahead_errors env;
-  error env (unexpected_error (lookahead.lex_token, lookahead.lex_value))
+  error env (get_unexpected_error (lookahead.lex_token, lookahead.lex_value))
 
 (* Consume zero or more tokens *)
 module Eat : sig
@@ -1852,7 +1852,7 @@ end = struct
       match lex_token with
       | T_IDENTIFIER -> Parse.identifier env, None
       | _ ->
-        let err = match Peek.token env with
+        let err = match lex_token with
         | T_FUNCTION
         | T_IF
         | T_IN
@@ -1908,7 +1908,7 @@ end = struct
         | T_ASYNC
         | T_AWAIT
         | T_DEBUGGER ->
-            Some (lex_loc, unexpected_error (lex_token, lex_value))
+            Some (lex_loc, get_unexpected_error (lex_token, lex_value))
         | _ ->
             error_unexpected env;
             None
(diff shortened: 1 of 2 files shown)


src/parser/parser_flow.ml
```diff
@@ -1846,9 +1848,11 @@ end = struct
...
      *)
     and identifier_or_reserved_keyword env =
-      match Peek.token env with
-      | T_IDENTIFIER -> Parse.identifier env
-      | _ -> (match Peek.token env with
+      let { lex_token; lex_value; lex_loc; _ } = lookahead env in
+      match lex_token with
+      | T_IDENTIFIER -> Parse.identifier env, None
+      | _ ->
+        let err = match Peek.token env with
```
gabelevi: You could `s/Peek.token env/lex_token/` right?

src/parser/parser_flow.ml
```diff
@@ -105,6 +105,16 @@ let rec filter_duplicate_errors acc = function
     filter_duplicate_errors acc rl
 | x :: rl -> filter_duplicate_errors (x :: acc) rl
 
+let unexpected_error = function
```
gabelevi: The naming is a little unfortunate with `error_unexpected` right below. Maybe `make_unexpected_error` or `get_unexpected_error` or something to help the dyslexics among us (me)?

samwgoldman: Haha. I was definitely being a bit cute here, but I can see how this is confusing. `get_unexpected_error` works for me.

src/parser/test/hardcoded_tests.js
```diff
@@ -2590,5 +2590,22 @@ module.exports = {
         }
       }
     }
+  },
+  'Valid export/import reserved words': {
```
gabelevi: Wow this is complicated! Took me a few minutes to grok what was and wasn't allowed and why!

Issue #817, opened 2015-09-18
disjoint inference gets confused when entering a new function


gregwebs:
``` javascript
/* @flow */
type Action = Input | Save
type Input = { type: "Input"; input: number }
type Save  = { type: "Save";  save: number }

function update(action: Action): Function {
  if (action.type === "Save") {
      // type error here
      // fix by unwrapping before the thunk
      // var save = action.save
      return function() { return action.save.toString() }
  } else {
      return function(){}
  }
}
```

This is expected to typecheck but as per the comment fails

```
disjoint-union.js:11:41,44: property `save`
Property not found in
disjoint-union.js:3:14,45: object type

Found 1 error
```

samwgoldman:
Thanks for providing a small reproduction!

Say you exported the `update` function, and I imported it from somewhere else:

```javascript
var action = { type: "Save", save: 0 };
var func = update(action);
action = { type: "Input", input: 0 }; // this is legal, because `action : Action` still, but...
func(); // ...this would blow up, if we allowed the definition above
```

This is actually true of many refinement scenarios, and isn't specific to disjoint union refinement. There are some scenarios where Flow is overly pessimistic, but I don't think this is one of them. The workaround in the comment of your example does not suffer from the same aliasing problem, and is the recommended fix.

Would you take a moment and let me know if we can improve the [documentation of this behavior](http://flowtype.org/docs/dynamic-type-tests.html#caveats)?

gregwebs:
Ahh, makes sense that the possibility of mutation makes this tricky.

Still, the ideal would be to favor immutability and give the error where mutation occurs. In general for a disjoint union mutation of a piece of data in the object in the union could be common, but actual mutation from one type in the union to a different type in the union is going to be very rare (and error prone), so it should probably just be illegal. The normal use case is that a type in the union is created once and it needs to be tracked throughout the program; a new union type will mean a new object. Do you know of real-world use case where union type switching via mutation is desired? I am sure someone is doing it somewhere, but that code can probably live without tagged unions or live with the possibility that flow will not stop them from creating errors when mutating (the ideal would be to have a warning that can be ignored and silenced).

It would be good to move the disjoint union documentation from the blog post to the main site, or maybe I missed that. The caveat documentation seems good, I seem to have missed that or forgotten it.

The other aspect of this issue is that the current error message is pretty bad and makes me assume that Flow lost track of the disjoint union. The error message I would like to see is:

```
disjoint-union.js:11:41,44: property `save`
[...]

PR #901, opened 2015-10-03
Don't havoc const bindings

samwgoldman:
Since the value of the binding can't change, the narrowed specific type
can't be made false.


samwgoldman:
This is unfortunately too easy to be true. Since heap refinements are separate (and would be havoced regardless of this change), I was thinking that this was safe, because we never refine mutable objects (like `Array<?A>` to `Array<A>`). However, disjoint union refinement does refine the object type itself, and those refinements can be invalidated regardless of `const`-ness (#817). Alas!

PR #903, opened 2015-10-05
Make flow opt-out by default, add gradual=true and @noflow options

### LINKED ISSUE #284: Replace /* @flow */ with entry in .flowconfig
When introducing Flow to legacy code, it's nice to specify on a per-file basis what files Flow should analyze. Thus, we have the `/* @flow */` or `/* @flow weak */` comments at the top of each file. Once Flow is widely used in a project, I imagine that this would become extremely tedious. Or worse, someone may accidentally fail to include the comment and then we'll have no protection on that file with no warning that we don't. I suppose `flow check --all` exists for these cases, but that's an extremely inflexible solution.

Couldn't all these cases (analyze a few files, most files, or all files from a project) be solved more simply by providing a way to specify a glob or regex in .flowconfig that determines what files to analyze? That's what I assumed the `[include]` option was for, but it looks like it still requires the files listed to include the `/* @flow */` comment (so it's more for checking files in a folder outside the one where .flowconfig is located).

mroch:
Flow is currently opt-in on a per-file basis (requires you to add `@flow` to each file, in addition to adding a `.flowconfig`). This is great for existing projects, but not great for new projects that want everything to be flow; people have resorted to adding lint checks, for example.

This diff makes Flow opt-out by default (requires you to add `@noflow` to files you want to ignore, or add them to the `[ignores]` in `.flowconfig`). It also adds a `gradual=true` option if you want the current opt-in behavior.

I chose to make `gradual=false` the default even though it's a breaking change because I think it's a much better long term place to be: it's more likely that flow is being used on new projects than added to existing ones; it feels more "pit of success"-y (can no longer forget to add `@flow` and think everything is all good); and it's easy to turn off.

Fixes #284

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -74,10 +74,7 @@ do
         # get config flags.
         # for now this is kind of ad-hoc:
         #
-        # 1. The default flow command is check with the --all flag. This flag
-        # can be overriden here with the line "all: false". Anything besides
-        # "false" is ignored, and the setting itself is ignored if a command
-        # besides check is run.
+        # 1. The default flow command is check.
         #
         # 2. The default flow command can be overriden here by supplying the
         # entire command on a line that begins with "cmd:". (Note: for writing
@@ -93,18 +90,12 @@ do
         # as argument ($1). We start the server before running the script and
         # stop the server after the script exits.
         #
-        all=" --all"
         shell=""
         cmd="check"
         stdin=""
         stderr_dest="$err_file"
         if [ -e ".testconfig" ]
         then
-            # all
-            if [ "$(awk '$1=="all:"{print $2}' .testconfig)" == "false" ]
-            then
-                all=""
-            fi
             # ignore_stderr
             if [ "$(awk '$1=="ignore_stderr:"{print $2}' .testconfig)" == "false" ]
             then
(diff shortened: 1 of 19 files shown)


gabelevi:
Hmm...I think it's a good idea in general but I think there's another missing piece. The problem with opt-out is the libraries. There is no good way to opt-out a library. You shouldn't modify a library, plus adding @noflow to each file would be painful. And [ignore] doesn't solve the problem, since there is a difference between a @noflow file and an ignored file. Flow knows about the former and doesn't know about the latter.

I think you could fix this by changing `gradual` to instead be a list of directories (maybe even with globs). Something like

    [options]
    gradual_dir=src/**/node_modules
    gradual_dir=some/other/path

It could also be its own section

    [gradual]
    src/**/node_modules
    some/other/path

The idea being that a file is assumed @flow unless its path matches a gradual path.

jeffmo:
Having an option is clearly a good idea, but gradual=off seems like the wrong default. Anyone who wants to just try flow out (or even just decide later in the game if they want it for a project they aren't sure about) probably wants to apply it to a part of an existing project rather than the whole thing.

This is likely the most common case, so we should probably default to opt-in.

mroch:
gonna be a while before I come back to this so I'll close it in the meantime and resurrect it later.

chyzwar:
@mroch Can you look at this again? 

I am looking for a way to include only src folder in my project root. There is no clean way of doing this now. This would also resolve a lot of issues people checking node_modules because .flowconfig is usually in project physical root. 

I am looking for something like gradual or ability to specify project root.

PR #907, opened 2015-10-05
Correct return and arg types for Stream::pipe

### LINKED ISSUE #906: Incorrect `void` return for `Stream::pipe`
According to Node [`Stream::pipe`](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options) documentation:

> This function returns the destination stream, so you can set up pipe chains ...

However, 0.16.0's [node.js library](https://github.com/facebook/flow/blob/v0.16.0/lib/node.js#L769) annotates the return value as `void`.

ssorallen:
[`Stream::pipe`](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options) both takes a Writable as its destination and returns that Writable so the call to `pipe` can be chained.

Fixes #906

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -766,7 +766,7 @@ declare module "readline" {
 }
 
 declare class stream$Stream extends events$EventEmitter {
-  pipe(dest: any, options?: any): void;
+  pipe(dest: stream$Writable, options?: any): stream$Writable;
 }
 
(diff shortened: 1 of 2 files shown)


mroch:
looks right to me, thanks

mroch:
@facebook-github-bot shipit

PR #921, opened 2015-10-09
Allow yield without an argument

### LINKED ISSUE #912: Flow does not allow `yield` without an expression; I think it should.
I think I should be able to omit a yield expression, but Flow doesn't agree. It's happy when I `yield undefined;` though.

image

samwgoldman:
The only tricky bit here is that yield* expressions must have an
argument. Personally, I would have preferred a sum type in the AST, but
the estree spec has `Expression | null`, so I wanted to stay close to
that.

Fixes #912

### DIFF
--- a/src/parser/estree_translator.ml
+++ b/src/parser/estree_translator.ml
@@ -486,7 +486,7 @@ end with type t = Impl.t) = struct
       )
     | loc, Yield yield -> Yield.(
         node "YieldExpression" loc [|
-          "argument", expression yield.argument;
+          "argument", option expression yield.argument;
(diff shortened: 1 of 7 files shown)


mroch:
what about with ASI? the trailing semicolon's not guaranteed is it?

samwgoldman:
Oof. Good point. Adding a handful of parser tests to cover that...

samwgoldman:
@mroch Latest commit should resolve ASI issue. Thanks for catching that!

mroch:
@facebook-github-bot shipit

PR #933, opened 2015-10-10
Key to createHmac can be a buffer

runn1ng:
If it's not a buffer, it's converted to a buffer anyway, at least in crypto-browserify

see
https://github.com/crypto-browserify/createHmac/blob/master/browser.js

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -291,7 +291,7 @@ declare module "crypto" {
   declare function createDiffieHellman(prime_length: number): crypto$DiffieHellman;
   declare function createDiffieHellman(prime: number, encoding?: string): crypto$DiffieHellman;
   declare function createHash(algorithm: string): crypto$Hash;
-  declare function createHmac(algorithm: string, key: string): crypto$Hmac;
+  declare function createHmac(algorithm: string, key: string | Buffer): crypto$Hmac;
   declare function creatSign(algorithm: string): crypto$Sign;
   declare function createVerify(algorithm: string): crypto$Verify;
   declare function getCiphers(): Array<string>;


gabelevi:
Seems to check out in node too

```JavaScript
> require('crypto').createHmac('sha1', 'abc').update('hi').digest('hex')
'755f776df0652923d90a05a1e1c7cbcf98ce05d6'
> require('crypto').createHmac('sha1', new Buffer('abc')).update('hi').digest('hex')
'755f776df0652923d90a05a1e1c7cbcf98ce05d6'
```

gabelevi:
@facebook-github-bot shipit

gabelevi:
Thanks for fixing!

runn1ng:
By the way, when we are at it, a lot of the crypto stuff have "any" as input/output in /lib/ when it should be, in my opinion, Buffer. 

But I am not sure how much should I charge in and start exchange "any"s to Buffers; I don't want to break anyone's working code, and I am not sure how good is the test coverage on node crypto

(well I could also write some more test then, but well, time is time :) )

gabelevi:
That's one of the problems with `any`...it's hard to tell the difference between intentional `any`'s and I'll-get-to-that-later `any`'s. For a particular repository, you can define `suppress_type`'s in the `.flowconfig`, so we often use `$FlowFixMe` for the I'll-get-to-that-later case.

My best guess is that a lot of the `any`'s you're seeing in the crypto code are I'll-get-to-that-later `any`'s. Yes, fixing them may be a breaking change, but the breakages should be real type mismatches. So if you have the time and knowledge to fix the types up that would be awesome!

PR #941, opened 2015-10-13
anyfun anyobj typeof refi

### LINKED ISSUE #936: typeof not correctly inferring the type
```js
function add(a: number, b: number): number {
  return a + b;
}

type A = {
  b: number | Function,
}

const obj: A = {
  b: 1,
};

// won't error
if (typeof obj.b !== 'function') {
  add(obj.b, 2);
}
```

Incorrect type error:
```
src/Types.js:26
 26:   add(obj.b, 2);
       ^^^^^^^^^^^^^ function call
 26:   add(obj.b, 2);
           ^^^^^ function type. This type is incompatible with
 12: function add(a: number, b: number): number {
                     ^^^^^^ number
```

Flow 0.17.0.

samwgoldman:
Before this change, the two added test cases resulted in a type error,
because the conditional did not result in a refinement. This is because
`Function` and `Object` annotations are represented as `AnyFunT` and
`AnyObjT` internally, respectively. These types were erroneously
excluded from the pattern match `is_function` and `is_object`,
respectively.

Fixes #936

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3983,8 +3983,8 @@ and filter cx trace t l pred =
 
 and is_string = function StrT _ -> true | _ -> false
 and is_number = function NumT _ -> true | _ -> false
-and is_function = function FunT _ -> true | _ -> false
-and is_object = function (ObjT _ | ArrT _ | NullT _) -> true | _ -> false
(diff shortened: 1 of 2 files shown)


gabelevi:
Ooh, nice! Thanks for fixing this!

gabelevi:
@facebook-github-bot shipit

PR #943, opened 2015-10-14
Add String.prototype.codePointAt, String.fromCodePoint to lib

zpao:
These are part of ES6.


samwgoldman:
Looks correct to me. Thanks for updating the tests, too!

zpao:
Sure thing. I even succeeded in not complaining about how frustrating it was! (well, almost :P).

PR #955, opened 2015-10-19
Declare both formats of crypto [pseudo]randomBytes

ssorallen:
[`randomBytes`][0] and [`pseudoRandomBytes`][1] have synchronous and
asynchronous versions determined by arguments. If a single argument is
passed, a `Buffer` is returned synchronously. If a second function
argument is passed, the `Buffer` is asynchronously passed to the
function argument.

[0]:
https://nodejs.org/docs/latest-v0.12.x/api/crypto.html#crypto_crypto_randombytes_size_callback
[1]:
https://nodejs.org/docs/latest-v0.12.x/api/crypto.html#crypto_crypto_pseudorandombytes_size_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -310,13 +310,19 @@ declare module "crypto" {
     iterations: number,
     keylen: number
   ): string;
+  // `UNUSED` argument strictly enforces arity to enable overloading this
+  // function with 1-arg and 2-arg variants.
+  declare function pseudoRandomBytes(size: number, UNUSED: void): Buffer;
   declare function pseudoRandomBytes(
     size: number,
     callback: (err: ?Error, buffer: Buffer) => void
   ): void;
+  // `UNUSED` argument strictly enforces arity to enable overloading this
+  // function with 1-arg and 2-arg variants.
+  declare function randomBytes(size: number, UNUSED: void): Buffer;
   declare function randomBytes(
     size: number,
-    callback?: (err: ?Error, buffer: Buffer) => void
+    callback: (err: ?Error, buffer: Buffer) => void
   ): void;
 }
 


samwgoldman:
I haven't tested this, but I think the single-parameter declaration needs to come after the two-parameter declaration, otherwise any invocation will match the former and return a `Buffer`, even when the second parameter is a function matching the signature specified in the latter.

ssorallen:
Aha, I didn't realize that. I will test it both ways.

ssorallen:
@samwgoldman Actually neither way seems to get the desired response from Flow. It looks like Flow combines the signatures with union types rather than selecting a specific signature, so it's happy if you treat the return value as `void | Buffer` regardless of the arguments.

I tried the signatures in both orders.

ssorallen:
It also seems to combine the arguments into a single, union'd signature in a way that I can't quite comprehend. `pseudoRandomBytes(1)` is considered legal, but so is `pseudoRandomBytes(1, 1)`.

samwgoldman:
@ssorallen Hmm, I think you're right. There's an unfortunate interaction between intersection types (the natural representation of overloads), and Flow's treatment of functions (which is in line with JavaScript's runtime semantics).

It's a bit involved to explain, but I'll do my best here. Consider the following code:

```javascript
function f(a, cb) {
  if (typeof cb === "undefined") {
    // just return the number
    return a;
  } else {
    // callback
    cb(a);
  }
}
```

We'd like to declare this function as returning a number given some arguments. If the `a` is a `number`, no second argument is required. Otherwise, we must pass `b`, which converts `a` to one.

```
declare function f(a: number): number;
declare function f(a: number, cb: (a: number) => void): void;
```

The declaration above is the same as the following intersection type:

```
declare var f: ((a: number) => number) & ((a: number, cb: (a: number) => void) => void);
```

When Flow sees a function call, like `f(0, 0)`, it tries each branch of the intersection type in order. If there is an error, Flow will catch it and try the next branch, until we're out of branches. If every branch errors, Flow outputs an error. Flow stops once it finds a function that applies without errors.

In the first branch, we try to apply `(a: number) => number` with arguments `0` and `0`. This is "too many" arguments, but JavaScript allows this and Flow models this behavior of JavaScript. However, this means we stop here and Flow thinks there is no error, but there absolutely is (bug).

If we reversed the branches, we'd try the form with a callback first, and flow would realize that `0` does not inhabit `(a: number) => void`, so we'd fall back to the second branch, which would succeed (same reason as above).

So, reordering doesn't actually buy us anything here. My bad.

I think this is a bug in Flow, and unfortunately a blocker for this PR. We've kicked around the idea of being strict about argument length when processing intersection/union, which might work. Another idea might be to have Flow look at all branches of an intersection/union instead of myopically, one at a time.

@jeffmo We've had this discussion before, but this is the first time I've realized our current behavior makes typing certain functions impossible. Is this on anyone's radar internally?

ssorallen:
Thanks for the thorough explanation, @samwgoldman.

It does seem valuable to merge this still and be lenient on the signature rather than enforce an incorrect signature. Without this change, Flow shows an error for `pseudoRandomBytes(10)`, for example, which is a supported call by the API. A false negative seems better than a false positive.

avikchaudhuri:
One way to force Flow to be strict on function arity at calls is to have the function take an extra parameter of type void. Then only undefined args can be passed to such a function.

In this particular case it actually matches the code, too! :) Let me know if that works.

samwgoldman:
@avikchaudhuri Smart! Now why didn't I think of that... :P

ssorallen:
@avikchaudhuri Clever! I'll try that.

ssorallen:
The trick unfortunately doesn't work because adding a second parameter makes the functions have the same arity.

```
declare function foo(size: number, UNUSED: void): Buffer;
declare function foo(size: number, callback: () => mixed): void;
```

```
foo(5, () => 5);
> ✗ Error arrow function. This type is incompatible with number
> ✗ Error number. This type is incompatible with undefined.
```

So you could add *another* bogus argument to ensure the declarations have different arity, but this seems rather goofy.

```
declare function foo(size: number, UNUSED: void): Buffer;
declare function foo(size: number, callback: () => mixed, UNUSED: void): void;
```

```
foo(5, () => 5);
> ✓
```

What do you all think?

samwgoldman:
@ssorallen I just tried your first example and I didn't get the same error, nor can I see why you would have gotten the error you did. Can you double check that?

ssorallen:
@samwgoldman You're right. The test I was trying wasn't correct. The error from Flow is unfortunately difficult to understand with the intersection types, but it is correctly returning an error.

**lib.js**:
```
declare function foo(size: number, UNUSED: void): string;
declare function foo(size: number, callback: () => mixed): void;
```

**test.js**:
```
const fakeStr = foo(10, function() { return false; });
if (fakeStr.length > 5) {
  console.log('howdy');
}
```

**`flow check` output**:
```
test.js:1
  1: let fakeStr = foo(10, function() { return false; });
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function call. Function cannot be called on
  1: let fakeStr = foo(10, function() { return false; });
                   ^^^ intersection type
```

ssorallen:
@samwgoldman check out the diff again when you get a chance. I pushed an update.

samwgoldman:
@ssorallen Great! Looks good to me, although a FB employee will need to do the merge.

You're totally right that the error message is hard to understand. I got into a lot more detail in [a comment for another PR](https://github.com/facebook/flow/pull/837#issuecomment-147611103), if you're curious about that behavior.

mroch:
@facebook-github-bot shipit

PR #960, opened 2015-10-20
Added Number.isFinite to core.js

TiddoLangerak:
The [`Number.isFinite`](http://www.ecma-international.org/ecma-262/6.0/#sec-number.isfinite) function was missing from the `Number` class declaration, this PR adds it.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -78,6 +78,7 @@ declare class Number {
     toPrecision(precision?: number): string;
     toString(radix?: number): string;
     valueOf(): number;
+    static isFinite(value: any): boolean;
     static isInteger(value: any): boolean;
     static isNaN(value: any): boolean;
     static isSafeInteger(value: any): boolean;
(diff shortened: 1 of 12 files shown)


mroch:
thanks! could you update the tests that failed? they are all probably just failing because their line numbers got bumped by one, so a quick fix is to run the tests (`./runtests.sh bin/flow`) and then copy all of the `.out` files to `.exp`. I do that like this: `find tests/ -name "*.out" | perl -pi -e 's/(.*).out/cp $1.out $1.exp/g' | bash`

lib/core.js
```diff
@@ -80,6 +80,7 @@ declare class Number {
     valueOf(): number;
     static isInteger(value: any): boolean;
     static isNaN(value: any): boolean;
+    static isFinite(value: any): boolean;
```
mroch: nitpick: move this to line 81 to alphabetize it

TiddoLangerak:
I've updated the tests and moved the line up, it should build properly now!

mroch:
thanks!

mroch:
@facebook-github-bot shipit

PR #964, opened 2015-10-20
Add test for default destructuring parameter

### LINKED ISSUE #736: Flow doesn't look at destructuring assignment's default value in function parameter.
```js
/* @flow */

function foo(a : number = 'lol') {} // ok error

function foo2({n} : {n : number} = 3) {} // no error

function foo3({p} = 'lol') {} // no error
```

Comes from [here](https://github.com/facebook/flow/blob/0b3e461cb2121b75dd46cbd320309fc6af8826fc/src/typing/type_inference_js.ml#L5641-L5647), Flow doesn't typecheck default value if it's a destructuring assignment.

Eyyub:
Default parameter value for destructuring was not checked by Flow.

If there is a `default` value, be sure that it correctly typechecks and if not we do nothing to allow this : 
```javascript
function bar(a = 3) { }
bar('lol'); // no error

function foo({a} = {a:3}) { }
foo({a:'lol'}); // it seems that should not error too
```

### DIFF
--- a/tests/destructuring/destructuring.exp
+++ b/tests/destructuring/destructuring.exp
@@ -81,8 +81,16 @@ destructuring.js:45:6,8: property `baz`
 Property not found in
 destructuring.js:45:4,4: object literal
 
+destructuring.js:49:19,19: property `b`
+Property not found in
+destructuring.js:49:24,41: object literal
+
+destructuring.js:50:55,59: string
+This type is incompatible with
+destructuring.js:50:28,33: number
+
 destructuring_init.js:1:5,9: Destructuring assignment must be initialized
 
 destructuring_param.js:5:17,17: Strict mode function may not have duplicate parameter names
 
-Found 21 errors
+Found 23 errors
--- a/tests/destructuring/destructuring.js
+++ b/tests/destructuring/destructuring.js
@@ -46,4 +46,7 @@ function test() {
   (rest.baz: string); // no error, rest is unsealed
 }
 
+function shab({a, b} = {a : 2, c : 'foo'}) { }
+function toz({a, b} : {a : number, b : number} = {a : 'str', b : 0}) { }
+
 module.exports = corge;


samwgoldman:
I think #872 also handles this (incidentally). This is going to be a lot easier to add, I have no issues rebasing that on top of this, and we should definitely keep these tests, just FYI.

Eyyub:
As @samwgoldman 's PR is merged, this PR just add tests. (I can also open an other PR if rebasing like this is dirty)

jeffmo:
Awesome! No need to create another PR -- but it looks like the tests are failing... Maybe you need to rebase past samg's changes first?

(If another PR is easier for you, that's fine too...whatever causes you the least pain :) )

Eyyub:
Okey np, that was because of the branch name ! Actually the tests are failling because Flow doesn't  handle them correctly but samg will surely come with a magical PR soon to fix that :p

samwgoldman:
To recap some discussion from IRC:

```javascript
function shab({a, b} = {a : 2, c : 'foo'}) { }
```

I don't think `shab` should necessarily be an error. The default expr doesn't have a `b` key, so at runtime the value of `b` within the function will be `void`. We can either error or flow a `void` into the type of `b`.

```javascript
function toz({a, b} : {a : number, b : number} = {a : 'str', b : 0}) { }
```

This is definitely a bug, which the current code does not catch. I think I have a fix for this, but it depends on some code I added in #994.

mroch:
@Eyyub @samwgoldman what's the status here? I rebased this and neither line causes any errors. sounds like it's ok for the first one to do that, and the second one not erroring is a bug? can we file an issue to track that and commit the `shab` case to close this PR?

samwgoldman:
Haha—so, yeah. I had a fair amount of WIP on this, but I tossed it out and haven't revisited.

My take is that this PR is too hacky and we shouldn't merge it. We need to separate parsing of objects and parsing of destructuring. Right now we do this weird thing where we parse an object, then transform it. Object syntax doesn't have defaults, so that method no longer works.

Ultimately this just confuses parsing support for destructuring binding statements (`let { foo = bar } = obj`). It's actually significantly simpler to just handle the case for params, which happens to be where destructuring+default is most often found.

I can still loop back on this in the coming weeks, but if you are looking for something to do, please take it. If you do, I heartily recommend starting with the params case and punting on binding statements. That's just me, though.

samwgoldman:
Ack. I replied based on the email subject, but I think this is a separate issue from the one I was talking about...

If you want to commit the `shab` case, that sounds good, but would you add an expected error that shows that `b` is inferred as `void`? Also, `b` should still be widened by other callers and it'd be good to test that, too.

```javascript
function shab({b} = {}) {
  (b : string); // errors: void ~> string, number ~> string
}
shab({ b: 0 });
```

Eyyub:
Hi guys, just to let you know that I am not actually able to commit on this PR(no computer since 2 month now) :/

samwgoldman:
I'm going to close this out, but I'll be sure to test along this boundary as I land improvements to destructuring in the (near) future.

Eyyub, thanks a ton for your continued contributions. Hope you regain access to a computer so we can see more great stuff from you.

PR #974, opened 2015-10-23
Account for tabs and spaces in errors

### LINKED ISSUE #970: Error message code highlights don't work properly with tabs
In an error message where code is pointed out, the arrows or underlines are misplaced if the code is indented using tabs instead of spaces.

```javascript
// @flow
function works(){
    return x.y;
}

function doesntWork(){
	return x.y;
}
```

Outputs:
```
test.js:3
  3:     return x.y;
                ^ identifier `x`. Could not resolve name

test.js:7
  7: 	return x.y;
             ^ identifier `x`. Could not resolve name
```

kkirby:
Instead of depending on spaces, the code has been updated to take a substring of the errored line, replacing all non-whitespace with spaces. This way the line's original whitespace is kept intact.

Fixes #970

### DIFF
--- a/src/common/errors_js.ml
+++ b/src/common/errors_js.ml
@@ -244,8 +244,10 @@ let print_file_at_location ~root stdin_file main_file loc s = Loc.(
         then highlight_error_in_line code_line c0 c1
         else [source_fragment_style code_line]
       in
-      let padding_size = (String.length line_number_text) + c0 in
-      let padding = String.make padding_size ' ' in
+      let padding = String.make
+        (String.length line_number_text) ' ' ^
+        Str.global_replace (Str.regexp "[^\t ]") " " (String.sub code_line 0 c0)
+      in
       let underline_size = if l1 == l0
         then max 1 (c1 - c0)
         else 1


mroch:
thanks! leaving the whitespace alone and replacing all the other characters is a clever solution.

I'm fine with your solution. the only other solution I can think of is to replace `\t` with `  ` (2 spaces) in the line that gets printed, which doesn't seem too bad especially since we're only showing a single line of context (making indentation largely irrelevant). I think the only reason for that is it would make sure that lines take up a predictable amount of space in the console.

I'll wait a bit to merge it in case you like the above and want to try that instead, but like I said I think this is a good solution. Let me know :)

samwgoldman:
Another possibility would be to insert a tab into the padding string where there is a tab in the code, right?

mroch:
@samwgoldman yeah, that's basically what this is doing

samwgoldman:
Oh, hah. I like that. While we're in here, what about zero-/half-/double-width unicode characters?

mroch:
notime

samwgoldman:
Haha... on second thought, I think the parser only supports ASCII anyway.

mroch:
i'm just gonna go with this. you guys can send another PR if you want to change it again.

thanks @kkirby

mroch:
@facebook-github-bot shipit

PR #982, opened 2015-10-24
Allow null type annotation. (Fix #375)

### LINKED ISSUE #375: Null-only type annotation
I've got a generic class that I want to parametrize by `null`:

```
class List<T> {
  get(index: uint32): T {  throw new Error("Not implemented"); }
  // Map, For Each, Reduce, etc. (`core/lib`'s `Array` analogues)
}
```

```
class NullList extends List<null> { // NG
  get(index) {
    if (this.isOutside(index)) throw new RangeError();
    return null;
  }
}
```

I'll grant that the `NullList` is a boring edge case.  For now I'm vacillating between a `NullList` implementation with `T=void` + `return undefined` and defering to the base class's `get`.  Have I missed a mechanism to specify a null-only type?

Eyyub:
Allow `null` type annotation, ensure `number | null | void` is equivalent to `?number`.

```javascript
function bar(lol : ?number) { }

function foo(lol : number | null | void) { }

function foobar(lol : ?number) : number | null | void { return lol; }
function barfoo(lol : number | null | void) : ?number { return lol; }

var n : ?number = 3;;
var k : number | null | void = 3;
var p : number | null | void = n;

foo(n);
foo(p);

bar(k);
bar(p);
```

### DIFF
--- a/src/parser/estree_translator.ml
+++ b/src/parser/estree_translator.ml
@@ -828,6 +828,7 @@ end with type t = Impl.t) = struct
     match t with
     | Any -> any_type loc
     | Void -> void_type loc
+    | Null -> null_type loc
     | Number -> number_type loc
     | String -> string_type loc
     | Boolean -> boolean_type loc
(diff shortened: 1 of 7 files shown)


mroch:
this looks good to me. not sure if there's some reason why we didn't have a null annotation so I'm going to run this by the team first

mroch:
@facebook-github-bot import

PR #984, opened 2015-10-24
Add string constraint to object keys for unsealed objtypes

samwgoldman:
Before this change, `Object.keys(o)`, where `o` is not a sealed object,
would return effectively `Array<any>`, which is definitely unsound.

I ran into this when I tried to use `Object.keys` on an object-as-map.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -2059,12 +2059,17 @@ let rec __flow cx (l, u) trace =
           (reason_op, reason_o)
       )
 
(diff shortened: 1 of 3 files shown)


mroch:
@facebook-github-bot shipit

samwgoldman:
What... what happened here?

mroch:
cc @vjeux 

the bot freaked out. I eventually deleted the shipit comment to make it stop... it didn't even stop after the PR was closed.

PR #991, opened 2015-10-26
Object indexer should not imply callable signature

samwgoldman:
Callable signatures are implemented (in a kind of hacky way) by adding a
property named "$call" to the propmap of an object type. When an object
is called as a function, Flow looks for the "$call" property on the
object.

When the object has an indexer with a string key, this lookup would
succeed. I do not think it should succeed, because the property lookup
is part of the hacky nature of callable objects. Imagine if the propmap
stored the callable signature separately (as it should be -- what if I
want to make an object with a $call property?).

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3478,6 +3478,13 @@ and is_object_prototype_method = function
   | "valueOf" -> true
   | _ -> false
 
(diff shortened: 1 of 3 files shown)


mroch:
@facebook-github-bot shipit

PR #992, opened 2015-10-26
Docs: Add discord flow channel

hzoo:
Dono if we want to add it since it's not as big but yeah


gabelevi:
Are there people discussing Flow on Discord? I'm not sure if we should add various forums to the readme unless we intend to actively monitor them. For example, we registered a slack channel but don't advertise it because we don't really check it.

samwgoldman:
I've checked it out. Maybe ~10 messages in the last week. I'd prefer everyone stick to a single thing, and I like IRC, but I guess a lot of people consider IRC difficult.

hzoo:
Ok will close for now.

PR #994, opened 2015-10-26
Use heap refinements in destructuring assignment

### LINKED ISSUE #975: null typecheck refinement doesn't work on object destructuring
If you do a null check on a property, and then destructure that property and use it, Flow will error saying that you cannot do an operation on a possibly undefined value.

```javascript
/*  1 */ // @flow
/*  2 */ type foo = {
/*  3 */         x?: number,
/*  4 */         y?: number
/*  5 */ };
/*  6 */ 
/*  7 */ function doesntWork(a:foo){
/*  8 */     if(a.x){
/*  9 */         let {x,...others} = a;
/* 10 */         let calc = x * 12;
/* 11 */     }
/* 12 */ }
/* 13 */  
/* 14 */ function works(a:foo){
/* 15 */     if(a.x){
/* 16 */         let x = a.x;
/* 17 */         let calc = x * 12;
/* 18 */     }
/* 19 */ }
/* 20 */  
/* 21 */ doesntWork({x:1});
/* 22 */ works({x:1});
```

Output:
```
test.js:10
 10: 		let calc = x * 12;
     		           ^ undefined. This type is incompatible with
 10: 		let calc = x * 12;
     		           ^^^^^^ number
```

samwgoldman:
Basic premise is `{p} = o` is the same as `p = o.p`, so the same heap
refinement logic can apply.

Refinements are keyed on the member expression chain, so this code
builds up such an expression as it recurses through the pattern.

The destructuring helper now depends on the Refinement module, so I
pushed the module definition up. It is otherwise unchanged.

The TODO in the test is for this other issue I found, but did not
address here, where annotations for destructuring assignments are
checked against the RHS expression type instead of the destructured type
-- which may now be refined & different, but could always be rebound &
different:

```javascript
var o: { p: string } = { p: "" };
var { p: x } : { x: string } = o; // unexpected error
```

Fixes #975

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -292,133 +292,6 @@ let nameify_default_export_decl decl = Ast.Statement.(
   | _ -> decl
 )
 
-(**
- * Given a LHS destructuring pattern, extract a list of (loc, identifier-name)
- * tuples from the pattern that represent new bindings. This is primarily useful
- * for exporting a destructuring variable declaration.
- *)
-let rec extract_destructured_bindings accum pattern = Ast.Pattern.(
-  match pattern with
-  | Identifier(loc, n) ->
-    let name = n.Ast.Identifier.name in
-    (loc, name)::accum
-
-  | Object(n) ->
-    let props = n.Object.properties in
-    List.fold_left extract_obj_prop_pattern_bindings accum props
-
-  | Array(n) ->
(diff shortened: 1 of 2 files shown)


samwgoldman:
I fully accept that this might be too much ceremony for too little gain. If "proper" heap refinements are even a little bit close to landing, especially.

gabelevi:
@facebook-github-bot import

Issue #916, opened 2015-10-08
Flow does not support " export * as module from './module' "


lapanoid:
I am getting 

```
export * as module from './module'
         ^^ Unexpected identifier
```

jeffmo:
There is a [recent/pending proposal](https://github.com/leebyron/ecmascript-more-export-from) for `export * as` for some future ES.next (but note that this is not actually part of ES6).

This is on our todo list in any case.

lapanoid:
@jeffmo could this be suppresed?

I tried this

```
/* $FlowIssue */
export * as api from './api'
```

```
[options]
suppress_comment=\\(.\\|\n\\)*\\$FlowIssue
```

But still have the same error.

lapanoid:
This is basically the same and valid for flow 

```
import * as _api from './api'
export const api = _api;
```

jeffmo:
> @jeffmo could this be suppresed?

Ah, yea that's actually a separate bug where parse errors aren't properly suppressed by suppression comments (cc @gabelevi )

aintnorest:
@jeffmo is there another way to suppress errors?

jeffmo:
We don't currently have a way to suppress parse errors. This syntax is still a stage1 proposal so it's very new. We do intend to support it, it's just not been implemented yet.

I'll try to hop on this over the holidays if I find myself with some downtime -- it shouldn't be too difficult to add.

jeffmo:
FYI: `export * as ` went out in 0.21 behind the 'esproposal.export_star_as=enabled' .flowconfig option

https://github.com/facebook/flow/commit/57b41d09774dd658778349374fd16afd6f3a333f

JCMais:
@jeffmo is the issue mentioned above 

> Ah, yea that's actually a separate bug where parse errors aren't properly suppressed by suppression comments 

fixed?

nerdess:
sorry to step into this already closed issue!

i've got the flag

`esproposal.export_star_as=enable`

set in my .flowconfig 

but sth like

`export * as bla from 'blubb';`

gives me an "Syntax error: Unexpected token" (on the "as")...

btw: i am not familiar with flow, i am just trying to get my head around some code that was given to me.

nerdess:
ok i fixed the code now by changing this

`
// @flow
export * as bla from 'blubb';
`

to this

`
// @flow
/* eslint-disable import/first */
import * as bla from 'blubb';
export {bla};
`

[...]

Issue #606, opened 2015-07-06
ES7 decorator support or a way to ignore the error


bsr203:
Hi,

ES7 decorators cannot be used with flow. Wish there is a way to continue using `flow` but may just ignore this feature. Currently it stops the server with error.

```
Unexpected token ILLEGAL

```

Error is pointing to the `@` character

```
@connectToStores(getState, TestStore)
export default class MyComponent extends React.Component {
```

nmn:
https://github.com/facebook/flow/issues/454#issuecomment-119524118

majhork:
try this:
```js
class Foo {
    /*::_(){`*/ @deco /*::`;}*/
    bar() {
    }

    /*::_(){`*/ @long_deco(
      params // ...
    ) /*::`;}*/
    baz() {
    }
}
```

nrako:
:fearful: oh my gosh, that is definitely not a viable solution.

majhork:
It isn't pleasant but it should be easy to find/replace when they support decorators later and it works right now.

nrako:
I agree that is an increasing pain point which at the moment hold me back from using Flow.

But for the sake of DX, I would rather see decorator ignored by the Flow parser than adding more flow comments...

Anyway, in the long term that doesn't solve the problems, however what is discussed in #454 might. I just hope we'll start to see a clear implementation details of that.

nmn:
I think for Flow's sake, it should get out of the business of implementing ES6 features entirely and offload that task to Babel.

This would mean that flow should switch to the AST that Babel generates, and do it's thing on that. Babel is already way ahead implementing new ES6 features. It would be a considerable amount of work in the short term, but would free up time for more important stuff in the future.

I'm not sure how difficult it will be to integrate the JS-based Babel into the OCaml based Flow.

AgentME:
@nmn ES6 is set in stone at this point iirc, so there's no longer an issue of Flow chasing a moving target. (And Babel is pretty far along in ES6 support.) A lot of ES6 features would be pretty complicated to support correctly if Flow only operated on the transpiled source. Babel renames scoped let/const variables when there are conflicts, and classes come out pretty mangled. Flow could be made to recognize the specific way that classes get transpiled, but then it probably would fail to recognize the way that classes with decorators get mangled unless it was changed to specifically recognize that. But then if there needs to be code added to Flow to recognize decorators, it might as well be made to just recognize (or ignore) the decorators directly.

[...]

PR #1001, opened 2015-10-28
Add failing test for class-level esproposal_decorators

idolize:
The `esproposal_decorators` option does not currently include decorators applied to the class itself.

This change includes a **FAILING test** of class-level decorators to ensure proper support is added.

### DIFF
--- a/tests/esproposal_decorators.ignore/test.js
+++ b/tests/esproposal_decorators.ignore/test.js
@@ -1,10 +1,11 @@
 /* @flow */
 
+@decorator1
 class Foo {
-  @decorator1
+  @decorator2
   method1() {}
 
-  @decorator2
   @decorator3
+  @decorator4
   method2() {}
 }


jeffmo:
We didn't include this syntax in the flag because it's still not clear if it will be part of the actual proposal. There has been discussion about it, but no clear decisions have been made about how to address the potential caveats involved. Specifically: (1) That it would be possible to decorate a class declaration, but not a function decoration and (2) That toplevel decorators would require an entirely different param/return signature than "property" signatures currently do.

So until there's further discussion that makes it clear that this is likely to move forward with the rest of the decorators proposal we should at least gate this behind a separate flag.

Moreover, if we go the route of enabling this (even if behind a separate flag) we need to add support for it to the parser and inference code in addition to adding it to a test.

cc @wycats to chime in on thoughts about a sub-flag

idolize:
@jeffmo Thanks for the response! I realize this PR is woefully incomplete, but I wasn't aware of the contention surrounding class-based decorators vs. function decorators, so thanks for elaborating on that!

As an aside: I totally understand the reasoning to hold off on adding this as a feature given the long-term uncertainty, but the JS (and specifically React) community is already utilizing decorators on class declarations pretty heavily at the moment.

So, being pragmatic about it, having some kind of simple escape hatch here would go a *long way* towards unblocking use of flow with existing ES7 + babel codebases in the short term. (Especially given that [right now we can't even use `suppress_comment`s](https://github.com/facebook/flow/issues/916#issuecomment-146563896) to stop the flood of errors).

jeffmo:
Totally! I'm not opposed to adding an ignore flag for it like property decorators -- I just want to get a grip on how likely it is to ship with the existing proposal vs defer to a secondary/additive proposal (and thus deserve its own flag).

As for suppress_comments not suppressing parse errors -- that's certainly a known bug that we need to address regardless of what we do here. (cc @gabelevi)

stephanos:
@idolize are you working on the ignore flag for class decorators or did you find another workaround?

idolize:
@stephanos Neither, unfortunately. I don't know enough OCaml to write it myself, and I was unable to find a suitable workaround in the meantime.

STRML:
As decorators are coming into wider use in React projects (Redux uses it everywhere, and users still on mixins use it with react-mixin), it would be really great to have at least an ignore flag for this.

cheapsteak:
Are there any updates on this, or has anyone found a workaround?
Was hoping to use Flow on our new project but not being able to use decorators is a fairly large deal

vjpr:
@cheapsteak See #606. Its ugly but I've found its not too bad, and I've bound it to a surround live template in IntelliJ.

And Flow is working really well for me now, and catching tons of little bugs :)

```
/*::`*/ @deco /*::`*/
class Foo { ...
```

samwgoldman:
Parsing support for decorators is absolutely on my radar. Clearly the community has decided to adopt this, so I think it's a reasonable time commitment to build in parsing support.

This PR is in a weird position, because we can't accept it—it will break the build. I'm going to close this for that reason. On the bright side, the change here is quite simple, and we will certainly add this back in once we add parsing support.

TL;DR, message received. This is on the list. Thanks for providing your feedback, everyone.

tomatau:
Is there an issue/branch we can track to be notified when this is addressed?

PR #1010, opened 2015-10-30
Dump param types

samwgoldman:
This makes it possible to look up the type of a param using infer-type
or dump-types. With this change, vim-flow and Nuclide are able to show
the type of a function parameter, which I find to be incredibly useful.


mroch:
@facebook-github-bot shipit

PR #1011, opened 2015-10-30
Define WeakSet in lib/core.js

### LINKED ISSUE #832: WeakSet is not defined in lib/core.js.
However, `WeakMap` is...

paulcapron:
Closes #832.

I’m sorry I didn’t add tests, but I didn’t know where to put them. There are some tests for `Set` in `tests/iterable`, but a `WeakSet` is *not* enumerable, so this wasn’t a right place. And there are no test for `WeakMap`. Maybe `tests/weakstructs` should be created?

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -386,6 +386,12 @@ declare class Set<T> {
     values(): Iterator<T>;
 }
 
+declare class WeakSet<T: Object> {
+    add(value: T): WeakSet<T>;
+    delete(value: T): boolean;
+    has(value: T): boolean;
+}
+
 /* Promises
    cf. https://github.com/borisyankov/DefinitelyTyped/blob/master/es6-promises/es6-promises.d.ts
 */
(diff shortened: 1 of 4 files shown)


lib/core.js
```diff
@@ -386,6 +386,12 @@ declare class Set<T> {
     values(): Iterator<T>;
 }
 
+declare class WeakSet<T: Object> {
```
mroch: i think you need to add a constructor, like

```
<V>(iterable: ?Iterable<V>): WeakSet<V>
```

not sure how that works when the param is missing... hopefully we leave it unresolved and the first call to `add` will lock it down. cc @avikchaudhuri

samwgoldman:
I noticed the missing constructor, too. I looked around and realized that `Set` and `WeakMap` are also missing constructors.

mroch:
@facebook-github-bot import

mroch:
no worries, I added the constructor(s) and some tests. thanks @paulcapron

PR #1024, opened 2015-11-02
Tweak testing instructions

dmnd:
`Skipping directory` isn't in the test output, but `SKIP` is.


mroch:
@facebook-github-bot shipit

Issue #475, opened 2015-05-28
Ignores React prop types for createClass between modules


mhagmajer:
Flowtype only checks prop types for pre-0.13 React classes defined in the same file.
See this short and simplified example below for details.

`welcome.jsx`
```javascript
/* @flow */

var React = require('react');

var Welcome = React.createClass({
  propTypes: {
    name: React.PropTypes.string.isRequired,
  },

  render: function(): any {
    return <b>{this.props.name}</b>;
  },
});

module.exports = Welcome;
```

`main.jsx`
```javascript
/* @flow */

var React = require('react');

var Welcome = require('./welcome');

var WelcomeLocal = React.createClass({
  propTypes: {
    name: React.PropTypes.string.isRequired,
  },

  render: function(): any {
    return <b>{this.props.name}</b>;
  },
});

type PropTypes = {
  name: string,
};

class WelcomeLocalClass extends React.Component<any, PropTypes, any> {
  render() {
    return <b>{this.props.name}</b>;
  }
};


var Main = React.createClass({
  render: function() {
    return <div>
      <Welcome />
      <WelcomeLocal /> {/* only this one fires */}
      <WelcomeLocalClass />
    </div>;
  },
});
```

`flowtype`
```
[...]/main.jsx:32:7,22: React element: WelcomeLocal
Error:
[...]/main.jsx:8:3,11: property name
Property not found in
[...]/main.jsx:32:7,22: type parameter D of React element: WelcomeLocal

Found 1 error
```

Expecting errors should include all three components: `Welcome`, `WelcomeLocal` and `WelcomeLocalClass`.

mhagmajer:
CC @avikchaudhuri, who I believe worked on React support for Flow

samwgoldman:
For the imported Welcome component, I can confirm that prop checking stops working across module boundaries (it does work within the same file).

Regarding your WelcomeLocalClass implementation, I am able to get type checking with this equivalent construction:

```javascript
class WelcomeLocalClass extends React.Component {
  props: PropTypes;
  static defaultProps: {};

  render() {
    return <b>{this.props.name}</b>;
  }
};
```

Curiously, the React.Component subclass does type check across module boundaries. Not sure what's going on there.

mhagmajer:
Thanks @samwgoldman for your response! Prop types in components based on ES6 classes do in fact type check correctly with the syntax you suggested. Interestingly enough if that `defaultProps` declaration is required for this to work.

[...]

PR #1025, opened 2015-11-02
Test proptypes typecheck across module boundaries

dmnd:
I'm trying to get proptypes to typecheck across module boundaries. But
despite https://github.com/facebook/flow/issues/475 it doesn't seem to
work for me. I'm not sure if I'm doing something wrong, or if there has
been a regression since that discussion, or what.

I created these test cases to try to work out what is happening. Note
that only the locally defined components typecheck and correctly report
that the name prop is missing.

The identical components from a sibling file don't. Why?

### DIFF
--- a/tests/react_modules/createclass-callsite.js
+++ b/tests/react_modules/createclass-callsite.js
@@ -1,6 +1,6 @@
 /* @flow */
 var React = require('react');
-var Hello = require('./createclass-callsite');
+var Hello = require('./createclass-module');
 
 var HelloLocal = React.createClass({
   propTypes: {
--- a/tests/react_modules/createclass-callsite.js
+++ b/tests/react_modules/createclass-callsite.js
@@ -1,0 +1,26 @@
+/* @flow */
+var React = require('react');
+var Hello = require('./createclass-callsite');
+
+var HelloLocal = React.createClass({
+  propTypes: {
+    name: React.PropTypes.string.isRequired,
+  },
+
+  render: function(): React.Element {
+    return <div>{this.props.name}</div>;
+  }
+});
+
+var Callsite = React.createClass({
+  render: function(): React.Element {
+    return (
+      <div>
+        <Hello />
+        <HelloLocal />
+      </div>
+    );
+  }
+});
+
+module.exports = Callsite;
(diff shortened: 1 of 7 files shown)


tests/react_modules/createclass-callsite.js
```diff
@@ -0,0 +1,26 @@
+/* @flow */
+var React = require('react');
+var Hello = require('./createclass-callsite');
```
samwgoldman: Should this require `./createclass-module' instead?

dmnd: (>ლ) Ugh, yes, sorry, it should be.

I just pushed a new commit that corrects that and another silly import mistake. Everything seems to work as I expect, except that an ES6 class with proptypes (instead of type annotations) behaves differently locally vs being imported.

But I just hastily did this so maybe there's another silly mistake, I'll try to check this more thoroughly in a few hours!

Thanks for your time.

dmnd:
Sorry again about those silly import mistakes. As far as I can tell there might still be an issue though. I get an error for `HelloLocal` inside `es6class-proptypes-callsite.js` which is great. Should I expect the same error for the imported `Hello` component?

I'm asking about this because I'm trying to decide if I should start using flow in an existing React codebase. I already have lots of proptypes, so Flow being able to use them as types is nice. But I'm running into problems like this and haven't found many docs about using proptypes as flow types.

gabelevi:
This PR is almost a year old. To celebrate it's birthday, let's merge it!

@dmnd - I have no idea why neither I nor anyone else on the team has merged this test thus far. My best guess is that, as our oldest PR, everyone just assumed there was a good reason it was just waiting around 😞 . I'm really sorry for this, and I'll get this tested and landed ASAP.

gabelevi:
Also, at the core of this PR was the question "why doesn't Flow complain about the error in `es6class-proptypes-callsite.js` when `<Hello />` was missing properties.

While it's hard to comment on how Flow acted a year ago, nowadays the behavior comes from writing

```
class Hello extends React.Component { ... }
```

instead of

```
class Hello extends React.Component<void, {name: string}, void> { ... }
```

It is a problem that Flow silently swallows errors when `React.Component` is missing its type parameters. @samwgoldman has been thinking about this, and hopefully we'll have a solution in the near future.

dmnd:
Nice, I'm glad this turned out to be useful in the end.

Is there an issue that tracks flow silently swallowing errors when `React.Component` is missing type parameters?

PR #1027, opened 2015-11-02
Some small connection fixes

gabelevi:
tests/connect_race was failing sporadically, so I investigated. This fixes a
few things that I found.

1. `flow start` opens a pipe to the server it spawns. It's possible for that pipe to die before the child dies. Adding a 1 second sleep resolves this race.
2. It was wrong for `commandConnect.ml` to call `connect_once` when it fails to start a server. The right thing to do is to consume a retry.
3. The whole `/tmp` -> `/private/tmp` symlink thing is confusing. `FlowConfig` should make sure it normalizes everything whenever it's asked for a path.


mroch:
@facebook-github-bot shipit

PR #1028, opened 2015-11-02
Add missing React SyntheticEvent methods…

gf3:
* `isDefaultPrevented(): boolean`
* `isPropagationStopped(): boolean`

See [SyntheticEvent interface](https://facebook.github.io/react/docs/events.html#syntheticevent)

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -190,6 +190,8 @@ declare class SyntheticEvent {
     currentTarget: EventTarget;
     defaultPrevented: boolean;
     eventPhase: number;
+    isDefaultPrevented(): boolean;
+    isPropagationStopped(): boolean;
     isTrusted: boolean;
     nativeEvent: Event;
     preventDefault(): void;


samwgoldman:
Sorry for unrelated sniping here, but does React even expose runtime bindings for `SyntheticEvent` and it's subclasses? If it does, shouldn't the type be scoped to the `react` module? If it doesn't, I have a few concerns:

1. It should be declared as an interface, so Flow doesn't introduce a runtime binding.
1. The type should still be scoped to the module.
1. How is this type useful? We can't do a type test (`e instanceof SyntheticEvent`). No other part of the declared React API returns or consumes them.

gf3:
@samwgoldman AFAICT the `SyntheticEvent` constructor is not directly exposed anywhere (one could theoretically access it through `instance.constructor` though)

I, personally, can't say why a type would be more useful than an interface in this context as I don't have enough experience with Flow.

gabelevi:
Yeah, it was probably wrong of me to make `SyntheticEvent` global...maybe React should export it as a type or something like that, so you have to do

```JavaScript
import type { SyntheticEvent } from 'React';
```

Anyway, that can be future work. This PR looks good to me. Thanks @gf3 for adding these methods in!

gabelevi:
@facebook-github-bot shipit

PR #1036, opened 2015-11-04
callable interfaces

samwgoldman:
Before this commit, interfaces with callable signatures couldn't be
called.

For the included test, the pre-change failure was:

```
primitives.js:12:1,8: function call
Function cannot be called on
primitives.js:12:1,5: ICall
```

After this change, the expected failure correctly complains about the
parameter type.


gabelevi:
Short and sweet! Nice!

gabelevi:
@facebook-github-bot shipit

PR #1041, opened 2015-11-05
Pluralize "parenthesis"

stevenpetryk:
"Parenthesis are important" sounds really silly.


stevenpetryk:
Also, "parenthesis" on the next line could be changed as well, but it makes grammatical sense depending on one's interpretation.

jeffmo:
Hehe, TIL...

PR #1048, opened 2015-11-06
Fix the build

gabelevi:


gabelevi:
@facebook-github-bot shipit

PR #1051, opened 2015-11-06
fix live example in 02-running

m90:
The live coding example at http://flowtype.org/docs/running.html is broken due to the linked react version having been bumped to 0.14

Unfortunately I am unable to install the missing jekyll dependencies on my ubuntu machine (I gave up after an hour), so I could only test this fix by manually converting the markdown into HTML. If it needs changes to work in the jekyll env let me know.

On a side note, the page is also 404ing on `showdown.js` but I did not want to pull it in from some CDN as I don't know about your policies regarding this.


gabelevi:
Tried it locally with jekyll (I also had some dependency issues with json-1.7.7...not really sure how I fixed it) and it works for me! Thanks a ton for fixing!

PR #1058, opened 2015-11-08
Fix build priorities.o

Eyyub:
It seems to be the second round :p 

I don't know if it useful to add priorities.o to hack/build.ocp and hack/Makefile since it's managed by HHVM.


gabelevi:
@facebook-github-bot shipit

gabelevi:
Thanks for fixing this one too!

PR #1062, opened 2015-11-09
[Parser] Combine run_*_tests.js into run_tests.js

gabelevi:
`run_esprima_tests.js` and `run_hardcoded_tests.js` were basically the same
script, so I'm combining them. This will make it a little easier to add a babel
test suite. This also means that `run_all.sh` is no longer needed.


Eyyub:
Was `src/parser/test/.run_tests.js.swp` wanted ?

PR #1064, opened 2015-11-09
Add missing React Component method, getDefaultProps

gf3:
* [`static getDefaultProps(): DefaultProps;`](https://facebook.github.io/react/docs/component-specs.html#getdefaultprops)

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -46,6 +46,7 @@ declare class ReactComponent<DefaultProps, Props, State> {
   static childContextTypes: any;
   static displayName: string;
   static defaultProps: DefaultProps;
+  static getDefaultProps(): DefaultProps;
 }
 
 /**


samwgoldman:
Do class-based components have this method? For legacy `createClass` components, we already support `getDefaultProps`.

gf3:
no they do not, but it is my understanding that `React.createClass` is not deprecated

samwgoldman:
@gf3 You're correct, `createClass`-based components are still supported and we already implement the `getDefaultProps` function for them. Typing for `createClass` components is hardcoded into Flow currently, so changes to this file will not affect them. If you try to use `getDefaultProps` on a `createClass`-based component, you should see that it already works.

gf3:
hmm, i'm not finding that to be the case:

```
src/components/cost-breakdown.js:64
 64:   getDefaultProps() {
                        ^ return. Missing annotation
```

same with a few other methods like `render` and `shouldComponentUpdate`, but not others like `componentWillReceiveProps` and `getInitialState`:

```
src/components/cost-breakdown.js:99
 99:   shouldComponentUpdate(nextProps: Object, nextState: Object) {
                                                                  ^ return. Missing annotation

src/components/cost-breakdown.js:127
127:   render() {
               ^ return. Missing annotation
```

samwgoldman:
The `missing annotation` error means that Flow thinks you should add an annotation (in this case, of the return type), not that Flow doesn't understand the function. Flow asks that you annotate the parameter types and the return type of exported functions.

samwgoldman:
(Also, it looks like Avik relaxed this constraint in a recent commit, so the return type of `getDefaultParams` is actually inferred, and therefore doesn't need to be annotated. This change is already on master and will be in the next release.)

gf3:
splendid, thanks sam!

PR #1067, opened 2015-11-09
declaration_files_incremental_* tests should use strip_root

gabelevi:


mroch:
gonna land this because it fixes the linux tests and seems right, but the mac tests seem to still be failing for probably some other reason.

mroch:
@facebook-github-bot shipit

PR #1070, opened 2015-11-10
Fix declaration file incremental tests (PR attempted #2)

gabelevi:
I found a whole slew of problems. This fixes most of them and makes travis happy again. The fixes include

- Avoid using `SSet.of_list` (this exposed the fact that our travis linux builds don't actually test with the right version of ocaml. I have a fix ready for that separately).
- There seem to be some connection issues that are reproing on travis due to the long name of the test. I'll look into it tomorrow. For now, making the test name shorter helps.
- One of the tests has mistakes in which .js files it was moving and where
- There as an actual bug in the declaration file code. When using haste, if `foo.js` `@providesModule Foo`, but `foo.js.flow` doesn't have an `@providesModule`, then `foo.js.flow` will look at `foo.js` to infer which module it provides. This means when `foo.js` changes we need to recheck `foo.js.flow`. For now I made it so that declaration files are always rechecked when the implementation files change.


gabelevi:
@facebook-github-bot shipit

PR #1074, opened 2015-11-10
Fix parallel runtests.sh for OSX and Travis Linux containers

gabelevi:
Two main issues:

1. OSX uses an old version of Bash and thus `declare -A` doesn't exist
2. Travis Linux containers kill processes when too much memory is used.


gabelevi:
@facebook-github-bot shipit

PR #1092, opened 2015-11-13
Remove call to external commands

OCamlPro-Bozman:
- Replace the call to `find`, by `Find.find` function.
- Replace `diff` equivalent on Windows.

Note: there isn't -u/--label option on Windows, we just do the diff without this extra arguments.


OCamlPro-Bozman:
I close this PR, please see [#1182](https://github.com/facebook/flow/pull/1182).

PR #1093, opened 2015-11-13
File embedding on Windows

OCamlPro-Bozman:
This reused the same mechanis than Hack and this remove dependency
over the exernal command `tar`.


mroch:
@facebook-github-bot shipit

PR #1094, opened 2015-11-13
Use the Path module for better compatibility

mroch:
For instance use Path.null_path instead of "/dev/null".


mroch:
@facebook-github-bot import

mroch:
@facebook-github-bot shipit

PR #1095, opened 2015-11-13
Deactivate signals on Windows

mroch:
We replace `Sys.set_signal` by `Sys_utils.set_signal` that is a noop
on Windows.


mroch:
@facebook-github-bot shipit

PR #1096, opened 2015-11-13
Fix `Module_js.file_exists` on Windows

mroch:
The current test for detecting the root path does not on Windows.
Now, we detect the root path by testing : `path = dirname path`.


mroch:
@facebook-github-bot shipit

PR #1098, opened 2015-11-13
array rest destructuring

### LINKED ISSUE #743: Flow throws "unsupported destructuring" on array rests.
I have this:

```javascript
/* @flow */                                                                                                 

var [head,...tail] = [1,2];
```

And flow dies with:

```
/test.js:3:11,17: 
unsupported destructuring
```

I was unable to find an existing issue for this, so if this is a duplicate, my apologies.

samwgoldman:
Rest elements in an array pattern can only appear at the end, so we can
just lop off a bit of tuple information to get the correct type.

This was a lot more straightforward than I expected, so let me know if
I'm just ignoring some complexity.

Fixes #743


samwgoldman:
Hmm, should `is_use` be `true` for `ArrRestT _`?

samwgoldman:
I figured out the answer to my `is_use` question. Most recent commit adds a test for the expected `err_operation` behavior.

mroch:
@facebook-github-bot shipit

PR #1099, opened 2015-11-13
Replace Unix.system by Sys.command and Unix.create_process.

mroch:
Unix.system cannot be used on Windows because it relies unix fork.


mroch:
@facebook-github-bot import

mroch:
@facebook-github-bot shipit

PR #1100, opened 2015-11-13
WebSocket statics

samwgoldman:
It's unclear to me whether these attributes should actually be constants by the [spec](https://html.spec.whatwg.org/multipage/comms.html#websocket), but browsers seem to hang these values off the WebSocket class.

While I'm in here, we can be more specific about the type of these constants, thanks to singleton types.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -209,6 +209,10 @@ declare class CloseEvent extends Event {
 }
 
 declare class WebSocket extends EventTarget {
+    static CONNECTING: 0;
+    static OPEN: 1;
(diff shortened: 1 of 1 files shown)


mroch:
unclear to me too, but at least the d.ts file [does the same thing](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/34ae8420ec14158c1015882f876bbe99ad7110af/ws/ws.d.ts#L14)

mroch:
@facebook-github-bot shipit

PR #1104, opened 2015-11-15
Preserve tuple information in Function.prototype.apply

### LINKED ISSUE #64: flow doesn't understand Function.prototype.apply
This should typecheck
```javascript
/* @flow */
'use strict';

function test(a: string, b: number) {
}

var args = ['a', 2];
test.apply(this, args);
```

Instead I get 
```
/Users/xxx/git/xxx/flow2.js:8:1,22: rest array of element of arguments
  /Users/xxx/git/xxx/flow2.js:4:18,23: string

/Users/xxxx/git/xxx/flow2.js:8:1,22: apply call of method apply
  /Users/xxx/git/xxx/flow2.js:8:1,4: function
```

samwgoldman:
Before this change, the overall element type of the array was passed
into all arguments of the target function. This made it impossible to
use apply for functions that took arguments of multiple parameters.

The added ApplyT type holds on to the target FunT until the arguments
array is resolved, and then calls the function. Because we wait for the
arguments array to be resolved, we can pull out the tuple types to use
as parameter types.

Fixes #64


mroch:
we have a report internally of a similar problem using spreads, e.g.

```
let args = [0, 'foo', true];
function foo(x: number, y: string, z: boolean) {}
foo(...args);
```

the apply case already uses a `RestT`, so I'm wondering if the issue is actually in the way we call any function with a `RestT` arg, so maybe you could move this change out of the apply case and solve this more generically.

samwgoldman:
Yeah, it might very well be more useful to carry around the `ArrT` inside the `RestT`, or at least carry some kind of tuple information in the `RestT` itself. I can look into this in the coming days/week.

mroch:
should we land this one so it makes the next release? seems good, even if we iterate/expand the scope.

samwgoldman:
If it looks correct on its own, I think the future work we discussed would be backwards compatible. Should be OK to land.

mroch:
i just saw `ArrRestT` from your #1098, seems like that might be exactly what you need here?

samwgoldman:
Hmm. I'm not sure that would solve this problem, but might be useful in the
future work. If you want to hold off to the next release that's OK. There
might well be a more orthogonal feature set here. These two PRs are
admittedly somewhat ad hoc.

src/typing/flow_js.ml
```diff
@@ -3066,20 +3066,14 @@ let rec __flow cx (l, u) trace =
...
-         reason_op,
-         (_, "apply"),
-         ({params_tlist=[o2;tinsArr2]; _} as funtype)
-       ))
-      -> (* TODO: closure *)
-
-      let reason = replace_reason "element of arguments" reason_op in
-      let elem = mk_tvar cx reason in
-
-      rec_flow cx trace (tinsArr2, ArrT(reason,elem,[]));
```
mroch: i was thinking that doing `rec_flow cx trace (tinsArr2, ArrRestT(reason, 0, elem)` here would mean that `elem` would wait for `tinsArr2` to be resolved and let you access the `ts` you need from the `ArrT`. but I think you'd still need another type like `ApplyT` to build the `ts @ [RestT t]` param list, so I'm not sure how my suggestion would work either... and `ApplyT` is already waiting for it to be resolved so using `ArrRestT` would be useless.

mroch:
@facebook-github-bot shipit

PR #1105, opened 2015-11-16
Allow declare type when there is an exports entry.

### LINKED ISSUE #880: Module declarations: Export default polymorphic function along with types
I'm trying to create a module declaration for a library that exports a polymorphic function, along with some types, but currently I'm not sure how to achieve that. I've seen in the issues and in some of the core libs declaration in the source that to define a default exported function you use `declare var exports`, so I tried to do that here:

```js
import type { Middleware, MiddlewareAPI, Dispatch, Store } from 'redux';

declare module 'redux-thunk' {
  declare type Thunk<State, Dispatchable> = (dispatch: Dispatch<Dispatchable>, getState: () => State) => ?Dispatchable;
  declare type ThunkMiddlewareDispatchable<State, Dispatchable> = Thunk<State, Dispatchable> | Dispatchable;

  declare var exports: {
    <State, Dispatchable>(
      api: MiddlewareAPI<State, ThunkMiddlewareDispatchable<State, Dispatchable>>
    ) : (next: Dispatch<Dispatchable>) => Dispatch<ThunkMiddlewareDispatchable<State, Dispatchable>>
  };
}
```

When I do this I have the following type errors:

```
someFile.js:6:15,19: "Thunk" export of "redux-thunk"
Property not found in
someFile.js:6:57,69: exports of "redux-thunk"

someFile.js:6:22,48: "ThunkMiddlewareDispatchable" export of "redux-thunk"
Property not found in
someFile.js:6:57,69: exports of "redux-thunk"
```

If I remove the exports declaration, the types are recognized again. It seems that the `declare var exports` is like overriding whatever was first declared, and I cannot find the correct syntax to move the type declarations into the `declare var exports` block.

Is there a way to correctly achieve this?

Eyyub:
As discussed with @samwgoldman `declare type` should be available when there is an `exports` entry in a declared module.

I didn't allow this before because of the TODO comment, but it seems there is no collision possible.

Fix #880.

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -2468,8 +2468,14 @@ and statement cx type_params_map = Ast.Statement.(
     let for_types, exports_ = Scope.(Entry.(
       match get_entry "exports" module_scope with
       | Some (Value { specific = exports; _ }) ->
-        (* TODO: what happens when other things are also declared? *)
-        SMap.empty(* ???? *), exports
+        (* TODO: what happens when other things(i.e. Value) are also declared? *)
+        let for_types = SMap.filter (fun _ ->
+          function
+          | Value _ -> false
+          | Type _ -> true
+        ) module_scope.entries in
+
+        for_types, exports
 
       | Some _ ->
         assert_false (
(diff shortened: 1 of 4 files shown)


Eyyub:
I don't know why it's failing :o

Eyyub:
On my debian the test `integration` is failing randomly, maybe because of tests runned in parallel ? (tests have passed on travis since yesterday)

Issue #900, opened 2015-10-02
Errors involving core libraries are difficult to debug


pasha-mf:
Suppose I made an error and now Flow reports that my code is not compatible with some core type definitions:

```
var elem: HTMLInputElement = document.createElement("button");

js/flowtest.js:29:30,61: call of method createElement
Function cannot be called on
[LIB] dom.js:235:5,17: Document
```

Now, sometimes the cause of the error may be obvious; but other times -- it won't, and you'll want to look at the actual definition related to the error. And there is no easy way of doing so: I can go online and read https://github.com/facebook/flow/blob/master/lib/dom.js, but those definitions are updated all the time, and my system is not using the most up-to-date version. I can also restart Flow server and it will tell me where the log files are being stored, and if I go there I'll see a dozen or so cryptic directories, each containing core libraries. It's not exactly clear which one is currently in use, but there is a workaround: just delete all of them, restart the server and then come back.

In other words, it's a big hassle.
What would be more convenient is to have an option in .flowconfig file which would instruct the flow server where to store those core libraries. For example I'd store them in my project's lib/flow/core/ and then will be able to easily check those definitions at any time.

PS: I have a strong feeling that definition of Promise.catch in core.js is incorrect, since it has a different signature than the second parameter in Promise.then.

mroch:
from your example it looks like you're using `--strip-root`. it should give you the full path to dom.js if you don't use that.

samwgoldman:
Ultimately, this error message is bad because `createElement` is heavily overloaded, overloaded functions are implemented as intersection types, and intersection types (like union types) do not provide very helpful error messages.

This is a very well known problem, but unfortunately the solution is not as well known. Thanks for providing yet another example of this so we can test possible solutions against real world use cases.

@pasha-mf Regarding the issue with `Promise.catch`—can you open a separate issue, so we don't lose track of it?

samwgoldman:
IMO this error is now much improved. Today Flow outputs this:

```
Error: test.js:3
  3: var elem: HTMLInputElement = document.createElement("button");
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ HTMLButtonElement. This type is incompatible with
  3: var elem: HTMLInputElement = document.createElement("button");
               ^^^^^^^^^^^^^^^^ HTMLInputElement
```

PR #1108, opened 2015-11-16
Core libs updates

pasha-mf:
The first commit fixes a problem I mentioned in https://github.com/facebook/flow/issues/900;

The second one adds a bunch of missing properties to HTMLTextAreaElement. I copied the definitions from the official spec, however ignoring the "readonly" flags. It is possible to implement these by converting the properties into getters, but it hasn't been done for any other HTML*Element class...

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -407,7 +407,7 @@ declare class Promise<R> {
 
     then<U>(
       onFulfill?: (value: R) => Promise<U> | U,
-      onReject?: (error: any) => Promise<U> | U
+      onReject?: (error: any) => ?Promise<U> | U
     ): Promise<U>;
 
     catch<U>(
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -945,20 +945,40 @@ declare class HTMLButtonElement extends HTMLElement {
   checkValidity(): boolean;
 }
 
+// http://dev.w3.org/html5/spec-preview/the-textarea-element.html
 declare class HTMLTextAreaElement extends HTMLElement {
+  autofocus: boolean;
   cols: number;
-  defaultValue: string;
+  dirName: string;
   disabled: boolean;
-  form: HTMLFormElement;
+  form: ?HTMLFormElement;
+  maxLength: number;
   name: string;
+  placeholder: string;
   readOnly: boolean;
+  required: boolean;
   rows: number;
-  type: string;
-  value: string;
+  wrap: string;
 
+  type: string;
+  defaultValue: string;
+  value: string;
+  textLength: number;
+
+  willValidate: boolean;
+  validity: Object; // Too be more precise, a ValidityState object
+  validationMessage: string;
   checkValidity(): boolean;
+  setCustomValidity(error: string): void;
+
+  labels: NodeList<HTMLLabelElement>;
+
   select(): void;
-}
+  selectionStart: number;
+  selectionEnd: number;
+  selectionDirection: "forward"|"backward"|"none";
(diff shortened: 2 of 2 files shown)


lib/core.js
```diff
@@ -407,7 +407,7 @@ declare class Promise<R> {
 
     then<U>(
       onFulfill?: (value: R) => Promise<U> | U,
-      onReject?: (error: any) => Promise<U> | U
+      onReject?: (error: any) => ?Promise<U> | U
```
samwgoldman: Can you provide a reference for this change? I'm not super familiar with all the intricacies of the Promise API, but I'm a little suspicious of this change.

pasha-mf: The spec is rather hard to read, but here they state that `promise.catch(onReject)` is exactly equivalent to `promise.then(undefined, onReject)`:
http://www.ecma-international.org/ecma-262/6.0/#sec-promise.prototype.catch

Also check https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/es6-promise/es6-promise.d.ts -- they explicitly allow the "reject" part to return nothing. This is actually the most common use case:
```
promise.then((data) => {
  return transform(data);
}).catch((error) => {
  console.log("There was an error handling the promise. Abort!");
  // Returning nothing 
});
```

samwgoldman: Interesting. I think `Promise<U> | U | void` would be more precise here. `?Promise<U> | U` also allows `null`.

pasha-mf: Ok, I'm starting to believe that my proposed change is incorrect: the `onReject` callback in `then` should remain as is, whereas the callback in `catch` must be fixed (converted to non-nullable).
The idea is that if a promise fails, and there is a catch block that returns nothing, then the after the error is caught the promise is assumed successful, returning the result of the last onReject callback.

Here's a test case:
```
/* @flow */

function promiseMeMore(promise: Promise<number>): Promise<number> {
  return promise.then((x) => 2*x,
                      () => console.log("You broke your promise!"));
}
function boast(x: number) {
  console.assert(typeof(x) === "number", "Oops, got "+typeof(x)+" instead of a number");
  console.log("You promised me: " + (1*x));
  return x;
}

var promise1: Promise<number> = Promise.resolve(100).then(boast);
var promise2: Promise<number> = promiseMeMore(promise1).then(boast);
var promise3: Promise<number> = Promise.reject().then(boast);
var promise4: Promise<number> = promiseMeMore(promise3).then(boast);
```
In this test `promise4` is invalid, trying to execute it will hit an assertion in `boast()`. For whatever reason I can't make Flow to catch this problem, with either old or new definitions...

lib/dom.js
```diff
@@ -945,20 +945,40 @@ declare class HTMLButtonElement extends HTMLElement {
...
   rows: number;
+  wrap: string;
+
   type: string;
+  defaultValue: string;
   value: string;
+  textLength: number;
 
+  willValidate: boolean;
+  validity: Object; // Too be more precise, a ValidityState object
```
kastermester: There's a spelling mistake/grammar error here. Comment should read: `To be more precise, a ValidityState object`

mroch:
I'm going to try to partially land this

mroch:
k, merging the dom.js changes in https://github.com/facebook/flow/pull/1254 (fixed the typo and some minor things that have changed since this was put up, like `SelectionDirection` and that `form` attributes are `HTMLFormElement | null`, since they can't be undefined).

I'm going to close this PR. let's discuss the Promise change in a new PR or an issue.

thanks!

PR #1109, opened 2015-11-17
Path2D: Expose shared API method on Path2D object

tp:
These are shared between Path2D and CanvasRenderingContext2D

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -540,6 +540,19 @@ declare class Path2D {
     addPathByStrokingText(text: string, styles: CanvasDrawingStyles, transformation: ?SVGMatrix, x: number, y: number, maxWidth?: number): void;
     addText(text: string, styles: CanvasDrawingStyles, transformation: ?SVGMatrix, path: Path2D, maxWidth?: number): void;
     addPathByStrokingText(text: string, styles: CanvasDrawingStyles, transformation: ?SVGMatrix, path: Path2D, maxWidth?: number): void;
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -540,6 +540,19 @@ declare class Path2D {
...
     addPathByStrokingText(text: string, styles: CanvasDrawingStyles, transformation: ?SVGMatrix, path: Path2D, maxWidth?: number): void;
+    
+    // CanvasPathMethods
+    // shared path API methods
+    closePath(): void;
+    moveTo(x: number, y: number): void;
+    lineTo(x: number, y: number): void;
+    quadraticCurveTo(cpx: number, cpy: number, x: number, y: number): void;
+    bezierCurveTo(cp1x: number, cp1y: number, cp2x: number, cp2y: number, x: number, y: number): void;
+    arcTo(x1: number, y1: number, x2: number, y2: number, radius: number): void;
```
mroch: in order to force flow to pick the other `arcTo` when you pass extra args, I _think_ you need to pass dummy void arguments to this one to make their arities the same, e.g. `..., radius: number, _1: void, _2: void): void;`. this is because their args are otherwise the same, so if flow sees `arcTo(1, 2, 3, 4, 5, 6, 7)` it'll match this signature (plus some extra args, which flow allows) and not try the next signature.

mroch:
@facebook-github-bot import

PR #1124, opened 2015-11-23
Fix runtests.sh when path to flow is absolute, or has spaces

mroch:
tried running `./runtests.sh /Library/Application\ Support/Flow/flow-v0.18.1/flow`, and it blew up for two reasons:

- `FLOW=$(pwd -P)/$1` yielded `/Users/mroch/flow//Library/Application\ Support/...` which doesn't exist (OS X only problem). fixed by `cd`-ing into the directory and using `pwd -P` from there to resolve any symlinks.
- after fixing that, the unquoted `$FLOW`s didn't work on paths with spaces (should be a problem on all platforms)

### DIFF
--- a/runtests.sh
+++ b/runtests.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 if [[ "$OSTYPE" == "darwin"* ]]; then
-  FLOW=$(pwd -P)/$1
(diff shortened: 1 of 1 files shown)


jeffmo:
@facebook-github-bot shipit

PR #1126, opened 2015-11-25
Better error messages for erroneous imports

jeffmo:
This change improves the error messages shown when a user tries to `import` a non-existant export from another module. This also includes edit-distance hinting when it seems like the referenced name in a named-import is "close" to one of the remote module's named exports.

The fuzzy-match hinting will also try to help guide the user for the (not-that-rare) cases where one accidentally attempts to use `import foo from "bar";` to import a named export from another module and becomes confused by the error message (not realizing that the correct syntax for this is `import {foo} from "bar";`).


src/typing/flow_js.ml
```diff
@@ -1760,6 +1760,67 @@ let rec __flow cx (l, u) trace =
...
+          | Some t -> t
+          | None ->
+              let msg = 
+                spf "This module has no named export called `%s`." export_name 
+              in
+
+              let known_exports = SMap.keys exports_tmap in
+              let suggestions = spellcheck known_exports export_name in
+              let msg = if List.length suggestions = 0 then msg else msg ^ (
+                spf " Did you mean `%s`?" (List.hd suggestions)
```
gabelevi: Could you add a test for this case?

gabelevi:
This is great! The error messages are miles better and the spellcheck suggestion is nifty!

src/typing/flow_js.ml
jeffmo: added

PR #1132, opened 2015-11-28
Correcting pbkdf2 behavior

runn1ng:
Salt and password can be both string or Buffer in pbkdf2 (and pbkdf2sync). 

I have checked it in both Node source and browserify-crypto code.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -302,13 +302,15 @@ declare module "crypto" {
     salt: string | Buffer,
     iterations: number,
     keylen: number,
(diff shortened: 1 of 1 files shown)


runn1ng:
(I have added some more related changes directly to PR)

runn1ng:
ok now pbkdf2 finally works correctly :)

should I rebase and squash the commits, or is it OK like this?

jeffmo:
@facebook-github-bot shipit

PR #1133, opened 2015-11-28
adding getAuthTag for ciphers

runn1ng:
getAuthTag makes sense only for ciphers with GCM mode; however, that would be a little hard to encode into type system, so I am adding it for all cipher modes.


runn1ng:
When I get to it (later today maybe), I will try to add setAuthTag for decryption.

runn1ng:
...done. That wasn't hard.

jeffmo:
@facebook-github-bot shipit

jeffmo:
Woo! Thanks :)

PR #1135, opened 2015-11-30
Combine canvas tests into DOM tests

mroch:
`tests/dom/` will contain tests for all of the classes in `lib/dom.js`.


mroch:
@facebook-github-bot shipit

PR #1136, opened 2015-11-30
Fix annotation on JSON.stringify

### LINKED ISSUE #189: Tighten the type definition for the `space` param to `JSON.stringify()`
Currently, `JSON.stringify()` is declared as:

```
declare class JSON {
  static stringify(
      value: any,
      replacer?: ?((key: string, value: any) => any) | Array<any>,
      space?: any
  ): string;
}
```

but I believe the final parameter should be declared as `space?: number | string`.

mroch:
Fixes #189


mroch:
@facebook-github-bot shipit

Issue #1143, opened 2015-12-02
Regression with maybe types and Promise.all


nmote:
```
/* @flow */

async function foo() {
  const x: Array<Promise<?string>> = [];
  const y: Promise<Array<?string>> = Promise.all(x);
  const z: Array<?string> = await y;
}
```

This code checks in v0.18.1, but not v0.19.0.

In v0.19.0 we get:

```
test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
  4:   const x: Array<Promise<?string>> = [];
                               ^^^^^^ null. This type is incompatible with
  4:   const x: Array<Promise<?string>> = [];
                      ^^^^^^^^^^^^^^^^ Promise

test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
  4:   const x: Array<Promise<?string>> = [];
                               ^^^^^^ string. This type is incompatible with
  4:   const x: Array<Promise<?string>> = [];
                      ^^^^^^^^^^^^^^^^ Promise

test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
  4:   const x: Array<Promise<?string>> = [];
                               ^^^^^^ undefined. This type is incompatible with
  4:   const x: Array<Promise<?string>> = [];
                      ^^^^^^^^^^^^^^^^ Promise

test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                               ^^^^^^ null. This type is incompatible with
  4:   const x: Array<Promise<?string>> = [];
                      ^^^^^^^^^^^^^^^^ Promise

test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                               ^^^^^^ string. This type is incompatible with
  4:   const x: Array<Promise<?string>> = [];
                      ^^^^^^^^^^^^^^^^ Promise

test.js:5
  5:   const y: Promise<Array<?string>> = Promise.all(x);
                                          ^^^^^^^^^^^^^^ call of method `all`
[...]

PR #1159, opened 2015-12-05
Some lib definitions mishandle Array covariance

gabelevi:
Arrays are not covariant, which means you cannot pass an `Array<string>` to a type
annotation that expects `Array<string | number>`.

Luckily generics can save the day!

Fixes https://github.com/facebook/flow/issues/1143

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -145,7 +145,7 @@ declare class Array<T> {
     @@iterator(): Iterator<T>;
     toLocaleString(): string;
     // concat creates a new array
-    concat<S>(...items: Array<Array<S> | S>): Array<T | S>;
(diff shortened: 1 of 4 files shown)


gabelevi:
@facebook-github-bot import

jpochyla:
> cannot pass an `Array<string>` to a type annotation that expects `Array<string | number>`

Can I ask if this is by design or an accidental consequence of how Flow handles arrays?

gabelevi:
@jpochyla - It is by design. If we allowed that, then someone could put a `number` into your `Array<string>`! If `Array` was immutable then we wouldn't have to worry, but since you can change an `Array` after I pass it to you, we have to disallow this kind of subtyping.

PR #1168, opened 2015-12-09
add react-dom and react-dom-server interface defs

### LINKED ISSUE #1101: Library definition for "react-dom" module
Unless I'm missing something, doing `require('react-dom')` in a module checked by Flow doesn't seem to be possible right now. To fix that, we'd want to add a `declare module 'react-dom'` block in https://github.com/facebook/flow/blob/master/lib/react.js, right?

graue:
fixes #1101

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -184,6 +184,32 @@ declare module React {
     declare var exports: $Exports<'react'>;
 }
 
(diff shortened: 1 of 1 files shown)


jeffmo:
@facebook-github-bot shipit

jeffmo:
Awesome, thanks!

PR #1177, opened 2015-12-11
Add missing declarations to builtins.

marudor:
I've added some missing ES6 functions to the declarations.

I've also added the finally method to Promise. It is non standard but as there were already NonStandard Methods declared I've added it as it is included in a lot of popular Libraries for Promises.


runn1ng:
I would be against the inclusion of non-standard API like "finally", when it's not in ES6.

marudor:
Actually I think you are right.
I've removed the non-standard Promise API from the libs. Including the existing ones.
Now it is all standard ES6.

If someone needs non-standard API that person should make its own lib.

mroch:
thanks! this looks good to me if you can resolve the conflicts.

cc @bhosmer, are you ok with removing the nonstandard Promise stuff since we can move it into our own lib now? and also, I'm sure you'll like the even-bigger RegExp union :)

marudor:
I've merged my PR with latest upstream to resolve the conflicts.

bhosmer:
@mroch yup, removing the nonstandard Promise stuff should be fine - we shouldn't even be seeing the core def. 

A bigger RegExp union? Delightful!

lib/core.js
```diff
@@ -431,15 +439,6 @@ declare class Promise<R> {
     static reject<T>(error?: any): Promise<T>;
     static all<T, Elem: Promise<T> | T>(promises: Array<Elem>): Promise<Array<T>>;
     static race<T, Elem: Promise<T> | T>(promises: Array<Elem>): Promise<T>;
-
-    // Non-standard APIs common in some libraries
```
jeffmo: Hmm... So these are indeed non-standard, but they're here because they're commonly understood in the places they do exist (and thus there's not a great reason for Flow to *not* understand them).

It's one thing for libraries or polyfills to ship these and claim standard-conformance, but it's another for Flow to have a libdef so it understands what'll happen if someone does happen to use them.

mroch: The reason I vote to remove is not that this declaration is incompatible //if it exists//, it's that it might not exist at all. `.done()` should error unless you're in an environment that polyfills `.done()`.

jeffmo: > it's that it might not exist at all

This is a real problem and something we're in a good position to address so I agree with your basis, but it's no different an issue than having APIs that match a different version of Node than what you're using or an ES6 API that you aren't polyfilling/isn't supported by some platforms yet.

We do need better ways of toggling these kinds of things but "copy/paste the promise libdef and hack it with your own changes" is a pretty unfortunate one. Especially considering how pervasive `done()` is and the fickleness of typing `Promise` APIs. We seem to tweak and refine Promise a lot in ways that often aren't very intuitive (i.e. the need for things like `Promise<T> | T`...which is still not even quite right).

samwgoldman: @mroch @jeffmo If we can't find consensus, can we find a "winner" here so we can merge/close this PR/move on with life?

runn1ng: Well, typescript 1.8 has now [augmenting to modules](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#augmenting-globalmodule-scope-from-modules) that would, if implemented in Flow, solve this issue nicely, in my opinion :)

But I am not helping here much :)

samwgoldman:
Ping? I'm inclined to take this as-is. The Promises spec is what it is. We can document the workaround for people using non-standard polyfills.

samwgoldman:
That is, I am committing to writing said docs, and will import this in a few days if nobody speaks up ;)

jeffmo:
@samwgoldman: As long as we have an easy answer for the people who were depending on this before that we can put in the "breaking changes" section of the Changelog, I'll concede in the name of moving forward.

Given that we don't have a good way to augment global objects in Flow, I don't know that we have a great answer though.

samwgoldman:
While augmenting builtins would be nice, the workaround is (as you know) to redefine the class entirely. Promise isn't very large. Will amend docs and put up a diff for review.

@facebook-github-bot import

marudor:
One big use case is probably people using bluebird.
For that you could inform them about https://github.com/marudor/flowInterfaces/ (and https://github.com/flowtype/flow-typed once that is fully set up)

lib/dom.js
```diff
@@ -395,6 +395,7 @@ declare class Element extends Node {
     scrollWidth: number;
     tagName: string;
 
+    closest(): void;
```
samwgoldman: `closest()` isn't a standard DOM method. It's experimental, but I don't think that's enough to add to the core library definition.

lib/dom.js
```diff
@@ -410,6 +411,7 @@ declare class Element extends Node {
     removeAttribute(name?: string): void;
     removeAttributeNS(namespaceURI: string, localName: string): void;
     removeAttributeNode(oldAttr: Attr): Attr;
+    requestFullscreen(): void;
```
samwgoldman: `requestFullscreen()` isn't a standard DOM method. It's experimental, but I don't think that's enough to add to the core library definition.

lib/dom.js
marudor: experimental basically just tells you it is fairly new. But as the DOM Spec is a living standard it won't be removed again.
It is also implemented by the major Browser now. (Latest addition was Safari with Safari 9)

samwgoldman:
I rebased these changes for you and the commit is incoming and should land soon. I did revert the two experimental features added here. While it's true that DOM is a living standard, we can't safely accept these definitions internally because we don't currently polyfill these methods. Currently no version of IE supports Element.prototype.closest() for example.

Once flow-typed starts coming together, I'm hoping that it will be a lot easier to pick-and-choose your own particular flavor of these APIs, depending on your environment. Until then, I think we need to draw a slightly more conservative line.

Thanks so much for your continued contributions!

PR #1182, opened 2015-12-16
Remove call to external commands and various fixes on Windows

OCamlPro-Bozman:
This PR include fixes as:
 
    - replacing all the external commands like `diff` and `find`;
    - replacing the call to `Unix.system` by `Sys.command` and `Unix.create_process`.

Note: for the `diff` command, there isn't -u/--label option on Windows, we just do the diff without this extra arguments.


gabelevi:
@facebook-github-bot import

PR #1184, opened 2015-12-16
Restore state on slave after the fork/exec.

OCamlPro-Bozman:
Following change in Hack that use 'Daemon.spawn' to create slave, the
'global references' have to be copied "manually" on slave. This is done
in server/serverWorker.ml.

This is tied with D45423.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -80,10 +80,12 @@
  * 'caml_' */
 #define CAML_NAME_SPACE
 #include <caml/mlvalues.h>
-#include <caml/unixsupport.h>
+#include <caml/callback.h>
 #include <caml/memory.h>
 #include <caml/alloc.h>
 #include <caml/fail.h>
+
+#include "handle.h"
 
 #include <assert.h>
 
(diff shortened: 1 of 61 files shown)


gabelevi:
@facebook-github-bot import

hack/heap/hh_shared.c
```diff
@@ -264,31 +314,204 @@ struct timeval log_duration(const char *prefix, struct timeval start_t) {
...
     win32_maperr(GetLastError());
-    uerror("VirtualAlloc2", Nothing);
+    raise_out_of_shared_memory();
   }
+}
+
+#elif defined(__APPLE__)
+
+static void memfd_reserve(char * mem, size_t sz) {
+  // TODO ??
```
gabelevi: Noticed this TODO :)

gabelevi:
Sorry, just followed the trail to this PR. Let me test it out.

> About commit OCamlPro@97c51d7 Is there any reason not to immediatly take the lock ?

I think grabbing the lock earlier should be fine. 

> Is there a reason for 'ocamlbuild' not to build Flow with -DOSS_SMALL_HH_TABLE_POWS ?

Hmm, interesting point. I wasn't actually aware of that option. My guess is that it's probably fine.

> Could you confirm that your internal CI use a linux kernel < 3.17 and owns a too "small" '/dev/shm' ? In that case, does setting HH_SHMDIR=/tmp fix the problem ? if /tmp is not too small also.

My kernel is 3.10 at the moment, but I have a 71G /dev/shm. When testing with the previous PR, I did notice that some hack tests passed individually but failed when run together, so it could be an out of space thing.

gabelevi:
Ok! This has finally been merged as 7 commits:

1. https://github.com/facebook/flow/commit/e60b1bd7549182342faa105afaa31a8cecc1e1ce
2. https://github.com/facebook/flow/commit/51f0bfb4d298821d011ddbb693c26dd13d36e9c8
3. https://github.com/facebook/flow/commit/fa6d3e87fad4e091a7cf11b9fa689f7f220d0918
4. https://github.com/facebook/flow/commit/8a737f43e6deb107ada8cac9c6e4168cc8a6739e
5. https://github.com/facebook/flow/commit/5c0b0485f1a6571ed501a50e20a436af7433c0c5
6. https://github.com/facebook/flow/commit/c94194c6c04e3fb1680abd8fa0a9fafdc2cab280
7. https://github.com/facebook/flow/commit/77569bc430707b2faebdb2d1b414e8d8ec2a18f8

Yay!

michael-go:
Hi,

I think I stumbled upon an error that is related to this issue.
when `flow check` is executed as part of a Makefile  or from `tmux` it fails with exit status 15.
after some digging with `strace -ff` I saw that a call to `fallocate` on an `/run/shm/fb_heap-..` file fails with `EINVAL`:
```
fallocate(3, 0, 18446703599741734912, 138) = -1 EINVAL (Invalid argument)        
37099 exit_group(15) 
```
I guess it has something to do with the fork/exec or with signal handlers and the deamon creation done in these commits above, cause what else can be so different for a process that is executed from a Makefile...

thanks!

PR #1186, opened 2015-12-16
Fix `float_of_string` on Windows.

OCamlPro-Bozman:
This PR import some C code from a PR of OCaml repository which fixes the `float_of_string` issue on hexadecimal notation on Windows.


gabelevi:
I'm afraid this PR is going to be tricky to merge for licensing reasons.

How bad is the bug? How big a deal is this? Can we not parse any strings to floats if they use hex, or is it only certain strings?

Slayer95:
Hello. Is this the only remaining issue holding #6?

PR #1187, opened 2015-12-16
Update build system and some fixes on the test-suite on Windows 

OCamlPro-Bozman:
This PR:

-  add build.ocp and 00_config.ocp to compile Flow with ocp-build.
- fix in the test-suite for Windows
  - Fix `temp_dir` test on Windows
  - Socket files are not recognized as on Linux, so we cannot use '-S' option in bash file. On Windows we juste check as a regular file ('-f').
  - We deactivate some tests on Windows which involve symlinks.


mroch:
@facebook-github-bot shipit

PR #1200, opened 2015-12-20
Remove call to external commands.

mroch:
- Replace the call to find, by Find.find function.

- Replace diff equivalent on Windows.

  Note: there isn't -u/--label option on Windows, we just do the diff
  without this extra arguments.


mroch:
@facebook-github-bot import

PR #1202, opened 2015-12-21
Add KeyboardEvent type

JustinTulloss:
Extends UIEvent with the keyboard event attributes as defined by MDN.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -136,6 +136,22 @@ declare class MessageEvent {
     source: Document;
 }
 
+declare class KeyboardEvent extends UIEvent {
+  altKey: boolean;
+  charCode: number;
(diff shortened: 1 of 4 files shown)


mroch:
i think the test failures actually are relevant. the tests just run `flow check` and compare the current output to the expected output, which includes line numbers. since this shifts stuff down in `dom.js`, the expected line numbers changed. you can `cp tests/dom/dom.out tests/dom/dom.exp` (and similar for the other failing test) to update the expected values.

JustinTulloss:
Oh I see, that's confusing!

I'll update.

mroch:
@facebook-github-bot shipit

mroch:
thanks!

PR #1207, opened 2015-12-22
Fix null example

andy-hanson:


samwgoldman:
Oops! Thanks for catching this. Will merge as soon as you sign CLA.

Issue #1137, opened 2015-11-30
Add https definitions to lib/node.js


mroch:
Flow is missing definitions for the [https module](https://nodejs.org/api/https.html). Also the [http module](https://nodejs.org/api/http.html) is missing some definitions.

jamesgpearce:
Isn't it basically exactly the same as http?

mroch:
yeah, except `options` has a bunch of extra options for TLS

jamesgpearce:
...meaning a manager could do it? :see_no_evil:

PR #1209, opened 2015-12-23
Fill out and fix some node interfaces 

henridf:
I'm curious to know if there is any way to avoid the redundancy described in the commit message of 0f574c0.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -653,6 +653,20 @@ declare module "fs" {
   declare function createWriteStream(path: string, options?: Object): WriteStream;
 }
 
+declare class http$IncomingMessage extends stream$Readable {
+  headers: Object;
+  httpVersion: string;
+  method: string;
+  trailers: Object;
+  setTimeout(msecs: number, callback: Function): void;
+  socket: any;  // TODO net.Socket
+  statusCode: number;
+  url: String;
+}
+
+declare class http$ClientRequest extends stream$Writable {
+}
+
 declare module "http" {
   declare class Server extends net$Server {
     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server;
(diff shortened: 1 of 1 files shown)


henridf:
#1137

lib/node.js
```diff
@@ -353,9 +353,11 @@ declare module "crypto" {
   ): void;
 }
 
+type address = {address: string; family: string; port: number};
```
mroch: everything in libs is global, so we need to give this a more specific name. `dgram$Socket$address`, perhaps?

lib/node.js
```diff
@@ -687,7 +698,28 @@ declare module "http" {
...
+    listen(handle: Object, callback?: Function): Server;
+    close(callback?: Function): Server;
+    setTimeout(msecs: number, callback: Function): Server;
+    timeout: number;
+  }
+
+  declare class ClientRequest extends http$ClientRequest {}
+  declare class IncomingMessage extends http$IncomingMessage {}
+
+  declare function createServer(options: Object, secureConnectionListener?: function): typeof Server;
```
mroch: `typeof Server` returns the `Server` class, not an instance of `Server`. I think you want to drop the `typeof`?

lib/node.js
henridf: ah yes, fixed

lib/node.js
henridf: Fixed. 

I was a little fuzzy on the meaning of `typeof` and did this by imitation of other places (e.g. `tls.Server.createServer`) that appear to be making the same error.  
(I'll try to make a pass to fix in a separate PR at some point).

mroch:
@facebook-github-bot shipit

PR #1212, opened 2015-12-23
changing vars to ES6 const (and let)

runn1ng:
I don't *actually* expect this pull request to be merged as-is... however, I made it as an idea/discussion starter/...?

Flow now can do ES6 const/let, and it would be nice to "teach" users to use them in real code, since they are really useful.

So I just changed all `vars` to `const`, where possible, and where not possible, to `let`s.

### DIFF
--- a/docs/01-arrays.md
+++ b/docs/01-arrays.md
@@ -14,8 +14,8 @@ the type `Array<T>` describes arrays whose elements are of type `T`.
 
 {% highlight javascript linenos=table %}
 /* @flow */
-var a = [1, 2, 3];
-var b: Array<number> = a.map(function(x) { return x + 1; });
+const a = [1, 2, 3];
+const b: Array<number> = a.map(function(x) { return x + 1; });
 {% endhighlight %}
 
 In this code, we create an array with the literal `[1, 2, 3]`, and call a method map on it, getting another array whose type we annotate as `Array<number>`.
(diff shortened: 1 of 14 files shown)


runn1ng:
I did *not* test if all those examples are actually working with const, since again, I assumed this will not be accepted anyway. But I can test it of course.

samwgoldman:
Hey, so if you do the legwork and test these examples, I'm prepared to merge.

If you could whip up a script that parses out the code examples and runs them through flow, you will earn my eternal gratitude and I will merge this.

Otherwise, I think this should be closed.

runn1ng:
I could, but I am not sure what should be the result. Since some of the examples are *supposed* to fail, while some are *supposed* to succeed.

But it will be useful anyway so I will try

samwgoldman:
@runn1ng So we're actually migrating our docs to a cool new system. The new docs use a `.doc.js` extension, run through Flow for errors, and are translated to markdown with error messages inline. The new docs are also in the master branch, so we can more easily commit features and docs together.

Unfortunately, that means this PR is targeting an outdated set of docs. Once we've finished the migration, I'm definitely open to this kind of PR.

Thanks a lot for your contribution. Sorry we weren't able to land this one.

runn1ng:
No problem, the system you describe sounds great :)

PR #1217, opened 2015-12-27
Quick fix for tools/js_test_files.js

gabelevi:


gabelevi:
@facebook-github-bot shipit

PR #1218, opened 2015-12-27
Remove Object.assign() from parser tests

gabelevi:
Older versions of node didn't like it. Tested this change with Node v0.10.26
and v4.1.2


gabelevi:
@facebook-github-bot shipit

PR #1221, opened 2015-12-28
Add HTML focus management APIs

mroch:
See https://html.spec.whatwg.org/multipage/interaction.html#focus-management-apis


mroch:
@facebook-github-bot shipit

PR #1223, opened 2015-12-28
[docs] Document `function exports` trick

frantic:


docs/01-declarations.md
```diff
@@ -94,3 +94,13 @@ declare module "path/to/M" {
   ...
 }
 {% endhighlight %}
+
+## Modules that export single function
```
jeffmo: So this feature isn't necessarily specific to functions -- it works with any value (classes, objects, numbers, etc). It's also specific to CommonJS modules.

Maybe you could clarify this to `CommonJS modules with a single export`?

jeffmo:
Merged (with clarification) over in https://github.com/facebook/flow/commit/2147008b4e473aaf2e8e18be3c7f68fd01b1c670#diff-0860be763605afea620fa7394916af55

Thanks for sending this over!

frantic:
awesome, thanks!

PR #1247, opened 2016-01-07
typing custom element lifecycle callbacks

AprilArcus:
I've added types for the four custom element lifecycle callback methods.

If I've read the spec correctly, ideally we would also like to be able to stipulate that

  1. the value of `ElementRegistrationOptions.prototype` should be an object which includes either `HTMLElement` or `SVGElement` in its prototype chain.
  2. the return value of `document.registerElement()` will be a constructor for a subclass of `HTMLElement` or `SVGElement` or the descendent named in (1).

Does anyone know how this could be expressed in Flow?

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -255,7 +255,18 @@ declare class HTMLCollection {
 
 // from http://w3c.github.io/webcomponents/spec/custom/#extensions-to-document-interface-to-register
 type ElementRegistrationOptions = {
-    prototype?: any;
+    prototype?: {
+      // from http://w3c.github.io/webcomponents/spec/custom/#types-of-callbacks
+      createdCallback?: () => void;
+      attachedCallback?: () => void;
+      detachedCallback?: () => void;
+      attributeChangedCallback?: (
+        attributeLocalName: string,
+        oldAttributeValue: ?string,
+        newAttributeValue: ?string,
+        attributeNamespace: ?string
+      ) => void;
+    };
     extends?: string;
 }
 
--- a/tests/dom/dom.exp
+++ b/tests/dom/dom.exp
@@ -3,13 +3,13 @@ CanvasRenderingContext2D.js:11:5,24: call of method `moveTo`
 Error:
 CanvasRenderingContext2D.js:11:16,18: string
 This type is incompatible with
-[LIB] dom.js:708:15,20: number
+[LIB] dom.js:719:15,20: number
 
 CanvasRenderingContext2D.js:11:5,24: call of method `moveTo`
 Error:
 CanvasRenderingContext2D.js:11:21,23: string
 This type is incompatible with
-[LIB] dom.js:708:26,31: number
+[LIB] dom.js:719:26,31: number
 
 eventtarget.js:9:6,40: call of method `attachEvent`
 Function cannot be called on possibly undefined value
--- a/tests/dom/registerElement.js
+++ b/tests/dom/registerElement.js
@@ -1,0 +1,68 @@
+// @flow
+
+let tests = [
+  // should work with Object.create()
+  function() {
+    document.registerElement('custom-element', {
+      prototype: Object.create(HTMLElement.prototype, {
+        createdCallback: { value: function createdCallback () {
+        }},
+        attachedCallback: { value: function attachedCallback () {
+        }},
+        detachedCallback: { value: function detachedCallback () {
+        }},
+        attributeChangedCallback: {
+          value: function attributeChangedCallback (
+            attributeLocalName,
+            oldAttributeValue,
+            newAttributeValue,
+            attributeNamespace
+          ) {
+          }
+        }
+      })
+    })
+  },
+  // or with Object.assign()
+  function() {
+    document.registerElement('custom-element', {
+      prototype: Object.assign(Object.create(HTMLElement.prototype), {
+        createdCallback () {
+        },
+        attachedCallback () {
+        },
+        detachedCallback () {
+        },
+        attributeChangedCallback (
+          attributeLocalName,
+          oldAttributeValue,
+          newAttributeValue,
+          attributeNamespace
+        ) {
+        }
+      })
+    })
+  },
+  // should complain about unrecognized callbacks
+  function() {
+    document.registerElement('custom-element', {
+      prototype: Object.create(HTMLElement.prototype, {
+        bogusCallback: { value: function bogusCallback () {
+        }}
+      })
+    })
+  },
+  // should complain about invalid callback parameters
+  function() {
+    document.registerElement('custom-element', {
+      prototype: Object.create(HTMLElement.prototype, {
+        createdCallback: { value: function createdCallback (bogusArg) {
+        }},
+        attachedCallback: { value: function attachedCallback (bogusArg) {
+        }},
+        detachedCallback: { value: function detachedCallback (bogusArg) {
+        }}
+      })
+    })
+  },
+];


AprilArcus:
I am a little confused about the failing tests. Can anyone give me a hint?

AprilArcus:
Here's a [gist](https://gist.github.com/AprilArcus/7024adb7ec765f669e41) of my attempt to enforce inheritance from `HTMLElement` with an intersection type. It seems that Flow is unaware that the value for the `prototype` key has inherited HTMLElement's methods through the call to `Object.create()`.

mroch:
some of them look like bogus race conditions so I'm re-running the tests. the dom, taint and taint-lib failures are because the tests work by comparing flow's output (`tests/testName/testName.out`) to the expected output (`tests/testName/testName.exp`), which includes line numbers, and your change adds lines so it pushes everything down. just `cp` .out to .exp if the changes are legit.

mroch:
they could also be because dom.js has changed a couple times, so make sure you rebase

mroch:
@facebook-github-bot import

mroch:
I'm rebasing this and updating the tests. I also changed the `void` return type to `mixed`. We've been trying to be better about doing that since most APIs don't look at the return value, and it's sometimes convenient to return a value because of arrow functions (e.g. `doStuff(() => somethingThatReturnsAValue())`)

Also, to be really pedantic, I think the `attributeChangedCallback` should actually be overloaded since `oldAttributeValue` and `newAttributeValue` can't both be null at the same time, and the `?string` args should probably be `string | null` since values are always strings, and namespaces are [`null | string`](https://dom.spec.whatwg.org/#concept-attribute-namespace). but I'm not even sure offhand how you overload object properties like that. Feel free to send a followup PR if this sounds worth looking into.

AprilArcus:
On `void` vs. `mixed`: this is a really strange API with a lot of footgun potential, and the reason I wanted to type it was to put a very restrictive fence around it.
  <ul><li>these callbacks are intended to be methods, not functions. They perform side effects, rather than returning values; they should never be called by the user and nobody should ever depend on their return type.</li>
  <li>the calling convention is strange compared to ES6 classes, and will look like either:
```
document.registerElement(
  'custom-element-name',
  prototype: Object.create(
    HTMLElement.prototype, {
      createdCallback: { value: function createdCallback() {...} },
      attachedCallback: { value: function attachedCallback() {...} },
      detachedCallback: { value: function detachedCallback() {...} },
      attributeChangedCallback: { value: function attributeChangedCallback() {...} },
    }));
```

or 

```
document.registerElement(
  'custom-element-name',
  prototype: Object.assign(
    Object.create(HTMLElement.prototype), {
      createdCallback() {...},
      attachedCallback() {...},
      detachedCallback() {...},
      attributeChangedCallback() {...},
    }));
```
these callbacks must have a `this` for `document.registerElement()` to bind to the DOM element instance, so using an arrow function would be semantically wrong.</li></ul>

On `string | null` vs `?string`: agreed.

On the signature of `attributeChangedCallback`: I agree that there is an interdependency between the parameters that isn't captured by this annotation, but I have trouble imagining errors that could be prevented by a more precise type or ergonomics that the inference engine could extract. That said, and if we want to try anyway, maybe a union type would work?

```
attributeChangedCallback: void | (
  attributeLocalName: string,
  oldAttributeValue: string,
  newAttributeValue: string,
  attributeNamespace: string | null
) => void | (
  attributeLocalName: string,
  oldAttributeValue: null,
  newAttributeValue: string,
  attributeNamespace: string | null
) => void | (
  attributeLocalName: string,
  oldAttributeValue: string,
  newAttributeValue: null,
  attributeNamespace: string | null
) => void;
```

Do you have any thoughts on the matter of enforcing the relationship between the return type of `document.registerElement()` and the prototype of the `ElementRegistrationOptions` object?

samwgoldman:
@AprilArcus Thanks for sending this PR and especially thank you for explaining the choices you made so lucidly.

On the `void` vs. `mixed` discussion, I think you make a very good case. To paraphrase, the reason to go with `mixed` is because we get a lot of feedback that brace-less arrow functions should be allowed, even though they have a non-`void` return type, because the caller ignores the return value. It seems that in this case, arrow functions would be a bad idea in the first place, since we want to use the re-bound `this`. I could go either way on it, but I just wanted to note my appreciation of your perspective here.

I was curious about the overloaded signature for attribute changed, so I wrote up a small test to see how Flow might behave:

```javascript
/* @flow */

function foo(delegate: {
  attributeChanged:
    ( ((oldValue: string, newValue: string) => void)
    & ((oldValue: string | null, newValue: string) => void)
    & ((oldValue: string, newValue: string | null) => void) )
}) {
  delegate.attributeChanged("old", "new"); // ok
  delegate.attributeChanged(null, "new");  // ok
  delegate.attributeChanged("old", null);  // ok
  delegate.attributeChanged(null, null);   // error
};

foo({
  attributeChanged: function(oldValue, newValue) {
    if (oldValue == null) {
      (newValue: string); // not that smart -- we still think newValue can be null
    }
  },
});
```

As you can see, the overloaded signature is only really useful if you're going to **call** the method, which nobody would do here, since it's library code. It doesn't help guide implementations, so I think moving to an overloaded signature would actually be worse, due to the consequent degradation in error messages.

@mroch Let me know if I can help push this through.

AprilArcus:
I added some tests that should fail (passing an unrecognized callback, or attempting to use invalid params), but they are currently passing (maybe due to `Object.assign()` or `Object.create()` upcasting to `any`?)

zertosh:
By looking at the test fixtures I feel that this ([flowtry](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjGBDAzrsAMTjjAFMAPAFzIDsATAgCQBUBZAGQFEYyBbOlTABvVGDDYqVAE4BLAEYBXGgGEAFtloBzMvRXYYMedgwBrABRjxYeBgMA5bAIBcYXDNnaANFfFwY9ABqBq7uct5gwMBgXNLScNKuLGqyBHyyWmpC8mRgtIqGvnlkCMEwoR4RUTFxCUkpaRlZYDl5BTBFtE5kuAAOJmQV4VoAlCIAvqiTqPRwGIoCtFQAdNJkWqk00jz8guYA5FAk+15EJCMA3EA)) should work:

```js
class Foo extends HTMLElement {
  attributeChangedCallback(
    localName: string,
    oldVal: string, // Error: This might be null
    newVal: string, // Error: This might be null
    namespace: string) {}
}

document.registerElement('foo', Foo);
```

However, flow 0.28.0 is reporting no errors.

But this ([flowtry](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjGBDAzrsAMTjjAFMAPAFzIDsATAgCQBUBZAGQFEYyBbOlTABvVGDDYqVbBgAWZegGFsMGACMZAawAUASgBcYPgEsKCkQF9UV1PTgYArgNpUAdACcyAc2O4a7nn5BbQByKBIQgBoRAAd3OCoEgE8YskNiOFc4hOTUi10AblQgA)) reports an odd error:

```js
class Foo extends HTMLElement {
  attachedCallback(): mixed {}
}

document.registerElement('foo', {prototype: Foo.prototype});
```

```
4:   attachedCallback(): mixed {}     ^ method `attachedCallback`. This type is incompatible with
/static/flowlib/dom.js:372:     attachedCallback?: () => mixed;
                                                   ^ undefined
```

PR #1250, opened 2016-01-07
implementation of net$Socket class in lib/node.js file

d14b3l:
net$Socket is now implemented accordingly to nodejs documentation, tests are passing locally


mroch:
rebased and landed in #1255, thanks!

PR #1254, opened 2016-01-09
Adding definition for HTMLLabelElement, and extending definition of HTMLTextareaElement to support all HTML5 properties

mroch:


mroch:
@facebook-github-bot shipit

PR #1264, opened 2016-01-11
Add os.homedir to node lib

kittens:
https://nodejs.org/api/os.html#os_os_homedir


mroch:
@facebook-github-bot shipit

PR #1285, opened 2016-01-18
adds a missing mandatory step in opam how-to

monperrus:
Otherwise, one gets a dirty:

    # + ocamlopt   -ccopt -DNO_LZ4 -c hack/hhi/hhi_elf.c
    # hack/hhi/hhi_elf.c:26:18: fatal error: gelf.h: No such file or directory
    # compilation terminated.
    # Command exited with code 2. 
    # Makefile:131: recipe for target 'build-flow-native-deps' failed  
    ### stderr ###
    # abort: there is no Mercurial repository here (.hg not found)
    # make: *** [build-flow-native-deps] Error 10

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ brew install flow
 You can also build and install flow via the OCaml [OPAM](https://opam.ocaml.org) package manager with one command:
 
 ```
+apt-get install libelf-dev # (required for gelf.h)
 opam install flowtype
 ```
 


gabelevi:
Oh, good point! Looking at opam-repository (where we publish each version of Flow), we do [list the external dependencies](https://github.com/ocaml/opam-repository/blob/master/packages/flowtype/flowtype.0.19.1/opam#L22). But it does look like we're missing an [additional step](https://opam.ocaml.org/doc/FAQ.html#Somepackagefailduringcompilationcomplainingaboutmissingdependenciesquotm4quotquotlibgtkquotetc)

So perhaps the steps should be

```Bash
opam install depext
opam depext --install flowtype
```

Hopefully this will keep working if we add more external dependencies. I haven't used `opam depext` before, but I booted up a debian/jessie64 box and tried it out and it seemed to work. 

I'll update the readme. Thanks again for the report!

PR #1289, opened 2016-01-19
Update opam instructions to use depext

gabelevi:
@monperrus pointed out the issue in https://github.com/facebook/flow/pull/1285


gabelevi:
@facebook-github-bot shipit

PR #1291, opened 2016-01-20
Use a custom opam repo to download from github instead of caml.inria.fr

mroch:
OPAM points at caml.inria.fr to download compilers. This is an additional point of failure that can break our Travis builds, so we have here a minimal OPAM repository that instead points at GitHub tarballs for the versions of ocaml that we build for. As a plus, it also uses SSL.

This also works on Macs, even though we currently use homebrew for them.

Tested (on a mac!) with:

```
rm -rf ~/.opam && TRAVIS_OS_NAME=linux OCAML_VERSION=4.01.0 OPAM_VERSION=1.1.1 bash -ex .travis-ci.sh
rm -rf ~/.opam && TRAVIS_OS_NAME=linux OCAML_VERSION=4.02.1 OPAM_VERSION=1.1.1 bash -ex .travis-ci.sh
rm -rf ~/.opam && TRAVIS_OS_NAME=linux OCAML_VERSION=4.01.0 OPAM_VERSION=1.2.0 bash -ex .travis-ci.sh
rm -rf ~/.opam && TRAVIS_OS_NAME=linux OCAML_VERSION=4.02.1 OPAM_VERSION=1.2.0 bash -ex .travis-ci.sh
```


mroch:
@facebook-github-bot shipit

PR #1293, opened 2016-01-20
Use opam on OS X in travis

mroch:
Also some changes to let this run more than once with an existing `~/.opam`, which was helpful for testing (and hopefully a future diff can enable travis caching!)


mroch:
makes the builds super slow to build from opam all the time. i want to figure out caching and then come back to the mac builds.

PR #1301, opened 2016-01-22
Adding refresh to plugin API

runn1ng:
Plugins are old and slowly-but-surely unsupported technology, but we still need that API for some legacy apps.


mroch:
@facebook-github-bot shipit

PR #1305, opened 2016-01-24
dom level 2 traversal apis

AprilArcus:
annotating the [DOM Level 2 Traversal API](https://dom.spec.whatwg.org/#traversal)

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -349,20 +349,20 @@ declare class Document extends Node {
     querySelectorAll(selector: string): NodeList<HTMLElement>;
 
     // tree traversal
-    createNodeIterator(root: Node, whatToShow: 2, filter: ?NodeFilter): NodeIterator<Attr>;
-    createNodeIterator(root: Node, whatToShow: 4, filter: ?NodeFilter): NodeIterator<Text>;
-    createNodeIterator(root: Node, whatToShow: 128, filter: ?NodeFilter): NodeIterator<Comment>;
-    createNodeIterator(root: Node, whatToShow: 256, filter: ?NodeFilter): NodeIterator<Document>;
-    createNodeIterator(root: Node, whatToShow: 512, filter: ?NodeFilter): NodeIterator<DocumentType>;
-    createNodeIterator(root: Node, whatToShow: 1024, filter: ?NodeFilter): NodeIterator<DocumentFragment>;
-    createNodeIterator(root: Node, whatToShow: ?number, filter: ?NodeFilter): NodeIterator<Node>;
-    createTreeWalker(root: Node, whatToShow: 2, filter: ?NodeFilter, entityReferenceExpansion: ?boolean): TreeWalker<Attr>;
-    createTreeWalker(root: Node, whatToShow: 4, filter: ?NodeFilter, entityReferenceExpansion: ?boolean): TreeWalker<Text>;
-    createTreeWalker(root: Node, whatToShow: 128, filter: ?NodeFilter, entityReferenceExpansion: ?boolean): TreeWalker<Comment>;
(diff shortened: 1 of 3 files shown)


lib/dom.js
```diff
@@ -1285,6 +1289,57 @@ declare class TrackDefault {
...
+  // NodeFilter.FILTER_ACCEPT or
+  // NodeFilter.FILTER_REJECT or
+  // NodeFilter.FILTER_SKIP
+  acceptNode(node: Node): number;
+}
+
+declare class NodeIterator {
+  root: Node;
+  whatToShow: number;
+  filter: NodeFilter;
```
samwgoldman: I think this should be `NodeFilter | null`, per [the spec](https://dom.spec.whatwg.org/#interface-nodeiterator)

lib/dom.js
```diff
@@ -1285,6 +1289,57 @@ declare class TrackDefault {
...
+  pointerBeforeReferenceNode: boolean;
+  detach(): void;
+  previousNode(): Node | null;
+  nextNode(): Node | null;
+}
+
+declare class TreeWalker {
+  root: Node;
+  whatToShow: number;
+  filter: NodeFilter;
```
samwgoldman: I think this should be `NodeFilter | null`, per [the spec](https://dom.spec.whatwg.org/#interface-treewalker)

samwgoldman:
Thanks! Just a few small things. The test breakages are line number related, so solving them is just replacing the `.exp` files with `.out` files.

If you're motivated, it would be awesome to replace the `number` annotations with the actual constant values. (Unfortunately, we don't support hex format for literal numbers, but the decimal representation should work.)

If you don't want to support this PR, let me know. I'll make the necessary fixes and push it through. Don't worry—you'll still be attributed in the commit!

AprilArcus:
Thanks! I added the numerical annotations along with some type parameters make use of them in common cases. Is there a way to annotate the fact that despite being class-like, `NodeIterator`, `NodeFilter` and `TreeWalker` do not have constructors?

samwgoldman:
Woah! Very cool changes. I'm especially intrigued by the type parameter stuff. It's sophisticated enough that I'm worried we might run into a Flow bug. Do you have any examples of code that you'd like to typecheck? There are some tests under `tests/node_tests` that specifically exercise these libdefs to make sure they're right. I think we can take this without tests, but they'd be nice.

For annotating class-like things which don't have constructors, a few things come to mind. Are `NodeIterator`, `NodeFilter`, and `TreeWalker` runtime values at all, or are they purely interfaces? That is, can I write `typeof foo === NodeIterator`? If not, you might consider declaring interfaces instead of classes.

AprilArcus:
They are non-callable objects that behave like constructors as far as instanceof is concerned.

```
> typeof NodeFilter;
< "object"
> NodeFilter.hasOwnProperty('SHOW_TEXT');
< true
> typeof NodeIterator;
< "object"
< typeof TreeWalker;
< "object"
> var i = document.createNodeIterator(document.body);
> i instanceof NodeIterator
< true
> var w = document.createTreeWalker(document.body);
> w instanceof TreeWalker
< true
```

samwgoldman:
Right right. Thanks for understanding me, even though I mixed up typeof and instanceof :P

samwgoldman:
I don't think we have a great way to type that. I think for now it's OK to leave them as classes. Maybe add a note that the constructors aren't callable, and we can come back in and fix that once we support that use case.

AprilArcus:
Hey @samwgoldman - I added a two small tests against the behavior I expect from these generics. They aren't currently working as I expect. Do you mind taking a look?

samwgoldman:
Man, I was going to wait for Travis to do the work for me, but I'll take a look manually tomorrow to see what's up. Thanks a ton for your continued work on this!

AprilArcus:
withdrawing in favor of https://github.com/facebook/flow/pull/1339

Issue #776, opened 2015-09-03
Singleton number types do not support negative numbers


pasha-mf:
```
var star_rating: 0 | 1 | 2 | 3 | 4 | 5 = 0; // ok
var direction: 1 | 0 | -1 = 1; // fails
```
Saying `+1` instead of `1` will also generate an error.

PR #1312, opened 2016-01-26
Add readline definitions

kittens:


lib/node.js
```diff
@@ -896,8 +896,34 @@ declare module "querystring" {
...
+    ctrl?: boolean,
+    shift?: boolean,
+    meta?: boolean
+  }): void;
+}
+
 declare module "readline" {
-  // TODO
+  declare var Interface : typeof readline$Interface;
+  declare function clearLine(stream: stream$Writable, dir: number /* -1 | 1 | 0 */): void;
```
kittens: Flow wouldn't let me put a negative number literal here which I assume is a bug since positive ones are allowed?

mroch:
hmm, negative singleton number types should work: #776

kittens:
Looks like it's working now, must have done something wrong. Updated.

PR #1313, opened 2016-01-26
Add punycode definitions

kittens:


lib/node.js
```diff
@@ -870,7 +870,15 @@ declare module "path" {
 }
 
 declare module "punycode" {
-  // TODO
+  declare function decode(string: string): string;
+  declare function encode(string: string): string;
+  declare function toASCII(domain: string): string;
+  declare function toUnicode(domain: string): string;
+  declare var ucs2: {
+    decode: (str: string) => string,
```
samwgoldman: It looks like `punycode.usc2.decode` returns an array of code points, not a string. can you confirm?

kittens:
Updated.

samwgoldman:
@facebook-github-bot import

PR #1314, opened 2016-01-26
Add vm definitions

kittens:


lib/node.js
```diff
@@ -1093,8 +1093,31 @@ declare module "util" {
...
+  columnOffset?: number,
+  displayErrors?: boolean,
+  timeout?: number
+};
+
+declare class vm$Script {
+  constructor(code: string, options: Object): void;
+  runInContext(contextifiedSandbox: vm$Context, options?: vm$ScriptOptions): void;
+  runInNewContext(sandbox?: Object, options?: vm$ScriptOptions): void;
+  runInThisContext(options?: vm$ScriptOptions): void;
```
samwgoldman: It seems like the return value for `runIn*` methods should by `any`, like the `runIn*` functions below. We don't know what the code will eval to, but I'm pretty sure it's returned from this API.

lib/node.js
```diff
@@ -1093,8 +1093,31 @@ declare module "util" {
...
+}
+
+declare class vm$Context {}
+
 declare module "vm" {
-  // TODO
+  declare var Script : typeof vm$Script
+  declare function createContext(sandbox?: Object): vm$Context;
+  declare function isContext(sandbox: any): boolean;
+  declare function runInContext(code: string, contextifiedSandbox: vm$Context, options?: vm$ScriptOptions): void;
```
samwgoldman: Ditto above. Won't this return whatever `code` evals to?

kittens:
Yeah it will. Updated.

PR #1315, opened 2016-01-26
Add string_decoder definitions

kittens:


lib/node.js
```diff
@@ -986,8 +986,13 @@ declare module "tty" {
   declare var WriteStream : typeof tty$WriteStream
 }
 
+declare class string_decoder$StringDecoder {
```
samwgoldman: I think this is missing the constructor, which according to the docs "takes a single argument, `encoding` which defaults to `'utf8'`."

That `encoding` arg might be more complex than `string`, too. Can you add the constructor and ensure we capture the most common types here? Thanks.

kittens:
Ah my bad, updated. Found a list of encodings [here](https://github.com/nodejs/node/blob/master/lib/string_decoder.js#L24-L43).

PR #1316, opened 2016-01-26
Fix Hash.prototype.update definition

kittens:
https://github.com/nodejs/node/blob/master/lib/crypto.js#L71


samwgoldman:
@facebook-github-bot import

samwgoldman:
...Bueller?

Issue #1272, opened 2016-01-13
Support `extends null`


bmeck:
``` javascript
class X extends null {}
```

results in:

```
class X extends null {
                          ^^^^ null. Expected polymorphic type instead of possibly null value
class X extends null {
                          ^^^^ null
```

used for non-coercible types.

samwgoldman:
Can you give an example of how such a class might behave?

bmeck:
@samwgoldman in what sense of the word, it is essentially an `Object` without `Object.prototype` in the chain.

```javascript
x = Object.create(null);
x+'' // throws cannot be converted to string
String(x) // throws no .toString
Number(x) // throws no .valueOf
+x// throws // throws cannot be converted to number
```

samwgoldman:
So, Flow doesn't catch the errors in your example today. I wasn't aware of that behavior, so thanks for bringing it to my attention. I'll see what we can do there.

However, I don't think your recommended syntax will work. The `extends null` part would be stripped off before runtime, and the runtime behavior of any class instance is that it's coercible. Flow can't change that.

Can I re-title this issue to target the `Object.create` syntax instead, or do you think I'm mistaken about classes here?

bmeck:
@samwgoldman `extends null` should work in the JS spec: https://tc39.github.io/ecma262/#sec-runtime-semantics-classdefinitionevaluation step 6.e.

samwgoldman:
Haha, oops. I'm totally wrong. `extends` is absolutely runtime syntax.

You're right. We get this wrong (and `Object.create`). Thanks for the issue report.

adrianheine:
#1332 might be related

SamChou19815:
The support is now explicitly banned as Flow becomes stricter.

PR #1332, opened 2016-01-28
Fix Object.create definition

adrianheine:
The prototype argument to `Object.create` is nullable, see [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-object.create).

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -33,7 +33,7 @@ declare class Object {
     static getPrototypeOf(o: any): any; // compiler magic
     static getOwnPropertyDescriptor(o: any, p: any): any;
     static getOwnPropertyNames(o: any): Array<string>;
-    static create(o: any, properties?: any): any; // compiler magic
+    static create(o?: any, properties?: any): any; // compiler magic
     static defineProperty(o: any, p: any, attributes: any): any;
     static defineProperties(o: any, properties: any): any;
     static seal(o: any): any;


samwgoldman:
Thanks, @adrianheine!

Can you share some code that exemplifies the issue that this PR solves? I quickly tested `Object.create()` and `Object.create(null)` and Flow accepts both already, but maybe I'm missing something.

AprilArcus:
The spec you linked states "If Type(O) is neither Object nor Null, throw a TypeError exception", so the type annotation we want for that first parameter is

`static create(o: Object | null, properties?: any): any;`

`static create(o?: any, properties?: any): any;`

allows the first parameter to be `undefined`.

samwgoldman:
I should also note that the libdef is somewhat for show. We special-case `Object.create` in the type checker in order to implement more of its behavior, since the type declaration language isn't (yet) sophisticated to describe it completely.

That said, happy to learn more about the background for this change.

adrianheine:
Hi, thanks for the great feedback :) I suppose this PR is quite useless. The problem I tried to solve can be seen in the following minimal example:

```javascript
/* @flow weak */
const options = Object.create(null)

function f(opt) {
  for (opt in options) {
    options[opt]
  }
}
```

Gives me (with 0.20.1):
```
options.js:2
  2: const options = Object.create(null)
                     ^^^^^^^^^^^^^^^^^^^ Object.create. This type is incompatible with
  2: const options = Object.create(null)
                                   ^^^^ null


Found 1 error
```

I didn't investigate and thought this issue had a trivial and obvious cause, but I realized that this issue is not too much about passing `null` to `Object.create`.

samwgoldman:
@adrianheine Do you think the right course of action is to close this issue, then?

adrianheine:
Sure, I'd like to get my issue triaged, though. Maybe it's handled in an already-open issue?

samwgoldman:
I think #1272 should cover it.

PR #1339, opened 2016-01-30
DOM Traversal, Take 2

AprilArcus:
I decided to start over from https://github.com/facebook/flow/pull/1305 using a more TDD approach. For now this just adds two new methods on `document` and a couple of tests to go with them. They are currently failing on my local build with this message:

```
traversal.js:6:15,32: property `createNodeIterator`
Property not found in
traversal.js:6:6,13: Document

traversal.js:10:15,30: property `createTreeWalker`
Property not found in
traversal.js:10:6,13: Document
```

I'm doing nothing at all fancy with the types here, so this is probably due to a misunderstanding of Flow's testing infrastructure on my part.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -316,7 +316,7 @@ declare class Document extends Node {
     importNode(importedNode: Node, deep: boolean): Node;
     createComment(data: string): Comment;
     getElementsByTagName(name: string): HTMLCollection;
-    createDocumentFragment(): Node;
+    createDocumentFragment(): DocumentFragment;
     getElementsByName(elementName: string): HTMLCollection;
     createAttribute(name: string): Attr;
     createTextNode(data: string): Text;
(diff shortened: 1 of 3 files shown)


samwgoldman:
Welcome back!

It looks like the tests failed, but with a different reason from the one in your comment. The Travis failure is a simple one—adding lines to `lib/dom.js` changes the output of existing test expectations. The fix for these errors is to simply update the expectations.

The failure in your comment indicates that Flow doesn't know about the declarations you added to `lib/dom.js`. It's our fault for not documenting this process well enough (TODO), but there's an extra step required between editing the files in `lib` in the source tree and testing flow—you need to build the binary.

To build the binary, you simply need to run `make`, but doing so successfully might require that you first install some dependencies. Check out the [instructions on the README](https://github.com/facebook/flow#building-flow) for some clues. If you run into any snags, feel free to respond here and I'll see if I can help troubleshoot.

samwgoldman:
Thanks again for your contribution. Let me know if there's anything I can do to help.

AprilArcus:
I will loop back to this tonight.

AprilArcus:
Okay, this is ready to go.

gabelevi:
@facebook-github-bot import

gabelevi:
@AprilArcus Thanks for the contribution and sorry for the huge delay before we took a look! I'm importing and taking a look!

gabelevi:
Hi @AprilArcus - spoke to @samwgoldman and he pointed out that Flow currently doesn't statically compute bitwise operations on known constants. That means even though Flow knows `1` has the value `1` and `2` has the value `2`, it doesn't know that `1 | 2` has the value `3`. This certainly can be done in the future, though for now it might cause trouble with this definition.

We're thinking for now we'd take the PR, but first merge the giant overload into a single function, which returns the most general type. It's not perfect, but it's simpler and the overload isn't too useful until bitwise operations are available. Also, the error messages for such a huge overload would be pretty messy if the types don't match.

Does that sound good to you? There's no action required on your side (I can update the PR as I merge it), I just wanted to check in with you before making these changes.

And thanks again for doing all this work!

AprilArcus:
Up to you. I do think the overload can be useful without support for bitwise operations, since often a single constant will be used. My main (only) use case for NodeIterator to quickly & easily obtain all the text nodes in a DOM subtree, which is impossible with `querySelectorAll()`, and in that case I would like Flow to know that I my iterator returns all and only nodes of type `Text | null`.

gabelevi:
Ok, here's what I'm doing:

Ok, here's what I've done

1. Kept all the overloads that you painstakingly and awesomely enumerated
2. Made whatToShow non-optional in all those overloads. The optional case needs to be handled separately
3. Added cases for when the literal value of whatToShow is any number. It defaults to the SHOW_ALL case, which seems like the most general return type.
4. Added cases for when whatToShow is omitted. Per the docs, it defaults to the SHOW_ALL case

PR #1341, opened 2016-01-31
Update year to 2016 in footer layout

zaynetro:
Going back to the future


gabelevi:
whatyearisit.jpg

Thanks for fixing this!

PR #1343, opened 2016-01-31
Add Object.setPrototypeOf definition

kittens:


kittens:
Although maybe this requires some extra Flow magic under the hood like `Object.getPrototypeOf`.

kittens:
Unsure if CI failure is due to this PR.

samwgoldman:
I think we might need some Flow magic to make this work. Can you share some test cases that would exercise this behavior? I think we might run into some issues swapping out prototypes.

Test failures are consequent of this change, but trivially. The test expectations record line numbers; the added line shifts some line numbers around. Solution is to update expectations. We can sort that out easily later, though. More interesting is the test cases.

kittens:
```javascript
var obj = { foo: "foo" };
Object.setPrototypeOf(obj, null);
obj.hasOwnProperty; // flow error

var obj2 = { bar: "bar" };
obj2.foo; // flow error
Object.setPrototypeOf(obj2, obj);
obj2.foo; // not a flow error
```

FWIW I don't actually need this behaviour. I just want to set an objects `__proto__` to `null` and Flow is complaining when I do `foo.__proto__ = null;`.

vkurchatkin:
It would be nice to have explicit support for prototype types, something like this:

```js
type TypeWithProto = $Obj<{ ownProp: string }, { inherited: () => void }>

type OwnType =  $Own<TypeWithProto>; // { ownProp: string },
type ProtoType = $Proto<TypeWithProto>; // { inherited: () => void }
```

Then this declaration could be just:

```js
static setPrototypeOf<T, P>(o: T, p: P): $Obj<$Own<T>, P>;
```

samwgoldman:
If `Object.setPrototypeOf` returned a fresh object, we could make this work with "magic" support within Flow (although we do need to increase the expressivity of object types, as @vkurchatkin points out).

However, `Object.setPrototypeOf` actually mutates its argument, which means we'll run afoul of heap invariance. Naive support would require that all uses of the object type must be compatible with *both* prototypes.

That might not be a problem in practice, if the intended use is to set the prototype to `null` at object creation time, but the naive solution would require that the object be used as if it had a `null` prototype for its entire lifetime.

mroch:
gonna close this since it sounds like we want to do the more advanced internal support. lmk if we should reopen.

Issue #1344, opened 2016-02-01
Inconsistent docs: Objects / Adding properties


amakhrov:
The [docs say](http://flowtype.org/docs/objects.html#adding-properties):
````
file.js:3:16,16: undefined
This type is incompatible with
  file.js:8:7,12: string
````
> In other words, Flow knows enough to infer that whenever the x property of o does exist, **it is a number**, so a string should not be expected.

So the text piece is not consistent with the code sample / output. Also it's not illustrative enough, as the order of `foo()` and `bar()` calls differs from the previous code snippet

gabelevi:
Closing this out, [per my comment on #1345](https://github.com/facebook/flow/pull/1345#issuecomment-203247954). Thanks for the report!

amakhrov:
@gabelevi looks like the new snippet of the code still lacks the `return` statement:
````
function foo(p) { p.x = 42; }
function bar(q) { q.f(); }
````
here the `bar` function doesn't return anything (return type is always underfined). My assumption is that it's not exactly what you wanted to illustrate by this snippet, considering the explanation below:

> Flow knows enough to infer that whenever the x property of o does exist, **it is a number**, so a string should not be expected.

gabelevi:
Gah, yeah, I see it. I'll patch it quickly.

PR #1345, opened 2016-02-01
Fix the code example in the Objects / Adding Properties (#1344)

amakhrov:
Addresses #1344

### DIFF
--- a/docs/01-objects.md
+++ b/docs/01-objects.md
@@ -160,12 +160,12 @@ For example, the following code typechecks:
 {% highlight javascript linenos=table %}
 /* @flow */
 function foo(p) { p.x = 42; }
-function bar(q) { q.f(); }
+function bar(q) { return q.f(); }
 
 var o = { };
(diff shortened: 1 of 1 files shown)


gabelevi:
Hi @amakhrov 

First of all, I'm terribly sorry it took us 2 months to look at this pull request. I'd love to shift the blame to someone else, but it's primarily my fault. We will try to do better in the future!

We've recently revamped the docs, and I believe [the new docs](http://flowtype.org/docs/objects.html#adding-properties) correct the issue you noticed in #1344.

I'm going to close this and #1344 since I believe the issue is resolved, but please re-open if I'm missing something.

amakhrov:
Hi @gabelevi 
Thanks for the feedback. I love the new 'interactive' (with tooltips) format of the docs.

However, the issue is still present in the sample code - will re-open the issue and add more details

gabelevi:
@amakhrov I feel dumb. I see the issue.

PR #1351, opened 2016-02-02
fix mistakes in code examples

jedwards1211:
Saying that `b` is type `Foo` or "of type" `Foo` doesn't make any sense, all you can say is that `b` contains a property of type `Class<Foo>`.

Moreover these code samples were clearly intended to demonstrate that the type of `new b.f()` is `Foo` and as such can be assigned to a variable of type `Foo` but can't be assigned to a variable of type `Bar`.  However, they didn't actually demonstrate that at all.  The missing prop error caused by `new b.g()` is irrelevant to this explanation of `typeof`.

I've updated the code and comments to illustrate the effect of `typeof` on instantiating with `new` clearly.

### DIFF
--- a/docs/01-typeof.md
+++ b/docs/01-typeof.md
@@ -70,28 +70,24 @@ whether that is a class, module or some other construct.
 {% highlight javascript linenos=table %}
 /* @flow */
 class Foo { }
+class Bar { }
 // b ends up being a Foo type, since f evaluates to Foo
 var b: { f : typeof Foo } = { f : Foo };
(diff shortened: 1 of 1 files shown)


mroch:
oops, we moved all these files from gh-pages to master last weekend. i'll take care of rebasing and getting this in.

gabelevi:
@facebook-github-bot import

gabelevi:
Oh, I see what @mroch meant about moving branches. Let me see if I can get this sorted out...

PR #1360, opened 2016-02-03
[RFC] `flow-lib`: JS interfaces for Flow APIs

jeffmo:
This is a first-pass on an idea for an npm package, called `flow-lib`, that allows JS tooling consumers to interface easily with Flow APIs. Pending more experimentation, my hope is that this can become a stable way for JS tooling to integrate with Flow via a simple, well-versioned npm dependency.

To start out I've added JS APIs for `flow check-contents`, `flow ast`, and Flow type definitions for the AST. I also intend to add `type-at-pos`-like APIs as well (all of the tooling needs I can imagine right now need this more than anything else) -- but before that I need to find a good way to represent types that isn't coupled too tightly with Flow implementation details. I'm still working on this and will update when I have something more here.

**NOTE: This is only an RFC while I look into more concrete use cases. I am putting this out for discussion, but I don't necessarily expect to merge it until we see it prove useful or promising for some *specific* tooling needs**


phpnode:
Exciting! It'd be really awesome if we could access the inferred types from the AST as well, although possibly out of scope for this PR.

jeffmo:
@phpnode: Yea, having a way to normatively access and refer to types (inferred and otherwise) is something I'd like to solve before moving landing this idea.

Also worth noting that this is something I intend to hack on in my free time for now, so I don't expect this to move too quickly (unless we can make a strong case for it being important) -- but I am still casually playing around here.

gabelevi:
I would love to use well-written JS-interfaces for Flow in the JS scripts I write too! Definitely much better than my current solution of exec'ing Flow!

ryyppy:
Since the cli is quite deep-water with secret commands (AFAIK I only found out about certain features through either flow --help, reading vim-flow or other hidden resources) ... I would like to know which commands are actually missing to make this feature complete?

jeffmo:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3676384) to review internal test results.

jeffmo:
@ryyppy: Awesome! Let's get this landed then we can add commands on top from there.

The obvious top-priority commands after the ones added in this PR would be `flow status` (results of typechecking), `flow coverage`, `flow type-at-pos`, `flow get-def`... There are more, but we can add them as needed I guess.

ryyppy:
Okay let's do this!

PR #1378, opened 2016-02-05
createDocumentFragment returns a DocumentFragment

adrianheine:
See https://www.w3.org/TR/DOM-Level-3-Core/core.html#ID-35CB04B5


mroch:
good catch, thanks!

mroch:
@facebook-github-bot shipit

PR #1385, opened 2016-02-09
Remove extraneous type param on Promise

ForbesLindesay:


gabelevi:
Heh, yeah, that was pretty useless, wasn't it! Sorry for the sloppiness and thanks for fixing!

We ended up deleting `done()` from our lib anyway, since it was non-standard. Anyone who needs to use a non-standard version of `Promise` can put a lib definition in their project and point the flowconfig at it. So I'm going to close this out.

Thanks again, and sorry for the slow response!

PR #1386, opened 2016-02-09
Adding whatwg fetch

runn1ng:
I have added whatwg fetch to declarations.

The specs are not finished yet, so it might be different in the end, but I added stuff from MDN, plus some other things. 

https://stackoverflow.com/questions/35296664/can-fetch-get-object-as-headers

(I am not sure if I understand `mixins` correctly, but I think I do :))

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -354,3 +354,121 @@ declare class PositionOptions {
     timeout: number;
     maximumAge: number;
 }
+
+
+// this part of spec is not finished yet, apparently
+// https://stackoverflow.com/questions/35296664/can-fetch-get-object-as-headers
+type HeadersInit = Headers | {[key: string]: string};
+
+
+// TODO Heades and URLSearchParams are almost the same thing.
+// Could it somehow be abstracted away?
+declare class Headers {
+    @@iterator(): Iterator<[string, string]>;
+    constructor(init?: HeadersInit): void;
+    append(name: string, value: string): void;
+    delete(name: string): void;
+    entries(): Iterator<[string, string]>;
+    get(name: string): string;
+    getAll(name: string): Array<string>;
+    has(name: string): boolean;
+    keys(): Iterator<string>;
+    set(name: string, value: string): void;
+    values(): Iterator<string>; 
+}
+
+declare class URLSearchParams {
+    @@iterator(): Iterator<[string, string]>;
+    constructor(query?: string | URLSearchParams): void;
+    append(name: string, value: string): void;
+    delete(name: string): void;
+    entries(): Iterator<[string, string]>;
+    get(name: string): string;
+    getAll(name: string): Array<string>;
+    has(name: string): boolean;
+    keys(): Iterator<string>;
+    set(name: string, value: string): void;
+    values(): Iterator<string>; 
+}
+
+type CacheType =  'default' | 'no-store' | 'reload' | 'no-cache' | 'force-cache' | 'only-if-cached';
+type CredentialsType = 'omit' | 'same-origin' | 'include';
+type ModeType = 'cors' | 'no-cors' | 'same-origin';
+type RedirectType = 'follow' | 'error' | 'manual';
+type MethodType = 'GET' | 'POST' | 'DELETE' | 'HEAD' | 'OPTIONS' | 'PUT' ;
+type ReferrerPolicyType =
+    '' | 'no-referrer' | 'no-referrer-when-downgrade' | 'origin-only' |
+    'origin-when-cross-origin' | 'unsafe-url';
+type ResponseType =  'basic' | 'cors' | 'default' | 'error' | 'opaque' | 'opaqueredirect' ;
+
+type RequestOptions = {
+    body?: Blob | FormData | URLSearchParams | string;
+
+    cache?: CacheType;
+    credentials?: CredentialsType;
+    headers?: HeadersInit; 
+    integrity?: string;
+    method?: MethodType;
+    mode?: ModeType;
+    redirect?: RedirectType;
+    referrer?: string;
+    referrerPolicy?: ReferrerPolicyType;
+}
+
+type ResponseOptions = {
+    status?: number;
+    statusText?: string;
+    headers?: HeadersInit
+}
+
+declare class Response {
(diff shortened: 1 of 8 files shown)


runn1ng:
The only thing I have differently from MDN is adding object directly to headers (it seems to work everywhere and the specs seem to imply that it's meant like that, see my linked stackoverflow.com question)

samwgoldman:
Woohoo! Thanks for doing the leg work here :)

Looks like the build was broken when you branched. Can you try rebasing to see if things are fixed now?

Also, would you mind adding a few tests? Even if the declarations are right, it's possible that we might run afoul of some bugs in Flow around union types, or just unexpected behavior.

Let me know if I can help or provide any more specific guidance.

runn1ng:
Haha, well I wrote it,  because I wanted to use fetch in one project :)

Ok, I will try, but probably won't get to all the edges.

samwgoldman:
I wouldn't aim for 100% coverage, but if you could hit the main use cases, that'd be really helpful.

runn1ng:
Oh, one thing I would like some help with. Some of the attributes are described as "read-only" in MDN. Is that somehow possible to write in flow declarations?

samwgoldman:
We don't currently support that, but it's definitely on our radar.

runn1ng:
One more thing. :) This is what MDN writes

> An object implementing `Headers` can directly be used in a `for...of` structure, instead of `entries()`: `for (var p of myHeaders)` is equivalent to `for (var p of myHeaders.entries())`.

How can this be done? Headers should have `mixins Iterator<[string, string]>`, right?

runn1ng:
Hm. This doesn't seem to work

`declare class Headers implements Iterator<[string, string]>`

neither does this

`declare class Headers mixins Iterator<[string, string]>`

samwgoldman:
We don't support `implements` on class declarations, although it's a feature we'd like to add in the future. `mixins` is designed to take a list of classes, and `Iterator` is an interface, not a class.

To implement an interface, you "simply" need to provide the methods that make your class compatible with the interface. From the docs, it looks like you want to support the `Iterable` interface. Technically, objects implement that interface by having a computed property with a key value of `Symbol.iterator`, but we are a bit behind on support for computed properties and symbols, so the temporary hack is to declare a property named `@@iterator` with an iterator value.

So, long story short, something like this:

```
declare class Headers {
  @@iterator: Iterator<[string, string]>;
}
```

Check out `lib/core.js` for some examples of this.

AprilArcus:
Hey, thank you for doing this! A couple minor comments:

* there is no constructor or other global object corresponding to `Body`, so it might be better to declare it as a type alias for an object rather than a class. Then rather than having `Request` and `Response` `mixins` or `extends` `Body`, you could intersect on it instead.

* to annotate that `Headers` implements the `iterator` interface, maybe you could add a `@@iterator(): Iterator<[string, string]>;` method, as `Map`, `Set` do?

* `RequestOptions` is missing the `referrerPolicy` field, which should be an enum:

```
type ReferrerPolicy =
  '' | 'no-referrer' | 'no-referrer-when-downgrade' | 'origin-only' |
  'origin-when-cross-origin' | 'unsafe-url';
```

* `Response::error()` and `Response::redirect()` [should be `static`](https://fetch.spec.whatwg.org/#response-class)

* `Response.type` should be an enum:

```
type ResponseType =
  'basic' | 'cors' | 'default' | 'error' | 'opaque' | 'opaqueredirect';
```

runn1ng:
Thanks for the feedback :)

runn1ng:
I am not sure how do you mean the intersection.

You cannot do this in Flow (vastly simplified)

```
type Body = {
    bodyUsed: boolean;
}

declare class Request {
    constructor(input: string): void;
    url: string;
}

type RequestBody = Request&Body;

var response: RequestBody = new Request("something"); //it doesn't implement Body!
```

How would you do it? The only thing I can think of is just copy-pasting the `Body` methods and attributes into Response and Request directly, and I just don't like copy-pasting large amount of lines.

Maybe I can try to rename `Body` to `_Body` and hopefully nobody will instantiate something with underscore? That's an ugly hack too.

AprilArcus:
I'm sorry; you're absolutely right about that — I wasn't thinking clearly when I made that suggestion :/

If this were a module declaration one could just decline to export the `Body` class, but if I understand correctly everything in /lib goes into the global type namespace.

runn1ng:
I have added some tests and your feedback.

Both the master branch and the PR is now failing on OS X with Travis. I cannot really do anything with that.

runn1ng:
...ok, now it tested. (I don't think the OS X test results are very deterministic. :) )

runn1ng:
hey, is there any problem with this? I would appreciate if it was in master :)

samwgoldman:
Sorry, just got lost in the shuffle. Will take a look today.

gabelevi:
@facebook-github-bot import

gabelevi:
Sorry for the delay! Will get this reviewed and checked in ASAP!

runn1ng:
I am not sure if I added this correctly.

I am not sure if `json(): Promise<Object>;` is correct. I was under the impression that it has to be object, since that's how I read _that resolves with a JSON object_ from the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Response).

However, apparently, "JSON object" doesn't mean "Object", but can actually be all this:

> JSON = null
>     or true or false
>     or JSONNumber
>     or JSONString
>     or JSONObject
>     or JSONArray

I am, however, not sure. I haven't tried that yet, since in all my code the APIs return Objects. Can someone else chime in?

runn1ng:
From the official spec:

> JSON
> Return the result of invoking the initial value of the parse property of the JSON object with the result of running utf-8 decode on bytes as argument. Rethrow any exceptions.

Well then it's not correct and should return the same thing as `JSON.parse` returns.

And when I look to JSON.parse library definition, [it just returns `any`](https://github.com/facebook/flow/blob/master/lib/core.js#L356). 

OK, I will make another PR that will correct this. I personally don't like writing `any` anywhere, but it cannot probably be helped here.

(In my opinion, `JSON.parse` should return `mixed` and not `any`, but that's beyond whatwg-fetch.)

runn1ng:
pull request here

https://github.com/facebook/flow/pull/1636

Issue #1117, opened 2015-11-19
in lib.dom.js: HTMLImageElement.crossOrigin and HTMLMediaElement.crossOrigin should be nullable


ckknight:
Currently HTMLImageElement and HTMLMediaElement define their crossOrigin property as `string`, despite the default value being `null`.

mroch:
[the spec](https://html.spec.whatwg.org/multipage/infrastructure.html#cors-settings-attribute) says that a _missing default value_ represents "no CORS" but an _invalid default value_ is equivalent to `"anonymous"`. I'm not sure, but I interpret it as `null` being an invalid value (it's a value that is not `"" | "anonymous" | "use-credentials"`), so it's actually different than it being `undefined`.

I think it should be `string | void` not `string | null`. even better would be for it to be `"" | "anonymous" | "use-credentials" | void`, though.

mroch:
but you're right, if I inspect an <img> on this page and do `$0.crossOrigin` I get `null`, not `undefined`. would love to find what justifies that behavior...

PR #1403, opened 2016-02-14
Allow `crossOrigin` to be nullable for HTMLImageElement, and HTMLMediaElement.

majjoha:
This pull request updates the definition of the `crossOrigin` property for HTMLImageElement, and HTMLMediaElement, so they are nullable as mentioned in https://github.com/facebook/flow/issues/1117.


mroch:
@facebook-github-bot shipit

PR #1404, opened 2016-02-14
Improve definitions for `Symbol`.

majjoha:
This updates the type definitions for `Symbol` as per [spec](http://www.ecma-international.org/ecma-262/6.0/#sec-symbol-objects).

After running `make test`, it seems that these changes break the tests. @samwgoldman or @mroch, would you mind providing some guidance on how I can fix them? As far as I can tell, it seems that the primary issue is a line offset in the various `*.exp` files.


mroch:
the tests work by comparing the `.out` file generated by running each test to the `.exp` file that's checked in. so to update the tests, run the tests to generate a new `.out`, then overwrite the `.exp` file with the `.out` file.

majjoha:
Thanks for clarifying, @mroch. I've updated all the relevant `.exp` files.

samwgoldman:
Thanks for doing this, @majjoha! We still need to add some more typing logic for symbol, but this is definitely a big improvement to the status quo.

@facebook-github-bot import

PR #1433, opened 2016-02-22
[docs] update of `flow --help` print

mkwtys:


mroch:
good catch, thanks!

mroch:
@facebook-github-bot shipit

Issue #1420, opened 2016-02-18
Flow is trying to parse JSON files as JS


grncdr:
A small repo to demonstrate: https://github.com/grncdr/flowtest, running `flow status` produces the following output:

```
src/wat.json:1
  1: []
     ^ Unexpected token [


Found 1 error
```

This happens on the current master: 079271416e89552d23f903ede11c93592ab5f149, I don't know when it was introduced. It currently breaks flow for most any project using a non-trivial amount of npm packages, as these often have static JSON files for test fixtures and whatnot that flow will now find and break on.

Other than that, the intersection & union type fixes in `master` are awesome! Keep up the great work on Flow! :heart: :heart:

grncdr:
Still happening with the 0.22 release.

arypurnomoz:
0.21 was okay i only got this problem on 0.22

jamiter:
A temporary solution is to ignore the JSON files:
```
[ignore]
.*/node_modules/.*\.json$
```

grncdr:
@jamiter unfortunately that seems to cause flow to ignore `package.json` files, which then causes any modules that make use  of the "main" field in package.json to be unresolvable.

grncdr:
for reference this was the regex I ended up with in my `[ignore]`:

```
.*\(lib\|src\|test\|builtin-modules\|binary-extensions\|faker\|spdx\).*\.json
```

(as far as I can tell Ocaml regexps don't support negative look-behind)

unscriptable:
Yep. 0.21 worked fine. 0.22 is b0rked.  The only work-around I can find is what @grncdr suggests: explicitly ignoring all of the json files in node_modules/ except package.json files.

jamiter:
@grncdr, thank you! That explained those unresolvable errors! (Duh) Finally have it all working again. This bug of course still needs to be fixed.

jeffmo:
FYI: I just landed https://github.com/facebook/flow/pull/1436 -- which lets Flow parse JSON files that have toplevel array structures (rather than just toplevel object structures).

Based on the error in the OP here, I think that should fix the error that caused that issue!

mroch:
so we're now compliant with [RFC 4627](https://tools.ietf.org/html/rfc4627#section-2) (the original JSON RFC), thanks!

but we're not with [RFC 7159](https://tools.ietf.org/html/rfc7159#section-2). the latter allows false / null / true / object / array / number / string.

[...]

PR #1436, opened 2016-02-23
Support JSON files with toplevel arrays

jeffmo:
There are plenty of JSON files in the wild (on npm, etc) that start with an array at the toplevel. This is a source of some moderate number of JSON file parsing errors that I've seen.

This fixes the json-parser function to handle both toplevel objects and arrays.


mroch:
the test failures are legit

zertosh:
This is great! But why not cover all the data types that can be toplevel values? string, number, boolean, null ?

mroch:
@zertosh see #1420, we're going to do that

zertosh:
Sorry, I missed that. Thanks!

PR #1335, opened 2016-01-29
Add type definitions for Object.entries and Object.values

ianobermiller:
Add type definitions for Object.entries and Object.values. See a test here:

http://tryflow.org/#5d9b08200a09cd3b88cae9fc59b5045b

PR #1450, opened 2016-02-24
Add Object.values to lib declaration

### LINKED ISSUE #1449: Missing Object.values in flow lib definitions: property `values`: Property not found in statics of Object
Noticed that I am getting a flow issue when using Object.values(): 

```js
const vals = Object.values({a: 1, b: 2});
```

##### Message
```
property `values`: /file.js:1
Property not found in
statics of Object
```

Didn't notice other issues yet for this, would be happy to add a lib declaration for this.

bassettsj:
Resolves #1449 

Changes:

- simply add [`Object.values()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values) signature to Object definition.


samwgoldman:
@bassettsj Thanks! There was some previous work done on this over in #1335. That PR includes a comment from me with tests around this behavior. Do you think you could base your change off that one, and include tests?

bassettsj:
@samwgoldman yeah, I think I can help. I have updated the current tests for the changes in the expected output from adding the two lines from #1335 and now I just need to work on covering the test cases from your comments. 

What are the steps for writing the tests here, do I just simply provide the example cases where I would expect type errors to be thrown and then update the expected output from flow in the tests? I didn't see much documentation around writing the unit test here :grin:

samwgoldman:
> do I just simply provide the example cases where I would expect type errors to be thrown and then update the expected output from flow in the tests?

Yes, precisely. Add the tests, verify the errors are what you'd expect, and update the `.exp` file (copy from `.out` or apply the `.diff`). If we make a change that alters the expected output, we check to see if that's a regression.

bassettsj:
Great, maybe we should add those steps to the readme in a future PR :-)

samwgoldman:
> add those steps to the readme

Yes!

Also, and I'm sorry for not being more clear about this, but the commits in #1335 are subtly wrong. See my comment there for all the details.

Thank you for contributing!

PR #1453, opened 2016-02-25
Add missing OCaml C runtime macros, occasional style changes as well

fxfactorial:
Signed the CLA as well.

I have half a mind to reimplement the command line with cmdliner, you guys are missing out on a beautiful API and auto generated man page and most of all much less code under command.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -195,7 +195,7 @@ typedef struct {
 static size_t shared_mem_size;
 
 /* Beginning of shared memory */
-static char* shared_mem = 0;
+static char* shared_mem = NULL;
 
 /* ENCODING: The first element is the size stored in bytes, the rest is
  * the data. The size is set to zero when the storage is empty.
@@ -232,17 +232,17 @@ static char* heap_init;
 /* This should only be used by the master */
 static size_t heap_init_size = 0;
 
-static size_t used_heap_size() {
+static size_t used_heap_size(void) {
   return *heap - heap_init;
 }
 
 /* Expose so we can display diagnostics */
-value hh_heap_size() {
+CAMLprim value hh_heap_size(void) {
   CAMLparam0();
   CAMLreturn(Val_long(used_heap_size()));
 }
 
-value hh_hash_used_slots() {
+CAMLprim value hh_hash_used_slots(void) {
   CAMLparam0();
   uint64_t count = 0;
   uintptr_t i = 0;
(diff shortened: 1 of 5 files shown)


mroch:
cc @int3 @jwatzman @dlreeves for the hack/ changes

jwatzman:
Not on the Hack team anymore, @int3 and @dlreeves should take a look. I did a *very* quick skim, it looks fine to me, except perhaps the changes to `src/embedded/flowlib_elf.c` -- not clear if those `free` calls are correct, would have to look at the libelf API again (ugh).

fxfactorial:
@jwatzman I can remove those `free` calls to play it safe, it did look like a memory leak on the face of it.

mroch:
the linux travis builds are segfaulting :(

fxfactorial:
@mroch Okay sorry about that, I will try building it as well and fix it.

fxfactorial:
Okay this is odd, I don't get a segfault on my ubuntu VM, albeit using later versions libs and compilers:

http://pastebin.com/VEnq118f

Also using a `nixpkg` version of `libelf`.  I will try to reproduce the build environment exactly.

fxfactorial:
@jwatzman @mroch @int3 @dlreeves I took away the three `free` calls in the `elf` function call and now the seg fault goes away. 

Sorry for delay.

gabelevi:
This looks good to me too.

mroch:
@facebook-github-bot shipit

mroch:
thanks, let's do this

fxfactorial:
I can't tell if this was actually merged? Should I close this PR?

samwgoldman:
The commit message will autoclose this when it lands. Looks like there are some build errors internally that are preventing this from landing.

fxfactorial:
@samwgoldman hmm, a little hard for me to fix them if I can't see them.

samwgoldman:
Yeah, I agree. We'll look into it ASAP.

hack/heap/hh_shared.c
```diff
@@ -592,7 +594,7 @@ void hh_add_dep(value ocaml_dep) {
   htable_add(deptbl, dep >> 31, dep);
```
mroch: ah, finally figured out what was wrong in our internal tests (and I'm still not sure why it doesn't repro in travis).

adding `CAMLparam1` here without also adding `CAMLreturn0` here and the `return` a couple lines up meant that `ocaml_dep` was never freed and it'd OOM.

I'm including that fix, no need to update. should be landing soon. thanks for your patience!

fxfactorial: @mroch Much appreciated, thanks! (Also found FB infer recently, awesome work guys! love all what you do)

PR #1459, opened 2016-02-26
Add support for flow prefixes in package.json keys

kittens:
WIP

This PR allows you to prefix keys in `package.json` with `flow:`. This is useful as you can trivially get type checking on a Flow project published to npm by just including the `src` directory and adding a `flow:main` field.

This is different to Flow comments as it preserves the original source and is different to the `.flow` extension as it doesn't require any convoluted build system commands.

Completely understand if you'd rather just have `.flow` so feel free to close but I think `flow:main` makes things substantially easier.

### DIFF
--- a/src/services/inference/module_js.ml
+++ b/src/services/inference/module_js.ml
@@ -157,12 +157,17 @@ module ReversePackageHeap = SharedMem.WithCache (StringKey) (struct
     let prefix = Prefix.make()
   end)
 
-let get_key key tokens = Ast.(
+let get_key' key tokens = Ast.(
   match SMap.get (spf "\"%s\"" key) tokens with
   | Some (_, Expression.Literal { Literal.value = Literal.String name; _ }) ->
       Some name
   | _ -> None
 )
+
+let get_key key tokens =
+  match get_key' (spf "\"flow:%s\"" key) tokens with
+  | Some x -> Some x
+  | _ -> get_key' (spf "\"%s\"" key) tokens
 
 let get_package_keys filename ast =
   let open Ast in
--- a/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/dir/index.js
+++ b/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/dir/index.js
@@ -1,0 +1,4 @@
+/* @flow */
+
+class Implementation {}
+module.exports.fun = (): Implementation => new Implementation();
--- a/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/dir/index.js.flow.ignored
+++ b/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/dir/index.js.flow.ignored
@@ -1,0 +1,4 @@
+/* @flow */
+
+declare class Definition {}
+declare export var fun: () => Definition;
--- a/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/package.json
+++ b/tests/declaration_files_incremental_node/node_modules/package_with_dir_flow_main/package.json
@@ -1,0 +1,2 @@
+{ "name": "package_with_full_main",
+  "flow:main": "dir" }
--- a/tests/declaration_files_incremental_node/node_modules/package_with_full_flow_main/code.js.flow.ignored
+++ b/tests/declaration_files_incremental_node/node_modules/package_with_full_flow_main/code.js.flow.ignored
@@ -1,0 +1,5 @@
+/* @flow */
+
+declare class Definition {}
+declare var fun: () => Definition;
+declare export { fun };
--- a/tests/declaration_files_incremental_node/node_modules/package_with_full_flow_main/code.js.ignored
+++ b/tests/declaration_files_incremental_node/node_modules/package_with_full_flow_main/code.js.ignored
@@ -1,0 +1,4 @@
+/* @flow */
+
+class Implementation {}
+module.exports.fun = (): Implementation => new Implementation();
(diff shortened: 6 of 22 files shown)


kittens:
Tests are failing, not sure what's wrong, I just see massive diff failures that look unrelated to what I changed.

gabelevi:
Woah, when did you switch to @kittens?

I think I'm not quite following your use case. Can you give an example? It doesn't sound like your talking about a situation with both a `main` and a `flow:main` like

```JavaScript
{
  "main": "src/transformed_index.js",
  "flow:main": "src/original_index.js",
  ...
}
```

since this doesn't seem to save you any build trouble. And it's not clear to me why you would have `flow:main` without a `main`.

Your OCaml looks good, though!

As for tests, I can look when Travis finishes. The incremental script runs [this bash script](https://github.com/facebook/flow/blob/48cd43de5c6198eaa202779799cf2cd469bbaa96/tests/declaration_files_incremental_node/test.sh) which moves the various files around, making sure that Flow's incremental mode follows the changing. That's probably one source of issues.

src/services/inference/module_js.ml
```diff
@@ -157,13 +157,18 @@ module ReversePackageHeap = SharedMem.WithCache (StringKey) (struct
     let prefix = Prefix.make()
   end)
 
-let get_key key tokens = Ast.(
+let get_key' key tokens = Ast.(
   match SMap.get (spf "\"%s\"" key) tokens with
```
gabelevi: Oh but you should probably get rid of this `spf` since you're doing it in `get_key`. The double quoting is probably not helping the tests.

Or only do the quoting in here. Your call!

kittens:
@gabelevi 

> Woah, when did you switch to @kittens?

Few weeks ago :stuck_out_tongue: It removes the confusion and accidental mentions with the other Facebook JavaScript React Sebastian.

> I think I'm not quite following your use case. Can you give an example? It doesn't sound like your talking about a situation with both a main and a flow:main like

I'm actually talking about that situation exactly. I recently published [kreporters](https://github.com/kittens/kreporters) which uses Flow types and there's only a few ways I can get third party consumers to use the types and they all aren't great:

* Flow comments: Doesn't support classes which I'm using and the transformations aren't complete lossless so there's a lot of information lost and Flow sometimes doesn't actually understand it.
* `.flow`: I'm not using a build system. I just use the Babel CLI like `$ babel src --out-dir lib`. I could switch to Gulp or Grunt to do my builds but that seems like overkill for such a small library. The equivalent shell command would probably be `$ find src -name ".js" -exec cp src/{} lib/{}.flow \;`, but that seems pretty gnarly and it wont work on Windows.
* Interface: Provide an interface file and make someone copy it to their project. I'd need to maintain this separately from the typed files and it's very manual and only suitable for projects that aren't typed.

kittens:
Also just want to highlight that having this ability means that libraries can provide type information without any changes to their build system. For example you just have to remove `src` from `.npmignore` and add `"flow:main": "src/index.js"` to your `package.json`.

gabelevi:
Ah ok, I get it. The requirement of putting `foo.js.flow` next to `flow.js` is inconvenient and it's much easier to tell node to read `lib/` and flow to read `src/`.

Yeah, this makes sense to me. I'd merge this once the tests pass :)

mroch:
how does this compare to the `typings` key TypeScript uses? is there really any value to being able to `flow:`-prefix every key? do other projects use : as namespaces?

an alternative would be `{ "flow": { "main": ... } }`. another thing we might put under the flow key would be something like .flowconfig's `module.name_mapper` so that you can map `require('foo/lib/derp')` to `foo/src/derp.js`

kittens:
@mroch 

> how does this compare to the `typings` key TypeScript uses?

You still need to point to an interface file as far as I know. This would allow you to point at your original source and have Flow point to that instead of some random location in `lib`.

> is there really any value to being able to `flow:`-prefix every key?

Not really, the only other key that's used is `name`. Figured it was more future proof to do it for any key to support it in automatically in the future.

> do other projects use : as namespaces?

`jsnext:main` is used by all the major module bundlers that support ES2015 modules. It's used to provide an alternate ES2015 build that they can consume while `main` still points at a CommonJS build. For example here are the [rollup docs](https://github.com/rollup/rollup/wiki/jsnext:main).

> an alternative would be `{ "flow": { "main": ... } }`

Keys used like that in `package.json` are historically used for build config. For example [Babel](https://github.com/babel/babel) uses a field called [`babel`](https://babeljs.io/docs/usage/babelrc/#use-via-package-json) and [ava](https://github.com/sindresorhus/ava) uses [`ava`](https://github.com/sindresorhus/ava#configuration). So it would seem more appropriate for the `flow` namespace in `package.json` to be analogous to `.flowconfig`. There's also precedence with `jsnext:main` to prefix `main` with a namespace.

kittens:
Actually looking into `typings` more, it does appear to be the same as `flow:main` as it can consume arbitrary typed files too. I think there's a compatibility hazard with using it though unless Flow supports all the TypeScript syntax.

gabelevi:
> an alternative would be { "flow": { "main": ... } }

This was my initial thought too, but I settled on the idea that `flow:foo` was good for overriding the value of `foo` for Flow while ` "flow": { ... }` would be good for flow specific stuff (like `.flowconfig` stuff) if we decide to allow any of it in the package.json

kittens:
Just an update, I'm still going to work on this PR. Should have addressed all the comments and have tests passing next week.

steelbrain:
Tara rum pum bump 😄

mroch:
I'm gonna close this to clean up the queue  since it's stale, but obviously feel free to resurrect it

STRML:
Shame to see this closed, we're using https://github.com/AgentME/flow-copy-source in some cases but for rollup builds et al it would be much better/easier to have a `flow:main`.

steelbrain:
I've opened https://github.com/facebook/flow/issues/2675 so this can be discussed and eventually fixed :)

PR #1463, opened 2016-02-26
Fix singleton print

### LINKED ISSUE #1440: type-at-pos fails with literal string enum types
e.g.
```
type MyType = "one" | "two"
let variable: MyType;
//    ^ type-at-pos here, to find out the type of `variable`, reports "unknown"
```

version: v0.22.0

(just tried literal number enums, and they also fail)

jaredly:
The type printer already has logic to print these types, but the
`is_printed_type_parseable` fn didn't seem to know that. This fixes #1440, and
now singleton strings, numbers, and bools don't foul up `type-at-pos`
reporting.

### DIFF
--- a/src/typing/type_printer.ml
+++ b/src/typing/type_printer.ml
@@ -256,6 +256,9 @@ let rec is_printed_type_parsable_impl weak cx enclosure = function
   | BoolT _
   | AnyT _
   | NullT _
+  | SingletonStrT _
+  | SingletonNumT _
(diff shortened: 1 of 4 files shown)


mroch:
d'oh, that `_ -> false` keeps biting us every time we add new types! thanks for the fix

mroch:
@facebook-github-bot shipit

samwgoldman:
Thanks, @jaredly!

PR #1464, opened 2016-02-26
Add dom definition for iframe

tasti:
This adds a type definition for the `iframe` element, since it isn't built-in already.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -293,6 +293,7 @@ declare class Document extends Node {
     createElement(tagName: 'canvas'): HTMLCanvasElement;
     createElement(tagName: 'div'): HTMLDivElement;
     createElement(tagName: 'form'): HTMLFormElement;
+    createElement(tagName: 'iframe'): HTMLIFrameElement;
     createElement(tagName: 'img'): HTMLImageElement;
     createElement(tagName: 'input'): HTMLInputElement;
     createElement(tagName: 'label'): HTMLLabelElement;
@@ -749,6 +750,20 @@ declare class HTMLFormElement extends HTMLElement {
     namedItem(name: string): any;
     reset(): void;
     submit(): void;
+}
+
+declare class HTMLIFrameElement extends HTMLElement {
+    allowFullScreen: boolean;
+    frameBorder: number;
+    height: number;
+    marginHeight: number;
+    marginWidth: number;
+    name: string;
+    scrolling: string;
+    sandbox: string;
+    src: string;
+    srcDoc: string;
+    width: number;
 }
 
 declare class HTMLImageElement extends HTMLElement {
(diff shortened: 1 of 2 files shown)


lib/dom.js
```diff
@@ -751,6 +752,20 @@ declare class HTMLFormElement extends HTMLElement {
...
 
+declare class HTMLIFrameElement extends HTMLElement {
+    allowFullScreen: boolean;
+    frameBorder: number;
+    height: number;
+    marginHeight: number;
+    marginWidth: number;
+    name: string;
+    scrolling: string;
+    sandbox: string;
```
mroch: https://html.spec.whatwg.org/multipage/embedded-content.html#the-iframe-element says this is a DOMTokenList but I don't know what that is

lib/dom.js
```diff
@@ -751,6 +752,20 @@ declare class HTMLFormElement extends HTMLElement {
     submit(): void;
 }
 
+declare class HTMLIFrameElement extends HTMLElement {
+    allowFullScreen: boolean;
+    frameBorder: number;
```
mroch: i think frameBorder, height, marginHeight, marginWidth, and width are all strings, per https://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-50708718 and also tried this in Chrome:

```
var frame = document.createElement('iframe');
frame.width // ""
frame.height // ""
// ...
```

lib/dom.js
tasti: Looks like this: https://dom.spec.whatwg.org/#interface-domtokenlist

It's basically a space-separated list of strings (tokens). Should I add `DOMTokenList` as a definition and use it instead?

mroch: yeah I think it's its own class, it has methods like `toggle` and `support` so it's definitely not just a string-compatible object. this seems pretty edge-casey so if you feel like adding it then go for it, otherwise I'd just leave a stub like

```
class DOMTokenList {
  // TODO: https://dom.spec.whatwg.org/#interface-domtokenlist
}
```

tasti: Wait a minute, `DOMTokenList` already exists: https://github.com/facebook/flow/blob/master/lib/dom.js#L400

I'll use that :)

tasti:
Failing test cases seem unrelated?

lib/dom.js
```diff
@@ -293,6 +293,7 @@ declare class Document extends Node {
     createElement(tagName: 'canvas'): HTMLCanvasElement;
     createElement(tagName: 'div'): HTMLDivElement;
     createElement(tagName: 'form'): HTMLFormElement;
+    createElement(tagName: 'iframe'): HTMLIFrameElement;
```
mroch: no, the test failures are related because this line pushed down all of the line numbers in the expected error messages. `cp tests/dom/dom.{out,exp}` to update the expected output.

tasti: Ahh I see, will update.

mroch:
@facebook-github-bot shipit

PR #1467, opened 2016-02-27
[docs] type annotations to use current flow cli output

nsfmc:
there is absolutely nothing special here, just updated the docs
with the output from the current flow cli since i'm going through
reading them as our team starts using it. thanks for everything!


samwgoldman:
Looks good to me! Thanks, @nsfmc!

mroch:
@facebook-github-bot shipit

PR #1476, opened 2016-02-29
Allow more trailing commas in type parameter lists.

majjoha:
With 39c156b9aacbd9593e6cc3d1ad11df12f188ad8f, we are now allowed to use trailing commas in type parameter lists when declaring our own types, e.g. `type foo<A,B,> = bar`. However, the example below still causes an `Unexpected token` error to be thrown.

```js
// @flow
class Baz { }
class Quux { }
class Bar<K,V> { }

class Foo extends Bar<Baz,Quux,> {
}
```

This pull request essentially just adds the same conditional introduced in 39c156b9aacbd9593e6cc3d1ad11df12f188ad8f to the `type_parameter_instantiation` function which resolves the issue. Please, note that I have no prior experience with the parser code, so in case I am misunderstanding something, just let me know, and I'll do my best to fix it.

### DIFF
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -705,7 +705,9 @@ end = struct
         | T_GREATER_THAN -> List.rev acc
         | _ ->
           Expect.token env T_COMMA;
-          params env acc
+          if Peek.token env = T_GREATER_THAN
+          then List.rev acc
+          else params env acc
 
       in fun env ->
           let start_loc = Peek.loc env in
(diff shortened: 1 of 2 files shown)


jeffmo:
Doh! I assumed these two things used the same util function... I should've tested both.

Thanks! And in the spirit of not making that mistake again: Would you mind adding some tests for classes (class-level and extends-level), interfaces, and functions?

`class Foo<A,B,> extends Bar<C,D,> {}`
`interface Foo<A,B,> {}`
`function f<A,B,>() {}`

(in case it's not already obvious: Parser tests run separately from the main flow tests and use the makefile under src/parser/ -- `cd src/parser; make test`)

jeffmo:
Thanks!

majjoha:
No worries!

PR #1479, opened 2016-03-02
Add declarations for ReactDOM.unstable_*

pvolok:


CentaurWarchief:
I don't think that those methods are relevant here. It isn't public API and they aren't supposed to be used by anyone. Can you explain why declaring it @pvolok?

pvolok:
@CentaurWarchief Sure. If I understand it correctly these functions are experimental but still public. There are real use cases like this: https://github.com/reactjs/react-modal/blob/master/lib/components/Modal.js#L70

Some people use these functions and I don't see why flow should complain about it.

CentaurWarchief:
Ah, I see. What I meant is that you shouldn't be using it as it is denoted `unstable_`, I didn't thought that someone was using it anyway.

But you're right. Makes sense. Sorry!

:beers:

lib/react.js
```diff
@@ -239,6 +239,22 @@ declare module 'react-dom' {
   ): React$Component<$DefaultPropsOf<Config>, $PropsOf<Config>, any>;
 
   declare function unmountComponentAtNode(container: any): boolean;
+
+  declare function unstable_batchedUpdates(
+    callback: Function,
+    a?: any,
```
gabelevi: I was skeptical that the args were named like this. [I was wrong](https://github.com/facebook/react/blob/62c9d999bef942383215f8dcaaf65fd06160eb60/src/renderers/shared/reconciler/ReactUpdates.js#L110) :stuck_out_tongue:

gabelevi:
@facebook-github-bot import

gabelevi:
Thanks for the pull request! I'll get this reviewed and checked in ASAP!

lib/react.js
pvolok: Now I understand that I should've linked this to make the review easier :)

PR #1481, opened 2016-03-02
DOM scrollIntoView(): Allow passing of object

danez:
The new specs for cssom allow passing an option-object to Element.scrollIntoView()

https://developer.mozilla.org/de/docs/Web/API/Element/scrollIntoView
https://drafts.csswg.org/cssom-view/#dom-element-scrollintoview

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -443,7 +443,7 @@ declare class Element extends Node {
     removeAttribute(name?: string): void;
     removeAttributeNS(namespaceURI: string, localName: string): void;
     removeAttributeNode(oldAttr: Attr): Attr;
-    scrollIntoView(top?: boolean): void;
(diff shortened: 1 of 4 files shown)


danez:
I now also added tests. I tried to also verify in the tests, that the argument-object only takes the defines properties, but for some reason in the test it is not failing when specifying property-keys which are not in the declaration of the object shape. I also tried with node.exec() tests and the also don't fail.


```js
declare class Element extends Node {
    scrollIntoView(arg?: (boolean | { behavior?: ('auto' | 'instant' | 'smooth'), block?: ('start' | 'end') })): void;
} 
```

```js
element.scrollIntoView({ invalid: 'some' })
```

From my point of view this should fail, as `invalid` is not specifyed

gabelevi:
@facebook-github-bot import

gabelevi:
Thanks for the PR and thanks for writing tests! This looks good to me! I'll get this checked in!

PR #1486, opened 2016-03-03
Adding Error.captureStackTrace

runn1ng:
Slight problem is that this is v8 only, so it makes sense only in Chrome or node.js or similar. Most importantly, it doesn't work in Firefox.


gabelevi:
I think this is a useful PR to merge. I'll check with the team and get this committed. I'll also rerecord the tests. Thanks for the PR and sorry for the slow response!

PR #1490, opened 2016-03-04
Documentation: add example

coot:
Add an example of a type declartion which will make flow complain when
a function is called with too many arguments.

### DIFF
--- a/website/docs/01-functions.md
+++ b/website/docs/01-functions.md
@@ -76,6 +76,14 @@ does not complain on most uses of builtin primitives like `Array`, whose
 methods are often called with closures that take fewer parameters than
 prescribed.
 
+If you want flow to complain when too many arguments are given you can use the
+following type declaration:
+{% highlight javascript linenos=table %}
+/* @flow */
+declare class NoMoreArgs {};
+declare type FunctionWithOneArgument = (arg: any, ...rest: Array<NoMoreArgs>) => any;
+{% endhighlight %}
+
 ## Function-based type annotations
 
 Since functions are first-class values in JavaScript (meaning they can be


samwgoldman:
Why not `...void[]`?

coot:
Using a dedicated type `NoMoreArgs` is more strict than using `void`, it will also complain when calling function with `undefined` arguments.

coot:
It's important when one tries to implement knockout observable type interface, where `ko.observable` returns a function which accepts 0 (then acts as a getter) arguments or 1 argument (acting as a setter).
```
declare class NoMoreArgs {}

declare interface KnockoutObservable<T> {
  (...rest: NoMoreArgs[]): T;
  (arg: T): void;
};
```

gabelevi:
Hi @coot, sorry for the delay in reviewing this.

I think I agree with @samwgoldman in this. Your `NoMoreArgs[]` example is a neat trick and definitely stricter than `void[]`, since it's much harder to construct a `NoMoreArgs` value than it is to construct a `void` value.

However, it is a bit of a trick, and it does require you declaring `NoMoreArgs` somewhere. Furthermore, the `void[]` paradigm feels "good enough". Expressions with the type `void` are somewhat rare...usually just the expression `undefined` or a value that has been refined with `=== undefined`. I'm having a hard time imagining someone accidentally calling a function with too many arguments that each have the type `void`.

However I think you are right that the documentation should show how to write a function that doesn't allow too many arguments. With your permission, I'd like to take this PR and update it to use `void[]` instead. Would that be ok with you?

And thanks for the contribution! We really do appreciate it!

coot:
Sure!

PR #1500, opened 2016-03-07
Add `getSelection` method to Document class and add Selection class

### LINKED ISSUE #1487: Document class is missing `getSelection` method
`Document.getSelection` in [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Document/getSelection)

vjefri:
Fixes #1487
Used the following resources to complete the task:
http://w3c.github.io/selection-api/#idl-def-Selection
https://developer.mozilla.org/en-US/docs/Web/API/Selection
https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection

Used the Range method as reference.
https://dom.spec.whatwg.org/#range

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -327,6 +327,7 @@ declare class Document extends Node {
     xmlVersion: string;
 
     registerElement(type: string, options?: ElementRegistrationOptions): any;
+    getSelection(): Selection | null;
 
     // 6.4.6 Focus management APIs
     activeElement: HTMLElement;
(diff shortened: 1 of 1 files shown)


vjefri:
Read over travis tests but don't know how to proceed. Please advise. Thank you.

gabelevi:
Thanks for the contribution and sorry for the delay in review! The tests probably just need to be re-recorded. I can do this as I merged the diff. 

I'll also make a few tweaks

1. Looks like this should also be available as `window.getSelection()`
2. I think a few of `Selection`'s methods have optional params. I'll update them.

vjefri:
Thanks for the review Gabe.

gabelevi:
My pleasure! Thanks for contributing!

PR #1510, opened 2016-03-08
add interface def for unstable_batchedUpdates in 'react-dom'

graue:
Tells Flow what this unstable React DOM function means so it won't complain.


gabelevi:
This is somewhat a dupe of #1479, but I like your use of generics, so I'll merge the two PRs together.

Thanks for the PR and sorry for the delay!

gabelevi:
Actually I think I'll take this PR and then take the other. That credits the authors better :)

PR #1516, opened 2016-03-09
Added support for streams.PassThrough

balupton:
Been around since node 0.10:
https://nodejs.org/docs/latest-v0.10.x/api/stream.html#stream_class_stream_passthrough

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -941,7 +941,6 @@ declare module "readline" {
 
 declare class stream$Stream extends events$EventEmitter {}
 
-type chunk = (Buffer | string);
 type readableStreamOptions = { highWaterMark? : number, encoding? : ?string, objectMode? : boolean };
 declare class stream$Readable extends stream$Stream {
   constructor(options?: readableStreamOptions): void;
@@ -953,8 +952,8 @@ declare class stream$Readable extends stream$Stream {
   read(size?: number): ?(string | Buffer);
   resume(): stream$Readable;
   unpipe(dest?: (stream$Writable | stream$Duplex)): void;
-  unshift(chunk: chunk): void;
-  push(chunk: ?chunk, encoding? : string): boolean;
+  unshift(chunk: (Buffer | string)): void;
+  push(chunk: ?(Buffer | string), encoding? : string): boolean;
   wrap(oldReadable: any): stream$Readable;
 }
 
@@ -963,21 +962,21 @@ declare class stream$Writable extends stream$Stream {
   constructor(options?: writableStreamOptions): void;
   cork(): void;
   end(
-    chunkOrEncodingOrCallback?: chunk | string | Function,
+    chunkOrEncodingOrCallback?: Buffer | string | Function,
     encodingOrCallback?: string | Function,
     callback?: Function
   ): void;
   setDefaultEncoding(encoding: string): boolean;
   uncork() : void;
   write(
-    chunk: chunk,
+    chunk: (Buffer | string),
     encodingOrCallback?: string | Function,
     callback?: Function
   ): boolean;
   _write(
-    chunk: chunk,
-    encodingOrCallback?: string | Function,
-    callback?: Function
+    chunk: (Buffer | string),
+    encoding: string,
+    callback: (error: ?Error, data?: chunk) => void
   ): boolean;
 }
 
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -1006,6 +1006,7 @@ declare module "stream" {
   declare var Writable : typeof stream$Writable
   declare var Duplex : typeof stream$Duplex
   declare var Transform : typeof stream$Duplex
+  declare var PassThrough : typeof stream$Duplex
```
mroch: it looks like this extends `stream.Transform`, so I think you'd want something like

```
declare class stream$Transform extends stream$Duplex {}
declare class stream$PassThrough extends stream$Transform {}
declare module "stream" {
  ...
  declare var Transform : typeof stream$Transform
  declare var PassThrough : typeof stream$PassThrough
}
```

saying that it's a Duplex seems accurate but not totally precise.

balupton:
Okay thanks will update later today 😊

balupton:
How's that?

balupton:
Okay, tested it all out locally, and working great with

- https://github.com/bevry/caterpillar/blob/9247e80f2cfe835423e2f77d4d7a5582ac5c0d23/source/transform.js
- https://github.com/bevry/caterpillar/blob/9247e80f2cfe835423e2f77d4d7a5582ac5c0d23/source/logger.js

lib/node.js
```diff
@@ -988,24 +993,41 @@ declare class stream$Duplex extends stream$Readable {
   constructor(options?: duplexStreamOptions): void;
   cork(): void;
   end(
-    chunk?: string | Buffer,
-    encodingOrCallback?: string | (data: any) => void,
-    callback?: (data: any) => void
+    chunk?: Buffer | string,
+    encodingOrCallback?: string | Function,
+    callback?: Function
```
balupton: The callback change here is because the callback is just a listener for the `finish` event, which receives nothing:

1. https://nodejs.org/dist/latest-v5.x/docs/api/stream.html#stream_writable_end_chunk_encoding_callback
2. https://nodejs.org/dist/latest-v5.x/docs/api/stream.html#stream_event_finish

lib/node.js
```diff
@@ -988,24 +993,41 @@ declare class stream$Duplex extends stream$Readable {
...
   setDefaultEncoding(encoding: string): boolean;
   uncork() : void;
   write(
-    chunk?: string | Buffer,
-    encodingOrCallback?: string | (data: any) => void,
-    callback?: (data: any) => void
+    chunk: Buffer | string,
+    encodingOrCallback?: string | Function,
+    callback?: Function
+  ): boolean;
```
balupton: The callback change here is actually a fix, because the callback is just a listener for the `flush` event, which has no documentation about any arguments:

1. https://nodejs.org/dist/latest-v5.x/docs/api/stream.html#stream_writable_write_chunk_encoding_callback

lib/node.js
```diff
@@ -962,17 +962,22 @@ declare class stream$Writable extends stream$Stream {
...
-    chunk: string | Buffer,
+    chunk: Buffer | string,
     encodingOrCallback?: string | Function,
     callback?: Function
   ): boolean;
+  _write(
+    chunk: Buffer | string,
+    encoding: string,
+    callback: (error: ?Error, data?: Buffer | string) => void
+  ): boolean;
```
balupton: This is part of Writable: https://nodejs.org/dist/latest-v5.x/docs/api/stream.html#stream_writable_write_chunk_encoding_callback_1

lib/node.js
```diff
@@ -962,17 +962,22 @@ declare class stream$Writable extends stream$Stream {
...
-    chunkOrEncodingOrCallback?: string | Buffer | Function,
+    chunkOrEncodingOrCallback?: Buffer | string | Function,
     encodingOrCallback?: string | Function,
     callback?: Function
   ): void;
   setDefaultEncoding(encoding: string): boolean;
   uncork() : void;
   write(
-    chunk: string | Buffer,
+    chunk: Buffer | string,
```
balupton: Consistency

lib/node.js
```diff
@@ -962,17 +962,22 @@ declare class stream$Writable extends stream$Stream {
   constructor(options?: writableStreamOptions): void;
   cork(): void;
   end(
-    chunkOrEncodingOrCallback?: string | Buffer | Function,
+    chunkOrEncodingOrCallback?: Buffer | string | Function,
```
balupton: Consistency

balupton:
Hate to push, however a few of my modules depend on this fix and I'm not sure how to shim this in the meantime... If there is anything I can do to speed this up, let me know! :+1:

gabelevi:
Sorry for the radio silence! This looks good to me! I'll get this reviewed and merged ASAP!

And yeah, originally we often typed callbacks with a `void` return type, like `callback: (arg: number) => void`. We intended to indicate the return type was ignored, in case people thought something like "if I return false then something will happen".

In reality, people often like to use arrow functions as callbacks, and often write arrow functions without curly braces. This meant their arrow functions were often returning non-void types. So we changed our minds and have started typing callback return types as `mixed` instead.

balupton:
Woohoo. Thank you!

PR #1530, opened 2016-03-12
Update flow.css to make "Edit on GitHub" links clickeable

montogeek:
Make "Edit on GitHub" links clickeable


vjeux:
Thanks for fixing what I broke! Unfortunately, I don't think that it's such a good idea to do inline-block on the headers, it happens to work because the next element is usually a block, but if it's not the case then the header would flow and not look well.

Here's a better fix: https://github.com/facebook/flow/pull/1539

PR #1446, opened 2016-02-24
minor doc fix

fabiomcosta:
valu -> value

PR #1532, opened 2016-03-12
Fix typo in react.doc.js

thangngoc89:


samwgoldman:
This came up before in #1446, but the typo is intentional to show a type error. We should show the errors inline, but it looks like that is broken right now, so it's totally reasonable to think that's an error on our part (it is, just a different kind of error). I'll look into why that's broken, but we should probably also add a comment on that line indicating the typo is expected.

thangngoc89:
Thanks. There should be some inline comments to prevent people (like myself) sending PR like this.

samwgoldman:
Yup! Sorry about the mix up. Totally our fault. Thanks for pitching in to fix. :)

PR #1536, opened 2016-03-14
Do not throw in normalize_type_impl for AbstractT.

pvolok:
Autocompletion for react components doesn't work with current master. It happens because `state` field has abstract type.

More concise example where autocompletion doesn't work:

```javascript
class A {
  state: $Abstract<{}>;
}

class B extends A {
  render() {
    this.<cursor here>
  }
}
```


mroch:
@facebook-github-bot shipit

mroch:
thanks :)

PR #1539, opened 2016-03-14
Fix edit on github link not clickable

vjeux:


mroch:
@facebook-github-bot shipit

PR #1542, opened 2016-03-16
Fixing command in getting-started.doc.js

janoskk:
The `npm run-script flow` command was not correct on some systems where the default shell is not bash (e.g. dash on Debian Jessie). The problem was the "[[ ]]" syntax that was replaced with a "test" expression in this fix.


samwgoldman:
cc @jeffmo

mroch:
@facebook-github-bot shipit

PR #1551, opened 2016-03-19
Add audioContext libdefs to core

marudor:
Fairly simple This includes Everything the AudioContext .
It does include some Editor Draft stuff.


jeffmo:
These are all browser APIs, right? If so, they probably belong over in `bom.js` instead

marudor:
Yes, you are right. They are Browser only.

samwgoldman:
This is really awesome work. I know I don't speak for everyone on this, but I'm pretty against pulling this in since it's not a standardized API. Ultimately I think this stuff belongs in flow-typed in an opt-in manner. I don't think it should even ship with flow-typed BOM, but should be a separate opt-in dependency for the Web Audio API.

marudor:
I think you are correct. It should be in flow-typed.
It might be a good Idea to get some guidelines going what the requirements are for flow included libs.
For me it looks like flow wants Standards only. Already has to be a Living Standard thing - no Editor Drafts or Working Drafts.
Yet flow understands async/await, Object.values, Object.entries (stage 3 on ES Next, so similar to a Working Draft?)
But also Object rest/spread (Stage 2)
And has basic support for class properties - although those are opt-in (Stage 1)

Hard to see where exactly you draw the line.

samwgoldman:
Since we took `fetch`, seems inconsistent to not take this too. Would like to see both move to flow-typed eventually. cc @gabelevi

gabelevi:
Wouldn't want to be inconsistent! Plus this is awesome work. I'll get this merged.

PR #1553, opened 2016-03-19
Clarify available array notations

### LINKED ISSUE #1537: [Docs] Clarification on Array Syntax needed
In the docs two versions to specify an array are used: 

- [`T[]`](http://flowtype.org/docs/syntax.html#function-declarations)
- [`Array<T>`](http://flowtype.org/docs/arrays.html#type-annotating-arrays). 

IMHO this could lead to confusion and only one of the two ways should be used. (Why are there even two ways of doing this?)

clentfort:
Mention the availability of `T[]` to annotated arrays in the docs.

Fixes #1537.

### DIFF
--- a/website/docs/01-arrays.md
+++ b/website/docs/01-arrays.md
@@ -7,8 +7,8 @@ next: classes.html
 ---
 
 Array types are simply instantiations of a special polymorphic Array class:
-the type `Array<T>` describes arrays whose elements are of type `T`. `T[]` can
-be used as a shorthand to annotated arrays.
+the type `Array<T>` describes arrays whose elements are of type `T`.  The
+shorthand syntax `T[]` is equivalent to `Array<T>`.
 
 ## Type Annotating Arrays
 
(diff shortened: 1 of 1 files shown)


website/docs/01-arrays.md
```diff
@@ -7,7 +7,8 @@ next: classes.html
 ---
 
 Array types are simply instantiations of a special polymorphic Array class:
-the type `Array<T>` describes arrays whose elements are of type `T`.
+the type `Array<T>` describes arrays whose elements are of type `T`. `T[]` can
+be used as a shorthand to annotated arrays.
```
samwgoldman: I like the idea of mentioning this syntax here, but I think it could be worded a bit more clearly.

How about "The shorthand syntax `T[]` is equivalent to `Array<T>`."?

gabelevi:
Looks good to me! Thanks for improving our docs! We really appreciate it!

@facebook-github-bot shipit

PR #1555, opened 2016-03-21
Added persist method in SyntheticEvent

bloodyowl:
ref http://facebook.github.io/react/docs/events.html#event-pooling


samwgoldman:
Nice. When I learned about this method, I had already lost a few hours trying to figure out what the hell was going wrong in my React app when I held a reference to a fired event. Thanks for adding this!

Test failure is unrelated flakiness.

@facebook-github-bot shipit

PR #1559, opened 2016-03-21
allow string in React.createElement

marudor:
Very simple change. It does allow something like this though:
```javascript
render() {
    const Cmp = true ? 'div' : Foo; // Foo is a React Component
    return (<Cmp/>);
  }
```
As JSX is basically React.createElement calls flow should allow strings - just like the documentation says.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -171,7 +171,7 @@ declare module react {
    */
   // TODO: React DOM elements
   declare function createElement<Config>(
-    name: ReactClass<Config>,
+    name: ReactClass<Config>|string,
     config: Config,
     children?: any
   ): React$Element<Config>;
--- a/tests/react/createElementRequiredProp_string.js
+++ b/tests/react/createElementRequiredProp_string.js
@@ -1,0 +1,22 @@
+// @flow
+import React from 'react';
+
+class Bar extends React.Component {
+  props: {
+    test: number,
+  };
+  render() {
+    return (
+      <div>
+        {this.props.test}
+      </div>
+    )
+  }
+}
+
+class Foo extends React.Component {
+  render() {
+    const Cmp = Math.random() < 0.5 ? 'div' : Bar;
+    return (<Cmp/>);
+  }
+}
--- a/tests/react/createElement_string.js
+++ b/tests/react/createElement_string.js
@@ -1,0 +1,11 @@
+// @flow
+import React from 'react';
+
+class Bar extends React.Component {}
+
+class Foo extends React.Component {
+  render() {
+    const Cmp = Math.random() < 0.5 ? 'div' : Bar;
+    return (<Cmp/>);
+  }
+}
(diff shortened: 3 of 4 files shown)


jeffmo:
cc @samwgoldman / @avikchaudhuri -- can you confirm that this shouldn't mess up any of the other React sorcery?

(travis results still pending, but I'm assuming they should show if any *obvious* problems arise)

samwgoldman:
Yeah, unfortunately this does cause issues, but unrelated to any of the recent changes. I ran into similar issues when attempting to add support for stateless functional component. In short, the union there causes errors to be deferred by the speculative match process, and sometimes the errors are deferred indefinitely :(

marudor:
@samwgoldman Oh? I've changed it and uses that on our work project. Worked fine. (~400 React Components and several Services handling React Components)
Could you provide an example where the union messes stuff up?

samwgoldman:
I'll confirm in a bit, but it's not the kind of issue that would cause errors in a previously correct project. A simple way to repro is a simple JSX expression with an expected error. For example, `<Foo />` for a component class `Foo` that has a required parameter.

marudor:
Could this be fixed with the latest changes to react stuff? (React.Component<DefaultProps, Props, State> reworked to just class properties)
I've created a small test with 2 files. One Component with required Prop and one Prop that has a Cmp that is either a string or that Comp with required Prop.
```javascript
// @flow
import React from 'react';

export default class Bar extends React.Component {
  props: {
    test: number,
  };
  render() {
    return (
      <div>
        {this.props.test}
      </div>
    )
  }
}
```
```javascript
// @flow 
import React from 'react';
import Bar from './Bar';

export default class Foo extends React.Component {
  render() {
    const Cmp = Math.random() < 0.5 ? 'div' : Bar;
    return (<Cmp/>);
  }
}
```

Flow correctly throws an Error as Cmp does not have the required Prop. I've though that is how it is supposed to be. As Cmp might be a Component that requires the prop flow wants that prop. No way for flow to know if it needs it or not.

samwgoldman:
OK, so this change is fine. The stateless functional component change is still problematic, but this doesn't hit the same problem—it does if you put `string` first in the union.

Can you add a test? The snippet from the PR description seems like a good one.

marudor:
I've added a test with and one without a required paramter.
As you said this might be a problem it's probably safer to have both as tests.

gabelevi:
Thanks for the pull request! I'll get this imported, reviewed and checked in!

gabelevi:
I made a small change to this before merging. As @jeffmo suggested, by using `$Keys<$JSXIntrinsics>` instead of just `string`, Flow is able to complain if you set `$JSXIntrinsics` (like [this test](https://github.com/facebook/flow/blob/0a28d833ada5b803d7c533c52efecd06592ef28f/tests/jsx_intrinsics.custom/lib/jsx.js) does) and use a jsx intrinsic that hasn't been defined. By If `$JSXIntrinsics` isn't explicitly set, then this will basically work the same way.

PR #1564, opened 2016-03-22
Using mixed as input to parseInt/parseFloat

runn1ng:
I needed something like this recently, when I got input that can be number or string. I am using `parseInt` to convert it to integer, but Flow complains, because it wants string as an input.

The MDN documentation says:

> string: The value to parse. If string is not a string, then it is converted to a string (using the ToString abstract operation). Leading whitespace in the string is ignored.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -14,8 +14,8 @@
 declare var NaN: number;
 declare var Infinity: number;
 
-declare function parseInt(string: string, radix?: number): number;
-declare function parseFloat(string: string): number;
+declare function parseInt(string: mixed, radix?: number): number;
+declare function parseFloat(string: mixed): number;
 declare function isNaN(number: mixed): boolean;
 declare function isFinite(number: mixed): boolean;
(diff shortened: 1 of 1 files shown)


gabelevi:
I feel a little mixed ( :laughing: ) about this. On one hand I hate implicit conversions and wish people would explicitly convert. On the other hand, it's important for Flow to model JavaScript and not just impose the will of @gabelevi.

So yeah, I'll get this merged. Thanks for the contribution!

runn1ng:
Hm. I was using `parseInt` *as* a conversion.

I am not sure what I am getting from JSON; so I put it into parseInt, and I am 100% sure number comes out of it.

rauschma:
String is the correct type for the first parameter. With numbers, you get unexpected results:

```text
> parseInt(0.0000008, 10)
8
```

Why? `parseInt()` coerces its parameter to string:

```text
> String(0.0000008)
'8e-7'
```

PR #1565, opened 2016-03-22
include Element.matches

marudor:
Include the matches method on the Element Class.
It was implemented as matchesSelector and prefixed versions initially but since more recent Versions it got implemented as specified matches.


gabelevi:
Looks good to me! Thanks for adding this! I'll get it merged!

PR #1575, opened 2016-03-23
Fix false negative arising from generics and addition

### LINKED ISSUE #1574: The (+) operator incorrectly accepts generic types
In the following code:

```js
function identity<A>(x: A): A {
  return x + x;
}

// $ flow
// No errors!
```

I expected Flow to give me a type error in `x + x`, because the abstract type `A` is incompatible with the types expected by the `+` operator. However, Flow incorrectly says that there are no type errors in this code.

Other operators work as expected:

```js
function identity<A>(x: A): A {
  return x - x;
}

// $ flow
// id.js:4
//   4:   return x - x;
//               ^ A. This type is incompatible with
//   4:   return x - x;
//               ^^^^^ number
// 
// id.js:4
//   4:   return x - x;
//               ^^^^^ number. This type is incompatible with
//   3: function identity<A>(x: A): A {
//      ^ some incompatible instantiation of `A`
// 
// id.js:4
//   4:   return x - x;
//                   ^ A. This type is incompatible with
//   4:   return x - x;
//               ^^^^^ number
// 
// 
// Found 3 errors
```

samwgoldman:
Summary:
This fixes a bad interaction between polymorphic instantiation and the
addition rules. Given generic types A and B and values a:A and b:B, the
expression a+b should always evaluate to `mixed` because we don't know a
or b will be numbers or strings, and thus whether the resulting
expression will be a number or string.

This expectation follows from the existing behavior, but the rule for
ground_subtype that prunes flows with an empty lower bound prevented the
entire addition expression from becoming mixed.

This change allows (EmptyT _, AdderT _) to reach flow_addition and
handles the possible EmptyT internally. This ensures that the addition
expression always gets a lower bound (mixed) which eventually runs into
the empty instantiation of the return annotation.

Fixes #1574 

Test Plan:
added tests

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3450,6 +3450,9 @@ and flow_addition cx trace reason l r u =
      (NumT _ | SingletonNumT _ | BoolT _ | SingletonBoolT _ | NullT _ | VoidT _)) ->
     rec_flow_t cx trace (NumT.why reason, u)
 
+  | (EmptyT _, _)
+  | (_, EmptyT _) -> ()
+
   | (_, _) ->
     let fake_str = StrT.why reason in
     rec_flow_t cx trace (l, fake_str);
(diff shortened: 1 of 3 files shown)


samwgoldman:
Hrm, actually this doesn't fix the case in #1574 with a single type parameter: `function f<A>(a:A):A {return a+a}`

mroch:
resurrect this if you want

PR #1579, opened 2016-03-24
Update 02-advanced-configuration.md

rjrodd:
I found that the suggested `suppress_comment` regex didn't work for me until I inserted another backslash before the n.

### DIFF
--- a/website/docs/02-advanced-configuration.md
+++ b/website/docs/02-advanced-configuration.md
@@ -189,7 +189,7 @@ can be overridden with command line flags.
 - `suppress_comment` (regex): defines a magical comment that suppresses any Flow errors on the following line. For example:
 
     ```
-    suppress_comment= \\(.\\|\n\\)*\\$FlowFixMe
+    suppress_comment= \\(.\\|\\n\\)*\\$FlowFixMe
     ```
 
     will match a comment like this:


jeffmo:
Hmm, we have a repo that uses `\\(.\\|\n\\)*\\$FlowFixMe` (no extra backslash) without issue (and I can't seem to repro this there).

I won't have time to dig in to this until later, but just wanted to comment so we can understand what's different between your system and ours before we take this

gabelevi:
TL;DR - I'm not sure why the `suppress_comment` is not working for you, but I don't think this change should fix it.

The value of `suppress_comment` is an OCaml string that would be used as an OCaml `Str.regexp`. For example, if you write `suppress_comment=\\[0-9\\]*`, this is interpreted as

```OCaml
let suppressComment = Str.regexp "\\[0-9\\]*"
```

So there are two layers of backslash escaping at work here:

1. Escaping for the OCaml string
2. Escaping for the OCaml regex (documented [here](http://caml.inria.fr/pub/docs/manual-ocaml/libref/Str.html))

So the value of the string `"\\[0-9\\]*"` is `\[0-9\]*`and that's used by the regex.

For this PR, the old suppress_comment was `\\(.\\|\n\\)*\\$FlowFixMe`. This is interpreted as the string `"\\(.\\|\n\\)*\\$FlowFixMe"`, whose value after escaping is `\(.\|<LF>\)*\$FlowFixMe`. (where `<LF>` is the line feed character, generally written as `\n`)

This PR changes `\n` to `\\n`. This means the OCaml string is changed from containing a `<LF>` to containing `\n` (a backslash followed by an `n`). The OCaml regex `\n` matches the character `n`.

```OCaml
# Str.string_match (Str.regexp "\\n") "n" 0;;
- : bool = true   
```

PR #1581, opened 2016-03-25
[travis] use opam on OS X

mroch:
homebrew is actually running ocaml 4.02.3, and in the future will likely move to 4.03.

this installs opam on OS X like we do on linux. it also disables the opam 1.1 build on OS X since it's a waste of time; since no package manager on OS X ships with 1.1, there's no reason to test it. and it builds opam/ocaml in a version-specific directory so that future diffs can try to add Travis caching without the installs stepping on each other.


mroch:
@facebook-github-bot import

PR #1582, opened 2016-03-25
Re-encrypt S3 vars under facebook/flow instead of mroch/flow... oops.

mroch:


mroch:
@facebook-github-bot shipit

PR #1587, opened 2016-03-26
Add DataTransfer interface

adrianheine:
According to https://html.spec.whatwg.org/multipage/interaction.html#the-datatransfer-interface.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -53,6 +53,40 @@ declare class FileList {
     [index: number]: File;
 }
 
+/* DataTransfer */
+
+declare class DataTransfer {
+  dropEffect: string;
+  effectAllowed: string;
+
+  items: DataTransferItemList; // readonly
+
+  setDragImage(image: Element, x: number, y: number): void;
+
+  /* old interface */
+  types: Array<string>; // readonly
+  getData(format: string): string;
+  setData(format: string, data: string): void;
+  clearData(format?: string): void;
+  files: FileList; // readonly
+}
+
+declare class DataTransferItemList {
+  length: number; // readonly
+  [index: number]: DataTransferItem;
+  add(data: string, type: string): ?DataTransferItem;
+  add(data: File): ?DataTransferItem;
+  remove(index: number): void;
+  clear(): void;
+}
+
+declare class DataTransferItem {
+  kind: string; // readonly
+  type: string; // readonly
+  getAsString(_callback?: (data: string) => void): void;
+  getAsFile(): ?File;
+}
+
 /* DOM */
 
 declare class DOMError {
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
+declare class DataTransfer {
+  dropEffect: string;
+  effectAllowed: string;
+
+  items: DataTransferItemList; // readonly
+
+  setDragImage(image: Element, x: number, y: number);
+
+  /* old interface */
+  types: array<string>; // readonly
```
gabelevi: `Array<string>`

lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
+  getData(format: string): string;
+  setData(format: string, data: string);
+  clearData(format: ?string);
```
gabelevi: With optional things I've been doing `clearData(format?: string);`

lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
+  add(data: string, type: string): ?DataTransferItem;
+  add(data: File): ?DataTransferItem;
+  remove(index: number);
+  clear();
+};
+
+declare class DataTransferItem {
+  kind: string; // readonly
+  type: string; // readonly
+  getAsString(_callback: ?(data: string) => void);
```
gabelevi: I've been typing the return type of callbacks as `mixed` recently. People really like to use arrow functions as callbacks, like

```JavaScript
  item.getAsString(data => foo(data))
```

and if `foo()` doesn't return `void` then you would get a type error.

gabelevi:
Looks great! I have a few nits, but I can address them when as I merge this in. I'll get it imported, reviewed and merged ASAP. Thanks for the contribution!

lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
 
+/* DataTransfer */
+
+declare class DataTransfer {
+  dropEffect: string;
+  effectAllowed: string;
+
+  items: DataTransferItemList; // readonly
+
+  setDragImage(image: Element, x: number, y: number);
```
gabelevi: Oh, and return types are required for methods in `declare class`. So this kind of line should be

```JavaScript
setDragImage(image: Element, x: number, y: number): void;
```

or it won't parse

adrianheine:
Adressed some of the issues.

lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
+  files: FileList; // readonly
+}
+
+declare class DataTransferItemList {
+  length: number; // readonly
+  [index: number]: DataTransferItem;
+  add(data: string, type: string): ?DataTransferItem;
+  add(data: File): ?DataTransferItem;
+  remove(index: number): void;
+  clear();
```
mroch: here's another that's missing a return type. you can run the tests to double check, `./runtests.sh bin/flow dom`

lib/dom.js
```diff
@@ -53,6 +53,40 @@ declare class FileList {
...
+};
```
mroch: can you remove these `;`s? thanks!

lib/dom.js
adrianheine: They fail for me for unrelated reasons.

Issue #1224, opened 2015-12-29
Add IndexedDB lib definitions


mroch:
should go in `lib/indexeddb.js`. see https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API. typescript's starts here: https://github.com/Microsoft/TypeScript/blob/72bb1d3decb1b8a8daeab3523275b2d89065410b/lib/lib.d.ts#L10926

PR #1591, opened 2016-03-28
first version for indexedDB

marudor:
closes https://github.com/facebook/flow/issues/1224
Please review.

### DIFF
--- a/lib/indexeddb.js
+++ b/lib/indexeddb.js
@@ -1,0 +1,118 @@
+// Implemented by window & worker
+declare interface IDBEnvironment {
+  indexedDB: IDBFactory;
+}
+
+type IDBDirection = 'next' | 'nextunique' | 'prev' | 'prevunique';
+
+// Implemented by window.indexedDB & worker.indexedDB
+declare interface IDBFactory {
+  open(name: string, version?: number): IDBOpenDBRequest;
+  deleteDatabase(name: string): IDBOpenDBRequest;
+  cmp(a: any, b: any): -1|0|1;
+}
+
+declare interface IDBRequest extends EventTarget {
+  result: IDBObjectStore;
+  error: Error;
+  source: ?(IDBIndex | IDBObjectStore | IDBCursor);
+  transaction: IDBTransaction;
+  readyState: 'pending'|'done';
+  onerror: (err: any) => void;
+  onsuccess: (e: any) => void;
+}
+
+declare interface IDBOpenDBRequest extends IDBRequest {
+  onblocked: (e: any) => void;
+  onupgradeneeded: (e: any) => void;
+}
+
+declare interface IDBDatabase extends EventTarget {
+  close(): void;
+  createObjectStore(name: string, options?: {
+    keyPath?: string|string[],
+    autoIncrement?: bool
+  }): IDBObjectStore;
+  deleteObjectStore(name: string): void;
+  transaction(storeNames: string|string[], mode?: 'readonly'|'readwrite'|'readwriteflush'): IDBTransaction;
+  name: string;
+  version: number;
+  objectStoreNames: string[];
+  onabort: (e: any) => void;
+  onerror: (e: any) => void;
+  onversionchange: (e: any) => void;
+}
+
+declare interface IDBTransaction extends EventTarget {
+  abort(): void;
+  objectStore(name: string): IDBObjectStore;
+  db: IDBDatabase;
+  mode: 'readonly'|'readwrite'|'versionchange';
+  objectStoreNames: string[];
+  error: Error;
+  onabort: (e: any) => void;
+  onerror: (e: any) => void;
+  oncomplete: (e: any) => void;
+}
+
+declare interface IDBObjectStore {
+  add(value: any, key?: any): IDBRequest;
+  clear(): IDBRequest;
+  delete(key: any): IDBRequest;
+  get(key: any): IDBRequest;
+  getAll(query?: any|IDBKeyRange, count?: number): IDBRequest;
+  getAllKeys(query?: IDBKeyRange, count?: number): IDBRequest;
+  createIndex(indexName: string, keyPath: ?(string|string[]), optionalParameter?: {
+    unique: bool,
+    multiEntry: bool,
+  }): IDBIndex;
+  deleteIndex(indexName: string): void;
+  index(indexName: string): IDBIndex;
+  put(value: any, key?: any): IDBRequest;
+  openCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
+  openKeyCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
+  count(keyRange?: any|IDBKeyRange): IDBRequest;
+  indexNames: string[];
+  keyPath: any;
+  transaction : IDBTransaction;
+  autoIncrement: bool;
+}
+
+declare interface IDBIndex extends EventTarget {
+  count(key?: any|IDBKeyRange): IDBRequest;
+  get(key?: any|IDBKeyRange): IDBRequest;
+  getKey(key?: any|IDBKeyRange): IDBRequest;
+  openCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
+  openKeyCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
+  name: string;
+  objectStore: IDBObjectStore;
+  keyPath: any;
+  multiEntry: bool;
+  unique: bool;
+}
+
+declare interface IDBKeyRange {
+  bound(lower: any, upper: any, lowerOpen?: bool, upperOpen?: bool): IDBKeyRange;
+  only(value: any): IDBKeyRange;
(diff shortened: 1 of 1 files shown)


lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+type IDBDirection = 'next' | 'nextunique' | 'prev' | 'prevunique';
+
+// Implemented by window.indexedDB & worker.indexedDB
+declare interface IDBFactory {
+  open(name: string, version?: number): IDBOpenDBRequest;
+  deleteDatabase(name: string): IDBOpenDBRequest;
+  cmp(a: any, b: any): -1|0|1;
+}
+
+declare interface IDBRequest extends EventTarget {
```
gabelevi: It's too bad we don't support unions for interfaces. It would be cool to write

```JavaScript
declare interface IDBRequest extends EventTarget {
  readState: 'pending',
  error: Error,
  result: IDBObjectStore;
  ...
} | {
  readyState: 'done';
  error: Error;
  result: void;
} | {
  readyState: 'done',
  error: null,
  result: IDBObjectStore;
  ...
}

As written, this interface doesn't reflect the fact that `result` may be null and `error` may be undefined. Not a huge deal, but maybe worth representing in the interface.

File: lib/indexeddb.js [hunk 2]
```diff
@@ -0,0 +1,118 @@
...
+  result: IDBObjectStore;
+  error: Error;
+  source: ?(IDBIndex | IDBObjectStore | IDBCursor);
+  transaction: IDBTransaction;
+  readyState: 'pending'|'done';
+  onerror: (err: any) => void;
```
gabelevi: Nowadays we're mostly writing our callbacks as `(err: any) => mixed`. People really like their arrow functions, and the `void` return type was getting in the way.

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+
+declare interface IDBOpenDBRequest extends IDBRequest {
+  onblocked: (e: any) => void;
+  onupgradeneeded: (e: any) => void;
+}
+
+declare interface IDBDatabase extends EventTarget {
+  close(): void;
+  createObjectStore(name: string, options?: {
+    keyPath?: string|string[],
```
gabelevi: I think `keyPath` is also nullable. So maybe `null | string | string[]`?

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+    autoIncrement?: bool
+  }): IDBObjectStore;
+  deleteObjectStore(name: string): void;
+  transaction(storeNames: string|string[], mode?: 'readonly'|'readwrite'|'readwriteflush'): IDBTransaction;
```
gabelevi: Hmm...I think [the spec says](https://www.w3.org/TR/IndexedDB/#idl-def-IDBTransactionMode) `'versionchange'` rather than `'readwriteflush'`

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  onerror: (e: any) => void;
+  onversionchange: (e: any) => void;
+}
+
+declare interface IDBTransaction extends EventTarget {
+  abort(): void;
+  objectStore(name: string): IDBObjectStore;
+  db: IDBDatabase;
+  mode: 'readonly'|'readwrite'|'versionchange';
+  objectStoreNames: string[];
```
gabelevi: [I don't see this in the spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBTransaction)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  error: Error;
+  onabort: (e: any) => void;
+  onerror: (e: any) => void;
+  oncomplete: (e: any) => void;
+}
+
+declare interface IDBObjectStore {
```
gabelevi: I think this is missing `name` and `indexNames` ([link to spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBObjectStore))

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  add(value: any, key?: any): IDBRequest;
+  clear(): IDBRequest;
+  delete(key: any): IDBRequest;
+  get(key: any): IDBRequest;
+  getAll(query?: any|IDBKeyRange, count?: number): IDBRequest;
```
gabelevi: [I don't see this in the spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBObjectStore)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  getAllKeys(query?: IDBKeyRange, count?: number): IDBRequest;
```
gabelevi: [I don't see this in the spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBObjectStore)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  createIndex(indexName: string, keyPath: ?(string|string[]), optionalParameter?: {
```
gabelevi: I don't think keyPath is nullable ([spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBObjectStore))

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+    unique: bool,
```
gabelevi: When [the spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBIndexParameters) gives default values, does that mean the properties are optional?

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+    multiEntry: bool,
+  }): IDBIndex;
+  deleteIndex(indexName: string): void;
+  index(indexName: string): IDBIndex;
+  put(value: any, key?: any): IDBRequest;
+  openCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
+  openKeyCursor(range?: any|IDBKeyRange, direction?: IDBDirection): IDBRequest;
```
gabelevi: [I don't see this in the spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBObjectStore)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  count(keyRange?: any|IDBKeyRange): IDBRequest;
+  indexNames: string[];
+  keyPath: any;
+  transaction : IDBTransaction;
+  autoIncrement: bool;
+}
+
+declare interface IDBIndex extends EventTarget {
+  count(key?: any|IDBKeyRange): IDBRequest;
+  get(key?: any|IDBKeyRange): IDBRequest;
```
gabelevi: I don't think `key` is optional in `get()` and `getKey()` ([spec](https://www.w3.org/TR/IndexedDB/#idl-def-IDBIndex))

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+declare interface IDBKeyRange {
+  bound(lower: any, upper: any, lowerOpen?: bool, upperOpen?: bool): IDBKeyRange;
+  only(value: any): IDBKeyRange;
+  lowerBound(bound: any, open?: bool): IDBKeyRange;
+  upperBound(bound: any, open?: bool): IDBKeyRange;
+  lower: any;
+  upper: any;
+  lowerOpen: any;
```
gabelevi: [https://www.w3.org/TR/IndexedDB/#idl-def-IDBKeyRange](boolean)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  upperOpen: any;
```
gabelevi: [boolean](https://www.w3.org/TR/IndexedDB/#idl-def-IDBKeyRange)

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+declare interface IDBCursor {
+  advance(count: number): void;
+  continue(key?: any): void;
+  delete(): IDBRequest;
+  update(newValue: any): IDBRequest;
+  source: IDBObjectStore|IDBIndex;
+  direction: IDBDatabase;
```
gabelevi: Typo - think you meant `IDBDirection`

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  key: ?any;
```
gabelevi: I think `any` should be sufficient

lib/indexeddb.js
```diff
@@ -0,0 +1,118 @@
...
+  primaryKey: ?any;
```
gabelevi: Ditto...unless you have a reason to do `?any` that I'm not seeing.

gabelevi:
Wow! Thanks for making the effort to spec out this stuff! It's very much appreciated! I've compared your PR with https://www.w3.org/TR/IndexedDB and found a few little issues, but it generally looks really good! Nice job!

marudor:
I'm going to fix these issues today. Thanks for reviewing.

marudor:
Adjusted based on your review. Thanks.
Should be fine now.

PR #1592, opened 2016-03-29
DOM: Add new type HTMLLinkElement...

tp:
...and return that when invoking document.createElement('link')

Definition based on https://developer.mozilla.org/en/docs/Web/HTML/Element/link

## make test

Before:
Passed: 6, Failed: 229, Skipped: 2

After:
Passed: 6, Failed: 229, Skipped: 2

Though I am not sure what I could have broken (if anything really broke), and why there were this many failures even on the master branch.


lib/dom.js
```diff
@@ -1114,6 +1115,16 @@ declare class HTMLLabelElement extends HTMLElement {
     control: HTMLElement | null;
 }
 
+declare class HTMLLinkElement extends HTMLElement {
+    crossorigin: 'anonymous' | 'use-credentials';
```
mroch: i think this can also be `void`. I ran `document.querySelector("link").crossorigin` on flowtype.org and got undefined in Chrome.

lib/dom.js
```diff
@@ -1114,6 +1115,16 @@ declare class HTMLLabelElement extends HTMLElement {
...
+    href: string;
+    hreflang: string;
+    media: string;
+    rel: string;
+    sizes: string;
```
mroch: i got an array for this one

lib/dom.js
```diff
@@ -1114,6 +1115,16 @@ declare class HTMLLabelElement extends HTMLElement {
...
+    type; string;
```
mroch: typo: semicolon instead of colon

lib/dom.js
mroch: actually it's a DOMTokenList

lib/dom.js
mroch: sorry, one more thing: I think the DOM API is `crossOrigin` even though the attribute name is `crossorigin`. see the DOM interface section in https://html.spec.whatwg.org/multipage/semantics.html#the-link-element

tp:
@mroch Thanks for your feedback, should've spotted the typo.. I refined the typings with your input.

Though I am not sure what is the best practice for the mentioned "void" for `crossOrigin`.

Is adding another `| void` best, or should `?('anonymous' | 'use-credentials')` or even `crossOrigin?:` be preferred?

mroch:
elsewhere we've done `crossOrigin: ?string` except for one really old `crossOrigin?: string`. I think the latter is wrong, since the property is always there... `'crossOrigin' in link` is always true.

When I run `link.crossOrigin` in Chrome, I actually get `null` (I said `undefined` before, but that's because I was running `link.crossorigin` instead of `link.crossOrigin`). So, I think `| void` would be wrong.

I'd say either do `?string` to be consistent with the others, or `?('anonymous' | 'use-credentials')`.

tp:
Alright, thanks for the input!

I decided to stick with the actual possible values instead of `string` in order to catch bogus input etc.

tp:
I wanted to get to the bottom of the failing tests from CI to get this in a merge-able state, but even after merging in master I still get a lot of errors:

Passed: 6, Failed: 231, Skipped: 2

Any ideas what might be broken on my local Mac, that I get so much more errors than CI?

tp:
Apparently my test scores improve when I start it like this:

```sh
~/Projects/github/facebook/flow
=> bash runtests.sh ./bin/flow
```

Passed: 235, Failed: 2, Skipped: 2

Now I can investigate the actually failing tests.

tp:
Ok, now I see that the errors was caused by the new line.

So I replaced the expected test output with my last run, which for the most part just updated the line numbers correctly, but there are some subtle changes to "Member XX" which I do not understand.

Is there any better way to update the tests after such a simple change?

gabelevi:
Honestly, to record the tests I do something dumb like

1. `./runtests bin/flow`
2. `find tests/ -name "*.diff" | sed "s/\(.*\)diff/mv \1out \1exp/" | sh`

I have a new test runner that handles re-recording tests, but it's not checked in yet.

Anyway, thanks for the contribution! I'll get this imported, reviewed and checked in!

PR #1599, opened 2016-03-30
Add missing deps to parser test build

mroch:


mroch:
@facebook-github-bot shipit

PR #1600, opened 2016-03-30
Fix parser bug that prevented async class properties

nmote:
Tested via `make test-ocaml` in the parser directory and `make test` at the top-level.


gabelevi:
@facebook-github-bot shipit

PR #1602, opened 2016-03-30
Fix flowconfig parsing crash

nmote:
Before we would get the cryptic error `Fatal error: exception Failure("hd")`

Test plan:
- Added unit test
- Manually checked that it would properly reject if the config contained a different version
- Manually checked that it would run properly if the config contained nothing under [version] or the correct version.


mroch:
@facebook-github-bot shipit

PR #1603, opened 2016-03-31
[travis] cache opam and ocaml dependencies

mroch:
saves about 5 min (50%) on the linux builds. caching on open source mac builds isn't supported yet, so we'll have to do our own upload to S3 if we want to cache that.


mroch:
@facebook-github-bot shipit

PR #1613, opened 2016-04-01
Fix runtests.sh on OS X

vkurchatkin:
`mktemp` doesn't work without template on OS X.


avikchaudhuri:
@vkurchatkin Thanks for the quick PR, but a diff to fix this is already in flight and should land soon.

avikchaudhuri:
See f44e0fc

vkurchatkin:
:+1:

PR #1617, opened 2016-04-02
[travis] manually cache Mac builds to S3

mroch:
Travis doesn't support caching in open-source Mac builds, so we'll do it ourselves.

this does basically what travis's [casher](https://github.com/travis-ci/casher) does; it downloads a tarball from S3, extracts it into the cache directory, runs the build, then updates the tarball with the new contents of the cache directory. if anything changed, it uploads the new tarball back to S3.

this cuts the mac build time by about 10 minutes!


mroch:
@facebook-github-bot import

Issue #51, opened 2014-11-19
Support require which returns `any` if flow lost it


mizchi:
I want to ignore `Required module not found` and load it as `any`.
Type checking to extenal libraries are not serious problem.

I tried this

```
declare require: (s: string) => any;
```

but it overrides all require and lost other info.

iammerrick:
@tcoopman Can this be "resolved" with an ignore of all files ending in css or a particular extension?

iammerrick:
@avikchaudhuri See above comment :-)

gaearon:
+1. This is currently my blocker for Flow.

tcoopman:
Mine too. Maybe some way to tell flow that a type is what it is and that it should not check it any further?

```js
var image: !string = require('image.jpg');
var css: !any = require('main.css');
```

Comparable with maybe types, but instead of `?` you add `!`. This doesn't let you do `require('./Activity.less');` though, so maybe the suggestion of @gaearon is still necessary unless you could do: `reauire('./Activity.less'): !any`

arypurnomoz:
+1

syranide:
+1

thealjey:
+1

williamfligor:
+1

gcazaciuc:
+1

nktpro:
+1

syranide:
Perhaps the proper approach is to actually have an intermediate step where autogenerated files are output instead, so that flow can parse them. I'd imagine there are a bunch of benefits to it not only being part of the final bundling process.

tcoopman:
@syranide not sure what you mean. You mean that flow should parse the css or url?

syranide:
@tcoopman Rather than webpack (etc) generating js-files for css-files/etc only at the final bundle-stage, being able to output them to disk could make sense, then flow would have access to all the js-files it needs, no workarounds necessary. This would be useful for making many other transforms compatible with flow as well.

tcoopman:
@syranide ok I see. That would be nice but it also seems that would need more support from the build tools?
Wouldn't you then also need to transform your original files or create a mapping file so flow knows where to find them?

So

`var f = require(file.jpg)`
Outputs file.jpg minimised and hashed filename + a is file that has the hashed filename.

After the translation it would be `var f = 'file.hash.jpg'` and flow could type check that.

Again seems nice but in the meantime a solution where you can tell flow that a variable is of a type without type checking it would be a good start.

syranide:
[...]

Issue #1322, opened 2016-01-27
.flowconfig option module.name_mapper can't map to a lib definition


samwgoldman:
Repro:

.flowconfig
```
[lib]
lib
[options]
module.name_mapper='1Doesnt\(Exist\)' -> '\1s'
```

A.js
```javascript
/* @flow */
var m1 = require('./1DoesntExist'); // actual error: can't find module 1DoesntExist
var a_1: string = m1.numVal; // expected error: number ~> string
```

lib/Exists.js
```javascript
declare module 'Exists' {
  declare var exports: {
    numVal: number;
  };
};
```

Expected output:
```
A.js:3:19,27: number
This type is incompatible with
A.js:3:10,15: string
```

Actual output:
```
A.js:2:22,37: ./1DoesntExist
Required module not found
```

Workaround:
While libdefs aren't working, it's still possible to map to a node module. There's an npm module called `empty` that can be useful here. See [this gist](https://gist.github.com/lambdahands/d19e0da96285b749f0ef) for a complete example.

I think the issue is that we do name mapping specifically against file paths, so the code just doesn't attempt to handle this case. I think that it should. cc @jeffmo, who originally added this feature.

jeffmo:
The issue appears to be [here](https://github.com/facebook/flow/blob/6a42d83db9263afa272a6fdf803556d47cc7b56d/src/services/inference/module_js.ml#L437), where we assume that an import can be resolved on the FS (but neglect to deal with the fact that an import might be made available in a libdef's `declare module`).

gabelevi:
It also appears that you can't map to lib files. If you take the example from the docs and add `CSSModule.js.flow` to `[libs]` then you get an error. https://gist.github.com/a879ca861671edb16143d4b20ea8526e

MoOx:
Following #1621, here is my flow config https://github.com/MoOx/js-boilerplate/blob/4797f86049d87f013b9cab60480d5c954c1f24a8/.flowconfig
And the CSSModule definition is defined in `node_modules/flow-interfaces/interfaces/CSSModule.d.js` (it's this one https://github.com/TechnologyAdvice/flow-interfaces/blob/20fc1a5439803523fd3d3cf338619e58e81fc13b/interfaces/CSSModule.d.js)

MoOx:
Probably related https://github.com/facebook/flow/pull/1801#issuecomment-219719511

kevinSuttle:
[...]

PR #1621, opened 2016-04-04
[doc] Fix CSSModule example documentation

MoOx:
Currently, the documentation is not enough, or not clear.

By following it, I (and I think [I am](http://stackoverflow.com/questions/34921185/required-module-not-found-for-module-that-exists-in-node-modules) [not](https://github.com/facebook/flow/issues/51#issuecomment-143923503) [alone](https://gist.github.com/lambdahands/d19e0da96285b749f0ef)) get a `Required module not found` error for css file references.

Example

```
web_modules/ReactFunction/index.js:4
  4: import styles from "./index.css"
                        ^^^^^^^^^^^^^ ./index.css. Required module not found
```

I just following this gist https://gist.github.com/lambdahands/d19e0da96285b749f0ef#gistcomment-1703455 to find a solution.

I am new to flow so this workaround is probably bad, but I wanted to start a discussion about it, since [no issue is opened about cssmodule](https://github.com/facebook/flow/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+cssmodule).

### DIFF
--- a/website/docs/ref/modules.doc.js
+++ b/website/docs/ref/modules.doc.js
@@ -265,6 +265,7 @@ ES_DefaultExport.default();
   ```
   [options]
   module.name_mapper='^\(.*\)\.css$' -> '<PROJECT_ROOT>/CSSModule.js.flow'
+  module.system=haste
   ```
 
   **NOTE: You do not need to manually substitue anything for


gabelevi:
Hi @MoOx, thanks for the PR! So I tried the example from the docs verbatim and it seemed to work for me with Flow v0.22.0. In theory setting `module.system=haste` shouldn't be necessary.

When I look at the issue you linked, I notice that it was closed in favor of https://github.com/facebook/flow/issues/1322, which does highlight an bug with `module.name_mapper`. Looking into it a little more, I'm seeing another bug if `CSSModule.js.flow` is in the `[libs]` section.

Could you post a gist of your `.flowconfig`? If you do then I may be able to help figure out what's going wrong. But changing the module system should be orthogonal.

gabelevi:
Let's move this discussion to https://github.com/facebook/flow/issues/1322

PR #1623, opened 2016-04-04
Fix build by copying header files

gabelevi:
Summary:
The open source build was broken for two reasons:

1. There were two new .c files that needed to be listed in the Makefile
2. Hack now is using .h files and ocamlbuild doesn't copy these into _build

I did a little bit of googling to see if it is easy to get ocamlbuild to understand .h files, but [it didn't seem that easy](http://stackoverflow.com/questions/19644451/how-to-make-ocamlbuild-compile-and-link-a-c-file-into-an-ocaml-project). So I took the easy way out and just updated the Makefile to simply copy the .h files into _build.

Test plan:
make clean
make test
make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,6 +70,8 @@ MODULES=\
 NATIVE_OBJECT_FILES=\
   hack/$(INOTIFY_STUBS)\
   hack/heap/hh_shared.o\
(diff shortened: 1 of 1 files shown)


gabelevi:
@facebook-github-bot shipit

PR #1625, opened 2016-04-05
Fix travis by runing Travis linux builds with smaller tables

gabelevi:
The shared hash table size was recently doubled, which started breaking travis tests on their Linux containers, since they don't have enough memory.

The easiest fix for now is to build the travis linux binaries with smaller tables. This also allows us to run more tests in parallel.

At some point we should probably revisit how large the default table size should be for the open source flow binaries.


gabelevi:
@facebook-github-bot shipit

Issue #1605, opened 2016-03-31
Installing Flow on a system without glibc


jbrockett:
I am trying to install flow on an alpine linux docker image.  I am able to install node without issue, but can't get flow to compile.  In looking through the make code it should work on systems without glibc (alpine is based on musl), but I am unable to get make to build flow.  I have worked through the various steps and it seems that the only part of the make that fails is the installation of inotify.  Has anyone been able to successfully build flow on Alpine or other non-glibc system?

gsmoraes:
I'm also trying to do the same, and have also tried building it from source, and I get stuck on the _inotify_ problem.

Here's the Dockfile I'm using to build the binary:
```
FROM alpine:edge

RUN echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
RUN apk --no-cache add alpine-sdk ocaml libelf ncurses inotify-tools

WORKDIR /tmp
RUN git clone --branch v0.22.1 https://github.com/facebook/flow.git --depth 1
WORKDIR /tmp/flow
RUN make
```

And here's the output of _make_:

```
mkdir -p bin
tar czf bin/flowlib.tar.gz -C lib .
echo "const char* const BuildInfo_kRevision = \"$(git rev-parse HEAD 2>/dev/null || hg identify -i)\";" > hack/utils/get_build_id.gen.c
ocamlbuild -ocamlc "ocamlopt   -ccopt -DNO_LZ4"\
	hack/third-party/inotify/inotify_stubs.o hack/heap/hh_shared.o hack/hhi/hhi_elf.o hack/utils/files.o hack/utils/get_build_id.gen.o hack/utils/get_build_id.o hack/utils/handle_stubs.o hack/utils/nproc.o hack/utils/realpath.o hack/utils/sysinfo.o hack/utils/priorities.o hack/utils/win32_support.o hack/hhi/hhi_win32res_stubs.o src/embedded/flowlib_elf.o
+ ocamlopt   -ccopt -DNO_LZ4 -c hack/third-party/inotify/inotify_stubs.c
hack/third-party/inotify/inotify_stubs.c:41:28: fatal error: inotify_compat.h: No such file or directory
compilation terminated.
Command exited with code 2.
Compilation unsuccessful after building 1 target (0 cached) in 00:00:00.
Makefile:132: recipe for target 'build-flow-native-deps' failed
make: *** [build-flow-native-deps] Error 10
```

[...]

PR #1631, opened 2016-04-06
fix mktemp on busybox

mroch:
per discussion in https://github.com/facebook/flow/issues/1605, mktemp on busybox requires 6 X's in its template


mroch:
@facebook-github-bot shipit

gabelevi:
@facebook-github-bot shipit

PR #1636, opened 2016-04-07
Correcting fetch's `json()`

runn1ng:
Fetch's json() should use `Promise<any>`, not `Promise<Object>`, since `any` is what `JSON.parse()` returns, and `json()` just gets text and runs `JSON.parse()` on it.


samwgoldman:
Thanks!

@facebook-github-bot shipit

PR #1638, opened 2016-04-08
Class decorator

marudor:
I've made flow able to parse class Decorator.
As it is a stage 1 feature it is "hidden" behind the esproposal.decorators option.
I've added parser tests & flow tests.


tomatau:
:+1:

samwgoldman:
cc @gabelevi who is the parser wizard

gabelevi:
Thanks for doing this! I have this imported and I've updated the import to handle some of the error cases. I'll get someone else from the team to do a quick review and get this merged today.

Thanks again!

PR #1640, opened 2016-04-08
Make flow automatically load libs from `flow-typed`

splodingsocks:
We've been organizing an effort over at https://github.com/flowtype/flow-typed to build a repository of 3rd party library definitions.

As part of this effort, @jeffmo and I thought it'd be a good idea to have flow automatically load any definitions from the `flow-typed` folder if it exists.

### DIFF
--- a/src/common/flowConfig.ml
+++ b/src/common/flowConfig.ml
@@ -333,7 +333,7 @@ let parse_includes config lines =
   { config with includes; }
 
 let parse_libs config lines =
-  let libs = trim_lines lines in
+  let libs = "flow-typed" :: (trim_lines lines) in
   { config with libs; }
 
 let parse_ignores config lines =
--- a/tests/libflow-typed/.flowconfig
+++ b/tests/libflow-typed/.flowconfig
@@ -1,0 +1,7 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
--- a/tests/libflow-typed/flow-typed/dino.js
+++ b/tests/libflow-typed/flow-typed/dino.js
@@ -1,0 +1,1 @@
+declare type Dinosaur = "T-Rex" | "Apatosaurus";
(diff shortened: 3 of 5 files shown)


vkurchatkin:
@splodingsocks maybe a more generic approach would be better, `flow-typed` is not the only possible source of type definitions

splodingsocks:
@vkurchatkin Thanks for taking the time to look at my PR! I'd love to try a more generic approach, but I"m not sure which direction to head. Do you have any suggestions?

vkurchatkin:
I was thinking about scanning `node_modules` for some marker, maybe a field in `package.json` or some special file name.

jeffmo:
Sweet! Thanks for sending this over.

@vkurchatkin: So I think what you're suggesting is covered by [`.js.flow`](http://flowtype.org/blog/2015/12/01/Version-0.19.0.html#declaration-files) files already, right?

vkurchatkin:
@jeffmo I'm not sure, how could one use `.js.flow` inside of one module to provide types for other module?

jeffmo:
@vkurchatkin: Ah, yea you can't do that. I'm not sure why one module would want to supply libdefs for another module though? Would would versions be managed?

jeffmo:
@facebook-github-bot import

vkurchatkin:
> I'm not sure why one module would want to supply libdefs for another module 

Well, but that's what flow-typed does... If there are typings for a single package, then versions can be just mirrored to keep things simple

jeffmo:
> Well, but that's what flow-typed does...

Sort of -- flow-typed (the repository) doesn't work via modules, though. It's only a repository of library definitions.

There are plans for that repo to have a tool that pulls libdefs from the repo and puts them in this directory, but there's nothing that would stop someone from building a separate tool that does the same thing with the same directory.

The main thing that I like about this PR is that it just implements a "standard" conventional place to put libdefs in general. My hope is that this will reduce the need to specify custom folders in the `[libs]` section of your `.flowconfig` an edge-case/customization feature (still available, but not as necessary for most circumstances).

jeffmo:
Made two small tweaks before merging:

1) If the user explicitly states `flow-typed` in their config (say, *after* another item) then we won't implicitly add it to the list to preserve their ordering for overrides

2) If there is no `flow-typed` directory on disk, don't implicitly add it to the libs list (so we don't try to hit the disk for stuff that isn't actually there later on)

splodingsocks:
🎉

Issue #1604, opened 2016-03-31
Autocomplete partially not working in if statements


marudor:
Autocomplete only works for "first level" in if statements.
Example:
```javascript
// @flow
function test(foo: string) {
  if ()
}
```
flow autocomplete 3 7 works as expected here.
```javascript
// @flow
function test(foo: string) {
  if (foo.)
}
```
flow autocomplete 3 11 now delivers nothing. Although string obviously has functions flow knows about.
```javascript
// @flow
function test(foo: string) {
  foo.
}
```
flow autocomplete 3 7 works as expected.
As these short examples show the if statement somehow confuses flow.

PR #1649, opened 2016-04-11
Autocomplete in if statements

marudor:
fixes https://github.com/facebook/flow/issues/1604
I've added a test case for this and basically added the dispatch_member_hook in the conditional member expressions.


mroch:
d'oh, can't believe this didn't work. nice catch!

will merge once travis finishes.

mroch:
@facebook-github-bot shipit

PR #1661, opened 2016-04-14
adjust core libdef

marudor:
This adjusts the core libdefs to reflect the current specs. I've mainly added some missing stuff. Some is resorted to be alphabetically.
I've removed enumerate on Reflect as it will be removed in the the next spec (https://tc39.github.io/ecma262/#sec-reflection)

I've also added the well-known Symbols. I'm planning on doing a change later that changes the @@iterator behavior to use the well known symbol instead.


gabelevi:
@facebook-github-bot import

PR #1662, opened 2016-04-14
adjust bom libdef

marudor:
Adjusted bom libdef.
most changes were in Navigator.
I've removed the prefixed stuff. We did not include prefixed in other libdefs and should not do it for this either.


gabelevi:
@facebook-github-bot import

Issue #1667, opened 2016-04-15
Enable spread operator on strings


vaukalak:
ES6 spread operator can be applied to strings :
```
let chars = [...'abc'];
        // ['a', 'b', 'c']
```
Flow v0.22.1 falls with error:
```
 9: let chars = [...'abc'];
                     ^^^^^ string. This type is incompatible with
  9: let chars = [...'abc'];
                     ^^^^^ spread operand
```

samwgoldman:
Thanks for the report! This is a known issue, so I'm going to merge this with the original task over in #1059.

PR #1668, opened 2016-04-15
Add spread for strings and Iterables

marudor:
fixes https://github.com/facebook/flow/issues/1667
Also added a test for this case.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -27,6 +27,7 @@ open Constraint_js
 open Type
 open Flow_error
 
+
 (* The following functions are used as constructors for function types and
    object types, which unfortunately have many fields, not all of which are
    meaningful in all contexts. This part of the design should be revisited:
@@ -698,7 +699,7 @@ module Cache = struct
     let compare = Pervasives.compare
   end)
 
-  (* Cache that remembers pairs of types that are passed to __flow__. *)
+  (* Cache that remembers pairs of types that are passed to __. *)
   module FlowConstraint = struct
     let cache = ref TypePairSet.empty
 
@@ -3333,6 +3334,12 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       ->
       rec_flow cx trace (l, SetPropT(reason, (reason, "$key"), i));
       rec_flow cx trace (l, SetPropT(reason, (reason, "$value"), t))
+
+    | (StrT (r, _), UseT (ArrT _))
+      ->
+      let arr = ArrT (r, StrT (r, AnyLiteral), []) in
+      rec_flow cx trace (arr, u)
+
 
     (*************************)
     (* repositioning, part 2 *)
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -794,7 +794,9 @@ let not_expect_use_bound t =
 (** NOTE: Do not call this function directly. Instead, call the wrapper
     functions `rec_flow`, `join_flow`, or `flow_opt` (described below) inside
     this module, and the function `flow` outside this module. **)
-let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
+let rec __flow cx ((l: Type.t), (u: Type.use_t))
+
+ trace =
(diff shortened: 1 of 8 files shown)


mroch:
i suspect that this doesn't work when there's a tvar involved, since `arr` won't be a `StrT` yet. you can probably add a `(StrT _, ArrT _)` rule in `Flow_js.__flow` instead.
t
can you also add negative tests (that produces an error)? for example:

```
var z2: Array<'1' | '2' | '3'> = [...'123']; // error, flow doesn't actually break up the chars
var z3: Array<'123'> = [...'123']; // error, flow does know that the chars are spread
```

I don't think the `z3` case errors right now. you'd fix that by flowing `StrT (reason, AnyLiteral)`.

marudor:
@morch Thanks for the feedback.
I've adjusted it to have a StrT, ArrT case in flow_js.ml.
In the statement.ml I degrade the specific string to a unspecific string (AnyLiteral) and use that to decide if I can spread it or not. (Flow can only spread generic strings now, no String with a specific literal)
Otherwise the function declarations with spread signature will fail.
But it should be okay now.

If you've got something to add please tell me.

samwgoldman:
Ideally we could generalize this to any iterable, not just arrays/strings. I know there's an open issue for that, but I'm on mobile...

marudor:
@samwgoldman I've just did that.
New tests are in iterable_spread.
Everything that has @@iterable will be able to use spread.
Side effect is that the error message for wrong generator usage changed slightly. (See core_tests.exp changes)

samwgoldman:
Wow, nice job digging into this stuff and making progress here! I know it's not a simple code base to get oriented in :)

I think this logic is not quite correct as written, though. Adding a case in `Flow_js.flow` for `(InstanceT _, UseT ArrT _) when is_iterable l` means that the following code would also be allowed:

```js
var x: T[] = iterableOfT
```

The above isn't legal, so we need something a bit more than a subtyping rule.

marudor:
@samwgoldman 
Added a test for your case and fixed it.
Can you think of any other case that would be allowed now although it shouldn't?

If so - would checking the reason of OpenT be reasonable? (OpenT reason has spread operand)

gabelevi:
So if you'd like to make this work for any `Iterable`, you should probably be able to follow the [example set by `for(x of myIterable) { ... }`](https://github.com/facebook/flow/blob/0697ff6610884c4814f579865a5776e2da7419ea/src/typing/statement.ml#L1616).

It would look something like

```OCaml
 and array_element_spread cx type_params_map (loc, e) =
  let spread_element = expression cx type_params_map (loc, e) in
  let reason = mk_reason "spread operand" loc in
  Flow.mk_tvar_where cx reason (fun tvar ->
    let iterable = Flow.get_builtin_typeapp
      cx
      reason
      "$Iterable"
      [tvar; AnyT.at loc; AnyT.at loc] in
    Flow.flow_t cx (spread_element, iterable);
  )
```

* Disclaimer: I haven't tried this and I'm not 100% sure it will actually work. Just an idea :)

yinquanteo:
Ran into this issue when trying to spread an Array Iterator. Is this commit good to merge or are we planning to make this work for any Iterable before merging?

samwgoldman:
@facebook-github-bot import

GitHub won't let me see all the changes here, so importing for review. If memory serves, I don't believe this is quite ready to merge but I'll take a look today.

marudor:
Correct, still not ready.

gabelevi:
@marudor - I'm doing some late-fall spring cleaning of our pull request queue. I'm going to close this PR out for now, but please feel free to re-open if you decide to pick this back up!

jessebeach:
Still running into this in 0.38.0.

jessebeach:
Working as of 0.40.0. Thanks!

galkowskit:
`[flow] [flow] Set (This type is incompatible with spread operand See also: constructor call)`

On:

```
map1.set(string1, {
      ...object1,
      array1: [...set1]
})
```
Where `map1` is a `Map()`, etc.

On 0.40.0.

PR #1675, opened 2016-04-17
WIP: function bind

marudor:
Not done yet. But basic support is implemented.

Parser understands :: syntax and has tests.
It is not yet guarded behind esproposal.function_bind.
actual usage still has some rough edges. (Complaining about undefined context and stuff like that)


marudor:
My start on this is way to naive. Gonna need more time & a fresh start.

PR #1681, opened 2016-04-19
Type parameter defaults

gabelevi:
Motivation:

Currently we have some code like

```JavaScript
interface $Iterator<Yield,Return,Next> { ... }
type Iterator<T> = $Iterator<T,void,void>;
```

It would be nice to instead be able to write

```JavaScript
interface Iterator<Yield,Return=void,Next=void> { ... }
```

General idea:

  * Type parameter declarations for classes, interfaces, types, etc can now include defaults
  * Type parameter declarations for methods, functions, etc are unaffected
  * Type parameter instantiation can now take 0 args (`Foo<>`)

Status quo:

  * A generic type with N type parameters can be instantiated with 0 or N type arguments
  * When you instantiate with 0 type arguments, we use `any` for all the type params

Change:

  * If a generic type with N type parameters has defaults and is instantiated with less than N type arguments, then we use the defaults

### DIFF
--- a/src/parser/estree_translator.ml
+++ b/src/parser/estree_translator.ml
@@ -1050,6 +1050,7 @@ end with type t = Impl.t) = struct
       "name", string tp.name;
       "bound", option type_annotation tp.bound;
       "variance", option variance tp.variance;
+      "default", option _type tp.default;
     |]
   )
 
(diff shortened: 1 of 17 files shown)


gabelevi:
@facebook-github-bot import

PR #1699, opened 2016-04-22
Small fixes for force-recheck

gabelevi:
Two small fixes in this commit:

1. On OSX the server would have a little bit of log spew. Fixed.
2. Most `flow command path/to/file.js` would use `path/to/file.js` to infer the
   root. However `flow force-recheck` would always use the cwd unless `--root`
   was explicitly passed. Fixed.


gabelevi:
@facebook-github-bot import

PR #1703, opened 2016-04-23
Add parser support for access modifiers

samwgoldman:
This commit generalizes parsing the static modifier to also parse access
modifiers (readwrite, readonly, writeonly).

Access modifiers are supported in object types, object literals,
classes, interfaces, and declared classes.

This commit only adds parsing support. Type checking support will be
added in subsequent commits.


samwgoldman:
@facebook-github-bot import

samwgoldman:
Based on feedback from the team, we're going to explore a different syntax for this.

Issue #1454, opened 2016-02-25
property `on` not found in domain object


d14b3l:
I know that in nodejs documentation there is lack of description for domain.on() error handler but in the very same documentation there is example of usage for this method:
    var reqd = domain.create();
    reqd.add(req);
    reqd.add(res);
    reqd.on('error', (er) => {
      console.error('Error', er, req.url);
      try {
        res.writeHead(500);
        res.end('Error occurred, sorry.');
      } catch (er) {
        console.error('Error sending 500', er, req.url);
      }
    });

But when I use domain.on in my code flow complains that on property is not defined in object.
Is there any reason why on property was not defined in domain module or I can add it via pull request?

mroch:
rather than adding the `on` method individually, it looks like we should model Domain as a class instead of an object type, and the class should extend EventEmitter. please do send a pull request!

for others' reference, the docs are here: https://nodejs.org/dist/latest-v4.x/docs/api/domain.html
and our implementation is here: https://github.com/facebook/flow/blob/15b0be7165b288973740a7b92ff3b068567fa973/lib/node.js#L463

d14b3l:
Hi, finally I found time to fix this issue. 
pull request: https://github.com/facebook/flow/pull/1712

PR #1712, opened 2016-04-26
domain reimplemented as class

d14b3l:
domain$Domain changed from type declaration to class definition as suggested in this issue:
https://github.com/facebook/flow/issues/1454

domain definition moved down so it would be able to extend eventEmmiter definition

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -485,7 +485,7 @@ declare class events$EventEmitter {
 
 declare module "events" {
   // TODO: See the comment above the events$EventEmitter declaration
-  declare class EventEmitter extends events$EventEmitter {}
+  declare var EventEmitter: typeof events$EventEmitter;
 }
 
 declare class domain$Domain extends events$EventEmitter {
--- a/lib/node.js
+++ b/lib/node.js
@@ -482,15 +482,15 @@ declare class events$EventEmitter {
   setMaxListeners(n: number): void;
 }
 
-declare class EventEmitter extends events$EventEmitter {};
 
 declare module "events" {
   // TODO: See the comment above the events$EventEmitter declaration
-  declare var EventEmitter : typeof events$EventEmitter;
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -499,9 +482,27 @@ declare class events$EventEmitter {
   setMaxListeners(n: number): void;
 }
 
+declare class EventEmitter extends events$EventEmitter {};
```
mroch: this makes `EventEmitter` a valid global, which is not true is it? you have to import it from `events`?

would doing `declare class domain$Domain extends events$EventEmitter {` work?

d14b3l: yes, it works fine, correction delivered

mroch:
great, thanks!

now I'm wondering if `myDomain instanceof require('events').EventEmitter` would pass flow. I believe it's true at runtime. not sure how to handle this 100% accurately. I think we should merge this anyway since it's an improvement and instanceof is an edge case.

lib/node.js
```diff
@@ -499,11 +482,29 @@ declare class events$EventEmitter {
   setMaxListeners(n: number): void;
 }
 
+
 declare module "events" {
   // TODO: See the comment above the events$EventEmitter declaration
   declare class EventEmitter extends events$EventEmitter {}
```
mroch: can you change this to what you were changing it to before?

```
declare var EventEmitter : typeof events$EventEmitter;
```

that will fix the `instanceof` issue.

d14b3l: Hi, correction delivered accordingly to your comment.

mroch:
@facebook-github-bot shipit

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/545899388924725/int_phab) to review.

PR #1727, opened 2016-04-28
Use curl to download Travis cache

mroch:
the aws cli requires auth, even for anonymous/public downloads, and we don't want to use a token for PRs. https doesn't require auth, so use curl instead.


mroch:
@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/473388409519847/int_phab) to review.

PR #1733, opened 2016-04-30
Support subtyping on Class<> qualified annotations

popham:


gabelevi:
Are you trying to make it so that `Class<B>` is a subtype of `Class<A>` if `B` is a subtype of `A`?

popham:
Yes.  I've got a bigger PR coming up that includes this.  Closing....

PR #1749, opened 2016-05-05
Add 'ls' command to list files flow can see

mroch:
There's currently no good way to see which files flow can see based on the `[include]` and `[ignore]` sections.

This adds a basic `ls` command which just prints every file the file finder finds. It includes `--include` and `--ignore` flags, so you can play with those before touching your `.flowconfig`.

### DIFF
--- a/src/commands/commandUtils.ml
+++ b/src/commands/commandUtils.ml
@@ -166,26 +166,40 @@ type flowconfig_params = {
   libs: string list;
 }
 
-let collect_flowconfig_flags =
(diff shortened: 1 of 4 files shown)


mroch:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1545727529064588/int_phab) to review.

PR #1750, opened 2016-05-05
Don't add unparsed files to search index

mroch:
When a file fails to parse or is skipped because it isn't @flow, we were generating an empty AST and passing it to the search service (which, AFAIK, nobody even uses yet 😿 ) which then stores a bunch of empty lists in shared mem.

This is unnecessary since the filenames are unused by the search service; they're only stored by filename to prevent workers from clobbering each other.

Additionally, if you have enough non-@flow files, you can fill up shared memory even though they're all otherwise skipped. Projects using babel-6 seem to do this easily.

### DIFF
--- a/src/parsing/parsing_service_js.ml
+++ b/src/parsing/parsing_service_js.ml
@@ -33,16 +33,10 @@ module ParserHeap = SharedMem.WithCache (Loc.FilenameKey) (struct
     let prefix = Prefix.make()
   end)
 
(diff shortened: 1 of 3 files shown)


mroch:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1119878931366686/int_phab) to review.

PR #1752, opened 2016-05-06
adding-new-attr-to-dom-lib - Adding getAttribute to the Node type

mitermayer:
Adding getAttribute to the Node type

- Including getAttribute


mitermayer:
no longer needed

PR #1753, opened 2016-05-06
Fix Makefile rebuilding of object files

mroch:
I was having problems with `make` not picking up changes to `.c` files. I think the problem is that `build-flow` depends on `build-flow-native-deps` which does not actually depend on the `.c` files, so that step was getting skipped.

As I started tweaking stuff, I eventually just replaced the target that runs `ocamlbuild` on all `.o` files with a rule that directly runs `ocamlopt` on each `.c` file, letting `make` instead of `ocamlbuild` decide when it's dirty. this is faster.

Test Plan: changed hh_shared.c, ran `make`. before, it wasn't picking up my changes unless I deleted `_build/`. after, it works. running `make` again doesn't rebuild things that are clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ test: build-flow copy-flow-files
 test-ocp: build-flow-with-ocp copy-flow-files-ocp
 	${MAKE} do-test
 
-js: $(NATIVE_OBJECT_FILES)
+js: $(BUILT_OBJECT_FILES)
 	mkdir -p bin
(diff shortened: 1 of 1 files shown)


mroch:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1748567428688686/int_phab) to review.

Issue #1705, opened 2016-04-24
Class<> qualified interface constraints do not accept compatible classes


popham:
Given the following, Flow generates an error:

``` js
interface I {
  static fn(): number
}

class Impl {
  static fn() {
    return 5;
  }
}

function g(C: Class<I>): number {
  return C.fn();
}

var m: number = g(Impl);
```

The error looks like so:

``` js
 15: var m: number = g(Impl);
                     ^^^^^^^ function call
 11: function g(C: Class<I>): number {
                         ^ I. This type is incompatible with
 15: var m: number = g(Impl);
                       ^^^^ Impl
```

The `Class` qualifier smells funny here, but my intuition says that it should work.

asolove:
The provided code works in v53. So I think this can be closed. (/cc @calebmer)

Issue #1704, opened 2016-04-24
Class instances flow to interface constraints without validating statics


popham:
Given the following, Flow doesn't generate any errors:

``` js
interface I {
  static x: number;
  static fn(): number
}

class Impl {
  static x: string;
  static fn(): string {
    return "some string";
  }
};
Impl.x = "another string";

function f(c: I): number {
  return c.constructor.fn();
}

var i = new Impl();
var n: number = f(i);
```

Flow should flag both members of `Impl` as incompatible with the corresponding members of `I`, right?  (I stumbled on this when babel 6.5.2--latest npm version--choked on an interface that includes the `static` "unexpected token" before a static method's signature.  Before I open a bug report over there, I want to verify that interfaces are intended to include statics--I asked this over at #803 a while back, but it was never _really_ resolved.)

wbyoung:
@popham there seem to be some inconsistencies in what Flow is doing right now. Here's some strange behavior I'm observing in 0.42.0:

```js
/* @flow */

interface I {
  static x: number;
}

class Impl implements I {
  static y: number;
}

const iClass: Class<I> = Impl;
const implClass: Class<Impl> = Impl;

console.log(iClass.y);
console.log(iClass.x);
console.log(implClass.y);
console.log(implClass.x);
```

```
14: console.log(iClass.y);
                       ^ property `y`. Property not found in
14: console.log(iClass.y);
                ^ statics of I
17: console.log(implClass.x);
                          ^ property `x`. Property not found in
17: console.log(implClass.x);
                ^ statics of Impl
```

There should probably also be an error since `Impl` doesn't fully implement `I`, but Flow is noticing that `iClass` does have the static `x` while it doesn't have `y`.

Issue #1369, opened 2016-02-05
Is it possible to restrict type of `this` for a particular method?


rpominov:
``` js
class Foo<T> {
  _x: T;

  constructor( x:T ) {
    this._x = x
  }

  // this method works with any instance of Foo
  getX(): T {
    return this._x
  }

  // this method works only with Foo<number>
  getXMultBy( y:number ): number {

    // this will result in "T. This type is incompatible with number", 
    // but let's ignore that for the sake of example
    return this._x * y
  }
} 
```

In this example, can we somehow add a restriction for `getXMultBy` that `this` must be an instance of `Foo<number>`, so the following code wouldn't type check?

``` js
(new Foo('bar')).getXMultBy(10)
```

leonardfactory:
I don't know if you could do this in your project, however I think you should use inheritance here, like:

```js
/* @flow */
class Foo<T> {
  _x: T;

  constructor(x: T) {
    this._x = x;
  }

  getX(): T {
    return this._x;
  }
}

class FooBar extends Foo<number> {
  constructor(x: number) {
    super(x);
  }

  getXMultBy(y: number) {
    return this._x * y;
  }
}

(new Foo('bar')).getX(); // ok
(new FooBar(12)).getXMultBy(10); // ok
(new Foo(12)).getXMultBy(10); // error!
```

If this is not feasible, you could try using (invariant)[https://github.com/zertosh/invariant]

rpominov:
@leonardfactory Yeah, inheritance isn't very convenient in my case. Invariant also will work, but only at run time...

Some context: in [Fantasy Land](https://github.com/fantasyland/fantasy-land) some methods (like `ap` or `sequence`) work only when wrapped value has a certain type (`ap` needs it to be a function, and `sequence` needs it to be an Applicative). https://github.com/fantasyland/fantasy-land/issues/50#issuecomment-173400402

Inheritance indeed may be the right choice for this problem. Still I'm not sure it will be practical to use separate classes for some methods in FL code. This is not well explored though, and I'm not sure in either case.

Anyway what I originally asked maybe useful in some cases, IMO. Would be cool if it was considered, or maybe already possible somehow.

rpominov:
What I'm trying to say is that, if we forget about OOP for a moment, `this` for methods basically just another argument, and it would be cool to be able to add type restrictions for that argument as well.

rpominov:
Related: #452

calebmer:
[...]

PR #1761, opened 2016-05-08
Support `this` pseudo-params in classes, and enforce subtyping on structural class statics

### LINKED ISSUE #807: Abstract class support
I don't see abstract classes on the roadmap.  Any chance of support under Flow?

I want a class that looks like so:

``` js
class ReaderA {
  // Imagine properties here.

  constructor(arena: Arena, depth: uint, isOrphan: boolean, segment: SomeBuffer) {
    if (depth > arena.maxDepth) {
      throw Error("Depth exceeded");
    }
    // Populate properties here.
  }

  // Pure virtual static.  This doesn't work.
  static _deref(arena: Arena, pointer: Datum, depth: uint): Reader;

  static _fieldGet<T>(context: Struct, offset: uint, defaultArena: Arena): T {
    var pointer = {
      segment: context._segment,
      position: context._pointersSection + offset
    };

    if (pointer.position < context._end && !isNull(pointer)) {
      return this._deref(context._arena, pointer, context._depth+1); // NG
    } else {
      return this._deref(defaultArena, defaultArena._root(), 0); // NG
    }
  }
  // ...
}
```

My current work-around, I think, is to extract `_fieldGet` to an external function, `fieldGet`.  I create a class and an interface:

``` js
class Reader {
  // Imagine properties here.

  constructor(arena: Arena, depth: uint, isOrphan: boolean, segment: SomeBuffer) {
    if (depth > arena.maxDepth) {
      throw Error("Depth exceeded");
    }
    // Populate properties here.
  }
}

interface ReaderI {
  static _deref(arena: Arena, pointer: Datum, depth: uint): Reader;
  static _fieldGet<T>(context: Struct, offset: uint, defaultArena: Arena): T;
}
```
- For all implementors of `ReaderI`, `_fieldGet` calls my extracted `fieldGet`.
- In places where I would have used the type `ReaderA`, I can use `Reader & ReaderI`.  That will be a useful type alias:  `type ReaderA = Reader & ReaderI`.

popham:
This patch adds support for a `this` pseudo-param on class methods (static and non-static).  With the syntax `someMethod(this: this & I,...` users can impose requirement beyond "is subtype of the enclosing class," the current behavior.  Some goofiness becomes admissible too, e.g. `someMethod(this: {})`--just don't be goofy (maybe a warning is warranted if the pseudo-param's type is not a subtype of the enclosing class).  I have a use case where the value in this extension lies in chaining of static methods, not non-statics, so I've extended class subtyping to structural class statics (InstanceT ~> SuperT still ignores subtyping of statics).  I feel like I'm on solid ground after a chat with @jeffmo  and @avikchaudhuri  on IRC, although during that discussion I forgot to contrast ES5 and Node inheritance against ES6 inheritance.  I recall in the past that @samwgoldman  and I disagreed on whether class statics should subtype.

Node's inheritance (`util.inherit`) and the ad hoc inheritance of ES5 (`SomeClass.prototype = Object.create(ParentClass.prototype, {...})` *without* SomeClass.\__proto__ = ParentClass) do not chain class statics.  Subtyping statics is biased toward ES6 inheritance, whereas the status quo is biased toward Node and ad hoc inheritance.

Closes #807, #1369, #1704, #1705

### DIFF
--- a/src/parser/estree_translator.ml
+++ b/src/parser/estree_translator.ml
@@ -869,6 +869,7 @@ end with type t = Impl.t) = struct
     | Any -> any_type loc
     | Void -> void_type loc
     | Null -> null_type loc
+    | This -> this_type loc
     | Number -> number_type loc
     | String -> string_type loc
     | Boolean -> boolean_type loc
@@ -892,6 +893,8 @@ end with type t = Impl.t) = struct
   and void_type loc = node "VoidTypeAnnotation" loc [||]
 
   and null_type loc = node "NullTypeAnnotation" loc [||]
+
+  and this_type loc = node "ThisTypeAnnotation" loc [||]
 
   and number_type loc = node "NumberTypeAnnotation" loc [||]
 
--- a/src/parser/lexer_flow.mll
+++ b/src/parser/lexer_flow.mll
@@ -525,6 +525,7 @@
       "false",   T_FALSE;
       "number",  T_NUMBER_TYPE;
       "string",  T_STRING_TYPE;
+      "this",    T_THIS;
       "void",    T_VOID_TYPE;
       "null",    T_NULL;
     ]
--- a/src/parser/loc.ml
+++ b/src/parser/loc.ml
@@ -67,6 +67,12 @@ let btwn loc1 loc2 = {
   _end = loc2._end;
 }
 
+let start loc = {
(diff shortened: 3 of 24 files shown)


popham:
New PR for rebase

PR #1762, opened 2016-05-09
Fix hardcoded parser tests and get travis to run them

gabelevi:
Tests recently broke, partially because they were testing some broken behavior.

We run these tests on our internal CI, but Travis should run them too, so we can see when PRs break tests.


gabelevi:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1602166670096005/int_phab) to review.

PR #1775, opened 2016-05-12
Support `this` pseudo-params in classes

### LINKED ISSUE #807: Abstract class support

popham:
This patch adds support for a this pseudo-param on class methods (static and non-static). With the syntax `someMethod(this: this & I,...` users can impose requirement beyond "is subtype of the enclosing class," the current behavior. Some goofiness becomes admissible too, e.g. someMethod(this: {})--just don't be goofy (maybe a warning is warranted if the pseudo-param's type is not a subtype of the enclosing class). ~~I have a use case where the value in this extension lies in chaining of static methods, not non-statics, so I've extended class subtyping to structural class statics. I feel like I'm on solid ground after a chat with @jeffmo and @avikchaudhuri on IRC, although during that discussion I forgot to contrast ES5 and Node inheritance against ES6 inheritance. I recall in the past that @samwgoldman and I disagreed on whether class statics should subtype.~~

~~Node's inheritance (util.inherit) and the ad hoc inheritance of ES5 (`SomeClass.prototype = Object.create(ParentClass.prototype, {...})` without `SomeClass.__proto__ = ParentClass`) do not chain class statics. Subtyping statics is biased toward ES6 inheritance, whereas the status quo is biased toward Node and ad hoc inheritance.~~ (PR #1847)

Closes #807, #1369, ~~#1704, #1705~~

A lot of the material here is part of PR #1823. Some has been factored out into PR #1847. Pulls there and a rebase here would remove a lot of noise.  ~~The patch is a work-in-progress.~~  There's a few false negatives in the tests.  ~~Also, generic class declarations are broken.  I want to poll IRC on how I should approach that aspect.~~


src/typing/flow_js.ml
```diff
@@ -3291,9 +3323,12 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       rec_flow cx trace (arrt, u)
 
     | (ArrT (_, t, _), (GetPropT _ | SetPropT _ | MethodT _ | LookupT _)) ->
```
popham: Why does the `TypeAppT~>MethodT` case above get excluded from `TypeAppExpansion` handling? Is there some edge case there that has an analogue here?

src/typing/classy.ml
```diff
@@ -0,0 +1,252 @@
...
+      Flow.flow cx (super, Type.SuperT (reason, insttype))
+    ) |> ignore
+end
+
+module Class = Make(Class_sig.T)
+module Interface = Make(Iface_sig.T)
+module DeclClass = Make(Decl_class_sig.T)
+
+(* Processes the bodies of instance and static class members. *)
+let toplevels cx ~decls ~stmts ~expr x =
```
popham: I hacked this function to make it work. @samwgoldman had it so pretty--definitely needs tidying.

src/typing/statement.ml
```diff
@@ -68,6 +68,11 @@ let function_desc ~async ~generator desc =
   | false, true -> spf "generator %s" desc
   | false, false -> desc
 
+let reject_explicit_this cx msg = function
```
popham: I don't know whether `this`-pseudoparams just work on objects.  There's a single test case where I move a method from the class's prototype to a compatible object, and it just works.  The tests should be extended to validate behavior on objects, or `this`-pseudoparams should be blacklisted on objects.

src/typing/func_sig.ml
```diff
@@ -81,11 +102,13 @@ let field_initializer tparams_map reason expr return_t = {
   kind = FieldInit expr;
   tparams = [];
   tparams_map;
-  params = Func_params.empty;
+  params = Func_params.empty Flow.dummy_this;
```
popham: I'm uncomfortable with field_initializers.  Is this okay?

src/parser/spider_monkey_ast.ml
```diff
@@ -54,7 +54,14 @@ and Type : sig
       }
     end
 
+    module ThisParam : sig
```
popham: Is this data modelling consistent with the spec?

tests/declare_class/lib/test.js
```diff
@@ -2,7 +2,7 @@ declare class _C {
   foo(): number;
 }
 declare var _module: {
-  C: Class<_C>;
+  C: typeof _C;
```
popham: This may generate some traffic in the issue tracker.  *Type* `_C` is not `this`-abstract, so extending `_module.C` under the `Class<_C>` case is no good.  The `typeof` operator exposes *value* `_C` for the work around.  I understand that somebody is currently working toward a solution here....

src/typing/flow_js.ml
```diff
@@ -930,7 +953,8 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
     (* The sink component of an annotation constrains values flowing
        into the annotated site. *)
 
-    | _, UseT (use_op, AnnotT (sink_t, _)) ->
+    | l, UseT (use_op, AnnotT (sink_t, _))
```
popham: This is part of the PR's "enforce subtyping on structural class statics" clause. It should probably be exported and rebased below this PR.

src/typing/flow_js.ml
```diff
@@ -2426,15 +2450,22 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
+       UseT (_, InstanceT (reason_inst, static, super, {
          fields_tmap;
          methods_tmap;
          structural = true;
          _;
        })))
       ->
       structural_subtype cx trace l reason_inst
-        (super, fields_tmap, methods_tmap)
+        (super, fields_tmap, methods_tmap);
```
popham: This is part of the PR's "enforce subtyping on structural class statics" clause. It should probably be exported and rebased below this PR.

src/typing/flow_js.ml
```diff
@@ -3461,15 +3495,23 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
+       UseT (_, ExtendsT ([], l, InstanceT (reason_inst, static, super, {
          fields_tmap;
          methods_tmap;
          structural = true;
          _;
        }))))
       ->
       structural_subtype cx trace l reason_inst
-        (super, fields_tmap, methods_tmap)
+        (super, fields_tmap, methods_tmap);
```
popham: This is part of the PR's "enforce subtyping on structural class statics" clause. It should probably be exported and rebased below this PR.

src/typing/flow_js.ml
```diff
@@ -3687,6 +3729,7 @@ and ground_subtype = function
   | (VoidT _, UseT (_, VoidT _))
   | (EmptyT _, _)
   | (_, UseT (_, MixedT _))
+  | (_, UseT (_, ClassT (MixedT _)))
```
popham: I think that this is part of the PR's "enforce subtyping on structural class statics" clause. It should probably be exported and rebased below this PR.

popham:
I feel like I owe some documentation here.  Is there an unwritten policy where new features remain undocumented for a period of time before introducing them through flowtype.org?  Maybe for the purpose of allowing new features to stabilize though the dev community before exposing them to a broader audience?  If so, then maybe a wiki page?

popham:
A lot of the material here is part of PR #1823.  A pull there and a rebase here would remove a lot of noise.  I'm going to focus on #1823.

popham:
I just realized that my testing has neglected declared classes.  Class declarations with this-pseudoparams would allow library authors to impose requirements where a library class requires the user to provide method(s) on an child class.

tests/declare_class/declare_class.exp
```diff
@@ -28,5 +28,37 @@ declare_class.js:10
...
+ 15: var s1: string = x.absM(); // NG: X doesn't implement V
+                      ^ statics of X
+
+pseudo_this.js:16
+ 16: var s2: string = X.absSM(); // NG: X doesn't implement V
+                      ^^^^^^^^^ call of method `absSM`
+ 10:   static absSM(this: Class<this> & Class<V>): string;
+                                              ^ property `virtualSM` of statics of V. Property not found in
+  6: declare class X {
+     ^ statics of X
```
popham: This `loc` misses `X`.  Is it even the correct line?  See line 36 above for the nonstatic analogue.

tests/this_param/interface.js
```diff
@@ -0,0 +1,89 @@
...
+  static anotherStaticMethod() {
+    return 2;
+  }
+}
+
+var c = new C();
+var n3: number = c.method();       // OK
+var n4: number = C.staticMethod(); // OK
+
+// TJP TODO: In tests, generalize the following to a generic so that #1369 can be closed. Remove this note.
```
popham: Done.  Remove this note.

tests/interface/interface.exp
```diff
@@ -108,5 +108,41 @@ test3.js:6
...
+           ^ statics of A
+
+test4.js:9
+  9: var P: Class<A> & Class<I> = A; // NG
+                             ^ property `fn` of I. Property not found in
+  9: var P: Class<A> & Class<I> = A; // NG
+                                  ^ A
+
+test4.js:20
+ 20: var q: A & I = b; // OK
```
popham: Whoops

tests/interface/interface.exp
```diff
@@ -108,5 +108,41 @@ test3.js:6
...
+                ^ property `gn` of statics of I. Property not found in
+ 20: var q: A & I = b; // OK
+                    ^ statics of B
+
+test4.js:21
+ 21: var Q: Class<A> & Class<I> = B; // OK
```
popham: Whoops

tests/interface/test4.js
```diff
@@ -0,0 +1,21 @@
...
+class A {}
+var a = new A();
+var p: A & I = a; // NG
+var P: Class<A> & Class<I> = A; // NG
+
+class B extends A {
+  fn(): number {
+    return 1;
+  }
+  gn(): number {
```
popham: Should have been static :(

popham:
I'm incorporating this branch into a project currently, and I have an interface method that I want to bind with return type `this`, where `this` is the implementing class.  I assumed that I was screwed, but then `check`ed just in case (I recall `this` being out of bounds on interfaces).  It's working for a trivial test, dumb luck would have it.  Is there an argument for banishing?  Obviously I want `this` support on interfaces.  Adding tests to demonstrate....

tests/this_param/interface.js
```diff
@@ -0,0 +1,89 @@
...
+  static anotherStaticMethod(): number;
+}
+
+class A {
+  x: number;
+  static x: number;
+  method(this: this & I): number {
+    return this.x + this.anotherMethod();
+  }
+  static staticMethod(this: Class<I> & Class<this>): number {
```
popham: Under `this: Class<this> & Class<I>`, the `this`-fixing process leaves a tvar flowing to SpeculativeMatchT instead of a concrete type.  When the `Class<this>` finally goes concrete, the speculative testing setup has already unwound, and the error leaks.

Swapping the intersection members like they're configured here allows the method to be found on `Class<I>` without ever touching `Class<this>`, hence the absence of an error in the current configuration.

Quick and dirty hypothesis check--the following eliminates the spurious error:

```ocaml
    (*Above the (head_l, SpeculativeMatchT) case (calls `speculative_match head_l u`)*)
    | (ClassT instance, UseT (_, SpeculativeMatchT _)) -> (
      let reason = prefix_reason "statics of " (reason_of_t instance) in
      let tvar = mk_tvar cx reason in
      rec_flow cx trace (instance, GetPropT(reason, (reason, "statics"), tvar));
      rec_flow cx trace (tvar, ReposLowerT (reason, u))
    )
```

```ocaml
    (*Above the (ClassT, _) flows statics to u*)
    | (InstanceT _, UseT (k, ClassT instance)) ->
      let reason = prefix_reason "statics of " (reason_of_t instance) in
      let tvar = mk_tvar cx reason in
      rec_flow cx trace (instance, GetPropT(reason, (reason, "statics"), tvar));
      rec_flow cx trace (l, UseT(k,tvar))
```

tests/this_param/object.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+}
+
+var c = new C();
+var n3: number = c.method();
+
+// False negative
+// "`C`. This type is incompatible with `J & Class<this>`, some incompatible
+// instantiation of `this`". Smells like a `subst` problem.
+var n4: number = C.staticMethod(); // OK
```
popham: "some incompatible instantiation of `this`" just sounds like the catchall error message to my ear now.  False negative probably due to `ClassT` induced leakiness of speculative matching similar to the "commutativity" problem.

tests/this_param/object.js
```diff
@@ -0,0 +1,100 @@
...
+var s1: string = b.anotherMethod();
+var n3: number = b.method(); // NG
+var s2: string = B.anotherStaticMethod();
+
+// False negative
+// Although this is no good, an unnecessary error seems to be generated:
+// "`J & Class<this>` intersection. This type is incompatible with `B`". An
+// appropriate error is generated for the case, this is unneeded. The problem
+// may be rooted in ObjT lookups doing a unify--is some bound filter missing?
+var s3: number = B.staticMethod(); // NG
```
popham: The false negative sounds like ClassT leakiness again

avikchaudhuri:
@popham Apologies for not getting the chance to look at this earlier.

My sense is that unless we try doing this feature for all functions (i.e., recognize them as potential open methods that could appear in method bags and be "mixin"d a bunch of times before eventually being called), we'd not know what the entire range of issues are. Internally the type system already carries the type of `this` for all functions, so it should be possible to lift that up to the surface syntax. This is well-suited for apply/bind/call use cases. But alternative designs may exist. For example, it might turn out to be more advantageous to always restrict the appearance of such methods in bags that satisfy the `this` type of those methods, etc.

At this point we simply don't know what the general case might look like, and it is an often-enough requested feature that we think we'll be working on it at some point in the near future. Locking down part of the design before we've had the chance to think through the general case seems risky.

Overall the class use-case is simpler, and simpler solutions might exist. We already know a lot about `this` in classes. The example you mention could easily be handled if we allow declaration of abstract methods (if avoiding the error at definition but throwing the error on call is the main problem he's trying to solve). Relying on an abstract method when defining a method is OK. Calling through a type that only exposes an abstract method is not OK. Done.

I think the general case is much more interesting, and gets to the heart of what's going on with open methods. So I propose we hold off on this PR.

How do you feel about implementing abstract methods (as outlined above) instead? Does that not satisfy all your use cases?

popham:
Fair enough.  I expect some OO-dogmatist heads to explode at the idea of an abstract static method.  I'll do it initially behind the `abstract` reserved word.

(With Flow, I have to pause once in a while and remind myself which code makes it through to the run-time build.  I figured that having the abstract method syntax aligned with the eventual pseudo-param syntax on functions would make it easier to remember that these "methods" disappear, although the absence of a function body serves as a pretty good reminder.  I'll reclassify my fixation on the run-time build as a personal problem :).)

PR #1819, opened 2016-05-19
adjust MutationRecord lib definition

cloudkite:
Was recently using MutationObserver with flow and noticed that MutationRecord does not quite match the spec. https://dom.spec.whatwg.org/#mutationrecord

- `type` can only be one of three strings
- `previousSibling`, `nextSibling`, `attributeName`, `attributeNamespace`, `oldValue` should all be nullable
- for readability changed order of properties to be the same as in spec

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -177,15 +177,15 @@ declare class FormData {
 }
 
 declare class MutationRecord {
-    oldValue: string;
-    previousSibling: Node;
-    addedNodes: NodeList<any>;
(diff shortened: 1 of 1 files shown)


mroch:
@facebook-github-bot shipit

facebook-github-bot-2:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/236190150092052/int_phab) to review.

PR #1821, opened 2016-05-20
added documentation for HOCs

nmn:
Added some explanation and an additional example to talk about declaring types for Higher Order Components.

### DIFF
--- a/website/docs/ref/react.doc.js
+++ b/website/docs/ref/react.doc.js
@@ -387,11 +387,13 @@ const Timeless = injectDate(Timely);
 */
 
 // D = DefaultProps, P: Props, S: State;
-declare var injectDatesWithStatelessSupport: (
-  <D, P, C: React$Component<D, P, any>>(
-    Komponent: Class<C>
-  ) => Class<React$Component<D, $Diff<P, Injected>, any>>
-) & <P>(component: (props: P) => any) => Class<React$Component<void, $Diff<P, Injected>, any>>;
+type InjectDateToClass = <D, P, C: React$Component<D, P, any>>(
+  Komponent: Class<C>
+) => Class<React$Component<D, $Diff<P, Injected>, any>>;
+
+type InjectDateStateless = <P>(component: (props: P) => any) => Class<React$Component<void, $Diff<P, Injected>, any>>;
+
+declare var injectDateStateless: InjectDateStateless & InjectDateToClass;
 
 
 /*
--- a/website/docs/ref/react.doc.js
+++ b/website/docs/ref/react.doc.js
@@ -329,46 +329,56 @@ type TimelyProps = {
   date: Date,
   name: string,
   excited: boolean
-}
+};
 
 class Timely extends React.Component<void, TimelyProps, void> {
   render() {
-    const hours = this.props.date.getHours()
-    const timeOfDay = hours > 17 ? 'Evening' : hours > 12 ? 'Afternoon' : 'Morning'
+    const hours = this.props.date.getHours();
+    const timeOfDay = hours > 17 ? 'Evening' : hours > 12 ? 'Afternoon' : 'Morning';
 
-    return <div>Good {timeOfDay} {this.props.name} {this.props.excited ? '!' : ''}</div>
+    return (
+      <div>
+        Good {timeOfDay} {this.props.name} {this.props.excited ? '!' : ''}
+      </div>
+    );
   }
 }
 
 // $ExpectError
 <Timely />; // Timely didn't get the required props
+
 // $ExpectError
-<Timely name='John' />; //- Timely didn't get the required props
+<Timely name='John' />; // Timely didn't get the required props
+
 // $ExpectError
 <Timely name='John' excited={true} />; // Timely didn't get the required props
+
 
 const x = <Timely date={new Date()} name='John' excited={true} />
 // Only way to satisfy all prop requirements.
 type Injected = {date: Date}
 declare function injectDate<DefaultProps, Props, State, C: React$Component<DefaultProps, Props, State>>(
   Komponent: Class<C>
-): Class<React$Component<DefaultProps, $Diff<Props, Injected>, void>>;
+): Class<React$Component<void, $Diff<Props, {date: Date}>, void>>;
 
 const Timeless = injectDate(Timely);
+// $ExpectError
+<Timeless />; // props not satisfied.
 
 // $ExpectError
-<Timeless />; // -- Error. props not satisfied.
+<Timeless excited={true} />; // props not satisfied.
+
 // $ExpectError
-<Timeless excited={true} />; // -- Error. props not satisfied.
+<Timeless name='Sally' />; // props not satisfied.
+
 // $ExpectError
-<Timeless name='Sally' />; // Error. props not satisfied.
-// $ExpectError
-<Timeless name={1234} excited={true} />; // -- Error. name must be a string
+<Timeless name={1234} excited={true} />; // name must be a string
+
 
 <Timeless name='Sally' excited={true} />;
-// No Errors
 
-<Timeless name='Sally' excited={true} date={1234} />; // -- Error, date must still be a Date object
+// $ExpectError
+<Timeless name='Sally' excited={true} date={1234} /> // date must still be a Date object
 
 /*
   If you want to support stateless functions with your HOCs, the type definition gets
@@ -387,7 +397,7 @@ declare var injectDatesWithStatelessSupport: (
 /*
   Here, we essentially overloaded the definition for the HOC with the types for stateless functions.
   This is neccessary as statelss component don't have defaultProps, while classes do.
-  
+
   As the type definitions for Higher-Order-Components can often get complex, it's usually best to
   to their types in a file with the `.flow` extension next to your `.js` source file.
 
--- a/website/docs/ref/react.doc.js
+++ b/website/docs/ref/react.doc.js
@@ -197,6 +197,28 @@ function renderButton(container: HTMLElement, visited?: boolean) {
 
   Any function that returns a React element can be used as a component class in
   a JSX expression.
+
+*/
+const AsyncGreeter = loadAsync(Greeter, new Promise((resolve, reject) => {
+  setTimeout(() => {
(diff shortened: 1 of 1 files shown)


website/docs/ref/react.doc.js
```diff
@@ -246,3 +255,86 @@ class HOC<Config> extends React.Component<void, Props<Config>, State<Config>> {
     }
   }
 }
+
+/*
+  ## Higher-Order-Components II: React Class Decrators
```
ryyppy: Typo: React Class Dec`o`rators

website/docs/ref/react.doc.js
```diff
@@ -246,3 +255,86 @@ class HOC<Config> extends React.Component<void, Props<Config>, State<Config>> {
...
+  Use at your own risk.
+
+  Higher-Order-Components are often also functions that accept a ReactClass,
+  wrap it in another ReactClass and return the new ReactClass.
+
+  In such cases, the newly created ReactClass will usually accept the same `props`
+  as the original Component, but it may require a sub-set of the original props,
+  as it my inject some on it's own.
+
+  Here we will see a Class Decrator that will automatically inject the current Date
```
ryyppy: Same here ;-)

website/docs/ref/react.doc.js
```diff
@@ -246,3 +255,86 @@ class HOC<Config> extends React.Component<void, Props<Config>, State<Config>> {
...
+<Timeless name={1234} excited={true} />; // name must be a string
+
+
+<Timeless name='Sally' excited={true} />;
+
+// $ExpectError
+<Timeless name='Sally' excited={true} date={1234} /> // date must still be a Date object
+
+/*
+  As of right now, you will often have to use a declaration to delare the type of a
```
ryyppy: Typo: use a declaration to de`c`lare

website/docs/ref/react.doc.js
```diff
@@ -199,6 +199,15 @@ function renderButton(container: HTMLElement, visited?: boolean) {
...
 /*
   ## Higher-order components
+
+  Higher-Order-Components are a slightly overloaded term.
+
+  Firstly, there's React Components that accept other React Components as a prop.
+  In this example we're going to see a Component that manages the data for another component
+  using `state`.
+
+  Here the HOC componen's state a key `config` which happens to be the same value as the one
```
ryyppy: Typo: the HOC componen`t`'s state...

ryyppy:
Cool write-up 👍

jeffmo:
@facebook-github-bot import

acdlite:
Does this account for functional components as well as class components?

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/285558625118705/int_phab) to review.

nmn:
As of right now it only accounts for class components. But I can add another commit for functional components.

I've been trying to find a way to make it work without needing overloads

jeffmo:
cc @samwgoldman who has some thoughts about future direction here as well

nmn:
So I updated to flow 0.27, and my previous trick of using overloads for functional components no longer works. So I'm not sure what the way forward is anymore.

website/docs/ref/react.doc.js
```diff
@@ -246,3 +255,86 @@ class HOC<Config> extends React.Component<void, Props<Config>, State<Config>> {
...
+<Timeless name='Sally' excited={true} />;
+
+// $ExpectError
+<Timeless name='Sally' excited={true} date={1234} /> // date must still be a Date object
+
+/*
+  As of right now, you will often have to use a declaration to declare the type of a
+  higher order function.
+
+  Use a `.flow` file next to the files where you define youre Higher-Order-Components
```
valscion: Typo here: "youre" should be "your"

nmn:
So, I updated to Flow 0.28 and tried my old trick of overloading function types to support stateless components.

First when I wrote all the types altogether it still didn't work. But when I used `type Name =` syntax to define the cases separately, and later used their intersection everything worked.

So, finally, the stateless components documentation has been added.

I wonder if this will ever be merged!

minedeljkovic:
Thank you @nmn for this! This is extremely helpful for project I'm working on, because it heavily uses flow, but also stateless components and HOC's (primarily for pure rendering).

However, I came to one problem with using this solution. When results of HOC's are exported and used from another module, flow checks when class components are used, but doesn't check when stateless components are used.
What is even more curious, when result of HOC wrapping stateless component is used in module where it is defined, flow suddenly checks in module that is importing it!

I set up repo to reproduce it:
https://github.com/minedeljkovic/flow-hoc-export-problem

The curious case is marked with comments [here](https://github.com/minedeljkovic/flow-hoc-export-problem/blob/master/useTimeless.js#L9) and [here](https://github.com/minedeljkovic/flow-hoc-export-problem/blob/master/timeless.js#L55)

Is this some problem with flow itself, or with usage of proposed solution?

nmn:
@minedeljkovic Correct me if you're wrong, but you're saying that flow doesn't check Stateless Components. But if you wrap the Stateless Component in an HOC, it suddenly starts working?

Yes, that's a problem in flow itself. For some reason, Stateless components don't work in Flow right now. However, when you wrap the Stateless Component in an HOC, the type says that the exported is a class. So even if your HOC actually exports another stateless function, the types I've written tricks Flow into thinking that it exports a React Class. And so it starts working.

minedeljkovic:
No, @nmn . Stateless Component checks, but stateless component wrapped in HOC doesn't check (used from another module, of course).

```js
import { TimelessStateless, TimelessClass, TimelyStateless } from './timeless';
...
const err = <TimelessClass />; // Class component wrapped in HOC checks
...
const err1 = <TimelessStateless />; // This should error! Stateless component wrapped in HOC doesn't check, but it does if TimelessStateless is used in timeless.js!
...
const err2 = <TimelyStateless />; // Simple stateless component checks
```

I updated problem repo https://github.com/minedeljkovic/flow-hoc-export-problem
You can easily clone it and reproduce it with `flow check`, if interested.

I am also wondering if this is well defined case to fill and independent issue...

nmn:
thanks. I'll check it out.

website/docs/ref/react.doc.js
```diff
@@ -285,10 +307,105 @@ function loadAsync<Config>(
...
+) => Class<React$Component<D, $Diff<P, Injected>, any>>;
+
+type InjectDateStateless = <P>(component: (props: P) => any) => Class<React$Component<void, $Diff<P, Injected>, any>>;
+
+declare var injectDateStateless: InjectDateStateless & InjectDateToClass;
+
+
+/*
+  Here, we essentially overloaded the definition for the HOC with the types for stateless functions.
+  This is neccessary as statelss component don't have defaultProps, while classes do.
```
Kovensky: [SFC can have defaultProps](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions).

nmn: the .propTypes don't matter. But it is possible to adapt the types for default Props. I chose to leave it out for simplicity as I think most developers don't use that features with stateless functions. I can re-visit this and see if I can make it work.

website/docs/ref/react.doc.js
```diff
@@ -197,6 +197,28 @@ function renderButton(container: HTMLElement, visited?: boolean) {
 
   Any function that returns a React element can be used as a component class in
   a JSX expression.
+
+*/
+const AsyncGreeter = loadAsync(Greeter, new Promise((resolve, reject) => {
```
thejameskyle: This `loadAysnc` function is a bit mysterious and confusing. Is there a way we can make it a bit more clear what this is an example of?

nmn: Yeah, I was trying to come up with a common HOC example. I'll think of a better example.

Also, the types are a little weird when trying to also support stateless functions. I'm confused if I should focus on simplicity or resilience.

thejameskyle: What do you find weird about it?

I would hope the tradeoff isn't drastic between the two that it could be simple and type safe

nmn: The weird thing is:
1. The `ReactClass` types don't work in practice. They don't actually infer types correctly.
2. If I try to make my own type definition for both classes and functions the HOC types don't work.
3. So, I HAVE to use overloads to support both classes and functions.

But then functions may or may not have defaultProps. And again, if I try to combine both into one type, the types don't quite flow correctly. This means that there needs to be one MORE overload.

4. So with three function overloads for a single HOC, the types get quite hairy. This is why I started with just the case for class.

thejameskyle: @nmn Can you create a separate issue for that if one does not already exist and link it here? We should discuss that on its own

gcanti:
@minedeljkovic this should work (tested in your flow-hoc-export-problem repo)

```js
type StatelessComponent<P> = (props: P) => ?React$Element<any>;
type ComponentClass<D, P, S> = Class<React$Component<D, P, S>>;
type Component<D, P, S> = ComponentClass<D, P, S> | StatelessComponent<P>;

type TimelyProps = {
  date: Date,
  name: string,
  excited: boolean
};

type OwnTimelyProps = {
  name: string,
  excited: boolean
};

export function injectDate<D, S>(C: Component<D, TimelyProps, S>): StatelessComponent<OwnTimelyProps> {
  return (op) => {
    const p = Object.assign({}, { date: new Date() }, op)
    return <C {...p} />
  }
}

export const TimelessClass = injectDate(TimelyClass);
export const TimelessStateless = injectDate(TimelyStateless);
```

gabelevi:
I'm making a pass at merging or closing older PRs and I came across this PR. I'm having a bit of trouble rebasing it and making sense of it. It seems to try to split the HOC documentations into two parts

1. Document HOCs which are React components that take other React components as props
2. Document HOCs which are functions, that take 1 or more React components and return a React component.

As it stands, the current documentation only contains an example of 2. This PR, seems to add a second example for 2. So now we have no examples for 1 and two examples for 2. 

So the main goal of this PR is to add an example of a complicated HOC that always provides a particular required property. So I think I'm going to trim this PR down to just adding that example to the docs. Hope that is okay!

PR #1822, opened 2016-05-20
Stop GitHub forks from spamming IRC and email

gabelevi:
Based on a trick found here: https://github.com/travis-ci/travis-ci/issues/5063#issue-115477841

Long story short, `travis encrypt -r "facebook/flow" "flowtype@fb.com"` creates a value that is only decrypted properly in the `facebook/flow` repo. This way, the email and IRC notifications only work for `facebook/flow` and not for the GitHub forks.


gabelevi:
@facebook-github-bot import

PR #1823, opened 2016-05-20
Class_sig refactor

popham:
Functorized Class_sig. See `src/typing/classy.mli` for the interface. For a top down view, start at https://github.com/popham/flow/blob/class-sig/src/typing/classy.ml#L13-L152. ~~I've pulled the patch out of another work-in-progress, so there may be a couple of overly-abstracted spots (specifically, https://github.com/popham/flow/blob/class-sig/src/typing/class/iface_sig.ml#L129 and https://github.com/popham/flow/blob/class-sig/src/typing/class/decl_class_sig.ml#L62-L63 come to mind).~~


samwgoldman:
What motivated this refactoring? It will be easier to review if you share the benefits of this approach.

popham:
* Separating declared classes from interfaces. The declared class case, `decl_class_sig`, reaches out to `class_sig` or `iface_sig` to assemble its behavior (instead of using the `structural` flag).
* Higher-abstraction-level functions with less than 20 lines of code make it easier to get a top-down view of the code. I think this is much easier to grok. (https://github.com/popham/flow/blob/class-sig/src/typing/classy.ml#L43-L50, https://github.com/popham/flow/blob/class-sig/src/typing/class/iface_sig.ml#L124-L135, https://github.com/popham/flow/blob/class-sig/src/typing/class/decl_class_sig.ml#L57-L69, https://github.com/popham/flow/blob/class-sig/src/typing/class/class_sig.ml#L124-L138, etc.)
* Less surface area on the interface--5 functions versus 14 functions.
* Functions with only a single call site have been nested within their callers.

In my work in progress (PR #1775), I needed finer control (and I didn't want to use the `structural` flag), so this structure emerged. I think the "structural" name is just an "is_interface" flag--yuck. Another approach to eliminating the flag would be another data structure. That would introduce more interface surface area, though; I think that this is preferable. I decided to rebase my PR #1775 on top of the refactoring, and I figured it would take a little extra effort to create a tidy, separate PR. I vaguely remember somebody advising me that separate PRs of smaller scope were more likely to get pulled :).

popham:
Replaced last remnants of the `structural`/`is_interface` flags with modules (https://github.com/popham/flow/blob/a44029c80c5056d533dc57a3cff4cb8cb272e77a/src/typing/statement.ml#L43-L84)

popham:
Was setup for PR #1775

PR #1830, opened 2016-05-22
Apply `--strip-root` to verbose info

popham:
I've been diffing `--verbose` output from the same test run under two branches lately.  The `--strip-root` flag doesn't get applied to verbose output under `__flow`, making for long, noisy lines.  With indented verbose output, line wrapping makes deep flows especially difficult to scan.

Verbose, but not *too* verbose :)

### DIFF
--- a/src/typing/context.ml
+++ b/src/typing/context.ml
@@ -237,3 +237,21 @@ let merge_into cx cx_other =
   set_evaluated cx (IMap.union (evaluated cx_other) (evaluated cx));
   set_globals cx (SSet.union (globals cx_other) (globals cx));
   set_graph cx (IMap.union (graph cx_other) (graph cx))
+
+let dump_reason cx reason = if should_strip_root cx
+  then Reason_js.dump_reason (Reason_js.strip_root (root cx) reason)
+  else Reason_js.dump_reason reason
+
+let string_of_reason cx reason = if should_strip_root cx
+  then Reason_js.string_of_reason (Reason_js.strip_root (root cx) reason)
+  else Reason_js.string_of_reason reason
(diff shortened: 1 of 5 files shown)


mroch:
Thanks!

I wonder if these could be put in Debug_js instead? If not then this seems good.

popham:
I assume you mean the additions to the Context interface.  I'll get them migrated within a couple days.

popham:
Whoops. I removed the PID for my purposes on that branch and forgot to undo. Recomputing....

mroch:
agreed the PID is annoying to diff. I end up doing stuff like

`diff -u <(flow-old check --verbose-indent | perl -p -e 's/\[\d+\]//g') <(flow-new check --verbose-indent | perl -p -e 's/\[\d+\]//g')`

mroch:
@facebook-github-bot shipit

popham:
I was fantasizing about a template argument, but that sounds easier.  Is there a reason that flows to an `OpenT`s don't report ids?

mroch:
my guess is that the verbose output just prints `string_of_ctor` and it doesn't make total sense for that function to print tvar ids.

i think there's a function that's commented out for printing tvars with ids, though.

mroch:
yeah, https://github.com/facebook/flow/blob/32f0b2b4fa6d4ab8b149212bece274e3382f2131/src/typing/type.ml#L1113

popham:
Ah.  The comment says no ids because parallelism => non-deterministic.  I've been using `--max-workers 0`, reasoning that it eliminates the parallelism.  Thanks for the pointer.

Issue #1047, opened 2015-11-06
typeof X subclass


lukescott:
I have something like this:
```javasscript
import Node from "./node";
import Element from "./element"; // subclass of Node

function tester(NodeType: typeof Node) {
// ...
}

tester(Element);
```
I'm getting:
```
import Node from "./node";
^^^^^^^^^^^^^^^^^^^^^^^^^^ Node. This type is incompatible with
tester(Element);
       ^^^^^^^ Element
```

Is this supposed to be that way? If so, is there a way to specify that a subclass is ok?

samwgoldman:
Currently Flow doesn't implement subtyping for classes, only instances. See existing issue #814.

This isn't strictly related to `typeof`. Here's some example code:

```javascript
/* @flow */

class A {}
class B extends A {}

declare var aInstance;
declare var bInstance;

function test1(a: A) {};
function test2(a: typeof aInstance) {}

function test3(aClass: Class<A>) {}
function test4(aClass: typeof A) {}

test1(bInstance);
test2(bInstance);

test3(B); // error
test4(B); // error
```

samwgoldman:
There is at least one good reason not to allow such subtyping: subclasses can have differently shaped constructors.

```javascript
/* @flow */

class A {
  foo: string;
  constructor(foo: string) {
    this.foo = foo;
  }
}

class B extends A {
  constructor(foo: number) {
    super(foo.toFixed());
  }
}
```

Therefore, if we allowed passing `B` into a function expecting a value of type `Class<A>`, it wouldn't be safe to actually construct an instance with the passed in class value.

In some earlier related discussion on IRC, @gabelevi kicked around the idea of tracking subtype compatibility, and allowing this pattern when it would be safe to do so. That sounds kind of complicated to me and I'm not sure if anyone is actually working on that.

lukescott:
I updated the sample code above. In my case I'm trying to create a Repeater class that extends Node or a derivative of Node. The constructor always takes a single `Object`.

PR #1847, opened 2016-05-25
Statics subtyping (RFC: `Newable<.>` to indicate constructor subtyping)

### LINKED ISSUE #836: Given class `A`, `Class<A>` subtyping isn't analogous to `A` subtyping
This smells kinda like #814.

``` js
class A {}
class B extends A {}

var ap: A = new A();
var app: A = new B(); // Okay: instance of B <: instance of A
var Ap: Class<A> = A;
var App: Class<A> = B; // NG?: statics of B <: statics of A

/*
someFile.js:1:7,7: A
someFile.js:7:21,21: B

Found 1 error
*/
```

With ES6 class statics chaining up to `Function` parallel to prototypes chaining up to `Object`, shouldn't `statics of B <: statics of A`?
- This sharpens the distinction between inheritance under Node's `utils.inherits` and ES6's inheritance model (and every other random implementation out there):  [utils.inherits](https://github.com/nodejs/node/blob/75fe7739e6d3dc669435f9148a025ab39f507a21/lib/util.js#L697-L735) uses a `super_` attribute on the constructor function-object to expose the statics ancestry, while ES6 uses prototype chaining (`ctor.__proto__ = theSuperClass` emulates it under Node).
- Constructor refinement would bring some complexity, e.g. `new this(...)` within a static method (but this currently doesn't work [#822]).

### LINKED ISSUE #1704: Class instances flow to interface constraints without validating statics

### LINKED ISSUE #1705: Class<> qualified interface constraints do not accept compatible classes

### LINKED ISSUE #1852: Can't create a class within class methods
Flow chokes on the following:

``` js
class C {
  static Fn(): Class<C> {
    return class D extends C {}
  }
  Gn(): Class<C> {
    return class E extends C {}
  }
}
```

The pattern is especially interesting when you start intersecting the `Class<C>`s with interfaces.

popham:
* Fixes static subtyping on annotations (see `tests/class_type/test2.js`--it lights up like Christmas under master).  Fixes #1705
* Imposes subtyping on interface statics.  They should be banished at the parser level or subtyped, but currently they just sit there.  Fixes #1704 
* Allows functions to return class extensions.  Fixes #1852
* Distributes ConstructorT across intersections.
* Distributes `constructor` lookups across intersections.
* Excludes `Class<X>~>Interface` flows (`Class<X>`~>`Class<Interface>` *are* allowed).
* Closes #836, #1047, #1893

**RFC:**
I feel that the sticky point here is the lack of subtyping on class constructors.  Suppose class `B` extends class `A`, but `B` implements a constructor that doesn't subtype that of `A`.

```js
var K: Class<A> = B; //OK under this PR, NG under master
var a = new K(...);
```

Working to the signature of `A` within `...` implies a false positive, while working to the signature of `B` within `...` implies a false negative.  This same problem shows up in class methods that `new this.constructor` and static class methods that `new this`.  I think that this is easily handled for implicit annotations with a `newable` flag on `insttype`, but for explicit annotations the annotation's syntax would probably be ugly.

* In the example above, `newable=true` is necessary for 90% of use cases.  (Hence the unification rule within `ClassT~>ClassT`?)
* In class methods with an `a: A`, `newable=false` is sufficiently strong in 99% of use cases (1% generously for `new a.constructor`).
* In general for `a: A`, `newable=false` is sufficiently strong in 99% of use cases.
* In class methods with a `K: Class<A>`, `newable=false` is sufficiently strong in 50% of use cases(?).
* (My thinking here is biased toward a program that has been explicitly annotated in its entirety.  Do these numbers shift profoundly in minimally annotated programs?)

Does there currently exist a plan and/or demand to tighten things up here?  My quick and dirty thinking is the `newable` flag on `insttype` and a `Newable<.>` wrapper for explicit annotations that impose constructor subtyping in addition to the norm:

```js
var K: Class<Newable<A>> = B; //NG: constructor subtyping fails
var a = new K(...);
```

I think that this is preferable to some pretty-sounding variation of `NonNewable` because of the prevalence of the `a: A` use case.

(And currently, `Class<.>` annotations are kinda useless, aren't they?  If the annotation necessarily contains the class type, then why shuffle it around?  Why provide it as a parameter?  I thought ES6 was spec'd to avoid circular importing problems, eliminating that possibility(?).  I suspect that post-transpile, broken circular-imports may be the only credible use case.)


popham:
Currently chasing:

```js
interface I {
  anotherMethod(): number;
}

interface J {
  static anotherStaticMethod(): number;
}

class MS {
  anotherMethod(): number {
    return 3;
  }
  static anotherStaticMethod(): number {
    return 4;
  }
}

class A {
  static sm3(): Class<I> & Class<J> {
    return MS;
  }
}

var Y = A.sm3();
var y = new Y();
var n6: number = y.constructor.anotherStaticMethod(); 
```

`n6` should be okay but I get

```
 28: var n6: number = y.constructor.anotherStaticMethod();
                                    ^^^^^^^^^^^^^^^^^^^ property `anotherStaticMethod`. Property not found in
 28: var n6: number = y.constructor.anotherStaticMethod();
                      ^ statics of I
```

Smells like a leak in `lower_concretize...`.

popham:
I thought interfaces should forbid `constructor`s, but I want a particular use case:

```
class A{}
interface Extensions {}
function extender(): Class<A> & Class<Extensions> {
  return class E extends A {
    // satisfy `Extensions` interface
  }
}
```

Classes created by `extender` should export fine, where each has its class id erased, but is still pretty tightly typed.  The problem arises, however, what if I want `Extensions` to impose the constructor?  Interfaces need constructors :(.  My new thought is a `NewableClass<>` peer to `Class<>` and a `insttype.newable` flag (in a different PR).

src/typing/flow_js.ml
```diff
@@ -1848,6 +1848,24 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
         (List.hd ts,
          LookupT (reason, strict, (List.tl ts) @ try_ts_on_failure, s, t))
 
+    | IntersectionT (r, rep), GetPropT (reason_op, (reason_prop, "constructor"), t) ->
```
popham: Generalize with an `is_set_op_distributive` gate on `u`, and handle UnionT too.

gabelevi:
@popham - I'm making an effort to clean out our oldest pull requests. What's the status of this PR? Are you waiting for feedback? Planning some more changes? Or is this abandoned?

By default, I'm closing out old PRs that I can't merge. This does not reflect the quality of the PR. Please feel free to re-open this if you think it should still be open.

Issue #1340, opened 2016-01-30
Can't import variable named `of`


rpominov:
Not sure if this by spec or a bug, but I couldn't find any mention of `of` being somehow special case in import.

```
4: import {of} from 'foo'
           ^^ Unexpected token of
```

codemonkey800:
`of` is a reserved keyword for ES6. It's called a `for-of` loop It's used for iterating through the values in an iterable object:

```
const stuff = [1, 2, 3, 4, 5];
for (const value of stuff) {
  console.log(value);
}
```

You can read more about it here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of :smiley:

michaelficarra:
@codemonkey800 No it's not.

codemonkey800:
@michaelficarra Hmm, you're right. My mistake. Just went on MDN and confirmed it's not. Not sure what the issue is then.

samwgoldman:
Yup, looks like we're overstrict on this. Thanks for the bug report!

samwgoldman:
(By the way, I think the fantasyland project is great—is there an effort to integrate Flow? If so, keep me in the loop. I think it's a good motivating use case for us.)

rpominov:
@samwgoldman Unfortunately we accidently have made that integration harder by trying to fix name clashes issue (see 2. in https://github.com/fantasyland/fantasy-land/issues/122#issuecomment-177134562)

Issue #1233, opened 2015-12-31
Flow treats `async` as a reserved word—is it?


Gozala:
Here is an example:

## `scratch.js`
```js
/* @flow */

import {succeed, async} from "./src/task"

```

## `stdout`

```
scratch.js:3
  3: import {succeed, async} from "./src/task"
                      ^^^^^ Unexpected token async
```

AgentME:
ES2015 has `await` as a reserved word that can't be used as an identifier. http://www.ecma-international.org/ecma-262/6.0/#sec-reserved-words

samwgoldman:
Woah, is it really possible that the Async Functions proposal doesn't carve out `async` as a new reserved word?

@jeffmo I don't see anything in [the proposed spec](http://tc39.github.io/ecmascript-asyncawait/). Missing something?

mroch:
fixed by #1858

PR #1858, opened 2016-05-27
[parser] Allow `type` as an imported binding

nmote:
This also fixes a number of other bugs around reserved words. Notably, both `static` and `implements` are disallowed in strict mode:

http://www.ecma-international.org/ecma-262/6.0/#sec-keywords

### DIFF
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -1782,7 +1782,14 @@ end = struct
       let lex_value = Peek.value env in
       let lex_loc = Peek.loc env in
       match lex_token with
-      | T_IDENTIFIER -> Parse.identifier env, None
+      (* Anything that is a special token in Flow but not in the ES6 spec
+         should be here. *)
+      | T_ASYNC
+      | T_DECLARE
+      | T_IDENTIFIER
+      | T_OF
+      | T_TYPE
+        -> Parse.identifier env, None
       | _ ->
         let err = match lex_token with
         | T_FUNCTION
@@ -1806,7 +1813,6 @@ end = struct
         | T_CASE
         | T_CATCH
         | T_CONTINUE
-        | T_DECLARE
         | T_DEFAULT
         | T_DO
         | T_FINALLY
@@ -1830,14 +1836,11 @@ end = struct
         | T_PROTECTED
         | T_PUBLIC
         | T_YIELD
-        | T_TYPE
-        | T_OF
         | T_ANY_TYPE
         | T_BOOLEAN_TYPE
         | T_NUMBER_TYPE
         | T_STRING_TYPE
         | T_VOID_TYPE
-        | T_ASYNC
         | T_AWAIT
         | T_DEBUGGER ->
             Some (lex_loc, get_unexpected_error (lex_token, lex_value))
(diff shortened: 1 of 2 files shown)


mroch:
I'm not sure about this...

[`ClassElement`](http://www.ecma-international.org/ecma-262/6.0/#sec-class-definitions) -> [`MethodDefinition`](http://www.ecma-international.org/ecma-262/6.0/#sec-method-definitions) -> [`PropertyName`](http://www.ecma-international.org/ecma-262/6.0/#sec-object-initializer) -> `LiteralPropertyName` -> [`IdentifierName`](http://www.ecma-international.org/ecma-262/6.0/#sec-names-and-keywords), which _does_ allow reserved words.

maybe blame those tests? seems like it's intentionally testing that they work.

nmote:
Huh, I'll look into it more but maybe it's a bug to use `identifier_or_reserved_keyword` in both of these contexts? It seems that in an imported binding, `static` and `implements` are disallowed, but in a class property they are allowed? Is that your understanding?

mroch:
that'd be

[`ImportClause`](http://www.ecma-international.org/ecma-262/6.0/#sec-imports) -> `NamedImports` -> `ImportsList` -> `ImportSpecifier` -> `ImportedBinding` [1] -> [`BindingIdentifier`](http://www.ecma-international.org/ecma-262/6.0/#sec-identifiers) -> `Identifier` -> `IdentifierName` but not `ReservedWord`

[1]: but if using `as` then it's `IdentifierName as ImportedBinding` which would allow reserved words before the `as` but not after

so yep, I think you're right, they're illegal in `import { HERE } from ...` but legal in classes and `import { HERE as NOT HERE } from ...`

src/parser/parser_flow.ml
```diff
@@ -1938,7 +1938,7 @@ end = struct
     and identifier_or_reserved_keyword env =
       let { lex_token; lex_value; lex_loc; _ } = lookahead env in
       match lex_token with
-      | T_IDENTIFIER -> Parse.identifier env, None
+      | _ when Peek.identifier env -> Parse.identifier env, None
```
gabelevi: Hmm, maybe this should be

```JavaScript
| T_IDENTIFIER -> Parse.identifier env, None
| _ when Peek.identifier env ->          
    (lex_loc, Identifier.({
      name = lex_value;
      typeAnnotation = None;
      optional = false;
    })), None
```

Since I think `identifier_or_reserved_keyword` doesn't want to error here.

gabelevi: Probably could even do without the `T_IDENTIFIER ` case I suppose

gabelevi: Actually ignore me I think I'm saying dumb things. :P

mroch:
this probably fixes https://github.com/facebook/flow/issues/1233, can you add tests for those keywords too?

nmote:
Okay, how about this?

nmote:
@mroch, just updated to add the requested tests.

mroch:
travis is happy so i'm happy!

mroch:
@facebook-github-bot shipit

facebook-github-bot-1:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1012379435483288/int_phab) to review.

mroch:
Fixes https://github.com/facebook/flow/issues/1340

Issue #1870, opened 2016-05-31
Analysis does not match fs.readFileSync declaration


bryannaegele:
https://github.com/facebook/flow/blob/master/lib/node.js#L634 shows fs.readFileSync correctly returning type string when an encoding is passed.

Given this declaration:
```js
declare module 'html-frontmatter' {
  declare var exports: (input: string) => {};
}
```

and this implementation:
```js
import fs from 'fs'
import path from 'path'
import frontMatter from 'front-matter'
import objectAssign from 'object-assign'
import htmlFrontMatter from 'html-frontmatter'

let data
...
const html = fs.readFileSync(pagePath, 'utf-8')
data = objectAssign({}, htmlFrontMatter(html), { body: html })
```

Flow check raises this error
```
22:     data = objectAssign({}, htmlFrontMatter(html), { body: html })
                                 ^^^^^^^^^^^^^^^^^^^^^ function call
 22:     data = objectAssign({}, htmlFrontMatter(html), { body: html })
                                                 ^^^^ Buffer. This type is incompatible with
  2:   declare var exports: (input: string) => {};
                                    ^^^^^^ string. See lib: interfaces/html-frontmatter.js:2
```

and running flow suggest for the file resolves Buffer, as well.
```
-    const html = fs.readFileSync(pagePath, 'utf-8')
+    const html: Buffer = fs.readFileSync(pagePath, 'utf-8')
```

Are overloads working correctly?

Node 6.2.0
Flow 0.26

mroch:
Overloads are tried in order and the first one to match wins. Since extra params are also allowed, that first one will match and the second one won't. :/

I think reversing the order of the overloads in lib/node.js will help. I can try it when I get into the office.

bryannaegele:
Thanks for the insight and tackling that @mroch.

bryannaegele:
@mroch I'm still seeing this issue with v0.27.
 
I also tried the non-shorthand object options to readFileSync `fs.readFileSync(pagePath, { encoding: 'utf-8' })` but saw the same result.

pitops:
hey how was this resolved? its 2020 and have the same issue.

PR #1874, opened 2016-05-31
Fix type of fs.readFileSync

mroch:
Fixes https://github.com/facebook/flow/issues/1870

since the first override had fewer params, it would accept any of the latter branches when the latter branches' return types didn't match, rather than making the entire intersection fail.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -630,10 +630,10 @@ declare module "fs" {
     options: { flag?: string },
     callback: (err: ?Error, data: Buffer) => void
   ): void;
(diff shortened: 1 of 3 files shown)


mroch:
@facebook-github-bot import

facebook-github-bot-9:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/582952658547840/int_phab) to review.

bryannaegele:
Thanks for tackling that @mroch and thanks for the insight on how flow evaluates and determines precedence.

PR #1892, opened 2016-06-03
Use the 4.03.0 input_value workaround everywhere

gabelevi:
I noticed that some Travis runs ([like this one](https://travis-ci.org/gabelevi/flow/builds/134900465)) were failing due to this bug in 4.03.0. We had a workaround in one place, but I've moved it up into `Timeout.input_value`, which is used basically everywhere. Hopefully this should fix the race-condition bad behavior that the test was noticing.

### DIFF
--- a/hack/procs/worker.ml
+++ b/hack/procs/worker.ml
@@ -133,16 +133,7 @@ let slave_main ic oc =
     Daemon.output_string oc s;
     Daemon.flush oc in
   try
-    let Request do_process =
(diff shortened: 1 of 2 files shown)


gabelevi:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1012569902112046/int_phab) to review.

PR #1894, opened 2016-06-03
ES6 class statics

popham:
ES6 classes inherit static fields and methods, so subtyping should apply to statics too.


tests/class_statics/class_statics.exp
```diff
@@ -1,9 +1,3 @@
-test.js:1
-  1: class A {
-           ^ A. This type is incompatible with
- 48: (B: typeof A);
-      ^ B
```
popham: This cast is safe since `B` is a subclass of `A`

PR #1897, opened 2016-06-05
Update KeyboardEvent.code to be a a string

nsfmc:
KeyboardEvent.code is defined as a DomString whose

> unitialized value must be "" (empty string)

so this changes KeyboardEvent.code from `number` to `string`.

see also [mdn's `KeyboardEvent.code` page](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -202,7 +202,7 @@ declare class MessageEvent extends Event {
 
 declare class KeyboardEvent extends UIEvent {
   altKey: boolean;
-  code: number;
+  code: string;
   ctrlKey: boolean;
   isComposing: boolean;
   key: string;


nsfmc:
Hi there, i'm also not sure if this change is exhaustive (it was the only place i found the attribute, but if i've missed something i'm happy to touch up the pr)

mroch:
I don't know the history here but I agree with your interpretation. The spec definitely says it's a string, and it looks like it has always been that way. [`keyCode`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) is the one that is/was a number (and is deprecated).

mroch:
@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1722398941332873/int_phab) to review.

PR #1903, opened 2016-06-06
Implement server-side coverage command

nmote:
Running `flow coverage` on the file below, in the Nuclide codebase, occupies the Flow server for two minutes on my machine:

`Nuclide/pkg/nuclide-reprint-js/lib/printers/complex/printLiteral.js`

This is because coverage is implemented using the `dump-types` server command, which forces all of the types to be stringified and sent across the process boundary. The file above, for whatever reason, has types that are ridiculously long when stringified. To solve this, I've added a new `coverage` command to the server, which simply returns a boolean depending on whether the region is covered or not. With this change, running `flow coverage` on the file above is basically instantaneous, and the output is identical to the output without this change.

I will follow up with another PR to add detailed coverage information to the JSON output, so Nuclide can switch from dump-types to coverage for its coverage display. Then we can get away from this pathological behavior.

Of course, this PR doesn't fix `dump-types`. But it solves my immediate problem and seems like the right long-term call for `coverage` anyway.

### DIFF
--- a/src/commands/coverageCommand.ml
+++ b/src/commands/coverageCommand.ml
@@ -59,10 +59,8 @@ let handle_error ~json (loc, err) strip =
   );
   flush stderr
 
-let is_covered ctor = ctor <> "AnyT" && ctor <> "EmptyT"
-
-let accum_coverage (covered, total) (_loc, ctor, _pretty, _raw, _reasons) =
-  if is_covered ctor
+let accum_coverage (covered, total) (_loc, is_covered) =
+  if is_covered
   then (covered + 1, total + 1)
   else (covered, total + 1)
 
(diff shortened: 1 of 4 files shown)


mroch:
@facebook-github-bot shipit

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1845562442330102/int_phab) to review.

nmote:
FYI, I just ran `coverage` over all the files in the Nuclide `pkg` directory and summed up the results, and they were the same both before and after this command (although the before took about 2 hours to run, and after took about 2 minutes):

```
covered:   202364
uncovered: 19411
percentage: 91.25%
```

PR #1904, opened 2016-06-07
Include uncovered locations in the coverage JSON output

nmote:
Tested in a few cases by manually inspecting the output and by making sure that Nuclide displayed the same results as before for type coverage after I modified it to use this instead of `dump-types`


nmote:
Modified to only compute the uncovered ranges if we are going to use them

mroch:
@facebook-github-bot shipit

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/499073293619253/int_phab) to review.

PR #1916, opened 2016-06-09
Nullable check with class fields lost after first expression

danez:
When doing a nullable check on a class field from within the class like `if(this.field) {}`, the field is only considered to be not null in the first expression within the checked scope (the `if`). All subsequent usages of the field are considered to be potentially null again.

```js
if(this.obj) {
  this.obj.call(); // < works
  this.obj.apply(); // < fails "this.obj might be null/undefined"
}
```

At first I thought it might have to do that within `call()` I could set the field to null, but then the same thing should also not work for simple variables, but it does:

```js
var x:?{ call: Function, apply: Function } = null;

if (x) {
  x.call();
  x.apply();
}
```
Because here I could also potentially set x to null within x.call().

So I'm not sure if this is a bug, but to me it seems it should work as it is pretty common to do this kind of checks.

Added a failing test case.

### DIFF
--- a/tests/class_fields/nullable.js
+++ b/tests/class_fields/nullable.js
@@ -1,0 +1,15 @@
+class NullableCheck {
+
+  obj:?{ call: Function, apply: Function };
+
+  constructor(obj:?{ call: Function, apply: Function }) {
+    this.obj = obj;
+  }
+
+  apply() {
+    if(this.obj) {
+      this.obj.call();
+      this.obj.apply();
+    }
+  }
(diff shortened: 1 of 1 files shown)


gabelevi:
This behavior is expected and is documented here: https://flowtype.org/docs/dynamic-type-tests.html#caveats

Flow is scared that the function call might change the value of `this.obj` back to `null`. Local variables like `x` don't have this problem since the functions don't have access to them.

PR #1917, opened 2016-06-09
Update the declaration of cloneNode to use self instead of Node

### LINKED ISSUE #749: definition of `.cloneNode` is not quite right
Here is my code snippet that run into this error:

```js
class BeforeAppendAttribute extends Attribute {
  write(node/*:Element*/, present/*:AttributeValue*/, past/*:AttributeValue*/) /*:Element*/ {
    return super.write(node.cloneNode(true), present, past);
  }
};
```

This produces following error:

```js
low check                                                                                                                ✱

/Users/gozala/Projects/flowing/flowing.js:51:24,43: call of method cloneNode
Error:
/private/tmp/flow/flowlib_9d9d25d/dom.js:114:1,166:1: Node
/private/tmp/flow/flowlib_9d9d25d/dom.js:338:1,383:1: Element

Found 1 error
```

Presumably because [`cloneNode`](https://github.com/facebook/flow/blob/master/lib/dom.js#L119) is not polymorphic. I'm afraid I don't know how to make it such to submit a fix.

tadeuzagallo:
This fixes #749 - the original solution presented in the issue replaced more signatures of `Node`, but from looking at it I don't think it'd be right to assume that any of the other types are `self` rather than `node`.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -236,7 +236,7 @@ declare class Node extends EventTarget {
   rootNode: Node;
   textContent: string;
   appendChild(newChild: Node): Node;
-  cloneNode(deep?: boolean): Node;
+  cloneNode(deep?: boolean): self;
   compareDocumentPosition(other: Node): number;
   contains(other: ?Node): boolean;
   hasChildNodes(): boolean;


gabelevi:
Thanks for the contribution! I think the type is named `this`, but I can tweak this PR when I import it.

gabelevi:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1063059697095446/int_phab) to review.

PR #1927, opened 2016-06-10
fix ocaml 4.03.0 compile

marudor:
I've add the warn_error_d tag - this removes deprecations as fatal warning.
Quite some deprecations were added in one of the latest commits, breaking compile for newer ocaml versions.
This way the warnings will still show up but will not prevent compilation.

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,4 +1,4 @@
-<**/*.ml*>: ocaml, warn_A, warn(-4-6-29-35-44-48-50), warn_error_A
+<**/*.ml*>: ocaml, warn_A, warn(-4-6-29-35-44-48-50), warn_error_A, warn_error_d
 <hack/heap/*.ml*>: warn(-27-34)
 <hack/third-party/core/result.ml>: warn(-41)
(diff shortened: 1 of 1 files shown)


mroch:
interesting, cool idea! didn't know you could do that.

our internal build system will have to be updated to use the same flags, so it might take a bit to merge this.

I'm removing the use of `Char.uppercase` in the meantime... sounds like there might be more deprecations to contend with after that, though?

mroch:
@facebook-github-bot import

marudor:
Yep, most Char functions got changed to x_ascii instead of x.
And changing it to the _ascii method will bump up min ocaml version to 4.03.0

mroch:
i wonder if someone will make a `base-char` opam module like `base-bytes`

mroch:
i looked into adding this to our internal build system ([buck](http://buckbuild.com)) but the concern is that we would want to fix any warnings anyway rather than letting them create log spew, so they might as well error.

this is a good tip, though. thanks!

PR #1929, opened 2016-06-10
WIP: Newable<.> operator for constructor subtyping

popham:
Adds a `Newable<.>` wrapper that enforces constructor subtyping. With constructor subtyping, `new k.constructor` calls work properly across module boundaries (`k:K` => ng `new k.constructor` and `k:Newable<K>` => ok `new k.constructor`). Currently, `Class<.>` indicates newable--this PR will break these uses, requiring a `Class<Newable<.>>` or `Newable<Class<.>>` instead.

* I consider intersection/union distributivity outside of scope.
* I consider statics subtyping outside of scope (although trivial).
* The open problem that I'm working at for my next commit is the `this` type within class methods (the syntax from my [`this` pseudo-parameter PR](https://github.com/facebook/flow/pull/1775) was anticipating this).
  - If there exists a `new this.constructor` in a method or a `new this` in a static method, then the dispatching instance/class must satisfy constructor subtyping.
  - Introducing a non-subtyping constructor in a class implies that all of the inherited methods with `new this.constructor`/`new this` must be overridden.
* The non-newable~>newable errors are still rough. Grep "error gets blocked".
* The *~>ImportTypeT and *~>ImportTypeofT cases currently do not transition NewableUndefined to NewableFalse for exported newables. The corresponding tests belong in `newable/import.js`.


popham:
I'm abandoning this line of inquiry. Even if I scrape method bodies for `new this.constructor` calls and impose an is-newable requirement on dispatching instances, the system doesn't close cleanly around interfaces and class declarations. Consider

```js
interface I {
  a(): this;
}
class A {
  a(): this {
    // Scrape method body to determine that the dispatching instance must be newable
    return new this.constructor;
  }
}
class B {
  a(): this {
    // Non-newable dispatching instance is sufficient
    return this;
  }
}
function fn(i: I): I {
  // Is `this` on `I.prototype.a` newable or non-newable?
  return i.a(); 
}
```

This strategy doesn't provide sufficient granularity to impose constraints on the dispatching instance. I'm left reverting to my `this`-pseudo-param syntax so that I can specify `a(this: Newable<this>): this` or `a(this: this): this`.

PR #1930, opened 2016-06-11
Introduce intersection operator in docs with object types

popham:
This comes up kinda frequently in IRC....

* The intersection operator's docs have focused on a function declaration case. I feel that this is a 10%-or-less use case, so I've refocused the intersection's first introduction on two object types (IMO the 90%-or-more use case). The function case is left as an example, and the operator's interaction with functions is relegated to a note (no variations of the word "contravariance" are used).
* The Objects documentation contains a section titled "Reusable Object Types." I've mentioned there that object types can be added by intersection, linking to the "Union and Intersection Types" page.
* The type alias concept makes the exposition cleaner, separating the union type's definition from its use. I've moved the type alias introduction prior to the intersection and union type introduction so that the type alias is a familiar concept to the reader.

### DIFF
--- a/website/_data/nav_docs.yml
+++ b/website/_data/nav_docs.yml
@@ -44,10 +44,10 @@
     title: Maybe Types
   - id: destructuring
     title: Destructuring
-  - id: union-intersection-types
-    title: Union and Intersection Types
   - id: type-aliases
     title: Type Aliases
+  - id: union-intersection-types
+    title: Union and Intersection Types
   - id: typeof
     title: Typeof types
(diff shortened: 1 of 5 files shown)


gabelevi:
Thanks for the contribution! Sorry for the delay...I'll get this reviewed ASAP!

facebook-github-bot-1:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/487840648072504/int_phab) to review.

PR #1941, opened 2016-06-14
Avoid frozen build when `git` or `.git/`+`hg` are unavailable

popham:
This came up in IRC earlier. An [unofficial package](https://aur.archlinux.org/packages/flow/) maintainer for Archlinux had his/her build freezing at `ocaml -I scripts -w -3 unix.cma scripts/gen_build_id.ml _build/hack/utils/get_build_id.gen.c`. The build is from a tarball, so the `.git/` directory is unavailable.

The user didn't have Mercurial installed. After the `git` case under `gen_build_id.ml` failed, a read under the `hg` case from stdout/stderr would not stop blocking. It looks like `create_process` doesn't recognize exit 127 when `hg` isn't found, so an `input_line` call doesn't break. This PR substitutes the `open_process_full` function which recognizes that the command doesn't exist and interrupts the `input_line` call. I suspect that the same problem arises in environments without `git` installed regardless of whether `hg` is installed.

### DIFF
--- a/scripts/gen_build_id.ml
+++ b/scripts/gen_build_id.ml
@@ -13,24 +13,26 @@
 let () =
   let out_file = Sys.argv.(1) in
   let rev =
-    try read_process_output "git" [|"git"; "rev-parse"; "HEAD"|]
+    try read_process_output "git rev-parse HEAD"
     with Failure _ ->
-      try read_process_output "hg" [|"hg"; "id"; "-i"|]
+      try read_process_output "hg id -i"
       with Failure _ -> ""
   in
   let time =
-    try read_process_output "git" [|"git"; "log"; "-1"; "--pretty=tformat:%ct"|]
+    try read_process_output "git log -1 --pretty=tformat:%ct"
     with Failure _ ->
       try
-        let raw = read_process_output "hg" [|"hg"; "log"; "-r"; "."; "-T"; "{date|hgdate}\\n"|] in
+        let raw = read_process_output "hg log -r . -T \"{date|hgdate}\\n\"" in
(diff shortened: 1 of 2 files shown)


mroch:
one constraint here is that it needs to work on Windows. any idea if `Unix.open_process_full` works there?

cc @gabelevi

popham:
No freakin' idea. The `Unix` module is heavily annotated with whether each member works on Windows, and there's no exception listed from Windows, so I assumed. I don't have a Windows box to test on--on Linux I tested permutations:

(.git/ exists | .git/ doesn't exist) X (`hg` exists | `hg` doesn't exist)

(I didn't check (`git` exists | `git` doesn't exist), which I probably should have).

mroch:
@gabelevi can you patch this and find out?

mroch:
One option that should be Windows-safe is to look for the .git and .hg folders before calling those commands. I'd definitely merge that without actually testing on Windows.

popham:
Kinda--I think it would need to check existence of the executables instead of the directories.  

From the [ocaml sources](https://github.com/ocaml/ocaml/blob/0a577fb6799e141a44ef2620469cff6d3b37a889/otherlibs/win32unix/unix.ml#L860), Windows would need a `COMSPEC` environment variable.  That still may be insufficient though.

Let me try polling for program existence without any input reads as a preliminary step. If that doesn't work, I'll just use `Sys.unix`. When I get around to it, that is....

mroch:
i think this works around it: 72759d80a884a6d870cd98792952cb7b0ba66dbb

thanks for figuring out the cause!

PR #1957, opened 2016-06-16
Fix the the ocaml parser test for TravisCI

gabelevi:
Travis was failing since 

* Parser ocaml test was linking with hh_shared.c
* hh_shared.c uses `BuildInfo_kRevision'
* We don't run the get_build_id step for the parser ocaml test


gabelevi:
@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/155327121548999/int_phab) to review.

PR #1977, opened 2016-06-19
Code block-ify Objects as Map

vjeux:
I often want to refer to it since the syntax is not obvious to me but it is not a proper code block so it's harder to find than I hope. Should make it better


vjeux:
Ping @mroch

gabelevi:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/801658696602093/int_phab) to review.

Issue #1932, opened 2016-06-12
Missing nodelib declaration: Buffer.from


passcod:
Looks like [`Buffer.from`](https://nodejs.org/api/buffer.html#buffer_buffer_from_buffer_alloc_and_buffer_allocunsafe) (added in Node 3.0.0 and further signatures in 5.10.0) is missing:

```
index.js:85
 85:       stream.on('data', (chunk) => buflist.push(Buffer.from(chunk)))
                                                            ^^^^ property `from`. Property not found in
 85:       stream.on('data', (chunk) => buflist.push(Buffer.from(chunk)))
                                                     ^^^^^^ statics of Buffer
```

PR #1978, opened 2016-06-19
Add some missing Buffer declarations.

CarlRosell:
By reading the [Buffer documentation](https://nodejs.org/api/buffer.html) I found that some methods were missing.

Added: 
- entries(): Iterator<[number, number]>
- includes(value: string | Buffer | number, offset?: number, encoding?: string): boolean
- indexOf(value: string | Buffer | number, offset?: number, encoding?: string): number
- keys(): Iterator<number>
- lastIndexOf(value: string | Buffer | number, offset?: number, encoding?: string): number
- swap16(): Buffer
- swap32(): Buffer
- values(): Iterator<number>
- static allocUnsafe(size: number): Buffer
- static allocUnsafeSlow(size: number): Buffer
- static from(value: Array<number> | number | string | Buffer, encoding?: buffer$Encoding): Buffer

Fixes: #1932

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -34,7 +34,7 @@ declare class Buffer {
   includes(value: string | Buffer | number, offset?: number, encoding?: string): boolean;
   indexOf(value: string | Buffer | number, offset?: number, encoding?: string): number;
   inspect(): string;
-  keys(): Iterator,
+  keys(): Iterator<number>,
   lastIndexOf(value: string | Buffer | number, offset?: number, encoding?: string): number;
   readDoubleBE(offset: number, noAssert?: boolean): number;
   readDoubleLE(offset: number, noAssert?: boolean): number;
--- a/lib/node.js
+++ b/lib/node.js
@@ -86,7 +86,7 @@ declare class Buffer {
   static byteLength(string: string, encoding?: buffer$Encoding): number;
   static compare(buf1: Buffer, buf2: Buffer): number;
   static concat(list: Array<Buffer>, totalLength?: number): Buffer;
-  static from(value: Array<number> | number | string | Buffer, encoding?: buffer$Encoding): Buffer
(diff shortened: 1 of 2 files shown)


passcod:
What is the status of this?

mroch:
not sure what the bot will do with conflicts but let's see...

mroch:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1748382965438578/int_phab) to review.

lib/node.js
```diff
@@ -25,64 +25,70 @@ declare class Buffer {
...
   static alloc(size: number, fill?: string | number, encoding?: buffer$Encoding): Buffer;
-  static isEncoding(encoding: string): boolean;
-  static isBuffer(obj: any): boolean;
+  static allocUnsafe(size: number): Buffer;
+  static allocUnsafeSlow(size: number): Buffer;
   static byteLength(string: string, encoding?: buffer$Encoding): number;
-  static concat(list: Array<Buffer>, totalLength?: number): Buffer;
   static compare(buf1: Buffer, buf2: Buffer): number;
+  static concat(list: Array<Buffer>, totalLength?: number): Buffer;
+  static from(value: Array<number> | number | string | Buffer, encoding?: buffer$Encoding): Buffer;
```
mroch: i'm tweaking this a little bit while I import it:

- `number` isn't allowed: https://github.com/nodejs/node/blob/0b9e135c3eaf9b83671b1ecbaa1aa4ba700a8add/lib/buffer.js#L130

- `encoding` is only allowed when `value` is a string (except for the `ArrayBuffer` case which we don't implement yet AFAICT)

CarlRosell: Good catch 👍

PR #1983, opened 2016-06-20
node Buffer can be constructed from ArrayBuffer and Uint8Array

philikon:
Also provide declaration for new Buffer.from() which along with Buffer.alloc() is to replace the constructor

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -18,10 +18,13 @@ type buffer$NonBufferEncoding =
   'utf16le' | 'UTF16LE' | 'utf-16le' | 'UTF-16LE';
 type buffer$Encoding = buffer$NonBufferEncoding | 'buffer'
 
+type buffer$Source = Array<number> | string | ArrayBuffer | Uint8Array;
+
 declare class Buffer {
   constructor(
-    value: Array<number> | number | string | Buffer,
-    encoding?: buffer$Encoding
+    value: number | buffer$Source | Buffer,
+    encodingOrByteOffset?: buffer$Encoding | number,
+    length?: number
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -18,10 +18,13 @@ type buffer$NonBufferEncoding =
   'utf16le' | 'UTF16LE' | 'utf-16le' | 'UTF-16LE';
 type buffer$Encoding = buffer$NonBufferEncoding | 'buffer'
 
+type buffer$Source = Array<number> | string | ArrayBuffer | Uint8Array;
```
mroch: both places this is used also do `| Buffer`. does it make sense to add `Buffer` to this union?

mroch:
you'll need to re-record the node tests (or I can do it when I merge if you want)

```
./runtests.sh bin/flow node_tests  # will fail
cp tests/node_tests/*.{out,exp}
./runtests.sh bin/flow node_tests  # will pass now
```

mroch:
just also saw https://github.com/facebook/flow/pull/1978 which adds Buffer.from but not ArrayBuffer/Uint8Array. I'm going to take that one first

lib/node.js
philikon: I guess. I didn't know whether I could reference `Buffer` before it's declaration...

philikon:
> just also saw #1978 which adds Buffer.from but not ArrayBuffer/Uint8Array. I'm going to take that one first

Ah, fair point. I'll rebase and see if I still need to do anything here.

facebook-github-bot-5:

PR #1873, opened 2016-05-31
Add `CustomEvent` class declaration

gf3:
Adds the [`CustomEvent`](https://developer.mozilla.org/en/docs/Web/API/CustomEvent) class

PR #1992, opened 2016-06-23
Added CustomEvent to bom.js

Emile-Filteau:
After using a CustomEvent on a project, we found out it was missing from Flow and we had to declare it ourselves every time. I think it could be a good addition to bom.js.

doc : https://developer.mozilla.org/en/docs/Web/API/CustomEvent

Is there any test for bom.js ?

@mroch

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -223,6 +223,11 @@ declare var NodeFilter: {
     SHOW_DOCUMENT_FRAGMENT: number;
 };
 
+declare class CustomEvent extends Event {
+  static initCustomEvent(type: string, canBubble: boolean, cancelable: boolean, detail: any): CustomEvent;
+  detail: any;
+}
+
 declare class CloseEvent extends Event {
     code: number;
     reason: string;


lib/bom.js
```diff
@@ -223,6 +223,11 @@ declare var NodeFilter: {
     SHOW_DOCUMENT_FRAGMENT: number;
 };
 
+declare class CustomEvent extends Event {
+  static initCustomEvent(type: DOMString, canBubble: Boolean, cancelable: Boolean, detail: Object): CustomEvent;
+  detail: Object;
```
mroch: i think this should be `any`. it sounds like it could be non-objects like strings or null too

lib/bom.js
```diff
@@ -223,6 +223,11 @@ declare var NodeFilter: {
     SHOW_DOCUMENT_FRAGMENT: number;
 };
 
+declare class CustomEvent extends Event {
+  static initCustomEvent(type: DOMString, canBubble: Boolean, cancelable: Boolean, detail: Object): CustomEvent;
```
mroch: instead of the spec's `DOMString`, we just use `string`, and `boolean` instead of `Boolean` (the former are the literals `true` and `false`, the latter is an object boxing those values)

and like below, I think `detail` should be `any` here too

mroch:
thanks @Emile-Filteau 

there's not a test specifically for bom.js

mroch:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/472361926293391/int_phab) to review.

mroch:
hey, sorry to do this but I noticed that https://github.com/facebook/flow/pull/1873 added this class first but we neglected to merge it for a while.

I'd also like to add the `CustomEvent` constructor since that's apparently the non-deprecated way to do things now. so, we'll merge some combination of this PR and the other one plus some related fixes. thanks!

PR #1995, opened 2016-06-23
Delete lint.ml* files from GitHub

gabelevi:
We've stopped syncing them so now all that's left is to delete them from GitHub


gabelevi:
I'll wait for Travis and then merge this directly.

PR #2004, opened 2016-06-28
Split up hh-utils to fix ocp-build

gabelevi:
In theory this fix could go even farther and make sure each library depends on only the little bit of utils that it needs. But I'm being lazy and only doing the minimum work needed to get the build working again.

Also it's unfortunate that I need to disable globbing for utils to get this working :/

### DIFF
--- a/ocp_build_hack.ocp.fb
+++ b/ocp_build_hack.ocp.fb
@@ -25,10 +25,11 @@ begin library "hh-heap"
   files = begin fb-glob "hack/heap"
   end
   requires += [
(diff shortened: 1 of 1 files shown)


gabelevi:
@facebook-github-bot shipit

facebook-github-bot-7:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1097976730241261/int_phab) to review.

PR #2012, opened 2016-06-29
[try] Store last edit in localStorage for next load

jeffmo:
It always mildly irked me that /try doesn't remember the thing I wrote last time, so here we go.

(Note that I made this diff in the GitHub browser editor, so I have not tested it at all yet. I'll do that and follow up with confirmation in a little bit when I get to the office)

### DIFF
--- a/website/_assets/js/tryFlow.js.es6
+++ b/website/_assets/js/tryFlow.js.es6
@@ -135,7 +135,8 @@ function getAnnotations(text, callback, options, editor) {
 }
 getAnnotations.async = true;
 
-const defaultValue = `/* @flow */
+const lastEditorValue = localStorage.getItem('tryFlowLastContent');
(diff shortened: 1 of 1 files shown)


jeffmo:
Ok well I can't get jekyll running on my machine, so I give up.

Here's a brief summary of where things went wrong (though I've spent about 20mins trying to figure out how to fix various things and haven't gotten anywhere)

https://gist.github.com/jeffmo/3145b466b84721b0da7afcec580c8222

jeffmo:
Was able to move past the libyaml stuff and get here now: https://gist.github.com/jeffmo/f9efed5286418b2494ef26fc5b147141

mroch:
Check out resources/travis to see how the deps are installed for travis builds. I'll be at the office soon to help more if that doesn't work

PR #2013, opened 2016-06-29
Infer munged instance members.

pvolok:
The doc says that flow should be able to infer all types within a module. If `munge_underscores` is enabled, then private methods' params annotation should not be required I guess.

I find myself writing a lot on annotations instead of allowing flow to infer types like this:

```js
export default class Comp extends React.Component {
  render() {
    return this._render('hi');
  }

  _render(msg) {
    return <Button onClick={this._handleClick}>{msg}</Button>;
  }

  _handleClick = (event) => {};
}
```

I'm only starting to dive into flow. So this PR is just my best guess how to fix the problem :)

### DIFF
--- a/newtests/munged_class_member_inference/_flowconfig
+++ b/newtests/munged_class_member_inference/_flowconfig
@@ -1,0 +1,10 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
+all=true
+esproposal.class_instance_fields=enable
+munge_underscores=true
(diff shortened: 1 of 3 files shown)


gabelevi:
Interesting pull request! I'll get this reviewed by the team ASAP!

gabelevi:
@facebook-github-bot import

gabelevi:
Sorry for the delay! It took a little time to get this reviewed, but it looks good! Merging it now! Thanks a ton for the awesome PR!

PR #2066, opened 2016-07-11
Add definitions for console methods

thejameskyle:
This is based on the WhatWG Console Standard (https://console.spec.whatwg.org/).

Obviously not all these methods are available in every environment. So the handling of this comes down to how Flow typically handles standards that aren't implemented everywhere at the current time.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -669,4 +669,20 @@ declare function require(id: string): any;
 declare var exports: any;
 
 /* Commonly available, shared between node and dom */
-declare var console: any;
+declare var console: {
+  clear(): void;
+  log(...data: Array<any>): void;
+  info(...data: Array<any>): void;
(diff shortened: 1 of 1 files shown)


bhosmer:
Hi @thejameskyle - judging by the [test output,](https://travis-ci.org/facebook/flow/builds/144031277) it looks like the new definitions are provoking a parse error in the library definition. Should be easy to find/fix and rerun tests.

thejameskyle:
😓 fixed

bhosmer:
Cool 😁  @gabelevi the AppVeyor failure looks unrelated to me, but I'll hold off on merging until you've had a look, don't know if we need to kick it or what.

bhosmer:
@facebook-github-bot import

PR #2068, opened 2016-07-12
use Uint8ClampedArray in canvas ImageData

jansedivy:
I was having a problem with data field in ImageData class because its defined as `Array<number>` instead of `Uint8ClampedArray`. With `Array<number>` I can't use various methods defined on the typed array.

### Error example
```
imageData.data.set(buffer);
               ^^^ property `set`. Property not found in
imageData.data.set(buffer);
^^^^^^^^^^^^^^^^^^^^^^^^^^ Array
```

reference: https://developer.mozilla.org/en-US/docs/Web/API/ImageData

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1035,7 +1035,7 @@ declare class Path2D {
 declare class ImageData {
   width: number;
   height: number;
-  data: Array<number>;
+  data: Uint8ClampedArray;
 };
 
 declare class CanvasRenderingContext2D {


bhosmer:
@jansedivy Thanks Jan!

@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1650800385244705/int_phab) to review.

PR #2094, opened 2016-07-18
Add some Chrome definitions

runn1ng:
I am selfishly adding only the few APIs I am using in my apps/extensions. I do not really have time (and knowledge) to add more.

### DIFF
--- a/lib/chrome.js
+++ b/lib/chrome.js
@@ -1,0 +1,175 @@
+declare class chrome$chrome {
+  hid: chrome$hid;
+  runtime: chrome$runtime;
+  storage: chrome$storage;
+  app: chrome$app;
+  browser: chrome$browser;
+  sockets: chrome$sockets;
+}
+
+declare class chrome$sockets {
+  udp: chrome$udp;
+}
+
+declare type chrome$udpProperties = {
+  persistent?: boolean;
+  name?: string;
+  bufferSize?: number;
+}
+
+declare type chrome$udpCreateInfo = {
+  socketId: number;
+}
+
+declare type chrome$udpSendInfo = {
+  resultCode: number;
+  bytesSent?: number;
+}
+
+declare class chrome$onReceive {
+  addListener: (callback: (info: chrome$udpReceiveInfo) => void) => void;
+}
+
+declare type chrome$udpReceiveInfo = {
+  socketId: number;
+  data: ArrayBuffer;
+  remoteAddress: string;
+  remotePort: number;
+}
+
+declare class chrome$udp {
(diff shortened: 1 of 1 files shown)


runn1ng:
If you think having the definitions are OK, I will also add some tests

vkurchatkin:
Why not use `flow-typed` for this?

runn1ng:
I am not entirely sure where to put it in flow-typed, since it isn't a library (as I wrote there).

And there already are node and react definitions here.

avikchaudhuri:
@jeffmo WDYT? Where should these definitions go?

@runn1ng Depending on what @jeffmo thinks, the way you write tests might be different, so hold on.

jeffmo:
Yea, we don't have the machinery yet for stuff like this in flow-typed. At some point we'd like to (and there is a rough idea for adding a new namespace that sits next to the `definitions/npm` namespace), but in the meantime platform-APIs are probably best suited in the core libs.

fredemmott:
If these are committed, am I able to extend them to add additional APIs that I use that aren't in here, without rebuilding/packaging flow?

runn1ng:
Nope, you have to rebuild flow, AFAIK.

fredemmott:
Strongly against merging then unless this becomes complete. Merging would pretty much mean that only chrome extensions that happen to use the same APIs as you can use flow properly. There's at least one this would break that I'm aware of.

runn1ng:
@fredemmott since Chrome is adding features pretty fast, it will probably never be complete and it will always be outdated.

But I understand what you mean, and I have no idea what to do with it.

jeffmo:
@fredemmott: If we don't merge, aren't you still in the same situation?

To be clear: We want to address this general issue soon (via flow-typed) -- we just aren't equipped to do so quite yet. Still work to do to get there

jeffmo:
Also note that it is always possible to *override* built-ins (you can always copy them into your own `[libs]` or `flow-typed` directory and adjust) -- just not extend them in place.

fredemmott:
> If we don't merge, aren't you still in the same situation?

No, I'm just adding my own

> Also note that it is always possible to override built-ins (you can always copy them into your own [libs] or flow-typed directory and adjust) -- just not extend them in place.

That might mean this isn't a problem - no duplicate definition errors like Hack does?

That said, it could be less confusing to have `chrome is undefined` than `chrome.foo is not defined`.

gabelevi:
I'm cleaning out our old PRs. It looks like the ideal solution is for https://github.com/flowtype/flow-typed/issues/201 to be resolved. We also have been talking about trying to move some of the core flowlibs to flow-typed too. So I'm going to close this PR out for now.

If anyone disagrees or thinks that this PR should be merged, feel free to re-open and make the case.

PR #2106, opened 2016-07-20
Add respect-pragma flag to check-contents

nmote:
This behavior makes more sense for an IDE since the user isn't explicitly asking Flow to check the contents of the file.

See https://github.com/facebook/nuclide/issues/604

### DIFF
--- a/src/commands/checkContentsCommand.ml
+++ b/src/commands/checkContentsCommand.ml
@@ -35,11 +35,13 @@ let spec = {
     |> strip_root_flag
     |> json_flags
     |> verbose_flags
+    |> flag "--respect-pragma" no_arg
+        ~doc:"Respect the presence or absence of an @flow pragma"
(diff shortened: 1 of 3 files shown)


samwgoldman:
Thanks! This looks good to me. Can you add a test?

nmote:
I don't see any existing tests that exercise `check-contents`. I don't think it's reasonable to require a test along with this change when there weren't any to begin with. I did test it manually.

nmote:
Maybe I'm looking in the wrong place but `grep -r check-contents .` from the `tests` directory returns no results.

samwgoldman:
mkay

@facebook-github-bot shipit

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/505732066302459/int_phab) to review internal test results.

PR #2109, opened 2016-07-20
Fix Tail for Windows

gabelevi:
I'm not sure what this code was supposed to do, but reading the log file and using those contents as the name of the log file doesn't make much sense. Removing this special case seems to work, so let's go with that.


gabelevi:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/609802375856350/int_phab) to review internal test results.

PR #2110, opened 2016-07-20
Fix log rotating on Windows

gabelevi:
When we create a new log file `foo.log`, we move the existing `foo.log` to `foo.log.old`. This wasn't quite working on Windows. The problems were

1. On Windows, `Sys.rename` fails if `foo.log.old` already exists. Easy to fix - we just explicitly delete it.
2. `Sys.rename` would fail if someone had `foo.log` open. If the server is starting up because you ran `flow status`, the status command still had `foo.log` open to tail the log messages. This was easy to fix - we just stop tailing the log file before we start the new server.

Either way, if the log rotation fails, we soldier on.

### DIFF
--- a/src/commands/commandConnect.ml
+++ b/src/commands/commandConnect.ml
@@ -181,6 +181,10 @@ let rec connect env retries start_time tail_env =
   | Result.Ok (ic, oc) -> (ic, oc)
   | Result.Error CCS.Server_missing ->
       if env.autostart then begin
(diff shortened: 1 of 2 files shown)


gabelevi:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1075424275870728/int_phab) to review internal test results.

PR #2111, opened 2016-07-20
Fix Windows process ID/HANDLE mess

gabelevi:
This was pretty frustrating to debug. But I think I figured out what's going on.

Coming from Linux & OS X, I'm used to referring to processes by the "Process ID" (pid). Many functions return a pid, like `Unix.getpid()` and `Unix.create_process()`, and many processes take a pid, like `Unix.waitpid` and `Unix.kill`.

Windows also has Process IDs. Each Process has a pid. But in addition to this, there are also Process Handles. They feel a little bit like file descriptors. For example, you can get a Process Handle with a call like `OpenProcess(pid)`. Process handles look similar to pids (also a relatively small number), but are not portable. So if you want to refer to a process definitively, you need to use the pid.

For some reason that isn't clear to me, functions like `Unix.getpid()` and `Unix.create_process()` return Process Handles on Windows, and not always the same process handles. This is mostly ok, since functions like `Unix.waitpid` expect process handles, so it's mostly opaque. However, if you want to output a pid to stdout or a log, then you need to dereference the handle first.

### DIFF
--- a/hack/utils/daemon.ml
+++ b/hack/utils/daemon.ml
@@ -59,6 +59,7 @@ module Entry : sig
      interface. *)
 
   type ('param, 'input, 'output) t
+  val get_name: ('param, 'input, 'output) t -> string
   val register:
     string -> ('param -> ('input, 'output) channel_pair -> unit) ->
     ('param, 'input, 'output) t
@@ -77,6 +78,8 @@ module Entry : sig
 end = struct
 
   type ('param, 'input, 'output) t = string
+
+  let get_name name = name
 
   (* Store functions as 'Obj.t' *)
   let entry_points : (string, Obj.t) Hashtbl.t = Hashtbl.create 23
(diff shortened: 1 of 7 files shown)


gabelevi:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1133084130089937/int_phab) to review internal test results.

hack/utils/daemon.ml
```diff
@@ -78,6 +79,8 @@ end = struct
 
   type ('param, 'input, 'output) t = string
 
+  let get_name name = name
```
mroch: `get_name` isn't very ocaml-y. how about `name_of_entry`?

hack/utils/sysinfo.c
```diff
@@ -44,3 +48,36 @@ value hh_sysinfo_uptime(void) {
...
 }
+
+/**
+ * There are a bunch of functions that you expect to return a pid,
+ * like Unix.getpid() and Unix.create_process(). However, on
+ * Windows, instead of returning the process ID, they return a
+ * process handle.
+ *
+ * Process handles seem act like pointers to a process. You can have
+ * more than one handle that points to a single process (unliked
```
mroch: s/unliked/unlike/

gabelevi:
👋 👋 👋 👋 👋 👋 (those are waving hands, if it's not clear)

Made the changes you suggested on the internal version of this PR :)

PR #2112, opened 2016-07-20
Timeout should raise End_of_file instead of EPIPE

gabelevi:
The Windows implementation of Timeout handles input byte by byte. This means we can end up with different exceptional behavior. One difference I noticed was when a pipe broke while we're waiting to read from that pipe. The built in `input_value` would raise `End_of_file`, but this implementation was raising `Unix.Unix_error (Unix.EPIPE, "read", "foo")`


gabelevi:
@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/532336466945239/int_phab) to review internal test results.

PR #2137, opened 2016-07-26
Better error message on Windows for long paths

gabelevi:
Windows doesn't allow long paths. Previously we'd just output a message saying that we were skipping the paths. This error message is a little more clear.

I ran into this when I used npm2 instead of npm3, which led to some really long file names. Since this is an OS restriction, it's probably better for the user to just avoid long file names than for us to do something clever.


gabelevi:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1039436542758397/int_phab) to review internal test results.

PR #2144, opened 2016-07-27
add @flow comment to .js.flow style file

### LINKED ISSUE #2142: Clearify in doc that `.js.flow` style declaration files need `// @flow` comment
[foo.js]
```js
// @flow
export var a = '123'
```
[foo.js.flow]
```js
declare export var a: number
```
[bar.js]
```js
// @flow
import { a } from './foo'

const x = a / 1 
a.undefinedMethodInNumber()
```

In this case, flow cannot detect the error in `a.undefinedMethodInNumber()`.

Attaching `// @flow` comment in `foo.js.flow` will solve the issue. The problem is, we don't notice this issue in most case because the line `const x = a / 1` will be corrected.

Please change the sample code in [Documentation](https://flowtype.org/docs/declarations.html).

shinout:
fix #2142 
Attach `// @flow` comment to the sample code.

### DIFF
--- a/website/docs/ref/declarations.doc.js
+++ b/website/docs/ref/declarations.doc.js
@@ -232,6 +232,7 @@ export function isLeapYear(year: number): bool {
   that we also write the keyword `export`, like we would in an implementation
   file.
 */
+// @flow
 declare export function isLeapYear(year: string): bool;
 
 /*


jeffmo:
Thanks!

@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing.

PR #2154, opened 2016-07-28
Fix Windows build after memfd_reserve fixes

gabelevi:
We deleted a `memfd_reserve` call inside of `hh_collect` since it was causing trouble. However this broke the Windows build (thanks AppVeyor for noticing!). The problem is that Windows needs to reserve the memory inside of `hh_collect`.

So this is hopefully the last diff to address this. This restores the `memfd_reserve` behavior, but only for Windows. It also renames the function a little to make it clear that it has nothing to do with `memfd`.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -639,11 +639,20 @@ static void raise_out_of_shared_memory()
 
 #ifdef _WIN32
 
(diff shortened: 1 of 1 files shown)


gabelevi:
@facebook-github-bot import

PR #2177, opened 2016-08-02
Relax console.assert's first argument

jdelStrother:
It accepts truthy/falsey values, rather than a simple boolean.

We have quite a few instances of `console.assert(someObjectThatMightBeNull, 'msg')` in our codebase, which flow 0.30 started warning about.  Am I using console.assert wrong, or would this PR be an appropriate fix?

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -702,7 +702,7 @@ declare var exports: any;
 
 /* Commonly available, shared between node and dom */
 declare var console: {
-  assert(condition: boolean, ...data: Array<any>): void;
+  assert(condition: mixed, ...data: Array<any>): void;
   clear(): void;
   count(label: string): void;
   debug(...data: Array<any>): void;


facebook-github-bot-6:

facebook-github-bot-7:

lib/core.js
```diff
@@ -702,7 +702,7 @@ declare var exports: any;
 
 /* Commonly available, shared between node and dom */
 declare var console: {
-  assert(condition: boolean, ...data: Array<any>): void;
+  assert(condition: any, ...data: Array<any>): void;
```
jeffmo: Can we make this `mixed` instead?

jdelStrother: Sure, though I don't entirely follow the benefits of doing so. Can you give an example error that would pick up?

thejameskyle: https://flowtype.org/docs/builtins.html#mixed

It's just more specific about what the input is here (even if that doesn't catch any real world errors). In general I would just think of `any` as a quick lazy solution, and always prefer `mixed`.

jdelStrother: OK, updated.

avikchaudhuri:
@facebook-github-bot shipit

facebook-github-bot-2:
Thanks for importing.

PR #2181, opened 2016-08-02
Document the all .flowconfig option

gabelevi:


gabelevi:
@facebook-github-bot shipit-ninja

facebook-github-bot-3:
Normally, gabelevi, someone else should review and ship your pull request. However, we will trust that you know what you are doing. Shipping!

facebook-github-bot-0:
Thanks for importing.

jeffmo:
Can this include the caveats of using this? This rarely makes for a good user experience (and often so after its too late -- ie after publishing code)

thejameskyle:
How does it make for a poor user experience?

jeffmo:
If a user without this config option uses a library with it turned on, Flow will not know this and yield `any` For modules that are expected to have types.

The same goes for the inverse -- turning this on will attempt to check any 3rd party JS that is not opted in (which results in lots of errors)

jeffmo:
I had proposed using .js.flow files as a way of accomplishing the no-pragma goal and I still think that's a pretty good idea.

It's a no-config approach and it's not soft-optional (like a system that works the same with any extension). You don't have to use it (say with .js + @flow), but if you really don't like @flow you have an easy option.

thejameskyle:
Could we improve the interaction between flow and `node_modules` in terms of localizing configs in order to solve this?

jeffmo:
Having thought about this for a while, I don't think so -- not without nuance and caveats (the thing we're trying to remove).

We don't want to *not* check node_modules because they can contain important .js.flow files.

There's also a general problem with composing configs that just gets confusing: Some options compose, others do not, it's really difficult to know which can/can't.

For example: Name_mapper probably shouldn't compose because it can map to things outside a sub-library that weren't intended.

jeffmo:
I really think all=true is a tempting trap for most cases and will hurt the ecosystem due to its latent effects (publishers publish not considering that users won't have it on, bad UX ensues).

thejameskyle:
I feel like it can be appropriate but there needs to be a more specific workflow for publishers that want to distribute flow types with their libraries so that they are uniform.

jeffmo:
Let's chat about it tomorrow? I think this is sufficiently nuanced and tricky that it might be worth more discussion... I want to be sure we articulate the pros/cons well here.

I'm not opposed to documenting all=true, I just think we should express the issues clearly where we do document it. There are a lot of requests for it, and usually when people try it they acknowledge that it's not really what they thought it would be.

jeffmo:
(I also only have my cell phone at home right now because #comcastsucks :( -- so harder than normal to articulate thoughts in touchscreen tap form)

Issue #1996, opened 2016-06-23
No simple way to author and share flow packages.


Gozala:
Documentation does not seem to provide much insight on what is an idiomatic way to write packages like what file extension to use how to build and publish to allow others consume your packages without needing to customize their configurations.

I will just assume that setup used by [graphql-js](https://github.com/graphql/graphql-js) is an idiomatic way to go about it:

- Author source in `src/` directory as `module.js` files.
- Use babel to transpile to plain JS.
- Publish to npm.

Problem with such setup is that users of my library loose all of the type information that original source contained. There are ways to workaround it but they either require flow configuration on the consumer side or non trivial build step from the publisher side.

## Proposal

I would like to propose following things to improve current situation & grow number of flow typed packages distributed over npm.

1. Documentation should certainly include a guide on how to author and publish flow typed package.
2. I would suggest to make `module.flow` an idiomatic file extension for flow typed files. For following reasons:
  1. Flow extends JS syntax so sharing same file extensions does not really makes sense.
  2. If you author files flow code in `module.js` you can't really generate compiled JS files alongside of them.
3. In 0.19 introduced useful feature, that allows putting `module.js.flow` along the side of `module.js` in which case flow would give precedence to `module.js.flow` over `module.js` when importing `./module`. Unfortunately no attempt to load `module.flow` is made. It would actually make sense to alter algorithm slightly so that flow would try `module.flow` then `module.js.flow` and then `module.js` which in fact would match node's algorithm more closely which attempts to load `module` and then `module.js`.

This would enable users to author their packages in `module.flow` files and then compile all `module.flow` files to `module.js` files along the side them. Publishing such package into npm also would preserve all the type information for consumers, since flow would first look for `dep/module.flow` (which will be an original source) before `dep/module.js`.

Note: Technically you could achieve more or less same even if your flow code was authored in `module.js` files but that would require copying and renaming them to `module.js.flow` which as far as I can tell has no trivial cross platform solution & requiring extra scripts isn't great. Another alternative is to just author code as `module.js.flow` and generate `module.js` files, it's just authoring in `.js.flow` file extension seems awkward.

vkurchatkin:
It seems completely unnecessary. You can just have `index.js` and `index.js.flow` in root folder, no need to rename all source files

Gozala:
That assumes that everything used by user is exposed through index module which is not necessarily the case, definitely not the case for packages I work with

Typed on tiny keyboard

zerkalica:
[...]

PR #2184, opened 2016-08-03
`flow shadow-file`

jeffmo:
This adds a new command, `flow shadow-file path/to/file.js`, which generates the minimal `.js.flow` interface file that can be used for publishing compiled Flow source. Specifically: It generates only `declare` and `declare export` statements in order to omit all the implementation text.

Background: Currently the best practice for publishing Flow source is (roughly):

1. `cp src/foo.js dist/foo.js.flow`
2. `babel src/foo.js > dist/foo.js`

This mostly works, but has a few downsides:

1. The published contents of `dist/*` are much larger than they need to be -- they contain a pre-compiled copy of all the original source. This is basically just crufty bytes that need to be downloaded every time the package is `npm install`-ed.
2. The published `.js.flow` files contain implementation details -- which are much more susceptible to breaking changes across Flow versions. While breaking changes across Flow version are still possible even in `declare` statements, this happens *far* less often.
3. We also want to re-use this type -> code codegen infrastructure to generate libdefs for flow-typed at some point. This particular diff is only about `.js.flow` files (i.e. no `declare module ...` statements) -- but that can be added on in a follow up diff.

This diff includes a new `codegen.ml` file which exposes an API intended for generating code from various things (types for now, possibly ASTs at some point if that's useful/performant/worth the effort). It's minimal in terms of feature-set right now but we can expand/improve it later. I didn't use `type_printer.ml` because it only handles some types and many of the types it will print aren't actual code or easily composable with other bits of code. It's also used by lots of stuff that I didn't really want to investigate breaking changes for while building out this feature.

At some point it probably makes sense to either improve `Type_printer` enough to subsume `Codegen` or the other way around. I suspect the `Codegen` is going to be easier to generalize, but we'll leave that for a later time to look into.

### DIFF
--- a/hack/utils/tty.mli
+++ b/hack/utils/tty.mli
@@ -35,11 +35,11 @@ type color_mode =
   | Color_Auto
 
 (*
- * Print a sequence of colorized strings to stdout, using ANSI color escapes
- * codes.
+ * Print a sequence of colorized strings to stdout/stderr, using ANSI color
+ * escapes codes.
  *)
-val cprint : ?color_mode:color_mode -> (style * string) list -> unit
-val cprintf : ?color_mode:color_mode -> style ->
+val cprint : ?color_mode:color_mode -> ?out_channel:out_channel -> (style * string) list -> unit
+val cprintf : ?color_mode:color_mode -> ?out_channel:out_channel -> style ->
   ('a, unit, string, unit) format4 -> 'a
 
 (* These two functions provide a four-state TTY-friendly spinner that
--- a/src/commands/shadowFileCommand.ml
+++ b/src/commands/shadowFileCommand.ml
@@ -1,0 +1,76 @@
+(**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the "flow" directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ *)
+
+open CommandUtils
+
+let spf = Printf.sprintf
+
+let name = "shadow-file"
+let spec = {
+  CommandSpec.
+  name;
+  doc = "Given a filename, generate a shadow (.js.flow) file.";
+  usage = Printf.sprintf
+    "Usage: %s %s [OPTIONS] [FILE] [FILE] [FILE]...\n\n\
+      e.g. %s %s foo.js > foo.js.flow\n"
+    CommandUtils.exe_name
+    name
+    CommandUtils.exe_name
+    name
+  ;
+  args = CommandSpec.ArgSpec.(
+    empty
+    |> server_flags
+    |> root_flag
+    |> error_flags
+    |> strip_root_flag
+    |> anon "file" (required string)
+        ~doc:"The file for which a shadow file should be generated"
+  )
+}
+
+let main option_values root error_flags strip_root file () = (
+  let root = guess_root (
+    match root with
+    | Some root -> Some root
+    | None -> Some (expand_path file)
+  ) in
+  let filename = ServerProt.FileName (expand_path file) in
+  let (in_chan, out_chan) = connect option_values root in
+  ServerProt.cmd_to_channel out_chan (ServerProt.GEN_INTERFACES [filename]);
+  let response =
+    (Timeout.input_value in_chan: ServerProt.gen_interface_response)
+  in
+
+  match response with
+  | response::[] -> (
+    match response with
+    | Utils_js.Err (ServerProt.GenIface_TypecheckError (_, errors)) ->
+      let errors = Errors.to_list errors in
+      Errors.print_error_summary ~out_channel:stderr ~flags:error_flags ~strip_root ~root errors;
+      FlowExitStatus.exit
+        ~msg:"\nThere must be no type errors in order to generate a shadow file!"
+        FlowExitStatus.Type_error;
+    | Utils_js.Err (ServerProt.GenIface_UnexpectedError (file_path, error)) ->
+      FlowExitStatus.exit
+        ~msg:(spf "Error: %s: %s" file_path error)
+        FlowExitStatus.Unknown_error
+    | Utils_js.OK (_file, interface) ->
+      print_endline interface
+  )
+  | response ->
+    let msg = spf (
+      "Internal Error: Expected a single interface description from the " ^^
+      "server, but received %d interfaces!"
+    ) (List.length response) in
+    prerr_endline msg
+)
+
+let command = CommandSpec.command spec main
--- a/src/typing/codegen.ml
+++ b/src/typing/codegen.ml
@@ -1,0 +1,298 @@
+(**
+ * Copyright (c) 2013-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the "flow" directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ *)
+
+let spf = Printf.sprintf
+
+type codegen_env = {
+  buf: Buffer.t;
+  declared_classes: string IMap.t;
+  mutable next_declared_class_name: int;
+  flow_cx: Context.t;
+  indent: string;
+  tparams: Type.typeparam list;
+  applied_tparams: Type.t list;
+}
+
+let add_str str env = Buffer.add_string env.buf str; env
+let add_tparams tparams env =
+  {env with tparams;}
+let add_applied_tparams applied_tparams env =
+  {env with applied_tparams;}
+let find_props tmap_id env = Flow_js.find_props env.flow_cx tmap_id
+let next_declared_class_name env =
+  let id = env.next_declared_class_name in
+  env.next_declared_class_name <- id + 1;
+  id
+let resolve_type t env = Flow_js.resolve_type env.flow_cx t
+let to_string env = Buffer.contents env.buf
+let with_indent indent fn env =
+  let orig_indent = env.indent in
+  let env = {env with indent = indent ^ orig_indent;} in
+  let env = fn env in
+  {env with indent = orig_indent;}
+
+let mk_env flow_cx buf = {
+  applied_tparams = [];
+  buf;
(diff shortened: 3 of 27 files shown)


jeffmo:
@facebook-github-bot import

facebook-github-bot-9:
Thanks for importing.

jeffmo:
@facebook-github-bot update

mroch:
might I suggest the name `flow declare` or `flow gen-decl` or `flow gen-def` or `flow interface` or something like that? I think it should a) be a verb if possible, and b) not introduce new terminology that isn't used elsewhere ("shadow file").

src/commands/shadowFileCommand.ml
```diff
@@ -0,0 +1,66 @@
+open CommandUtils
```
mroch: missing copyright header

src/services/interfaceGenerator/InterfaceGenerator.ml
```diff
@@ -0,0 +1,240 @@
+let spf = Printf.sprintf
```
mroch: - missing copyright header
- filename should be interface_generator.ml and the module should be `Interface_generator`

src/typing/codegen.ml
```diff
@@ -0,0 +1,288 @@
+let spf = Printf.sprintf
```
mroch: missing copyright header

src/services/interfaceGenerator/InterfaceGenerator.mli
```diff
@@ -0,0 +1 @@
+val generate_interface: Context.t -> string
```
mroch: missing copyright header

hack/utils/tty.mli
```diff
@@ -38,8 +38,8 @@ type color_mode =
  * Print a sequence of colorized strings to stdout, using ANSI color escapes
```
mroch: this comment is rotten given `stderr`.

you could alternatively pass an output channel and then call `Tty.cprint ~oc:stderr` or `Tty.cprint ~oc:stdout` instead of a boolean argument.

src/typing/codegen.ml
thejameskyle: Is there no lint rule for this on Fb projects?

jeffmo: I did this locally, so I definitely wasn't running any kind of ocaml lint

jeffmo:
> might I suggest the name `flow declare` or `flow gen-decl` or `flow gen-def` or `flow interface` or something like that? I think it should a) be a verb if possible, and b) not introduce new terminology that isn't used elsewhere ("shadow file").

So I'd like us to get away from using `declare` and `def`/`definition` to describe `.js.flow` files. I would like to update this in the docs as well (and I can & should do that in this diff if we can agree on `shadow` or something else).

Specifically: There's a lot of confusion right now around the differences between `.js.flow` files (things that shadow `.js` files) and "libdefs" (things that go into the `[libdefs]` section of your flowconfig). 

I wrote up some sample documentation here that I'd like to move into docs soon (maybe in this diff): https://github.com/facebook/flow/issues/1996#issuecomment-230919868

I'm not necessarily tied to `shadow`, but it makes the most sense to me: They're files that shadow adjacent/same-named `.js` files.

jeffmo:
I should be able to circle back and update the "libdef" and "shadow files" docs either this weekend or Monday.

jeffmo:
* copyright headers added
* cprintf comment updated

jeffmo:
Switched `cprintf` functions to take `out_channel` instead of a bool flag

mroch:
Thanks for the updates.

I'm still not convinced on the name "shadow-file". You've previously suggested that users should write `.js.flow` files and compile them to `.js` files. The term "shadow" makes it feel like a second-class thing when it's really the other way around. Also, my understanding is that this command generates `declare` interfaces, not a complete shadow file as you've defined that term in the docs you linked to. flow-copy-source does what I think `flow shadow-file` should do based on its name.

jeffmo:
> The term "shadow" makes it feel like a second-class thing when it's really the other way around.

I've sort of gone back and forth on this, too. At the moment I'm stuck on the "it's not perfect, but not that unclear" side (you write a flow file that shadows the real, compiled, JS file).

> my understanding is that this command generates `declare` interfaces

That's right. Depending on how you interpret "shadow", this still seems to make sense to me -- but maybe I've just gotten used to it after thinking about it for a little while now.

I'm still pretty open to names other than "shadow" if you have any suggestions, though? "shadow" is just the best I was able to come up with so far.

The only things I want to avoid are the terms "lib"/"def" (too confusingly similar to "[libdefs]"), and "decl"/"declare" (doesn't adequately distinguish libdefs, which have `declare`s, from these files, which have `declare`s).

I suppose we could name the command `flow gen-flow-file`?

jeffmo:
cc @marudor @ryyppy for some outside perspectives: Any thoughts on what we should call ".js.flow" files?

(the files that get published to npm and sit next to .js files to "shadow" them -- for lack of a better term at the moment :) )

src/typing/codegen.ml
```diff
@@ -0,0 +1,298 @@
...
+ *)
+
+let spf = Printf.sprintf
+
+type codegen_env = {
+  buf: Buffer.t;
+  declared_classes: string IMap.t;
+  mutable next_declared_class_name: int;
+  flow_cx: Context.t;
+  indent: string;
```
jeffmo: unused right now

ryyppy:
When I initially saw the PR with `shadow-file`, I knew what this is about before I even saw the PR details (although I might be biased).

Usually when I explain libdefs / js.flow files to others, I usually called them `header` files (like in C) on either lib or file level... but I guess this is even more confusing...

mroch:
I don't think we're discussing what to call all kinds of .js.flow files here, only what to call .js.flow files which happen to contain only `declare` statements.

I think we should call these declaration files, because they only contain declarations. "libdefs" from the builtins or `[libs]` are just global declaration files, and we could start calling them such (and should rename them to `.js.flow`).

I'd call all forms of `.js.flow` files something like "Flow files."

samwgoldman:
This is great. I am especially happy that this encoding will prevent issues where config differences between lib and consumer cause errors, as you pointed out in your description. 

Does this work for files that import types/values from other modules? Can you include some tests for generating shadow files for modules with dependencies?

Since you're generating the declarations from types, I assume suppression types just come out as any, correct? Type aliases/annots are substituted by their definition as well, yes?

What about generating shadow files for many modules at once? I think that could be a common use case, for example fbjs's build step copies over all files. Is there any perf gain to be had with a batch mode?

src/typing/codegen.ml
```diff
@@ -0,0 +1,298 @@
...
+let add_tparams tparams env =
+  {env with tparams;}
+let add_applied_tparams applied_tparams env =
+  {env with applied_tparams;}
+let find_props tmap_id env = Flow_js.find_props env.flow_cx tmap_id
+let next_declared_class_name env =
+  let id = env.next_declared_class_name in
+  env.next_declared_class_name <- id + 1;
+  id
+let resolve_type t env = Flow_js.resolve_type env.flow_cx t
```
samwgoldman: On mobile so forgive an unresearched comment: I think Flow_js.resolve_type will return `any` for open tvars, so ideally this should happen after merge. Does it?

jeffmo:
> only what to call .js.flow files which happen to contain only declare statements.

This is the level of nuance we want to avoid. The mental model needs to be simple, and currently it is: flow files always shadow js files and can be anything a js file can be. Libdefs are global and registered in the config.

People seem to understand this model once described in this way, so I want to avoid breaking things down into any further categories of files. TS's story around interface/decl/typings files is all over the place with terminology and I've always found it confusing in conversations about them -- so this has motivated a strong need to keep things simple and minimal as we build out our story here.

jeffmo:
> Can you include some tests for generating shadow files for modules with dependencies?

Good call, I'll add some tests. In the beginning I expect these imported types to simply be flattened into the declared types, but once we work out our story around imports between libdefs we can take that approach.

> Since you're generating the declarations from types, I assume suppression types just come out as any, correct? Type aliases/annots are substituted by their definition as well, yes?

That's right, but I should add tests to prove it

> What about generating shadow files for many modules at once?

Right now this only takes a single file on the command line and prints the results to stdout (first-version simplicity). The server command is intentionally set up to accept multiple files and return multiple results to support multi-file generation though. It might be a little better to do this in a follow-up.

samwgoldman:
These files can include imports though, right? They're JS files including declarations, not lib defs iiuc. In fact, you probably need that in order to use class types (nominal) from other modules.

jeffmo:
Oh, good point -- you're right. I'll need to work that in then

src/typing/codegen.ml
jeffmo: Right, the expectation is that the `cx` passed into `Codegen.mk_env` is fully merged.

mroch:
> This is the level of nuance we want to avoid. The mental model needs to be simple, and currently it is: flow files always shadow js files and can be anything a js file can be.

So I agree with that mental model, but this command doesn't generate all types of shadow files, it generates a very specific subset: the declarations/interface of some real JS file.

Why does a command called `shadow-file` or `gen-flow-file` only generate interfaces? From the name and your description of shadow files, why doesn't it just copy the .js file to .js.flow (maybe even adding missing types, subsuming the `suggest` command)? That would also be a valid shadow file.

All I'm suggesting is that the name should reflect what the command does.

jeffmo:
> Why does a command called shadow-file or gen-flow-file only generate interfaces?

The purpose of the command is to generate a `.js.flow` file that is semantically identical to the `.js` file -- so whether this command does `cp` or whether it generates `declare`s is basically an implementation detail in terms of that goal. The command chooses `declares` because they offer the best bang for the buck in that goal (compat, minimal bytes, etc).

Truthfully though, I don't think it's necessary to be completely precise in the name as long as it's not incorrect. In this case, generating a shadow/`.js.flow` file *is* what we're doing here -- it's just left as an implicit detail (easily notable in the `--help` text and docs) that we're generating the minimal form.

If it's any consolation: If we ever do discover a good reason to generate non-`declare`-only `.js.flow` files, we can always add a flag that instructs the command to codegen alternative forms (i.e. `--fullImpl`, etc).

Issue #1887, opened 2016-06-02
.flowconfig file too complex


xaviervia:
Hi,

I love the project, but after trying to set it up in several different applications I'm finding out the `.flowconfig` file to be pretty complex to use. I'm trying to pitch adoption for adopting flowtype in my team and I can see how this is going to turn into a persistent pain point that will make adoption more unlikely. The problem are the OCaml regular expressions.

Given that I love the tool I'm willing to go through the process of learning how to do that configuration, but I find somewhat surprising that a tool specifically aimed at JavaScript developers doesn't provide a more familiar interface for configuration, such as glob patterns or JS regular expressions. If this is not viable I would suggest adding a complete real life example to the documentation–the way I managed to fix mine was by reading the [`.flowconfig` file from the React Native repository](https://github.com/facebook/react-native/blob/master/.flowconfig).

Thanks!

mroch:
agreed, i'm hoping to switch to PCRE/JSCRE (https://github.com/facebook/flow/issues/564). i think the best time to do it would be while changing the entire .flowconfig format to JSON or some other format (https://github.com/facebook/flow/issues/153), otherwise we won't be able to tell which format is being used.

xaviervia:
Thanks for your reply! Good to see that this is on the radar already.

Do you think it would be possible to add the real life example to the documentation for the time being? I fully understand that changing the configuration file format will imply non trivial effort, I think what's more valuable to me is to have some way of doing these configurations easily and without having to dig into the OCaml regular expressions documentation. I think the more common issues and gotchas (one undocumented issue I got for example is a npm linked dependency being processed even when the node_modules folder was being ignored) could be listed in there and that would be a big help 😃

calebmer:
Improving the format of `.flowconfig` is on our radar. Thanks for opening this issue 👍

Issue #564, opened 2015-06-25
Support negative lookaheads in .flowconfig excludes


mroch:
There's currently no good way to exclude an entire directory *except* for a single subdirectory. For example, say you have

```
foo/.flowconfig
foo/bar/a/
foo/bar/b/
foo/bar/...
foo/bar/z/
```

and you want to exclude `foo/bar/` *except* `foo/bar/b/`.

since excludes win over excludes, I think the only way right now is to list all of the other dirs in `excludes`.

it'd be nice if we supported negative lookaheads, like `bar/(?!b/)`. using `Pcre` would do it.

kyldvs:
:+1: 

This will be nice for including certain flow definitions from node_modules without pulling in all the errors from modules you don't care about

bm-w:
+1

As a temporary workaround, I resorted to `.*\(\(\([^/]s\|[^s]\)r\|[^r]\)c\|[^c]\)/main.js` today…

Efimster:
+1
Does anybody could clarify why flowtype uses different approaches to define [ignore] and [include]. I can't see any advantages in using OCaml regular expressions and match it against absolute path.

mroch:
we use ocaml regular expressions simply because flow is written in ocaml. linking in PCRE or using [ocaml-re](https://github.com/ocaml/ocaml-re) adds dependencies that would make building flow more difficult. obviously that's not a very good reason which is why I want to change them as soon as we can figure out a migration path.

SamChou19815:
We already have `!`

PR #2186, opened 2016-08-04
require [version] in flowconfig

mroch:
NOTE: this PR is 2 commits. the first is a boring codemod that adds [version] to all of the tests' .flowconfigs. the second is the real diff.

In the near future, I think we're going to want to break backwards-compatibility in `.flowconfig` (e.g. #564, #1887).

Rather than having to deprecate a config, we can instead branch based on the `[version]` field. Deprecating names sucks long term because we have to use a suboptimal new name forever, and sucks short term because you can have both old and new names at the same time. so, if we're going to error about that, it's no different than the way erroring based on version would work.

So if we see

```
[ignore]
some-regex

[version]
0.30.0
```

we know to parse `some-regex` using ocaml regular expressions, but if we see

```
[ignore]
some-regex

[version]
; some future version where #564 is fixed
0.10000.0
```

then we know to use PCRE.

This PR modifies `flow init` to default to

```
[version]
>= currentversion
```

which has the added benefit of ensuring that other users don't run flow on an older version than you've tested it on. we could be stricter and default to `^` instead of `>=`, but I didn't want to lock people into specific versions unless they opt into that.


avikchaudhuri:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3680463) to review internal test results.

mroch:
will rebase and reopen this later

Issue #2130, opened 2016-07-25
Import in declaration file breaks import in regular files


Macil:
.flowconfig

```
[include]

[ignore]

[libs]
flow-libs/

[options]
```

flow-libs/lib.js

``` js
import {EventEmitter} from 'events';

declare class Foo extends EventEmitter {
  bar(): number;
}
```

test.js

``` js
/* @flow */
import {EventEmitter} from 'events';
const o = new Foo();
```

Flow gives an error:

```
test.js:2
  2: import {EventEmitter} from 'events';
             ^^^^^^^^^^^^ Named import from module `events`. This module has no named export called `EventEmitter`.


Found 1 error
```

Not great but changing flow-libs/lib.js to this makes Flow stop giving an error:

``` js
import {EventEmitter as _EventEmitter} from 'events';

declare class Foo extends _EventEmitter {
  bar(): number;
}
```

(I think I need to find a way to change my declaration files so that it doesn't leak globals like EventEmitter, but that's separate from this issue.)

avikchaudhuri:
I don't think Flow really supports imports in declaration files. Use `events$EventEmitter` to get more reliable behavior. cc @jeffmo

thejameskyle:
Yes, dependencies within Flow lib definitions are sort of missing right now. It would be great to improve it for this exact case.

However, as @avikchaudhuri mentioned, types within lib definitions are global across dependencies so you can reference stuff like `events$EventEmitter` directly. We have a nice naming scheme (`moduleName$ObjectType`) for this and won't break it on you (at least not until after we have a better solution).

vaukalak:
@thejameskyle, @avikchaudhuri  could you please provide an example of this ? Right now I have:
file `definitions/foo.js`:
```
// @flow
declare module 'foo' {
  declare type $Foo = { testFoo() : string };
  declare type $FooBaz = $Foo & baz$Baz;
}
```
file `definitions/baz.js`:
```
// @flow
declare module 'baz' {
  declare type $Baz = { testBaz() : number };
}
```
and `index.js`:
```
/* @flow */
import type { $Foo, $FooBaz } from 'foo';
import type { $Baz } from 'baz';

var baz : $Baz;

function test(fooBaz : $FooBaz) {
  const s : number = fooBaz.testFoo();
  const n : string = fooBaz.testBaz();
}
```
My `.flowconfig` is:
```
[libs]
definitions/foo.js
definitions/baz.js
```
When I run `flow` command I get following:
```
[...]

Issue #2102, opened 2016-07-20
Extending EventEmitter stops module class from checking 


runn1ng:
I have this in flow declaration file for an [external module](https://github.com/node-hid/node-hid)

```javascript
import {EventEmitter} from 'Events';

declare module 'node-hid' {
  
  declare type HIDDeviceDescription = {
    vendorId: number;
    productId: number;
    path: string;
  }

  declare function devices(): Array<HIDDeviceDescription>;
  
  declare class HID extends EventEmitter {
    constructor(path: string): void;
    write(buffer: Array<number>): void;
    pause(): void;
    resume(): void;
    close(): void;
  } 
}
```

This is in my code

```javascript
var device = new HID.HID("123");
device.foobar();
```

`device.foobar` should throw an error, but it doesn't. Removing `extends EventEmitter` fixes that, but then, I cannot use EventEmitter's functions.

runn1ng:
hm, it seems the definitions in flow-typed are using `extends events$EventEmitter`. I don't entirely understand why, but when it works for others, it should work for me :)

And yes, it does work. You can close this issue since it's probably a duplicate of another existing issue.

avikchaudhuri:
@runn1ng Are you saying that the above code typechecks differently when you use `EventEmitter` after import vs. the global (but intended to be private to the lib file) `events$EventEmitter`? I don't think we support importing from one lib file to define another yet. So it's kind of expected.

But this is also kind of a missing feature...

runn1ng:
Yep, that's what I am saying. And with EventEmitter after import, it checks
incorrectly.

So yeah it's probably just a missing feature, but then it should just throw
error and not silently accept.

Issue #1153, opened 2015-12-03
Flow 0.19 won't recognize EventEmitter class if an interface file references it too.


philikon:
Consider this small reproducible test case:

**.flowconfig**

```
[libs]
interface.js
```

**interface.js**

``` js
/* @flow */

import type { EventEmitter } from 'events';

declare module 'foobar' {
  declare class Foobar extends EventEmitter {}
};
```

**index.js**

``` js
/* @flow */
import { EventEmitter } from 'events';

class Example extends EventEmitter {}
```

Flow 0.18.x gave no errors, but with Flow 0.19.x I get:

```
index.js:2
  2: import { EventEmitter } from 'events';
              ^^^^^^^^^^^^ "EventEmitter" export of "events". Property not found in
  2: import { EventEmitter } from 'events';
                                  ^^^^^^^^ exports of "events"


Found 1 error
```

Am I doing something wrong? This seems like a regression to me.

philikon:
(FWIW, if the interface file is not part of the flow config, there are no errors...)

philikon:
This problem still persists, btw.

pronebird:
See #2201 and use correct import

PR #2201, opened 2016-08-06
[fix] require('events') isn't recognized as the 'EventEmitter' class

### LINKED ISSUE #604: Bug: require('events') isn't recognized as the `EventEmitter` class
In node, `require('events') === require('events').EventEmitter`

However, flow thinks that `require('events')` is just a module declaration with a single member, `EventEmitter`.

The following (perfectly good) code raises a flow error:

```js
var EventEmitter = require('events');

class Awesome extends EventEmitter {
}
```

```
Commonjs exports of "events"
this type is incompatible with
  module events
```

This code raises no errors, however:

```js
var RealEventEmitter = require('events').EventEmitter;

class Awesomer extends RealEventEmitter {
}
```

As seen on tryflow: http://tryflow.org/#0986989fbb749f0a3801a71d150130aa

zacharygolba:
Closes #604

--

```javascript
// Also this PR should fix the following issues if the type import is changed...

// From
import type { EventEmitter } from 'events';

// To
import type EventEmitter from 'events';
```

#1153, #2102, #2130

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -588,7 +588,11 @@ declare class events$EventEmitter {
 
 declare module "events" {
   // TODO: See the comment above the events$EventEmitter declaration
-  declare var EventEmitter: typeof events$EventEmitter;
+  declare class EventEmitter extends events$EventEmitter {
+    static EventEmitter: typeof EventEmitter;
+  }
+
+  declare var exports: typeof EventEmitter;
 }
 
 declare class domain$Domain extends events$EventEmitter {


lib/node.js
```diff
@@ -588,7 +588,11 @@ declare class events$EventEmitter {
 
 declare module "events" {
   // TODO: See the comment above the events$EventEmitter declaration
-  declare var EventEmitter: typeof events$EventEmitter;
+  declare class EventEmitter extends events$EventEmitter {
+    static EventEmitter: typeof EventEmitter;
```
avikchaudhuri: What's the line for?

zacharygolba: It's to ensure that `EventEmitter` has a circular dependency of `EventEmitter`. Without it, things like `require('events').EventEmitter` would not work. This mirrors the `EventEmitter` interface in node.

screen shot 2016-08-06 at 8 30 25 am

avikchaudhuri:
@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3725689) to review internal test results.

PR #1299, opened 2016-01-22
Adding KeyboardEvents to EventTarget

runn1ng:
Fixing #1298.

This is a solution that *works*, but I somehow feel it's *wrong*. I hope someone does something, that feels cleaner to me :)

I fixed it in a best way I can :)

PR #2205, opened 2016-08-07
Mouse Events

danharper:
I've added these to match the Keyboard events added in #1299.

(I'm not sure if there's tests I need to update for this, looking at the previous PR I think not).

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -90,21 +90,28 @@ declare class DOMError {
 
 type EventHandler = (event: Event) => mixed
 type EventListener = {handleEvent: EventHandler} | EventHandler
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks for the contribution! This looks right to me :)

Since line numbers are included in the expected test output, and this change pushed down a few lines, the expected output needs to be updated. Don't worry though, I'll take care of that on our side of things.

@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3682882) to review internal test results.

Issue #1694, opened 2016-04-20
Flow complains about missing props annotation for exported classes with constructors


mikelambert:
In this code, the `TestPasses` works fine, but flow complains about `TestBroken` with:
>  10:   constructor(props) {super(props);}
>                    ^^^^^ parameter `props`. Missing annotation

The only difference is in the `export` keyword. From my reading of the [docs](http://flowtype.org/docs/react.html), I'd assume both should work fine in the presence of the `props: {}`. The minimal repro case needs (1) to be exported, and (2) a constructor that takes props. It also errors on non-empty prop type definitions as well.

```
/* @flow */
'use strict';
import React from 'react-native';
class TestPasses extends React.Component {
  props: {};
  constructor(props) {super(props);}
}
export class TestBroken extends React.Component {
  props: {};
  constructor(props) {super(props);}
}
```

mikelambert:
Oh, and this is happening with Flow 0.23.0, and occurs on a fresh 'flow init'.

mikelambert:
Also occurs on this slightly more minimal test case (didn't know if it was special-casing React Components for the `props` type definitions):
```
/* @flow */
'use strict';
class TestPasses extends Object {
  props: {};
  constructor(props) {}
}
export class TestBroken extends Object {
  props: {};
  constructor(props) {}
}
```

mikelambert:
Unless this is not a bug, and I need to define the props as an actual Type, and then use it to annotate the constructor's props? (As seen in http://sitr.us/2015/05/31/type-checking-react-with-flow.html )

The official Flow docs page for React seemed to suggest my approach above was correct, though... ( ie, towards the bottom of http://flowtype.org/docs/react.html in the "Defining components as React.Component subclasses" section (which incidentally, has a header-formatting issue )

cknitt:
+1

rh389:
+1 

I'm also using react native, if that makes any difference, but I'm using `import React from 'react'` separate from the RN imports.

desmond1121:
+1

McFlurriez:
+1

cristian-sima:
+1

acusti:
[...]

PR #2213, opened 2016-08-08
Add method signature annotation for class example

acusti:
Addresses the confusion created by the current `React.Component` subclass example and described in #1694 

Could also create a type for `state` as well to keep it consistent. If this isn’t desired and it’s preferred to keep the example with the inline type declaration for the `props` instance property, I can instead add a note after the example warning users that any exported component will require this kind of annotation.

When I was trying to figure this out in my own project, I had hoped that there would be a way to reference the inline `props` annotation in the `constructor` method signature, but couldn’t figure it out. If that’s possible in a nice way, I think changing this example to use that technique would be best.

### DIFF
--- a/website/docs/ref/react.doc.js
+++ b/website/docs/ref/react.doc.js
@@ -130,13 +130,14 @@ const Counter = React.createClass({
   and more. With class-based components, you can specify the components'
   props, default props, and state using Flow's annotation syntax.
 */
+type Props = {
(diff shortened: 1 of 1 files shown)


facebook-github-bot-9:

facebook-github-bot-0:

thejameskyle:
@facebook-github-bot import

PR #2217, opened 2016-08-08
Limit length of type hints

nmote:
Run on this file here:
https://github.com/facebook/nuclide/blob/e679970e6a20af834e9862fb83e49a4abaf6549c/pkg/nuclide-reprint-js/lib/printers/complex/printBinaryExpression.js

`flow type-at-pos printBinaryExpression.js 43 13`

Before this change, Flow printed so much text that my tmux session was
overwhelmed for a couple minutes. When Nuclide called it, Atom would crash. Now,
it still takes 15 seconds to run but the output is smaller and Atom doesn't
crash when Nuclide calls `type-at-pos`. Flow features are unavailable while the
type is being computed, but that's better than the entire editor crashing.

### DIFF
--- a/src/common/utils_js.ml
+++ b/src/common/utils_js.ml
@@ -173,3 +173,9 @@ let typo_suggestions =
       | _ -> 3
     in
     fst (List.fold_left (fold_results limit name) ([], max_int) possible_names)
+
+let truncate_string str max_len =
+  if (String.length str) <= max_len then
+    str
+  else
+    String.sub str 0 max_len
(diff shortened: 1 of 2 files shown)


nmote:
The build is broken. Looks like line numbers changes in https://github.com/facebook/flow/commit/284994aaface54f17f44cb4bc4150617209d6b5b which is causing a test to fail. It's unrelated to this change.

avikchaudhuri:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3699028) to review internal test results.

nmote:
Abandoning in favor of 9ead7390d79d4690463a10581a46578671bffb7d

PR #2234, opened 2016-08-11
add Google Analytics script

lacker:
We already set up a Google Analytics account a long time ago, but it seems we aren't actually using it from the Flow website. This PR adds the right script tag.


thejameskyle:
Can I get access to this Google Analytics account?

thejameskyle:
Actually... can everyone on the Flow team get access?

mroch:
https://github.com/facebook/flow/blob/b7af0c689e6ff5690d4a55898a5daea95959128a/website/_assets/js/analytics.js

PR #2237, opened 2016-08-12
Fixes for compatibility with OCaml 4.04

damiendoligez:


gabelevi:
Thanks for the PR @damiendoligez ! It's awesome that OCaml 4.04 catches the `CAMLlocal1` misuse!

As for `warn_error_A`, I think I may avoid that part of the PR for now. I see that this has caused trouble for opam since we were missing an upper bound, which you've very kindly fixed - https://github.com/ocaml/opam-repository/pull/7189/files . Ideally, we'd have a development build that would error on warnings and a production build (which opam would use) which would not. It shouldn't be too hard to set up, I just don't have the time at the moment. I'll file an issue (https://github.com/facebook/flow/issues/2258). In the meantime, we need ocamlbuild to error on warnings, since our internal buck build does that and we'd like PRs to fail earlier rather than later.

gabelevi:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3722553) to review internal test results.

damiendoligez:
>Ideally, we'd have a development build that would error on warnings and a production build (which opam would use) which would not. It shouldn't be too hard to set up, I just don't have the time at the moment.

This is good practice. Don't forget to do it when you have the time.

PR #2238, opened 2016-08-12
Declare the correct return type for Buffer#toJSON in node. Fixes #1647

### LINKED ISSUE #1647: node-js Buffer.toJSON() type is Array<number> but should be an Object
In lib/node.js we can see that Buffer.toJSON is returning an Array<number> type:
https://github.com/facebook/flow/blob/master/lib/node.js#L41

but node-js documentation says that it should be an Object type:
https://nodejs.org/api/buffer.html#buffer_buf_tojson

aackerman:


gabelevi:
Looks good to me - tests need to be rerecorded but I can handle that

gabelevi:
@facebook-github-bot import

facebook-github-bot-1:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3722599) to review internal test results.

PR #2240, opened 2016-08-12
Fix `flow start` on Windows 7

### LINKED ISSUE #2183: 'flow.exe' unhandled unix exception
Thanks for supporting windows! Great job!

Unfortunately, I cannot start the flow server with the default/start command. When I try to run

`flow.exe` or `flow.exe start`

flow crashes with an unhandled unix exception:

> Unhandled exception: Unix.Unix_error(12, "set_close_on_exec", "")

Nevertheless, starting the server with

`flow.exe server`

works and after that calling `flow.exe` works as well.

OS is Windows 7 64bit.

gabelevi:
This is a somewhat unsatisfying fix for a strange problem. Here's the general idea:

1. `flow start` calls `Unix.create_process` to create the server. This api allows you to specify the stdio for the new process, so we're using the log file for both stdout and stderr
2. On OS X and Linux this works fine
3. On Windows 10 (and I think 8), `flow start` would leak its stdout/stderr to the new process. This meant when node execs `flow start` it would hang until the server died, since it never would get the `"close"` event (which happens when the process has closed stdio).
4. The fix for Windows 10 (and I think 8) was to use `Unix.set_close_on_exec` to avoid leaking stdout/stderr to the new process. This fixed the node issue and didn't break anything on Linux or OS X
5. For some reason, `Unix.set_close_on_exec` doesn't work on Windows 7. Digging into the C code didn't seem to clear things up. However, Windows 7 doesn't seem to leak stdout/stderr to the new process anyway. So just catching the error and doing nothing seems to work well enough

Fixes #2183

### DIFF
--- a/src/server/serverFunctors.ml
+++ b/src/server/serverFunctors.ml
@@ -385,9 +385,26 @@ end = struct
     let log_file = Path.to_string (Options.log_file options) in
     let log_fd = open_log_file options in
     let config_file = Server_files_js.config_file root in
-    (* Avoid leaking stdout and stderr to the server *)
-    Unix.(set_close_on_exec stdout);
-    Unix.(set_close_on_exec stderr);
+    (* Daemon.spawn is creating a new process with log_fd as both the stdout
+     * and stderr. We are NOT leaking stdout and stderr. But the Windows
+     * implementation of OCaml does leak stdout and stderr. This means any process
+     * that waits for `flow start`'s stdout and stderr to close might wait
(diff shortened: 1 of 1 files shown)


gabelevi:
@facebook-github-bot import

facebook-github-bot-7:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3710006) to review internal test results.

Issue #2219, opened 2016-08-09
Better support for promises


vkurchatkin:
There is lot of room for improvement, here is a list of things that comes to mind:
1. Support alternative promise implementations (https://github.com/facebook/flow/issues/2171)
2. Support subclassing promises
3. Support iterables in `Promise.all`

SamChou19815:
2 and 3 are supported, 1 might be possible but not recommended

PR #2248, opened 2016-08-14
Better support for promises, WIP

vkurchatkin:
General idea is to make promise type more permissive. The most common problem is using custom promise implementations (such as Bluebird), in particular with `async-await`.

Typedefs for popular promise libraries should probably be adjusted to properly implement `$Thenable`, for example, this won't work as is: https://github.com/marudor/flowInterfaces/blob/master/packages/iflow-bluebird/index.js.flow#L89

To-do list:

 - [x] - support `$Thenable` in `Promise.all`;
 - [x] - support iterables other than Array in `Promise.all`;
 - [ ] - support subclassing promises;
 - [ ] - figure out why `Promise<T> | $Thenable<T> | T` is required in `$await` (probably a bug in union type handling).

See: https://github.com/facebook/flow/issues/2219

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -494,29 +494,36 @@ declare class WeakSet<T: Object> {
    cf. https://github.com/borisyankov/DefinitelyTyped/blob/master/es6-promises/es6-promises.d.ts
 */
 
+interface $Thenable<+R> {
+  then<U>(
+    onFulfill?: (value: R) => $Thenable<U> | U,
+    onReject?: (error: any) => $Thenable<U> | U
+  ): $Thenable<U>;
+}
+
 declare class Promise<+R> {
     constructor(callback: (
-      resolve: (result: Promise<R> | R) => void,
+      resolve: (result: $Thenable<R> | R) => void,
       reject:  (error: any) => void
     ) => mixed): void;
 
     then<U>(
-      onFulfill?: (value: R) => Promise<U> | U,
-      onReject?: (error: any) => Promise<U> | U
+      onFulfill?: (value: R) => $Thenable<U> | U,
+      onReject?: (error: any) => $Thenable<U> | U
     ): Promise<U>;
 
     catch<U>(
-      onReject?: (error: any) => ?Promise<U> | U
+      onReject?: (error: any) => ?$Thenable<U> | U
     ): Promise<U>;
 
-    static resolve<T>(object: Promise<T> | T): Promise<T>;
+    static resolve<T>(object: $Thenable<T> | T): Promise<T>;
     static reject<T>(error?: any): Promise<T>;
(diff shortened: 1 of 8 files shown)


vkurchatkin:
Turns out `Promise.all` uses type of `$await` under the hood, so it just works

vkurchatkin:
Added hacky support for iterables, seems to work, but I'm not sure that implementation is acceptable

samwgoldman:
So cool! Thanks for taking this on. Small recommendation: I think it will be easier to break this up into multiple PRs if you can. In particular, subclassing Promises and supporting iterables in Promise.all should be orthogonal enough to supporting Promise-like things that we can land them all separately. Doing the work in smaller chunks makes it much easier to review :)

Regarding the issue where both `Promise<T>` and `$Thenable<T>` are required in the union—you're right that we shouldn't require this. If I understand the issue, if you omit `Promise<T>` then we will run into "cannot choose branch" errors which require annotations.

To summarize the issue, when a union appears as an upper bound, we block until there is enough type information to choose a path. The libdefs are not merged in yet, so if the information we need to decide is in a libdef, we actually can't make the choice and we ask for annotations.

This class of issue has come up before, but we have only come up with hacky ways to solve it. See 0a08916ceb2d37f742213715c438332886f7d123 for an example. I'm not sure if a similar hack would be suitable here, or if this is a forcing function to resolve the underlying issue instead.

cc @avikchaudhuri who might have some ideas for how to unblock you.

avikchaudhuri:
@samwgoldman The issue you link to IIRC relates to choosing the wrong branch (waiting for the wrong case to resolve is mistaken as success). @vkurchatkin What's the problem you're running into, exactly, when you just have `$Thenable<T> | T`? (You're of course right the extra case shouldn't be needed, by manual inspection `Promise<R>` is always a subtype of `$Thenable<R>`, yes?)

vkurchatkin:
> So cool! Thanks for taking this on. Small recommendation: I think it will be easier to break this up into multiple PRs if you can

Ok, I'll close this one and open separate PRs

> if you omit Promise<T> then we will run into "cannot choose branch" errors which require annotations

Exactly, that's what happens. Full error message:

```
union type. See lib: core.js:528
  Case 1 may work:
  type application of identifier `$Thenable`. See lib: core.js:528
  But if it doesn't, case 2 looks promising too:
   16:   return foo();
                ^^^^^ type parameter `T` of async return
  Please provide additional annotation(s) to determine whether case 1 works (or consider merging it with case 2):
   12:   return 42;
         ^^^^^^^^^^ async return
```

>  by manual inspection Promise<R> is always a subtype of $Thenable<R>, yes?

Yes, there is a test to verify this.

PR #2252, opened 2016-08-15
Builtin support for requiring resource files (css, jpg, etc)

gabelevi:
Often, people require resource files from their JavaScript. Until now, this has
mainly been handled by using the `module.name_mapper` option in the
`./flowconfig` like so:

```
module.name_mapper='^(.*).css$' -> 'flow/css'
module.name_mapper='^(.*).(jpg\|png\|gif\|eot\|svg\|ttf\|woff\|woff2\|mp4\|webm)$' -> 'flow/file'
```

Where `flow/css.js` would export a css mock and `flow/file.js` would export
`string`. This would let you write

```
const myCSS = require('./myCSS.css');
const myJPG = require('./myJPG.jpg');
```

This change teaches Flow that it's ok to import and require these resource
files. By default, requiring a resource file returns a `string`, except for css
which returns `void`. This hopefully will let code that requires resource files
to work with Flow without the need for configuration.

### DIFF
--- a/src/typing/type_inference_js.ml
+++ b/src/typing/type_inference_js.ml
@@ -168,3 +168,39 @@ let infer_lib_file ~metadata ~exclude_syms file statements comments =
   ));
 
   cx, SMap.keys Scope.(module_scope.entries)
+
+let infer_resource_file ~metadata ~filename ~module_name =
+  Flow_js.Cache.clear();
+
+  let cx = Flow_js.fresh_context metadata filename module_name in
+  let exported_module_name = Modulename.to_string module_name in
+
+  let reason_exports_module =
+    Reason.reason_of_string (
+      Utils.spf "exports of module `%s`" exported_module_name) in
+  let reason_loc = Loc.({ none with source = Some filename }) in
+
+  let filename = Loc.(match filename with
+  | ResourceFile filename -> filename
+  | SourceFile _
+  | LibFile _
+  | Builtins
+  | JsonFile _ -> failwith "Resource file is not a resource file?!") in
+  let reason, exports_t = match Utils.extension_of_filename filename with
+  | Some ".css" ->
+      let reason = Reason.mk_reason
+        "Flow assumes requiring a .css file returns undefined"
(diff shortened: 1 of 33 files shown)


gabelevi:
@facebook-github-bot import

facebook-github-bot-9:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3716923) to review internal test results.

thejameskyle:
cc @gaearon

src/typing/type_inference_js.ml
```diff
@@ -168,3 +168,39 @@ let infer_lib_file ~metadata ~exclude_syms file statements comments =
...
+  let filename = Loc.(match filename with
+  | ResourceFile filename -> filename
+  | SourceFile _
+  | LibFile _
+  | Builtins
+  | JsonFile _ -> failwith "Resource file is not a resource file?!") in
+  let reason, exports_t = match Utils.extension_of_filename filename with
+  | Some ".css" ->
+      let reason = Reason.mk_reason
+        "Flow assumes requiring a .css file returns undefined"
```
gaearon: Many people use CSS modules which export an object with string values for each key. Can we give a hint on how to configure those for people who want them?

thejameskyle: Maybe it makes sense to make `{ [key: string]: any }` the default since the people that will care about it will use it that way and the ones who don't won't use it at all.

gaearon: Agree.

gabelevi: Ah, just saw this. I'll switch to `Object` by default for `.css` then.

There was a little disagreement about the configuration, so as a compromise I landed this PR without any new configuration options. We'll figure those out afterwards.

`module.name_mapper` does still work, though, so if you need something special for css you can still do 

```
[options]
module.name_mapper='^(.*).css$' -> 'flow/css'
```

Where `flow/css.js` exports whatever type you'd like.

Issue #999, opened 2015-10-27
Move emacs plugin to a dedicated repository


dgellow:
Could you create an official repository for flow-types.el? Similar to the [facebook/vim-flow](https://github.com/facebook/vim-flow) plugin.
I think it would improve visibility and make it easier to work on it together.

FossiFoo:
this, please. i would contribute by making a proper mode and adding my company-flowtype auto-complete. having this in the root dir of flow makes it hard to properly submit PRs.

samwgoldman:
OK cool. If that's holding you back from contributing, let's do it.

FossiFoo:
nice :)

FossiFoo:
any progress? is there anybody i can bug with this?

avikchaudhuri:
I can take this up

dgellow:
@avikchaudhuri @samwgoldman So is there something in progress related to this issue?

avikchaudhuri:
Yeah, expect something to materialize later this week.

dgellow:
@avikchaudhuri ping?

FossiFoo:
bump

dgellow:
The emacs package has been moved to https://github.com/flowtype/flow-for-emacs/. I close the issue.
See https://github.com/facebook/flow/pull/2731 for details.

PR #2261, opened 2016-08-16
Emacs plugin rewrite

### LINKED ISSUE #724: Use --json in emacs plugin
The [emacs plugin](https://github.com/facebook/flow/blob/master/flow-types.el) should call `flow --json` instead of parsing the CLI output. Right now we have to maintain a special format for emacs, matching the [caml format](https://github.com/emacs-mirror/emacs/blob/cdd5e86967e0a3d23da3079dc5b17ace699e66b3/lisp/progmodes/compile.el#L159) built into emacs.

Instead, it would be nice if all plugins used `--json` and did their own formatting in whatever way made the most sense for them. Relatedly, see https://github.com/facebook/vim-flow/issues/15.

afonso360:
Instead of defining a few functions and using global-set-key possibly overriding users keybindings, I've created a proper minor mode with local keybindings and prefixes.

I also added company-mode support for autocompletion, it works but some work is still required.

As we init flow servers we keep track of their existance, this is still not used but eventually I would like to have a variable to enforce a flow server before executing each command and the possibility to kill all flow servers tracked by Emacs

Fixes #724 except for flow-types-fill-types

### DIFF
--- a/flow-types.el
+++ b/flow-types.el
@@ -48,7 +48,7 @@
 ;; this is used to cleanup on emacs exit
 (defvar flow-types--instances '())
 
-(defvar flow-types-mode-map
+(defvar flow-types-command-map
   (let ((map (make-sparse-keymap)))
     (define-key map (kbd "t") #'flow-types-type-at-pos)
     (define-key map (kbd "j") #'flow-types-jump-def)
@@ -64,10 +64,10 @@ with the flow static typechecker from facebook"
   ;; Mode line indicator
   " Flow"
   ;; keymap to be used
-  ;; TODO: This needs to be changed to have a prefix
-  :keymap flow-types-mode-map
+  ;; ;; TODO: This needs to be changed to have a prefix
+  ;; :keymap flow-types-mode-map
   :group 'flow-types
-  (local-set-key flow-types-keymap-prefix flow-types-mode-map))
+  (local-set-key flow-types-keymap-prefix flow-types-command-map))
 
 (define-globalized-minor-mode global-flow-types-mode flow-types-mode
   (lambda () (flow-types-mode 1)))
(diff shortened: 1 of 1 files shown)


facebook-github-bot-2:

facebook-github-bot-6:

avikchaudhuri:
Thanks @afonso360! I'll take a look. Meanwhile I modernized the existing plugin a bit, but didn't go as far as creating a minor-mode. This looks pretty cool. Would be nice to merge these efforts and launch a separate project to maintain the plugin, WDYT?

afonso360:
Sounds like a good idea, also closes  #999

FossiFoo:
yay. the more the merrier. this looks almost identical to the minor-mode we whipped up.
i'd be happy to contribute some various functionality too.

but could somebody from facebook, please, please, pretty please open up a repo for this? see #999

FossiFoo:
my current version is here btw: https://github.com/FossiFoo/.emacs.d/blob/master/site-lisp/flowtype-mode/flowtype-mode.el

avikchaudhuri:
We're in the process of creating a repo and inviting contributions. Thanks
for your patience!

avikchaudhuri:
@afonso360 Could you please share install instructions? I'm using GNU Emacs 24.5.1. For some reason after `M-x global-flow-types-mode` whenever I type the character `f` I get "Symbol's function definition is void: string-of-region." When I define that function I get the same error for `flow-binary`. Are you relying on the existing definitions in `flow-types.el` somehow?

avikchaudhuri:
@FossiFoo Could you also please share install instructions so that I can play around? There are a bunch of dependencies. Which version of emacs is your mode built for?

afonso360:
@avikchaudhuri Yes I was relying on those functions because I didn't want to remove existing functionality, I've pushed a commit that should fix the missing function definitions and prevent the keybindings from leaking. (they were being bound without the prefix).

As for install instructions, placing this on the config works, but it enables flow globally, which might not be the best solution.
```
(add-to-list 'load-path "~/git/flow/")
(require 'flow-types)
(global-flow-types-mode)
```

If you don't want flow enabled globally, replacing global-flow-types-mode with
```
(add-hook 'web-mode-hook 'flow-types-mode)
```
Will only enable flow-types on the major mode Web-mode.

FossiFoo:
Any progress on the repo?

I'm not aware of any special things that need to be done or any special emacs version required. I'm currently using 24.5.2 but it also works (maybe with patches?) with spacemacs and aquamacs. All "needed" packages can be fetched from MELPA, but it should run fine without flycheck or company for example. 

Of course this is so far not polished at all and for my private use, so i didn't clean up for a while. :)

gabelevi:
Sorry for the delay - I'm pinging @avikchaudhuri and seeing what's blocking us from publishing the emacs repository. Thanks for your patience!

FossiFoo:
cool. a repo with any start would still be a great thing... :)

gabelevi:
Ok! It's up! Let's all move to https://github.com/flowtype/flow-for-emacs/ !

As for this PR, I'm going to close it out. My apologies again for the delay in getting that repo published!

FossiFoo:
can you then please also remove the emacs file from this repo? it's just confusing to have that around if the official source is now elsewhere.

gabelevi:
@FossiFoo - sure! Just created https://github.com/facebook/flow/pull/2731. Thanks for the reminder!

PR #2265, opened 2016-08-16
Add support for iterables in Promise.all

vkurchatkin:


bhosmer:
@vkurchatkin thanks for this! One immediate comment, then I'll defer to @samwgoldman who built the original `Promise.all`/`TupleMapT` functionality for a deeper look. :)

src/typing/flow_js.ml
```diff
@@ -3696,6 +3696,15 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       ) in
       rec_flow_t cx trace (ArrT (reason, f t, List.map f ts), tout)
 
+    | _, TupleMapT (reason, funt, tout) ->
```
bhosmer: I wonder if this wildcard lower bound interacts badly with the assumptions made [here](https://github.com/facebook/flow/blob/master/src/typing/flow_js.ml#L2821-L2832).

vkurchatkin: Hmm, I guess it does, but not sure if it's a bad thing

bhosmer: Yeah, I think you're right. I was worried about the `VoidT` case being captured by the new rule instead of falling through, but looking at the tests I think it's handled properly - the new error is different but fine. 

Personally I'd ask if you could update (or simply remove) the comments in that `CustomFun` rule which are now out of date, and update the ones in `tests/promises/all` which say "expected array instead of", to reflect the newer more general behavior. Otherwise it looks good to me! @samwgoldman WDYT, good to go mod comments?

vkurchatkin: > the new error is different but fine.

It seems that it wasn't working as exepected anyway (no "(too few arguments)" part)

tests/promises/promises.exp
```diff
@@ -36,13 +36,17 @@ all.js:19
 
 all.js:24
  24: Promise.all(); // Error: expected array instead of undefined (too few arguments)
-     ^^^^^^^^^^^^^ call of method `all`. Expected array instead of possibly undefined value
+     ^^^^^^^^^^^^^ call of method `all`
+Library type error:. See lib: core.js:437
+property `@@iterator` of $Iterable. Property not found in possibly undefined value. See lib: core.js:437
  24: Promise.all(); // Error: expected array instead of undefined (too few arguments)
      ^^^^^^^^^^^^^ undefined (too few arguments)
```
bhosmer: Here's the "too few arguments" part, nicely unchanged. :)

bhosmer:
@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3759069) to review internal test results.

PR #2266, opened 2016-08-17
Add type declarations for web worker globals and the importScripts me…

### LINKED ISSUE #726: [REQUEST] Web workers
Hey,

I noticed that while I can test for basic browser existence, Flow has no knowledge of `importScripts` function for web workers.

aackerman:
…thod. Fixes #726 

I had issues getting the tests to build locally, so I'm unsure about the exact validity of the changes.

You can checkout the output of what I was running into here. I was able to build a few days ago but I can't after pulling master. https://gist.github.com/aackerman/eba4d0f3fa67baaf38c5bc306f0e2c35

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -51,31 +51,34 @@ type Gamepad = {
   mapping: string;
   timestamp: number;
 }
-declare class Navigator {
-    appCodeName: 'Mozilla';
+declare class NavigatorCommon {
     appName: 'Netscape';
     appVersion: string;
+    platform: string;
+    userAgent: string;
+    language: string;
+    languages: Array<string>;
+    onLine: boolean;
+    hardwareConcurrency: number;
+}
+
+declare class Navigator mixins NavigatorCommon {
+    appCodeName: 'Mozilla';
     buildID: string;
     cookieEnabled: boolean;
     doNotTrack?: any;
     geolocation: Geolocation;
     mediaDevices?: Object;
-    hardwareConcurrency: number;
     javaEnabled: Function;
-    language: string;
-    languages: Array<string>;
     maxTouchPoints: number;
     mediaDevices: Object;
     mimeTypes: MimeTypeArray;
-    onLine: boolean;
     oscpu: string;
     permissions: any;
-    platform: string;
     plugins: PluginArray;
     product: 'Gecko';
     productSub: '20030107' | '20100101';
     serviceWorker?: Object;
-    userAgent: string;
     vendor: '' | 'Google Inc.' | 'Apple Computer, Inc';
     vendorSub: '';
     getBattery?: () => Promise<Gamepad>;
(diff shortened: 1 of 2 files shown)


bhosmer:
Hi Aaron - thanks for this! Taking a look through the declarations now. (Sorry about the build issues - maybe a `libelf` version problem? Might be worth updating per the "Building Flow" section of the readme, but if the problem persists LMK and I'll dig further.)

lib/bom.js
```diff
@@ -261,6 +261,61 @@ declare class Worker extends EventTarget {
...
+    onerror: (ev: Event) => any;
+    onlanguagechange: (ev: Event) => any;
+    onoffline: (ev: Event) => any;
+    ononline: (ev: Event) => any;
+    onrejectionhandled: (ev: PromiseRejectionEvent) => any;
+    onunhandledrejection: (ev: PromiseRejectionEvent) => any;
+}
+
+declare class DedicatedWorkerGlobalScope extends WorkerGlobalScope {
+    close(): void;
```
bhosmer: Any reason why `close()` is declared in the subs rather than `WorkerGlobalScope` itself? (FYI I'm using [this](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope) as a reference)

aackerman: I was working from the [spec document](https://html.spec.whatwg.org/multipage/workers.html#the-workerglobalscope-common-interface) as a reference and it happened to have `close()` on the subs instead of the shared interface.  I think it would be fine to change though since they're the same signature. Is that something you would like to see changed?

bhosmer: Ha, interesting! Wonder why it's that way in the spec... it's not a big deal, but I'd say it's probably worth changing if you're in there doing other stuff.

lib/bom.js
```diff
@@ -261,6 +261,61 @@ declare class Worker extends EventTarget {
...
+    protocol: string;
+    host: string;
+    hostname: string;
+    port: string;
+    pathname: string;
+    search: string;
+    hash: string;
+}
+
+declare class WorkerNavigator {
```
bhosmer: Hey, would it be possible to factor these props into (say) `NavigatorCommon` and then have both this and `Navigator` extend it? It'd be cumbersome to actually match the spec and break them up into several different superinterfaces, but a single block of the common properties should be simple enough, I think... 

LMK if I'm missing a subtlety though - don't want to make things harder than they need to be. :)

aackerman:
I was able to get the build working again after wiping out my vagrant box and reinstalling deps FYI.

bhosmer:
Awesome, thanks Aaron! ...and good news on the build. :)

@facebook-github-bot import

facebook-github-bot-1:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3758393) to review internal test results.

PR #2276, opened 2016-08-17
Avoid killing Flow server on package.json changes

nmote:
Flow only cares about the `main` and `name` fields in `package.json`. However if any `package.json` changes at all, the Flow server dies. This is pretty frustrating and leads to a lot of lost time for Nuclide devs (since we have `package.json` files all over the place). This change makes it so the Flow server only dies if a relevant change is detected.

This works (I've tested it manually), but I'm not super confident that it's the right way to do it:

* I have to duplicate the options computation in order to call `reparse`. Is there somewhere a utility could live that would be like `reparse` except it grabs these options itself? Would that belong in `Parsing_service_js` or do you want to keep that from taking a dependency on `Options`?
* I'm not sure what the implications are of reparsing a file. Since it modifies global state it's hard to know. Is that a safe operation here?

Once we get the implementation strategy sorted out I'll look into adding automated tests.

CC @zertosh

### DIFF
--- a/src/server/server.ml
+++ b/src/server/server.ml
@@ -479,16 +479,32 @@ module FlowProgram : Server.SERVER_PROGRAM = struct
       FlowExitStatus.(exit Server_out_of_date)
     end;
 
-    (* Die if a package.json changed *)
-    let modified_packages = SSet.filter (fun f ->
+    let is_incompatible filename_str =
+      let filename = Loc.JsonFile filename_str in
+      let filename_set = FilenameSet.singleton filename in
+      let ast =
+        let _ = Parsing_service_js.reparse_with_defaults
+          options
+          (* workers *) None
+          filename_set
+        in
+        Parsing_service_js.get_ast_unsafe filename
+      in
+      Module_js.package_incompatible filename_str ast
+    in
+
+    (* Die if a package.json changed in an incompatible way *)
+    let incompatible_packages = SSet.filter (fun f ->
       (String_utils.string_starts_with f sroot ||
         Files.is_included options f)
-      && (Filename.basename f) = "package.json" && want f
+      && (Filename.basename f) = "package.json"
+      && want f
+      && is_incompatible f
     ) updates in
-    if not (SSet.is_empty modified_packages)
+    if not (SSet.is_empty incompatible_packages)
     then begin
       Flow_logger.log "Status: Error";
-      SSet.iter (Flow_logger.log "Modified package: %s") modified_packages;
+      SSet.iter (Flow_logger.log "Modified package: %s") incompatible_packages;
       Flow_logger.log
         "Packages changed in an incompatible way. Exiting.\n%!";
       FlowExitStatus.(exit Server_out_of_date)
--- a/src/services/inference/module_js.ml
+++ b/src/services/inference/module_js.ml
@@ -45,6 +45,9 @@ type info = {
 
 type mode = ModuleMode_Checked | ModuleMode_Weak | ModuleMode_Unchecked
 
+let relevant_package_keys = ["name"; "main"]
+let is_relevant_key key = List.mem key relevant_package_keys
+
 let choose_provider_and_warn_about_duplicates =
   let is_flow_ext file = Loc.check_suffix file Files.flow_ext in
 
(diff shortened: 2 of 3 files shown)


avikchaudhuri:
Looks promising! I'll read this more carefully later today. Thanks for fixing this glaring inefficiency!

zertosh:
This is going to save a ton of time for Nuclide devs 👍

nmote:
@avikchaudhuri ping 😄

avikchaudhuri:
@nmote Thanks for your patience. Requesting a couple of changes.

Here's a summary of what used to happen:

* ParserHeap maintains a file -> ast map for all files, including `package.json` files.
* PackageHeap maintains a package_file -> tokens map for only `package.json` files.
* Since we died on any change to `package.json` files, the invariant `Module_js.get_package_keys package_file (ParserHeap.find_unsafe package_file) = PackageHeap.find_unsafe package_file` would be maintained trivially for `package.json` files.

Here's a summary of what happens now:

* ParserHeap is updated with new asts when `package.json` files change.
* PackageHeap is *not* updated as long as important tokens (`name` and `main`) don't change when `package.json` files change.
* We die only on changes to important tokens in `package.json` files. So the invariant `Module_js.get_package_keys package_file (ParserHeap.find_unsafe package_file) = PackageHeap.find_unsafe package_file` no longer holds for `package.json` files, but the weaker invariant that focuses only on important tokens in that token map holds.

I'm not sure this invariant breakage is a big deal now, since we only read the important tokens from `PackageHeap`. But it might cause unexpected results if we read some other tokens in the future. To safeguard against that I'd recommend also filtering the tokens to only store the important ones in `PackageHeap`, with a comment (in Module_js.add_package).

The other comment is on efficiency. `re_parse`, like other batch operations, uses parallelism. If multiple package files change, it would be better to batch them up before calling `re_parse`, otherwise we might pay some constant set up cost for each call without utilizing any benefits from parallelization.

Looks pretty good otherwise!

nmote:
Thanks for the response. I'll make the requested changes.

The only question I have left is about calling `reparse`. It seems a little gross that at both callsites we have to repeat the boilerplate to get all the options. Can you recommend a place to put a function that doesn't require these arguments and instead gets them from the options? Would `Parsing_service_js` be an appropriate place for such a util?

avikchaudhuri:
@nmote Sure, feel free to put it is `Parsing_service_js` for now; we could move it some place else more common in the future if it's used for more than parsing.

src/server/server.ml
```diff
@@ -410,16 +410,47 @@ module FlowProgram : Server.SERVER_PROGRAM = struct
...
     end;
 
-    (* Die if a package.json changed *)
-    let modified_packages = SSet.filter (fun f ->
+    let is_incompatible filename_str =
+      let filename = Loc.JsonFile filename_str in
+      let filename_set = FilenameSet.singleton filename in
+      let ast =
+        (* Copied these options from types_js.ml, where the other reparse callsite is.
+           It's not really ideal so I'm open to suggestions. I'm also not sure what
```
bhosmer: I definitely think we want to have this gather-options-and-call-reparse block in only one place - it's full of details that'll be prone to drifting otherwise. But it should be simple enough to factor as a helper (in `types_js`, or wherever makes sense) that takes a `FilenameSet` that we can call from here, I think.

Perf-wise I'd be shocked if the time to reparse was perceptible. But if so, it should be pretty straightforward to record the fact that the ASTs left by this check in shared memory are in fact current.

bhosmer:
@nmote @avikchaudhuri - sorry, added the line comment before seeing the thread that had ensued. :) Feel free to disregard if appropriate.

bhosmer:
@avikchaudhuri @nmote ...nope, looks like it was still a comment worth making. :)

Echoing the invariant comment also - the safe thing is to guarantee that there's nothing out of date in PackageHeap, i.e. only storing keys that will be unchanged if the server stays up. If we wanted to store more volatile keys later, we'd need to introduce a "Building package heap" section to `recheck`, but that'll have to be done carefully.

nmote:
Okay, I've made sure that we only store the relevant keys in PackageHeap, and that the server dies if and only if some of those keys are changed. I also added the reparse_with_defaults utility in a separate PR.

Regarding batching calls to re_parse: I considered it, but I think the code is a bit simpler this way and this slight inefficiency is still far better than killing the entire server so I'm confident that this is a step in the right direction, even if some inefficiencies remain.

nmote:
One thing I'm still not sure about is automated tests. Do you think they are necessary for this functionality? If so where should I put them?

I have tested manually against a test project and Nuclide's codebase. Making changes to package.json files does not kill the server unless they are changes to the main or name fields.

avikchaudhuri:
@facebook-github-bot import

Issue #1512, opened 2016-03-08
Arbitrary number properties are allowed


vkurchatkin:
This type-checks:

``` js
type O = {};

function stuff(o: O) {
  const b = o[0].ff.ffa
}
```

`o[0]` behaves as it it was declared as `any`

vkurchatkin:
This is still an issue.

JohnyDays:
I think, if you want this to type-check you have to seal the object. E.g `type O = { [key: string]: void };`

vkurchatkin:
Nope, this is definitely a bug, for example changing `[0]` to `['0']` fixes the problem

gkz:
Fixed in 80ca16f77ba81a3475dd1f7a9f4d9d9e3d71bca8

Issue #380, opened 2015-04-13
"non-string literal property keys not supported"


redchair123:
```
var x = {
    1234:"a",
    5678:"b"
}
```

Flow generates an error per integer key.

According to [11.1.5 of ECMA-262](http://www.ecma-international.org/ecma-262/5.1/#sec-11.1.5) this is a perfectly valid construct.

EDIT: @jgrund 's solution in https://github.com/facebook/flow/issues/380#issuecomment-224380551 works well

fatboysuns:
Another example when I run it on a jquery javascript.
$.ajax({
                url: url,
                dataType: 'html',
                success: function(data) {
                   //do something
                },
                error: function(data) {
// something else
                },
// This is the line Flow complains
                statusCode: {
                    404: function() {
                         alert('404');
                    }
                }
            });

gabelevi:
Number literals are annoying, since we need to canonicalize them. We should probably just bite the bullet and implement the algorithm. CC @mroch

julianacipa:
Any update on this issue?

Zaba999:
:+1: Is there any progress with this issue? I really need this one :)

jgrund:
As a really awful hack:

```
var x = {
    [1234]:"a",
    [5678]:"b"
}
```

rosskevin:
ugh.  I too need this.

StreetStrider:
The issue still bothers. Tried to typecast to something like `{[n: number]: mixed}` — not working. Even if it did, it's such a crazy workaround for such a small thing.

Niggler:
Any progress?

danpantry:
Playing devils advocate here: All keys on an object will be implicitly converted to a string. I like that this error is explicit about that behaviour (even if the error is not very well defined).

lenovouser:
Any plans to fix this? This should't throw an error

arhtudormorar:
This is working for me:

```
type PropsType = {
	[key: string]: string
}

let currencyId = 1,
    currencyCode = "RON";

let currencies: PropsType = {
	[currencyId.toString()]: currencyCode
}
```

Niggler:
We've internally decided to move away from flow.  @jgrund 's array hack works well

StreetStrider:
@Niggler, please, reopen the ticket. There're plenty of people who still interested in this. Let maintainers see the interest.

Niggler:
[...]

PR #2293, opened 2016-08-19
Fix accessing number literal properties on objects

vkurchatkin:
This uses `raw` value as string representation of number literal. This is an oversimplification, but better than nothing.

The correct way is to implement this part of the spec: http://www.ecma-international.org/ecma-262/6.0/index.html#sec-tostring-applied-to-the-number-type

Fixes: https://github.com/facebook/flow/issues/1512

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3636,7 +3636,8 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       ->
       rec_flow cx trace (key, ElemT(reason_op, l, tout, Read))
 
-    | (StrT (reason_x, Literal x), ElemT(reason_op, (ObjT _ as o), t, rw)) ->
+    | (StrT (reason_x, Literal x), ElemT(reason_op, (ObjT _ as o), t, rw))
+    | (NumT (reason_x, Literal (_, x)), ElemT(reason_op, (ObjT _ as o), t, rw)) ->
       let reason_x = replace_reason (spf "property `%s`" x) reason_x in
       let u = match rw with
       | Read -> GetPropT (reason_op, (reason_x, x), t)
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -3636,8 +3636,16 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       ->
       rec_flow cx trace (key, ElemT(reason_op, l, tout, Read))
 
-    | (StrT (reason_x, Literal x), ElemT(reason_op, (ObjT _ as o), t, rw))
-    | (NumT (reason_x, Literal (_, x)), ElemT(reason_op, (ObjT _ as o), t, rw)) ->
+    | (StrT (reason_x, Literal x), ElemT(reason_op, (ObjT _ as o), t, rw)) ->
+      let reason_x = replace_reason (spf "property `%s`" x) reason_x in
+      let u = match rw with
+      | Read -> GetPropT (reason_op, (reason_x, x), t)
+      | Write -> SetPropT (reason_op, (reason_x, x), t)
+      in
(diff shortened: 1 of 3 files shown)


vkurchatkin:
Switched to `string_of_float_trunc`, it seems to work good enough

src/typing/flow_js.ml
```diff
@@ -3644,6 +3644,15 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       in
       rec_flow cx trace (o, u)
 
+    | (NumT (reason_x, Literal (n, _)), ElemT(reason_op, (ObjT _ as o), t, rw)) ->
+      let x = string_of_float_trunc n in
```
bhosmer: Hi Vladimir - it looks like the runtime behavior of e.g. `obj[n]` is to call `toString` on `n` and use the result as a property name, so `string_of_float_trunc` doesn't seem right here - instead we would need some `js_string_of_float` which replicated the specified behavior of JS's `(number).toString`.

(Aside: with that in hand, it should be pretty straightforward to support numeric property names in object literals as well!)

vkurchatkin: Related comment: https://github.com/facebook/flow/issues/380#issuecomment-111329261

> wait until we have a credible replica of toString before going forward on this.

 Spec wording vaguely indicates that final result could be implementation-specific, so I'm not sure that it's worth it.

`string_of_float_trunc` seems to work good enough. Another option is to only allow integer literals as property keys.

bhosmer: The problem with `string_of_float_trunc` is that it would fold distinct properties together, no? 
```
> let obj = {}
< Object {}
> obj[1] = "hey"
< "hey"
> obj[1.1] = 999
< 999
```

vkurchatkin: No, it should work for this case. It just truncates trailing `.` to format integers correctly

vkurchatkin:
@bhosmer ping

samwgoldman:
I'll take this over from Basil. Thanks for being so patient.

gabelevi:
This looks good to me. I'll import, rebase it past some internal refactors, get it reviewed, and merge it. Sorry for the delay!

gabelevi:
This has turned out to be a little more complicated than anticipated. Since we support types like

```
function foo(obj: { [key: number]: string }): string {
  return obj[1];
}
```

It's a problem if we automatically convert `1` into `"1"` as this PR does. Instead, we need to update  `GetPropT` and `SetPropT` to carry along the information that this is a number literal. @samwgoldman is going to attempt this within the next few weeks.

vkurchatkin:
@gabelevi haven't thought about this. I'll try to fix it myself, if you don't mind

samwgoldman:
Normally I would be hugely supportive but I'm pretty much done with the enabling refactor. Will coordinate with you regarding follow up steps once the diff lands!

vkurchatkin:
@samwgoldman got it!

samwgoldman:
OK, so the groundwork has landed in this revision: 1fd2a7242dfe427507110fc4b04ca37990fb092c

We now look up the prototype chain for all computed properties: https://github.com/facebook/flow/blob/b857236d9a010c63a0d8e19cfc7ae0a7618ddd3d/src/typing/flow_js.ml#L4007

To make this work, the trick will be to find all the places that a `propref` is looked up in an `ObjT` or `InstanceT`, see if the propref is a `Computed (NumT (_, Literal lit))` and try to convert to a named property. If not, continue propagating the computed propref up the chain.

Let me know if you have any questions, or if there is something else missing.

mrkev:
Hopping in real quick to see if you're still pursuing this @vkurchatkin 😅it's been a while

nmote:
It looks like this has been abandoned. Feel free to reopen if you pick this back up again.

PR #2297, opened 2016-08-19
add analytics in header directly

lacker:
AFAICT analytics.js is not imported from anywhere and isn't getting run right now.

This PR installs Google Analytics in the way most similar to how Google recommends - by just pasting an opaque blog right before </head>.


mroch:
@facebook-github-bot shipit

facebook-github-bot-5:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3746562) to review internal test results.

PR #2307, opened 2016-08-21
[RFC] GraphQL support

pvolok:
_Not for merge but for discussion._

Hi! I've been playing with flow for a couple of months trying to make it to analyze graphql queries. I'm new to both flow and ocaml, so the code here is certainly shit.

But now that I got some practice with this I'd like to discuss the possibility of adding a graphql support into flow.

I can think of the following features that would be possible:
- GraphiQL-like code assistance in ide
  - Query/fragment validation
  - Autocomplete
  - Type+doc under the cursor
- Object shape from queries
- Checking query variables
- Finding unused fields

This functionality would probably not be as lightweight in terms of the code in flow. And it's not obvious if the benefits would outweigh the burden of maintaining additional code. Also, I briefly discussed this idea on irc with @jeffmo a while ago, and he mentioned that flow could support plugins one day.

@sampepose said that the relay team typechecks graphql-queries by using codegen for now: https://github.com/facebook/relay/issues/1297

What do you think, should flow understand graphql?


samwgoldman:
Wow! This is pretty incredible 👍 

This will take a while to review and digest. I'll take a look this week and bring it up in our team meeting on Monday.

Thanks a ton for taking this on!

pvolok:
@samwgoldman no-no! This is not for review (if you are talking about the code). It should've been an issue actually :)

Basically, I have 3 questions:
- **Does graphql query/framgent analysis belong to flow?** For me it' analogous to jsx.
- **Does the flow team have time to review PRs for graphql support?** Flow team seems to be very busy. Well, it's totally understandable. If I will work on this feature, it will be many PRs created and questions asked. So I have to ask if you are ok with dealing with these PRs/questions or would rather prefer not to have this functionality for now.

Looking forward to hear from you after the meeting!

samwgoldman:
Hey, Pavel. I'm sorry for dragging my feet on this for so long. I hope you didn't close this out of frustration with the lack of communication from my end.

Let me try to answer your questions.

> Does graphql query/framgent analysis belong to flow?

Flow doesn't currently provide any kind of plugin feature, so if we want to do query/fragment analysis, it has to happen in Flow at the moment. I do think the opportunity here is large enough to warrant experimentation in the Flow codebase. Maybe down the line if we build out plugin support, this could be extracted, but I don't think we need to wait for that.

Also, you've done a very nice job of isolating the code here. I don't think there's much downside to this approach, thanks to that.

> Does the flow team have time to review PRs for graphql support?

I think the biggest hurdle here is that we don't have a lot of domain expertise on the team. We'd need to ramp up ourselves to really understand the implications of any work done here. Because of that, reviews may take even longer than normal (and the "normal" review interval is regrettably quite big these days...)

@mroch has been on leave, but I think he is best suited to be the point-person from the Flow team here. It would also be great if some FB GraphQL people would volunteer to help steer this effort. @leebyron, what do you think?

Again, sorry for the delay here. If you're still interested in working with us on this, I'd be delighted if you decided to re-open this. Thanks!

pvolok:
I've been busy lately, so I wouldn't have written any code anyway.

This sounds great. I'm definitely still interested in working on this. I'm going to have more time soon and will try to come up with small step PRs to make it easier to review. Thanks for the answers!

vkurchatkin:
I'm interested in this feature too. IMO this should be implemented as collection of magic types, like `$GraphQLSchema`, `$GraphQLQuery`, that could be composed. Relay-specific functionality could be build on top of that.

pvolok:
Yeah. I look at this feature as a general graphql support.

pvolok:
Made this and issue instead of PR: #2823

PR #2319, opened 2016-08-23
Adjusted callback error types for many methods in the fs module decla…

### LINKED ISSUE #1946: Some node methods' errors types are wrong
Node has a `SystemError` class that extends `Error`. It has additional fields such as `errno` and `code`. At the moment all type declarations use Error type. ([example](https://github.com/facebook/flow/blob/699253b9ba4d2ce9b214381cfa8521b73017d3cd/lib/node.js#L572)) As methods such as `exists` are going to be deprecated and methods like `stat` will be used instead it's quite important to type errors as `SystemError` because people will have to rely on a particular field like `code`

aackerman:
…rations for Node. Added declarations for mkdtemp and mkdtempSync. Fixes #1946 

The new error definition was used based on the TypeScript definition [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/324169800cb792855a96272f5d58f646e7b1d56d/node/node.d.ts#L235-L241).

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -8,6 +8,14 @@
  *
  */
 
+type ErrnoError extends Error {
+  errno?: number;
+  code?: string;
+  path?: string;
+  syscall?: string;
+  stack?: string;
+}
+
 type buffer$NonBufferEncoding =
   'hex' | 'HEX' |
   'utf8' | 'UTF8' | 'utf-8' | 'UTF-8' |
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks!

@facebook-github-bot import

facebook-github-bot-6:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3762301) to review internal test results.

PR #2335, opened 2016-08-25
Remove hh_realpath error from /try

spicyj:
https://twitter.com/danielchooper/status/768837086124183558

### DIFF
--- a/js/realpath.js
+++ b/js/realpath.js
@@ -1,6 +1,6 @@
 //Provides: hh_realpath
-//Requires: js_print_stderr
 function hh_realpath(path) {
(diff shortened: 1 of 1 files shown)


mroch:
"doesn't even run properly" isn't fair. it does run properly *right now*, but since realpath doesn't work in JS like it does in C, it's fragile and we should stop calling it. but it does run properly.

samwgoldman:
I'll leave it up to @mroch whether we take this or not. The console error isn't really doing us any good, is it? Could it just be a comment in the source?

mroch:
sure, leaving a comment works for me

PR #2344, opened 2016-08-26
Fix value of OscillatorNode.type

Maushundb:
OscillatorNode.type should be a string enum, not an AudioParam. See https://webaudio.github.io/web-audio-api/#the-oscillatornode-interface for details.


mroch:
looks like this was also fixed by https://github.com/facebook/flow/commit/5cfa7fa03a12be3f276cbec964042a39ce2113ac. thanks, though!

PR #2349, opened 2016-08-27
Add declaration for performance global. Fixes #2345

### LINKED ISSUE #2345: Missing performance global declaration
See https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVA3AhgJzAYywDsBVAZwFMAFCnKOHAW2PwoFkKsyBXHCsALyowYAC4BPAA4U4UMNLoNmRVmACEAgWADk3IgBMKUAJZEK+7WABkV4WKky5C+kxYUAdMxwBrQZp1QeviixnBEljZ2EtKy8rQuyqzu+DCcOGy43mR+WtqBKiFhEbYi0Y5xiq4qHoycPHw5AUGF4dYl9jFO8Upuyam4HFy8FNmaufnBoeEA3KhAA

screen shot 2016-08-25 at 8 43 30 pm

Should be implemented here: https://github.com/facebook/flow/blob/master/lib/bom.js

We're using it in React: https://github.com/facebook/react/blob/a229cdba7fd00799d82c1bab23704220d47e6331/src/renderers/shared/ReactDebugTool.js#L196-L201

aackerman:


lib/bom.js
```diff
@@ -136,6 +136,68 @@ declare class PluginArray {
...
+    requestStart: number;
+    responseEnd: number;
+    responseStart: number;
+    secureConnectionStart: number;
+    unloadEventEnd: number;
+    unloadEventStart: number;
+}
+
+declare class PerformanceNavigation {
+    type: 0 | 1 | 2 | 255;
```
erikdesjardins: Should the constants also be added?
https://www.w3.org/TR/navigation-timing/#sec-navigation-info-interface

e.g.
```js
if (performance.navigation.type === performance.navigation.TYPE_BACK_FORWARD) doSomethingWhenReturningToPage()
```

aackerman: I've gone ahead and added them, I must have missed them.

mroch:
this looks good to me but the `fetch` test needs to be re-recorded

PR #2353, opened 2016-08-27
[console.table] Support arrays of objects

vjeux:
Right now it fails: https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AznGAUwDoAXAQwCMiAKAbQG9yAuMARgF8AaMJ1gJg4BdAJQBuVEA

But

```js
console.table([{a: 1}, {a: 2}]);
```

is working correctly in Chrome:

screen shot 2016-08-27 at 12 01 13 pm

We're using it inside of React: https://github.com/facebook/react/blob/master/src/renderers/shared/ReactPerf.js#L332-L341

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -714,7 +714,7 @@ declare var console: {
   log(...data: Array<any>): void;
   profile(name: string): void;
   profileEnd(): void;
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks!

@facebook-github-bot shipit

PR #2356, opened 2016-08-28
Delete custom Travis CI OS X caching

gabelevi:
Travis CI's caching didn't used to work on OS X, so https://github.com/facebook/flow/commit/9505c1971b3aa2cfc9bf5c0838c970828a0da8bc added support for it. However, it has since been fixed, so this custom caching is no longer needed: https://blog.travis-ci.com/2016-05-03-caches-are-coming-to-everyone

I noticed this when looking into why our OS X Travis builds were failing. That was happening because we were caching opam and getting stuck with old opam info. Throwing away the cache fixes (which can be done from the Travis CI UI)


gabelevi:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3784374) to review internal test results.

mroch:
yay!

the custom caching on OS X was actually faster than the built in one on Linux. maybe we should instead use it all the time ;) (mostly kidding)

you can clear the custom cache from the AWS console by deleting the corresponding tar.gz in the cicache bucket.

PR #2361, opened 2016-08-29
Update node.js

jgrund:
According to https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback

the `childprocess.exec` callback is a string by default and can also be a buffer depending on args.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -137,7 +137,7 @@ declare class child_process$Error extends Error {
   signal: ?string,
 }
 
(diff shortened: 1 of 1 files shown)


facebook-github-bot-6:

facebook-github-bot-6:

mroch:
i'm merging this because it's definitely wrong for it to be just `Buffer`. thanks!

I also filed https://github.com/facebook/flow/issues/2422 to see if we can follow up to make it more accurate, because it depends on the encoding option; you shouldn't have to handle both cases in the callback.

PR #2365, opened 2016-08-30
Only print json if the --json flag is used with the coverage command

### LINKED ISSUE #2315: flow coverage with --json and --color outputs text and json
The flow coverage tool outputs both the colored source text as plain text and then the coverage metrics as json when the two flags are used in conjunction. It's odd so I decided to file an issue.

aackerman:
Fixes #2315

### DIFF
--- a/src/commands/coverageCommand.ml
+++ b/src/commands/coverageCommand.ml
@@ -123,7 +123,7 @@ let rec split_overlapping_ranges accum = Loc.(function
           in
           accum, rest
 
-        else if loc1._end.offset = loc2._end.offset then
+        else if loc1._end.offset <= loc2._end.offset then
           (* range 1 and 2 end at the same place, so split range 1 and consume
              the first part, which doesn't overlap *)
           let head_loc = { loc1 with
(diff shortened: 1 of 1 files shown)


mroch:
should it error if you use both `--color --json`?

aackerman:
@mroch that's probably clearer behavior. I'll go ahead and change that when I have the chance.

aackerman:
Probably not idiomatic OCaml, but I made changes to print out error information if the user attempts to use the `color` or `debug` flag in conjunction with the `json` flag.

avikchaudhuri:
@facebook-github-bot import

samwgoldman:
It looks like this changes the behavior of `--debug --color` to only output the debug information. Is it intentional that these options should be mutually exclusive? They weren't before and I don't think they need to be, as long as they are exclusive of `--json`.

aackerman:
@samwgoldman I went ahead and fixed that, I don't want to change the existing behavior aside from erroring when flags cannot be used in combination.

PR #2382, opened 2016-09-02
remove duplicated `mediaDevices` declaration

stryju:
`Navigator` in `bom.js` had two declarations of `mediaDevices`, one optional - removed the required one, as older browsers could not have it


avikchaudhuri:
@facebook-github-bot import

PR #2384, opened 2016-09-02
Get gen_flow_files working on Windows

gabelevi:
`./tool test` was broken on Windows because `flow gen_flow_files` didn't work on Windows. Rather than just fixing the test, I decided to fix the command for Windows, since Windows isn't intended to be a second-class citizen.


gabelevi:
@facebook-github-bot import

Issue #2388, opened 2016-09-04
Add option to error (or warn) on unresolved type imports


jamesreggio:
If an `import type` declaration fails to actually resolve, Flow will silently allow use of the unresolved imported type. Variables typed to an unresolved import appear to pass all type checks, which feels dangerous.

Here's a [brief illustration](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzACcBXbMAXjAAoBDIgcwC4wAjOOGASkoD4wA3mAC+AbnTBgYAHK18JIrRgwAngBow+ABYBLXGAAmcAKa5sAckLGAHnvzjJYLYmMA3Y0Q0BlOAFtjACoqAA7GYHpgBDrKYErwCMYGmnBsYSS4iQB0qDq+wXBEhPghYQI+-kGhwmBQRH5g5lC0ANbGAPq+cAYkMMbm4qiOAOpaKoYmZpaauvrBdd0YYbTkHnVEAPyYOAQ1HCzlgSWUgmISUgCCK0Rr4bPzJItJCFrG5MtgZAbGGDD0iZpHCLpLKDKQ+MDPMZGUwWQgAGR02DCAEYAAzEUw9QiI6ZhXC0fypLS0Vw6AqbLB4QisegsADy2g8lTCVAEp1BYACL0MywYHjAcCgkWir3wqhqtGiiIYuPCeQK2P02EQHIAkqkYDo3GFtHJphEMHFkmAAAakbAm26RWhQYwabRwEgMLQchKxIhhOCsUmO3Di5WEYK0XC4aXhN5sDi9ZbZc3UKAcLiiIA) of the problem.

Is there any way to strictly verify imports? Perhaps this could be a flag (or, better yet, default behavior) in the future?

jeffmo:
Yea, we've sort of gone back and forth on this in the past. I'm inclined to agree that it probably should be an error by default (it's *usually* a mistake or, at best, unintended to result in a loosely-typed `any`).

Having a config-option escape hatch for when it is intentional (rare) does seem like the right thing here.

For context: The reason we've gone back and forth on this is due to the fact that it's in line with how we treat other imports from untyped modules: Untyped modules present `any `-typed exports (this is the nature of gradual typing).

The thing that convinces me that it's reasonable to error-by-default is that an import of a type is usually done with the expectation that the type will be enforced (but when the module isn't typed, it can't be!).

[...]

PR #2392, opened 2016-09-05
Error when `import type` pulls from an untyped module

jeffmo:
Previously `import type {T} from "./untyped";` and `import type T from "./untyped";` would silently result in a local type, `T`, of type `any`. When importing from an untyped module, this is the best we can do really.

However, when importing a "type", it's usually the case that one expects the source module to be typed and it is often surprising when one discovers -- usually much later -- that the "type" you thought you were importing was `any` all along!

See https://github.com/facebook/flow/issues/2388 for the latest report on this. There have been other less formal reports of this in the past as well.

This PR makes `import type` error if the source module is not typed (i.e. the `ModuleT` is `AnyT`). In the rare case where a user is intentionally importing types from an untyped module and wants Flow to assume the implicit `any` behavior, they can simply use a `// $FlowFixMe` suppression comment.

### DIFF
--- a/newtests/import_type_from_untyped/_flowconfig
+++ b/newtests/import_type_from_untyped/_flowconfig
@@ -1,0 +1,7 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
--- a/newtests/import_type_from_untyped/test.js
+++ b/newtests/import_type_from_untyped/test.js
@@ -1,0 +1,20 @@
+/* @flow */
+
+
+import {suite, test} from '../../tsrc/test/Tester';
+
+export default suite(({addFile, addFiles, addCode}) => [
+  test('importing a type from an untyped module errors', [
+    addFile('untyped.js'),
+    addCode('import type {TNamed} from "./untyped";').newErrors(`
+      test.js:3
+        3: import type {TNamed} from "./untyped";
+                        ^^^^^^ Named import from module ${'`'}./untyped${'`'}. Importing a type from an untyped module is not safe! Did you mean to add ${'`'}// @flow${'`'} to the top of the module that exports ${'`'}TNamed${'`'}?
+    `),
+    addCode('import type TDefault from "./untyped";').newErrors(`
+      test.js:5
+        5: import type TDefault from "./untyped";
+                       ^^^^^^^^ Default import from ${'`'}./untyped${'`'}. Importing a type from an untyped module is not safe! Did you mean to add ${'`'}// @flow${'`'} to the top of ${'`'}./untyped${'`'}?
+    `)
+  ]),
+]);
--- a/src/commands/lsCommand.ml
+++ b/src/commands/lsCommand.ml
@@ -154,7 +154,7 @@ let get_ls_files ~subdir ~root ~strip_root ~ignore_flag ~include_flag =
     opt_ignore_non_literal_requires = false;
     opt_max_header_tokens = FlowConfig.(
       flowconfig.options.Opts.max_header_tokens
-    )
+    );
   } in
 
   let _, libs = Files.init options in
--- a/src/commands/serverCommands.ml
+++ b/src/commands/serverCommands.ml
@@ -351,7 +351,7 @@ module OptionParser(Config : CONFIG) = struct
       );
       opt_max_header_tokens = FlowConfig.(
         flowconfig.options.Opts.max_header_tokens
-      )
+      );
     } in
     Main.start options
 
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1784,12 +1784,34 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
 
     (* imports are `any`-typed when they are from (1) unchecked modules or (2)
        modules with `any`-typed exports *)
-    | (AnyT _ | AnyObjT _),
-        ( CJSRequireT(reason, t)
+    | (AnyObjT _, (
+        CJSRequireT(reason, t)
         | ImportModuleNsT(reason, t)
         | ImportDefaultT(reason, _, _, t)
         | ImportNamedT(reason, _, _, t)
-        ) ->
+      )) ->
+      rec_flow_t cx trace (AnyT.why reason, t)
+
+    | (AnyT _, (CJSRequireT(reason, t) | ImportModuleNsT(reason, t))) ->
+      rec_flow_t cx trace (AnyT.why reason, t)
+
+    | (AnyT _, ImportDefaultT(reason, import_kind, (_, module_name), t)) ->
+      if import_kind = ImportType then add_error cx (mk_info reason [
+        spf (
+          "Importing a type from an untyped module is not safe! Did you " ^^
+          "mean to add `// @flow` to the top of `%s`?"
+        ) module_name
+      ]);
+      rec_flow_t cx trace (AnyT.why reason, t)
+
+    | (AnyT _, ImportNamedT(reason, import_kind, export_name, t)) ->
+      if import_kind = ImportType then add_error cx (mk_info reason [
+        spf (
+          "Importing a type from an untyped module is not safe! Did you " ^^
+          "mean to add `// @flow` to the top of the module that exports " ^^
+          "`%s`?"
+        ) export_name
+      ]);
       rec_flow_t cx trace (AnyT.why reason, t)
 
     | ((PolyT (_, TypeT _) | TypeT _), AssertImportIsValueT(reason, name)) ->
--- a/tests/annot2/A.js
+++ b/tests/annot2/A.js
@@ -3,6 +3,7 @@
  * @flow
  */
 
+// $FlowFixMe
 import type T from "T";
 
 export default class {


newtests/import_type_from_untyped_true/test.js
```diff
@@ -0,0 +1,11 @@
+/* @flow */
+
+
+import {suite, test} from '../../tsrc/test/Tester';
+
+export default suite(({addFile, addFiles, addCode}) => [
+  test('importing a type from an untyped module errors', [
```
jeffmo: Need to update this copy/pasted test description

jamesreggio:
lol @ the commit message.

Otherwise, very much appreciate this.

avikchaudhuri:
I'd love to have this without the config option, i.e., on by default.

jeffmo:
@avikchaudhuri: By "on by default", do you mean "give an error if one uses `import type` on a non-@flow file"? Or do you mean "allow `import type` from a non-@flow file by default" (status quo)?

This PR does the former -- it switches the default to error, and it adds a config option as an escape hatch. I expect almost no one to use the escape hatch...but it exists as a backup.

jeffmo:
@facebook-github-bot import

jeffmo:
(still gotta fix travis)

gabelevi:
Hmm, I'm not sure about this. Looking at the PR, I see that @avikchaudhuri asked for this to be on by default and with no config. I feel the opposite way. I think I'd only be comfortable with this off by default and behind a config. Here is why I think this is not a good default behavior.

## Changing a @flow file to not @flow

I don't like how this can cause errors if you remove the `@flow` comment from a file. I think part of the gradual typing premise is that removing `@flow` should suppress errors, not cause more errors.

## Inconsistent with import

Why can you `import` from a untyped file but not `import type`? I see the argument "well import is ES6 but import type is Flow", but it doesn't feel like a strong argument. People who learn about `import type` probably expect it to work like `import`.

## Warnings vs errors
This feels more like a warning than an error. It's well established that untyped files are a source of unsoundness. This is more of a "hey, did you realize that foo.js is untyped" warning than a "hey, something is wrong here" error. And in the interest of providing high quality errors by default, I don't think warnings should be enabled by default.

The warning is "hey, we're importing this type as any! Are you sure that's what you intended?". I don't think warnings should be on by default. We should endeavor to only complain about real problems by default.

## Gradual typing

It sucks that you can't do gradual typing now without using suppression comments. As soon as you need to `import type Foo from './foo'`, you need to flowify `foo.js` too.

jeffmo:
> Changing a @flow file to not @flow

This is the most compelling argument against this diff and is a real caveat to doing this. It does suck that it could incur new errors, but on the other hand (devil's advocate) it also reminds me that I'm not just disabling types in the local file I'm removing the `@flow` from, but I'm also removing them from other files as well. There's a good chance that doesn't occur to me -- and adding FixMes is my way of acknowledging that risk if I'm willing to take it.

> Inconsistent with `import`

This doesn't really bother me for 2 reasons:

1. Inconsistency is bad when it's implicit or hard to discover (easy to make mistakes). In this case, it's pretty fast and easy to discover this difference (you get a clear message) and making a mistake is the opposite of enforcing this rules. I don't think inconsistency is really a bad word in this context.
2. You can run tests (automated and manual) on imported values to notice when they are wrong. You'll catch them later on -- or maybe not at all -- but you do have a non-perfect opportunity to catch them. With imported types, only Flow knows when they are or could be wrong. Flow is your only real chance to notice the problem outside of careful manual inspection (which rarely happens -- see www).

> Warnings vs errors

I also agree with this distinction and the general need for it (not just here). There have been loads of times where things have turned out to be "probably bad for you, but not bad frequently enough that we're willing to take on the noise in the signal".

What if we had a `flow --show-warnings` flag that lists the things you don't normally want to be spammed about? We can store both errors and warnings together, but toggle which are shown based on the flag that's passed. It would be a bit of an intrusive change throughout the system (have to go through and decide when to show either or both everywhere errors can be seen today) -- but probably warranted if we can't poke a significant enough hole in the idea.

> It sucks that you can't do gradual typing now without using suppression comments.

It's hard to disagree with this statement because "gradual typing" means so many things, so I think we need to be more specific. The common win you get from gradual typing is the ability to take your time while *adding* types to a codebase. You don't need suppression comments for this, of course.

Another win you get from gradual typing is the ability to turn them off. It turns out that turning them off has some often-overlooked side effects, and when they are overlooked it actually degrades general trust in the type system's "inference" capabilities (see various feedback about people feeling like they're not sure when Flow is checking something for them or not).

Going back to my argument above: Given that there are ways to catch value errors other than Flow (tests, running your code, etc) and that Flow doesn't "gradually value" things, it's easier to lean on a "trust me" contract for gradual typing on values. When you have types that can only be enforced by Flow, however, leaning on a "trust me" contract is much more fragile and degrades peoples' trust in types.

To me the caveat that you might need to add some `$FlowFixMe`s in order to untype something sometimes is worth the cost. It's a little more (cheap) work, but it comes with the benefit of knowing the non-local, unchecked effects you're incurring.

gabelevi:
> It's hard to disagree with this statement because "gradual typing" means so many things, so I think we need to be more specific. The common win you get from gradual typing is the ability to take your time while adding types to a codebase. You don't need suppression comments for this, of course.

This is the case I'm thinking of. Starting with an untyped codebase and gradually adding types. Consider the following untyped code

```js
import Bar from './Bar';

export default function fooOfBar(bar) { return bar.toFoo(); }
```

Before your change, we would add types like so

```js
import Bar from './Bar';
import type Foo from './Foo';

export default function fooOfBar(bar: Bar): Foo { return bar.toFoo(); }
```

After you change, you would be force to write

```js
import Bar from './Bar';
// $FlowFixMe - Foo is untyped
import type Foo from './Foo';

export default function fooOfBar(bar: Bar): Foo { return bar.toFoo(); }
```

Yes, you could go add types to `Foo`, but adding that kind of requirement can lead to a snowball where you need to add types to more and more files.

> What if we had a flow --show-warnings flag that lists the things you don't normally want to be spammed about? 

I do like the idea of having warnings off by default and allow the users to enable them with config/flags. I think you'll need a way to enable them all and a way to enable only some. We probably could start emitting a code for each warning type, which would help. 

> When you have types that can only be enforced by Flow, however, leaning on a "trust me" contract is much more fragile and degrades peoples' trust in types.

This point is well received. False negatives (when something is wrong and Flow doesn't complain) erode users trust. False positives (when something is not wrong but Flow complains) add friction to adoption. It's definitely a balancing act. But I think this adds too much friction to the default. I'm fine with this friction as an opt-in feature, but not as a default

avikchaudhuri:
cc @vjeux

vjeux:
The argument that when you remove `// @flow` breaks stuff is compelling and is definitely a downside.

For your example, I actually disagree that the person should be writing the following when `Foo` is not typed:

```js
import type Foo from './Foo';
export default function fooOfBar(bar: Bar): Foo { return bar.toFoo(); }
```

If `Foo` is not typed, I do believe that you want an error right away. As a user, if I'm writing `Foo` here, I expect flow to guarantee me that it's going to be properly typed, and not just use it as an annotation.

If I did want to say that it returns a `Foo`, I would do

```js
type Foo = any; // the real thing is not typed yet
```

vjeux:
I chatted a bit with @avikchaudhuri offline and some more thoughts:

The way I ran into this is because I added `@flow` to `Foo` and then it triggered errors on multiple files that used the type in a completely wrong way. I'm pretty sure that the author didn't even realize that it wasn't actually checked. As a user, my thought process is that I need to manually write `any` in order to get something not covered, but in this case it just happened to magically create an `any` even though I was in the flow world (as I wrote `import type`).

The danger with this is that once you start relying on wrong types, you may end up building a castle in fantasy land and the moment you try to actually type the file, it's a nightmare to fix all those. It would have been better to have them not typed in the first place than having wrong types.

One thing he mentioned as well is that 

```js
import Bar from './Bar';
export default function fooOfBar(bar: Bar): Foo { return bar.toFoo(); }
```

should probably also be disallowed since you are using a type from a file that is not flowified and therefore a `any` type is being created out of thin air.

jamesreggio:
Hi — I'm the original issue reporter (#2388).

@vjeux elegantly described the reality that led me to file this issue:

> The danger with this is that once you start relying on wrong types, you may end up building a castle in fantasy land and the moment you try to actually type the file, it's a nightmare to fix all those that accumulated. It would have been better to have them not typed in the first place than having wrong types.

@gabelevi summarized another aspect of this problem:

> False negatives (when something is wrong and Flow doesn't complain) erode users trust.

When reviewing a colleague's code on a GraphQL project, I found a bug that I thought Flow should've caught. My colleague meticulously annotated all of his methods, most which utilized types from GraphQL that he was pulling in with `import type`. It seemed odd that Flow wasn't throwing an error for an 'obvious' type violation.

Eventually, I discovered that the `graphql-js` module we were importing had no type data, and all of our imported GraphQL types were implicit `any`s. This eroded my trust, and also led to the unpleasant realization that whenever `graphql-js` actually exports types in its distribution code (https://github.com/graphql/graphql-js/issues/359), we'll be in for a world of hurt — because we certainly didn't annotate things perfectly.

I understand the desire to avoid introducing a large number of new errors in a new release of a tool, but I'd be willing to bet that most people don't realize that an implicit `any` is the behavior for importing from an untyped module.

jeffmo:
So I think it's clear what my position is here on the trade-off between "explicit suppression/explicit acknowledgement" and "implicit `any`/implicit acknowledgement" -- but just wanted to chime in on:

> I understand the desire to avoid introducing a large number of new errors in a new release of a tool, but I'd be willing to bet that most people don't realize that an implicit any is the behavior for importing from an untyped module.

It's worth mentioning that, when running this change on the FB codebase, I see a very high number of `import type`s that are importing from modules that have never been typed. The most popular untyped modules contributing to this (I checked the histories of the top-10 offending untyped modules) just haven't had types added to them yet. In other words: It's not that they have been disabled, they've just never been typed in the first place.

We could interpret this 2 ways: (a) People are intentionally importing a type expecting potential type errors in the future when the referenced module is typed, or (b) people aren't thinking that deeply about the `import type` they're writing (and Flow isn't erroring) so they're just moving forward. I'm more inclined to assume the 2nd scenario is most common since most of us tend to think more locally than globally while writing code and "is another file typed?" is a rare question to ask -- especially when nothing pushes you to ask the question.

To me this suggests that, although this change is certainly a different trade-off from the status-quo, by permitting untyped `import type` we're seeing more unintended `any` types than we would if we asked people to be explicit about their expectations.

**Conclusion: Requiring a suppression comment to explicitly dictate intent when untyped-import is intentional seems like a little work for a lot of win, numberlifically-speaking.**

gabelevi:
This PR is pretty old, so I'm going to close it out for now. Feel free to reopen if/when you want to actively continue this discussion!

ballardrog:
This was resurrected as part of the new Flow Linter project, and will be appearing as the untyped-type-import lint rule.

PR #2414, opened 2016-09-07
Add Windows to the list of supported platforms

jeffmo:


README.md
```diff
@@ -10,6 +10,7 @@ Flow works with:
 
 * Mac OS X
 * Linux (64-bit)
+* Windows (64-bit)
 
 There are binary distributions for Mac OS X and many variants of Linux; you can also build it from source on almost any 64-bit Linux variant.
```
mroch: this should be updated too, right?

also, the second clause is a bit weird... it implies it can't be built from source on OS X or Windows.

PR #2415, opened 2016-09-07
Introduce reparse_with_defaults

nmote:
As discussed in #2276, this will reduce code duplication.


src/parsing/parsing_service_js.mli
```diff
@@ -54,6 +54,12 @@ val reparse:
   FilenameSet.t ->          (* filenames to reparse *)
   FilenameSet.t * results   (* modified files and job results *)
 
+val reparse_with_defaults:
+  ServerEnv.genv ->
```
mroch: nothing else in this module depends on `ServerEnv` (nothing in `parsing/` at all IIRC). can you pass `options` instead, please?

nmote:
Makes sense, done.

Issue #2393, opened 2016-09-05
Missing AnimationEvent type in dom.js


leoselig:
We have a lot of different sub classes of `Event` in `https://github.com/facebook/flow/blob/master/lib/dom.js` (e.g. `MouseEvent`)

I think it's fair to put [`AnimationEvent`](https://developer.mozilla.org/en-US/docs/Web/API/AnimationEvent) in there as well.

Right now, I have to use a custom type to be handle it:

``` js
declare class AnimationEvent extends UIEvent {
  animationName: string;
  elapsedTime: number;
  pseudoElement: string;

  // deprecated

  initAnimationEvent: (
    type: 'animationstart' | 'animationend' | 'animationiteration',
    canBubble: boolean,
    cancelable: boolean,
    animationName: string,
    elapsedTime: number
  ) => void;
}
```

Or are there any concerns due to it being still marked as `experimental`?

jeffmo:
According to the doc you sent, "basic" support exists in all browsers -- so I think that's ok to add (feel free to send a PR).

It looks like `initAnimationEvent()` is deprecated, though -- so that probably doesn't make sense to include in core. And if we add `psuedoElement`, it should be marked as `optional` since it's not supported by most browsers. This way Flow would require you to use something like:

`if (e.pseudoElement) { console.log(e.pseudoElement); }`

leoselig:
@jeffmo I don't know why you closed this as I don't see a fix for this anywhere. Please re-open or be so kind to provide a reason

> It looks like initAnimationEvent() is deprecated, though -- so that probably doesn't make sense to include in core.

I included it anyway for consistency because the `init...Event()` methods are all included so far, featuring a `depreated` comment (e.g. https://github.com/facebook/flow/blob/master/lib/dom.js#L258)

I prepared a PR

thejameskyle:
I don't think he intended to close it, it's easy to press by mistake. Either way, we'll just move over to your PR

PR #2442, opened 2016-09-12
Add AnimationEvent to DOM types

leoselig:
All DOM event types with basic browser support should be added to lib/dom.js.
However, AnimationEvent was still missing, this adds support for it.

see https://github.com/facebook/flow/issues/2393

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -261,6 +261,22 @@ declare class KeyboardEvent extends UIEvent {
   which: number;
 }
 
+declare class AnimationEvent extends UIEvent {
+  animationName: string;
+  elapsedTime: number;
+  pseudoElement: string;
+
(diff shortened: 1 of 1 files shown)


leoselig:
I don't really understand the errors

**Travis CI:** The tests are failing on my local machine as well but I don't see how what I changed affects them

**AppVeyor:** What is `package_json_changes` about and what could be a reasons to fail with `failed to kill server 2024` 

Are there some more resources on contributing and the test setup?

thejameskyle:
For the errors in travis the tests just need to be re-recorded, line numbers have changed in the snapshots. AppVeyor seems fine though.

Don't worry about it, I'll import and fix up.

thejameskyle:
@facebook-github-bot import

PR #2444, opened 2016-09-12
Add `padStart` and `padEnd` to `String.prototype`.

pvolok:
It's at stage 4: https://github.com/tc39/proposal-string-pad-start-end


thejameskyle:
@facebook-github-bot import

PR #2448, opened 2016-09-12
[parser tests] fix assertion that an array is empty

mroch:
if you had a hardcoded parser test like

```
'foo': {
  'errors': [],
}
```

then we were not actually asserting that there were no errors in the ML version of the test runner. the JS version catches it, though.


mroch:
@facebook-github-bot import

PR #2471, opened 2016-09-14
FormData definitions

danharper:
Just filled out the definitions for [`FormData`](https://developer.mozilla.org/en/docs/Web/API/FormData).

I've updated the test which was failing from line numbers changing, and I've added a test for `FormData` (there was no `bom` dir - I assume this is the correct place for them?)

One concern there might be is I've tightened the types (on the sole existing method) from `any` to `string`, which seems to be correct as a string _is_ the correct value to give to these functions. Give it an object and you get `"[Object object]"` back.

``` diff
- append(name: any, value: any, blobName?: string): void;
+ append(name: string, value: string, blobName?: string): void;
```

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -241,7 +241,16 @@ declare class DOMParser {
 }
 
 declare class FormData {
-    append(name: any, value: any, blobName?: string): void;
+    constructor(form?: HTMLFormElement): void;
+    has(name: string): boolean;
+    get(name: string): ?string;
+    getAll(name: string): Array<string>;
+    set(name: string, value: string, blobName?: string): void;
+    append(name: string, value: string, blobName?: string): void;
+    delete(name: string): void;
+    keys(): Iterator<string>;
+    values(): Iterator<string>;
+    entries(): Iterator<[string, string]>;
 }
 
 declare class MutationRecord {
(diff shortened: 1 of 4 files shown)


thejameskyle:
@facebook-github-bot import

danharper:
@thejameskyle Oops, apparently the `value`s can also be a `File` type too. I'll update the definitions later today!

danharper:
`Blob` or `File` https://xhr.spec.whatwg.org/#create-an-entry - sorry about that!

danharper:
Ok, correct definitions seem to be:

```js
type FormDataEntryValue = string | File

declare class FormData {
    constructor(form?: HTMLFormElement): void;

    has(name: string): boolean;
    get(name: string): ?FormDataEntryValue;
    getAll(name: string): Array<FormDataEntryValue>;

    set(name: string, value: string): void;
    set(name: string, value: Blob | File, filename?: string): void;

    append(name: string, value: string): void;
    append(name: string, value: Blob | File, filename?: string): void;

    delete(name: string): void;

    keys(): Iterator<string>;
    values(): Iterator<FormDataEntryValue>;
    entries(): Iterator<[string, FormDataEntryValue]>;
}
```

I'll add tests for it tonight & update this PR.

(I've also noticed the existing `File`'s constructor definition is incorrect, so I'll send a separate PR for that too).

PR #2493, opened 2016-09-20
make Event polymorphic

stryju:
this allows to annotate "primitive" events better, like [iceconnectionstatechange](https://developer.mozilla.org/en-US/docs/Web/Events/iceconnectionstatechange), where the only difference is the type of `event.target: RTCPeerConnection`

---

example use-case:

```js
peerConnection.addEventListener(
  'iceconnectionstatechange',
  (ev: Event<RTCPeerConnection>): void => {
    doSomethingWith(ev.target.iceConnectionState);
  }
);
```

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -129,7 +129,7 @@ type Event$Init = {
   composed?: boolean,
 }
 
-declare class Event {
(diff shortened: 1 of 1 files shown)


stryju:
ugh - back to the drawing board :)

PR #2501, opened 2016-09-21
Add missing type "stdio" of child_process.fork option properties

glennreyes:
This adds `stdio` type `child_process.fork` in `options` property

According to the description in the [official Node.js API] they say:
> The array must contain exactly one item with value 'ipc' or an error will be thrown. For instance [0, 1, 2, 'ipc']

I'm sure this could be typed a bit smarter instead of `Array<any>`, but anyway better than not typed. I'd be very grateful for any help or suggestions 🙌

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -171,7 +171,7 @@ type child_process$forkOpts = {
   execPath?: string;
   execArgv?: Array<string>;
   silent?: boolean;
-  stdio?: Array<any>;
+  stdio?: Array<number | 'ipc'>;
   uid?: number;
   gid?: number;
 };
--- a/lib/node.js
+++ b/lib/node.js
@@ -171,6 +171,7 @@ type child_process$forkOpts = {
   execPath?: string;
   execArgv?: Array<string>;
   silent?: boolean;
+  stdio?: Array<any>;
   uid?: number;
   gid?: number;
 };


facebook-github-bot-5:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3926138).

gabelevi:
Thanks for the contribution! I'll fix up the tests and get this merged!

sindresorhus:
`stdio` can also be a `string`. And the array can contain any of the types: `string`, `number`, `Stream`, `null`, `undefined`.

https://nodejs.org/api/child_process.html#child_process_options_stdio

I don't know Flow that well, but imagine something like this would be more proper:

```
stdio?: string | Array<string | number | Stream | null | void>
```

When it's a string, it can only be one of the following values: `'pipe'`, `'ignore'`, `'inherit'`, so might be useful to enforce that too.

-

// @thejameskyle

glennreyes:
@sindresorhus 
The [docs specifically at `child_process.fork`](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options) says following on the `stdio` option:
> stdio <Array> Supports the array version of child_process.spawn()'s stdio option. When this option is provided, it overrides silent. The array must contain exactly one item with value 'ipc' or an error will be thrown. For instance [0, 1, 2, 'ipc'].

I was looking for better typing according to the last sentence.

sindresorhus:
@glennreyes Yeah, missed that this was only about `fork`. I'll do a follow-up PR with a more general improvement.

glennreyes:
@gabelevi Any updates about the tests?

PR #2506, opened 2016-09-21
[parser] Add esprima 3.0.0 tests

mroch:
Adds the [`test/fixtures/`](https://github.com/jquery/esprima/tree/3.0.0/test/fixtures) directory from Esprima 3.0.0, and a test runner (see `src/parser/test/run_esprima_tests.ml`) that runs the flow parser against these fixtures.

Here's the current output: https://gist.github.com/mroch/085d5e3f5fb5e08c85972fa5cffe4655

These fixtures work via pairs of files: a `.js` source file and a `.json` expected output file. So far, this supports `.tree.json` outputs, which is a dump of the AST. It doesn't yet support `.tokens.json` (just the list of tokens), `.failures.json` (expected errors), etc. and instead marks those tests as skipped.

There is still a lot of work to do both to run all of the tests, and also to fix Flow's estree compatibility, so this PR doesn't run these tests via `make test` yet. Run `make test-esprima-ocaml` instead.


mroch:
this is too big for our import scripts... will have to import it manually.

PR #2512, opened 2016-09-22
[tool] fix race condition with server restarts

mroch:
the `package_json_changes` test was racy. adding a file that should kill the server might not have been noticed by the server by the time the next action runs.

added a new action, `waitForServerToDie(timeout)`, which just sleeps until the server dies or the timeout occurs.

added a new assertion, `serverRunning(isRunning)`, which confirms that the same PID is still running (or not running).

### DIFF
--- a/newtests/package_json_changes/test.js
+++ b/newtests/package_json_changes/test.js
@@ -3,70 +3,67 @@
 
 import {suite, test} from '../../tsrc/test/Tester';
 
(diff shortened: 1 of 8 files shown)


facebook-github-bot-5:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3909447).

Issue #2503, opened 2016-09-21
AST: defaults should use AssignmentPattern


mroch:
https://github.com/estree/estree/blob/master/es2015.md#assignmentpattern

babel and esprima both include defaults as an AssignmentPattern in the `params` list, rather than as a separate `defaults` field on the function declaration.

PR #2516, opened 2016-09-22
[parser] use AssignmentPattern for function param defaults

mroch:
ESTree/esprima changed from storing function param defaults in an array parallel to the params, to storing them as `AssignmentPattern` params.

Fixes https://github.com/facebook/flow/issues/2503


facebook-github-bot-1:
Thanks for importing. If you are a Facebook employee, you can view this diff [on Phabricator](https://phabricator.intern.facebook.com/D3910336).

PR #2535, opened 2016-09-27
Add Make server.close callback more specific.

jgrund:
Change the callback of `server.close` from `Function`
to one that takes an optional error and returns void.

Signed-off-by: Joe Grund joe.grund@intel.com

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -822,7 +822,7 @@ declare module "http" {
     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server;
     listen(path: string, callback?: Function): Server;
     listen(handle: Object, callback?: Function): Server;
-    close(callback?: Function): Server;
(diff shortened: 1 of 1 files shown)


gabelevi:
Sorry for the delay! This fell through the cracks in our oncall rotation. This looks good to me and I'll get this reviewed and merged ASAP!

lib/node.js
```diff
@@ -822,7 +822,7 @@ declare module "http" {
     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server;
     listen(path: string, callback?: Function): Server;
     listen(handle: Object, callback?: Function): Server;
-    close(callback?: Function): Server;
+    close(callback?: (error: ?Error) => void): Server;
```
gabelevi: Nowadays we tend to use `mixed` as the return type for callbacks when the callback return value is ignored. I'll update this before logging!

PR #2556, opened 2016-10-01
Provide description for EmptyT

vjeux:
As mentioned in an internal post ( https://www.facebook.com/groups/flowtype/permalink/1145825575466056/ ), in case there's an impossible type, get-type-at-pos doesn't return anything meaningful because EmptyT description is "" instead of "empty". This is an attempt at fixing it.

Note that I wrote this from github and haven't compiled or tested the change, but I'd really love for this to be fixed inside of flow such that nuclide can show a proper typehint.

### DIFF
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -342,7 +342,7 @@ let rec string_of_desc = function
   | RString -> "string"
   | RBoolean -> "boolean"
   | RMixed -> "mixed"
-  | REmpty -> ""
+  | REmpty -> "empty"
   | RAny -> "any"
   | RVoid -> "undefined"
   | RNull -> "null"


vjeux:
Input:
```js
/* @flow */
function flatten(elemArray: Array<number>): number {
  elemArray.map(elem => {
    if (Array.isArray(elem)) {
      return flatten(elem);
    }
  });
  return 10;
}
```

When I mouse over `elem` inside of `flatten(elem)`, I receive:

> "{"type":"(unknown)","reasons":[],"loc":{"source":"test.js","type":"SourceFile","start":{"line":6,"column":22,"offset":163},"end":{"line":6,"column":25,"offset":167}},"path":"test.js","line":6,"endline":6,"start":22,"end":25} "

The `(unknown)` string is generated here: https://github.com/facebook/flow/blob/601791f8d587bd347dde0e09a0456f0bfe61e0e7/src/commands/typeAtPosCommand.ml#L62

PR #2560, opened 2016-10-02
Add type definitions for HTMLMetaElement

charliesome:
I was noodling around earlier today and needed to use `HTMLMetaElement` in some code.

Flow's `dom.js` doesn't provide type definitions for `HTMLMetaElement`, so I've added them.


mroch:
thanks! Travis hasn't finished running yet but I bet the `dom` test will fail since this bumped line numbers down.

can you run `./runtests.sh bin/flow dom; cp tests/dom/dom.out tests/dom/dom.exp` to update the test?

charliesome:
@mroch Done! Thanks for the quick reply 😊

PR #2568, opened 2016-10-03
Update types of node cluster

glennreyes:
This PR updates all the types of [`Cluster`](https://nodejs.org/dist/latest-v6.x/docs/api/cluster.html) to the current Node API.


gabelevi:
Sorry for the delay for this too! Importing...

PR #2572, opened 2016-10-04
Adds a `weak` option to the .flowConfig format

tberman:
Similar to the `all` option, this emulates passing --weak to the command line, but inside the config file.


gabelevi:
Sorry for the delay! I'll get this imported and reviewed ASAP! Since we have the `--weak` flag, it would make sense to add a `.flowconfig` option. Thanks for the contribution!

Issue #1483, opened 2016-03-03
isPrototypeOf property not found


hmeerlo:
I get an error from Flow on the following code:

```
133:       } else if (Date.prototype.isPrototypeOf(aprop) || Date.prototype.isPrototypeOf(bprop)) {
                                     ^^^^^^^^^^^^^ property `isPrototypeOf`. Property not found in
133:       } else if (Date.prototype.isPrototypeOf(aprop) || Date.prototype.isPrototypeOf(bprop)) {
                      ^^^^ Date
```

Isn't this supposed to be correct?

PR #2577, opened 2016-10-04
[core] add missing isPrototypeOf

mrtnbroder:
Adds support for https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/isPrototypeOf

Fixes https://github.com/facebook/flow/issues/1483

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -51,6 +51,7 @@ declare class Object {
     static setPrototypeOf(o: any, proto: ?Object): bool;
     static values(object: any): Array<mixed>;
     hasOwnProperty(prop: any): boolean;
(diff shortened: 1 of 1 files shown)


mroch:
Thanks! I edited the summary to note that this fixes https://github.com/facebook/flow/issues/1483.

One caveat to this is that Flow won't understand `isPrototypeOf` in refinements, so the example if statement in #1483 won't teach Flow that something is a `Date`. But obviously it's better to not refine than to say that `isPrototypeOf` doesn't even exist.

This breaks some tests because it bumps down a bunch of expected line numbers. We'll re-record them while merging it.

PR #2593, opened 2016-10-06
Fix typo

kandros:
As later snippets use the right name and does not use it as an example of undefined value


mroch:
good eye, thanks

PR #2609, opened 2016-10-11
Misc DOM method type definition fixes

loganfsmyth:
- Add some missing Attr properties
- Allow nullable inputs on more DOM methods
- Return the correct type for `Range#cloneContents`


gabelevi:
Sorry for the delay! This looks good to me! Went through the spec briefly and this seems to match up!

loganfsmyth:
Thanks!

PR #2610, opened 2016-10-11
Mark MessageEvent#data as mixed.

loganfsmyth:
For postMessage, `data` is anything that can be cloned using structured cloning, so it can be tons more things. `mixed` seemed like the best approach to it.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -271,12 +271,14 @@ declare class PromiseRejectionEvent extends Event {
   reason: any;
 }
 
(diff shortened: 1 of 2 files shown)


loganfsmyth:
Leaving the tests broken since they'll just conflict with https://github.com/facebook/flow/pull/2609.

gabelevi:
Sorry for the delay for this one too! I'm rebasing it on top of #2609 which just landed and will re-record the tests. Thanks for the contribution!

PR #2658, opened 2016-10-20
Explicitly mention scope of type aliases

Leeds-eBooks:
I have been quite confused about the use of `declare` before type aliases. I thought that `declare` made a type alias available in the project's global scope, but after reading [this tweet](https://twitter.com/lbljeffmo/status/788504721439031296), I realised that just putting `type T = {...}` was enough to make `T` available to flow throughout the project. Rather than asking this on issues and getting a confirmation, I thought I would make a PR in case it helps. This change I propose would have made it clear to me.

### DIFF
--- a/website/docs/01-type-aliases.md
+++ b/website/docs/01-type-aliases.md
@@ -21,8 +21,8 @@ var x: T = 0;
 ```
 
 We declare the new type `T` is an alias for the built-in type `number`.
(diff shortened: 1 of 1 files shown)


Leeds-eBooks:
This appears to be wrong actually, so I'm closing. The docs do need to be improved on this topic, but this change I proposed is incorrect.

PR #2662, opened 2016-10-20
Click and Double click are mouse events

### LINKED ISSUE #2661: click and dblClick are not MouseEvents (yet)
See https://developer.mozilla.org/en-US/docs/Web/Events.

These events are defined to have the MouseEvent type. I have reviewed the commit history and no mention has so far been made of any rationale for the exclusion of the events.

This makes it impossible to write:
addEventListener('click', function(event: MouseEvent)) { })

conartist6:
Summary: Add click and doubleClick to DOM MouseEventTypes.

Fixes #2661

Test plan: Run tests


mroch:
weird, these seem like the most obvious mouse events!

thanks!

PR #2670, opened 2016-10-22
Fix error message is not correct in Getting Started with Flow doc

### LINKED ISSUE #2669: Error message is not correct in Getting Started with Flow doc
In the example code in [Getting Started with Flow](https://flowtype.org/docs/getting-started.html), `str` is declared as `var`

`var str: number = 'hello world!';`.

But in the flow error message, the doc shows `let` instead.

```
index.js:3
  3: let str: number = 'hello world!';
                       ^^^^^^^^^^^^^^ string. This type is incompatible with
  3: let str: number = 'hello world!';
              ^^^^^^ number
```

It's just a minor typo. Let me fix this.

supasate:
Fix #2669

### DIFF
--- a/website/docs/getting-started.doc.js
+++ b/website/docs/getting-started.doc.js
@@ -75,9 +75,9 @@ console.log(str);
   > flow 2> /dev/null
 
   index.js:3
-    3: let str: number = 'hello world!';
+    3: var str: number = 'hello world!';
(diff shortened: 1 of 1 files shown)


gabelevi:
Thanks for fixing this! We really appreciate the help!

PR #2677, opened 2016-10-24
Add special cases for touch and wheel events in EventTarget class

straku:
Hi, I've stumbled upon an issue while adding flow annotations to existing module. Example:

```
el.addEventListener('wheel', handleWheel)

function handleWheel (event: WheelEvent) {
  const { deltaX, deltaY } = event
  console.log(deltaX, deltaY)
}
```

With that code I am not getting any errors for `deltaX` and `deltaY` because they exist in `WheelEvent`. I am getting error because there is a mismatch between declaration for `EventTarget` in `dom.js`, which defines that event handler should have parameter with `Event` type instead of `WheelEvent`.

If I replace `WheelEvent` annotation with `Event` or remove it completely I am getting errors that `deltaX` and `deltaY` are not found in `Event`.

If I add special cases for `EventTarget` declarations in `dom.js` like presented in this PR everything works fine. Unfortunately I have very little experience with flow and would be happy to know if this is a right direction or I am missing some essential knowledge about doing it with annotations and without the need of editing current declarations.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -94,25 +94,39 @@ type MouseEventHandler = (event: MouseEvent) => mixed
 type MouseEventListener = {handleEvent: MouseEventHandler} | MouseEventHandler
 type KeyboardEventHandler = (event: KeyboardEvent) => mixed
 type KeyboardEventListener = {handleEvent: KeyboardEventHandler} | KeyboardEventHandler
+type TouchEventHandler = (event: TouchEvent) => mixed
+type TouchEventListener = {handleEvent: TouchEventHandler} | TouchEventHandler
+type WheelEventHandler = (event: WheelEvent) => mixed
+type WheelEventListener = {handleEvent: WheelEventHandler} | WheelEventHandler
 
 type MouseEventTypes = 'contextmenu' | 'mousedown' | 'mouseenter' | 'mouseleave' | 'mousemove' | 'mouseout' | 'mouseover' | 'mouseup' | 'click' | 'dblclick';
 type KeyboardEventTypes = 'keydown' | 'keyup' | 'keypress';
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -94,25 +94,39 @@ type MouseEventHandler = (event: MouseEvent) => mixed
...
 type KeyboardEventListener = {handleEvent: KeyboardEventHandler} | KeyboardEventHandler
+type TouchEventHandler = (event: TouchEvent) => mixed
+type TouchEventListener = {handleEvent: TouchEventHandler} | TouchEventHandler
+type WheelEventHandler = (event: WheelEvent) => mixed
+type WheelEventListener = {handleEvent: WheelEventHandler} | WheelEventHandler
 
 type MouseEventTypes = 'contextmenu' | 'mousedown' | 'mouseenter' | 'mouseleave' | 'mousemove' | 'mouseout' | 'mouseover' | 'mouseup' | 'click' | 'dblclick';
 type KeyboardEventTypes = 'keydown' | 'keyup' | 'keypress';
+type TouchEventTypes = 'touchstart' | 'touchmove' | 'touchend' | 'touchcancel';
+type WheelEventTypes = 'wheel';
```
HitoriSensei: There are few more event types for WheelEvent triggers;
`type WheelEventTypes = 'wheel' | 'mousewheel' | 'DOMMouseScroll';`

gabelevi:
Thanks for the contribution! Tests will need to be updated, but I can do that for you. I'll get this imported, reviewed and merged ASAP

straku:
There are also `mousewheel` and `DOMMouseScroll` (non-standard) events that need `WheelEvent` type in the handler. Should I add them?

PR #2680, opened 2016-10-25
Overload addEventListener method of Worker.

jgrund:
A `Worker` instance that listens for a message event
should receive a `MessageEvent` object.

Overload `addEventListener` for `Worker`.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -325,6 +325,8 @@ declare class Worker extends EventTarget {
     constructor(stringUrl: string): void;
     onerror: (ev: Event) => any;
     onmessage: (ev: MessageEvent) => any;
+    addEventListener(type: 'message', listener: (evt: MessageEvent) => void, useCapture?: boolean): void;
+    addEventListener(type: string, listener: EventListener, useCapture?: boolean): void;
     postMessage(message: any, ports?: any): void;
     terminate(): void;
 }
--- a/tests/fetch/fetch.exp
+++ b/tests/fetch/fetch.exp
@@ -3,13 +3,13 @@ fetch.js:12
                                 ^^^^^^^^^^^^^^^^ function call
  12: const b: Promise<string> = fetch(myRequest); // incorrect
                       ^^^^^^ string. This type is incompatible with
-Response. See lib: bom.js:817
+Response. See lib: bom.js:819
 
 fetch.js:25
  25: const d: Promise<Blob> = fetch('image.png'); // incorrect
                               ^^^^^^^^^^^^^^^^^^ function call
-Library type error:. See lib: bom.js:817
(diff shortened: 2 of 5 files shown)


lib/bom.js
```diff
@@ -325,6 +325,8 @@ declare class Worker extends EventTarget {
     constructor(stringUrl: string): void;
     onerror: (ev: Event) => any;
     onmessage: (ev: MessageEvent) => any;
+    addEventListener(type: 'message', listener: (evt: MessageEvent) => void, useCapture?: boolean): void;
```
gabelevi: This could probably be `=> mixed` to be a little more lenient. I'll update it.

gabelevi:
Looks good to me! I'll import it and re-record the test! Thanks for the contribution!

gabelevi:
As an update: while testing this change I noticed that the time Flow spends initializing the builtin libraries regresses by about 20s on my machine. It turns out that this change hits some unoptimized edge cases for intersections. We're looking into optimizing those, and then will re-evaluate. Sorry for the delay!

jgrund:
@gabelevi Thanks for the update

philikon:
Any further updates? Seems like the lack of optimization is blocking at least this PR and #3077.

gabelevi:
I'm afraid not yet. Back when I showed this to @avikchaudhuri in November, he wrote

> This is an excellent case to try optimize. cc @samwgoldman who found out yesterday that the non-termination reported on GH a couple of days back is actually also a perf issue involving intersections, although probably needing a different fix than this one.
>
> I'm debugged enough of these to conjecture this can be fixed with some low hanging optimization (fail early on the first two cases by observing that string is not a subtype of either MouseEventTypes or KeyboardEventTypes). Separately, even after layers of optimization on type resolution used by union/intersection, we still probably blow up in funny ways by missing caching opportunities. I can look into this particular one after Nov 15, but will be happy to advise somebody else if they want to have a look first.

Unfortunately, we haven't tried optimizing this case yet. Other optimizations haven't helped with this. I'm also not sure if anyone on the team has free cycles to take this on at the moment. Sorry 😢

nmote:
I'm going to take another look at this to see if the blocking performance issues have been resolved in the intervening two years

nmote:
Unfortunately, it looks like this still causes significant problems. I don't think Flow is well-equipped to handle overloaded function declarations in subclasses. EventTarget already has a bunch of overloads for `addEventListener`.

This also causes type errors. Generally, a class that overrides a method must provide a method with a subtype of the parent class's method. However, the intent here is to overload, not override. I don't think Flow is well-equipped to express this.

I'm going to leave this open, because it's conceivable that this will change, but for now we can't take this PR.

PR #2700, opened 2016-10-28
Fix prev/next links in docs pages.

jonboiser:
https://flowtype.org/docs/

This PR fixes the links on the bottom of three docs pages to match the order implied in the Table of Contents. 

Here's how the docs currently flow (if they are to match TOC on left of the page). Below are what they should be (and what the PR tries to implement).

"quickreference" -> "builtins"
**should be**
"quickreference" -> "syntax"

"syntax" -> "arrays"
**should be**
"syntax" -> "builtins"

### DIFF
--- a/website/docs/ref/builtins.doc.js
+++ b/website/docs/ref/builtins.doc.js
@@ -4,8 +4,8 @@
 id: builtins
 title: Built-in Types
 permalink: /docs/builtins.html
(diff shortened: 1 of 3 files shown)


gabelevi:
Whoops! Thanks for fixing! We really appreciate it!

PR #2707, opened 2016-10-28
Document the existence of compiled-to-JS Flow parser on npm

wbinnssmith:
I figured this could be useful to promote more prominently on the main
README. Feel free to refine or reject.

Also, while setting up a build for the JS parser, I found I also needed
to install ocamlfind (used in the Makefile) which didn't come alongside
any of the other packages. If this is the case, I can add that to the
parser README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ Flow can also compile its parser to JavaScript. [Read how here](src/parser/READM
 
 While Flow is written in OCaml, its parser is available as a compiled-to-JavaScript module published to npm, named [flow-parser](https://www.npmjs.com/package/flow-parser). **Most end users of Flow
 will not need to use this parser directly** (and should install [flow-bin](https://www.npmjs.org/package/flow-bin) from npm above), but JavaScript packages which make use of parsing
-Flow-typed JavaScript can use this to generate Flow's syntax tree with inferred types attached.
+Flow-typed JavaScript can use this to generate Flow's syntax tree with annotated types attached.
 
 ## Running the tests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,6 +114,14 @@ This produces a `bin` folder containing the `flow` binary.
 
 *Note: at this time, the OCaml dependency prevents us from adding Flow to [npm](http://npmjs.org). Try [flow-bin](https://www.npmjs.org/package/flow-bin) if you need a npm binary wrapper.*
 
+Flow can also compile its parser to JavaScript. [Read how here](src/parser/README.md).
+
+## Using Flow's parser from JavaScript
+
+While Flow is written in OCaml, its parser is available as a compiled-to-JavaScript module published to npm, named [flow-parser](https://www.npmjs.com/package/flow-parser). **Most end users of Flow
+will not need to use this parser directly** (and should install [flow-bin](https://www.npmjs.org/package/flow-bin) from npm above), but JavaScript packages which make use of parsing
+Flow-typed JavaScript can use this to generate Flow's syntax tree with inferred types attached.
```
gabelevi: I'm afraid this isn't quite true. The parser will output an AST with Flow type annotations included, but inferred types are not represented in the AST. The `flow-parser` npm package contains only the Flow parser.

wbinnssmith:
Thanks for letting me know @gabelevi! Just replaced inferred with annotated so it's explicit which types are attached :)

Let me know if anything else is needed! Also, can you confirm that ocamlfind is needed for a build environment for the parser (at least the js target)? Would be happy to note that in the parser's README.

gabelevi:
Looks good to me! Thanks for updating!

And yeah, you're right, ocamlfind is needed to build the js version of the parser. IIRC, we use `ocamlbuild -use-ocamlfind -pkgs js_of_ocaml` to compile the code that provides the JavaScript API to the parser, and then use `js_of_ocaml` directly to create the JavaScript code.

Merging this now! Thanks again for the contribution!

PR #2708, opened 2016-10-28
bom: TextEncoder constructor parameter optional

happylynx:
... as defined in https://www.w3.org/TR/encoding/#interface-textdecoder


mroch:
agreed, thanks!

PR #2717, opened 2016-10-30
Array#pop and Array#shift should return nullables

AprilArcus:
Calling Array#pop and Array#shift on empty arrays returns `undefined`, so the return value of these functions should be annotated as nullable.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -198,7 +198,7 @@ declare class Array<T> {
     keys(): Iterator<number>;
     lastIndexOf(searchElement: T, fromIndex?: number): number;
     map<U>(callbackfn: (value: T, index: number, array: Array<T>) => U, thisArg?: any): Array<U>;
-    pop(): ?T;
+    pop(): T | void;
     push(...items: Array<T>): number;
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: Array<T>) => T,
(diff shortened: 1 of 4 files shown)


gabelevi:
Thanks for the contribution and sorry for the delay! This looks good to me! I rebased this branch and switched `?T` to `T | void` which is a little bit more precise. I also fixed the related Flow code. This change will cause lots of Flow errors in existing code, but it IS right so is probably worth it. I'll run it past the team, but I think they'll be onboard.

Thanks again!

gabelevi:
Ok, so I've discussed this with the team and we're not going to take this PR for now. The main reason is that there is a lot of code like

```
while(arr.length > 0) {
  doSomething(arr.shift());
}
```

or

```
arr.length > 0 && doSomething(arr.pop());
```

If we took this PR as is, Flow would make Flow users check for `undefined` in the result of `pop()` and `shift()`, even if the user properly guarded against empty arrays.

Future work could result in a smarter Array type that carries a non-empty bit. Then we could refine the type of arrays with length checks and support the common JavaScript idioms. But for now, I'm going to close out this PR.

Thanks for your patience and understanding. Sorry this took so long to resolve!

benlangfeld:
@gabelevi If this is considered along with indexing, there are two arguments that could be made:

1. The general case is not an empty Array, but the absence of a value at the requested index.
2. In some theoretical context where the interpreter is multi-threaded, your examples are not "properly guarded against empty arrays", and the type check should be explicitly on the return value. All of these cases should be errors to be truly safe:

```javascript
const foos: Array<number> = [1]
const foo: number = foos[1]

const bars: Array<number> = []
const bar: number = bars.pop()
const baz: number = bars.shift()
```

Particularly, I would like Flow to force me to write this:

```javascript
const foos: Array<number> = [1]
const foo: number | void = foos[1]

const bars: Array<number> = []
const bar: number | void = bars.pop()
const baz: number | void = bars.shift()
```

I don't think your suggestion of a non-empty Array type would work for all of these cases, would it? Would it instead be possible to make this nullable behaviour opt-in for those who want Flow to enforce this safety?

PR #2721, opened 2016-10-31
Update variance.doc.js

vyorkin:
The actual function name is `h`:
```
function h(o: {p: T1}): void {
}
```
But its mistakenly referenced as `f`.


mroch:
indeed, thanks!

PR #2735, opened 2016-11-02
Update declarations.doc.js

morozow:
Fixed error in "collocated" word


gabelevi:
Whoops, thanks for fixing this typo! We really appreciate it!

PR #2739, opened 2016-11-02
[flowtype.org] use Yarn instead of npm/bower

mroch:


mroch:
don't import yet, I need to unstack this from #2738. also shooting in the dark here, we'll see how travis likes this.

PR #2755, opened 2016-11-04
fix: typo in node defs: s/creatSign/createSign

keithamus:


keithamus:
In addition to this - Node also exports the `Sign` and `Verify` classes - but this flow definition does not export them. Can I make a (separate) PR to also export these classes?

jeffmo:
@keithamus: Definitely! Thanks for this :)

keithamus:
Thanks for mering @jeffmo. Could you also please point me in the right direction for fixing this on my local project? Can I somehow redeclare this function?

PR #2756, opened 2016-11-04
fix: add node.js crypto.Sign and crypto.Verify exports

keithamus:
Since Node 0.1.92, one has been able to `require('crypto').Sign` which gives you the class that the factory `createSign` instantiates. Same for `Verify`.

This adds the two classes in the type signature for the crypto module. Meaning `require('crypto').Sign` validates properly for flow users who chose to use it over `createSign` (and again, same for Verify).

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -465,6 +465,9 @@ declare class crypto$Verify extends stream$Writable {
 
 declare module "crypto" {
   declare var DEFAULT_ENCODING: string;
+  
+  declare class Sign extends crypo$Sign {};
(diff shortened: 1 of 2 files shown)


gabelevi:
Thanks for the contribution and sorry for the delay! There were a few syntax/typos, but I pushed fixes to the branch. Now this looks good to me! I'll merge ASAP! Thanks again!

keithamus:
Hah, sorry I did this a bit too quickly inside github. Thanks for making those changes, look forward to seeing this merged!

PR #2757, opened 2016-11-04
fix: Node.js definitions: supply static & constructor signatures for crypto$Verify & crypto$SIgn

keithamus:
The signatures for these class constructors differ very slightly to those of WritableStream. They require an algo as the first argument - and they also can be called without `new`. This PR adds declarations for both of these features.

([Source for verify](https://github.com/nodejs/node/blob/7537718460c7b964ffbbc0910b12eaff9cd8b7a8/lib/crypto.js#L293-L302), [Source for sign](https://github.com/nodejs/node/blob/7537718460c7b964ffbbc0910b12eaff9cd8b7a8/lib/crypto.js#L259-L266))

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -442,6 +442,8 @@ type crypto$Sign$private_key = string | {
   passphrase: string,
 }
 declare class crypto$Sign extends stream$Writable {
+  static(algorithm: string, options?: writableStreamOptions): crypto$Sign,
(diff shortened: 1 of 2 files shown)


gabelevi:
Alrighty, this looks good to me too! I'll get this merged! Thanks for the contribution and sorry for the delay!

PR #2766, opened 2016-11-06
Add missing ChildProcess#unref()

jeanlauliac:
Node.js' ChildProcess class has a barely-documented function 'unref', that does not have its own section but is described as part of the 'detached' spawn option: https://nodejs.org/api/child_process.html#child_process_options_detached.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -120,7 +120,8 @@ declare module "buffer" {
   declare var INSPECT_MAX_BYTES: number;
 }
 
-type child_process$execOpts = {
+type
+$execOpts = {
   cwd?: string;
   env?: Object;
   encoding?: string;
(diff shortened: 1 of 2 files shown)


stephencelis:
May want to add `ref(): void` while you're at it!

``` js
> ChildProcess.prototype
ChildProcess {
  spawn: [Function],
  kill: [Function],
  ref: [Function],
  unref: [Function] }
```

gabelevi:
Thanks for the contribution and sorry for the delay! This looks good to me! I'll get this merged!

stephencelis:
@gabelevi Any chance on getting `ref()` in at the same time?

gabelevi:
@stephencelis Sure!

PR #2767, opened 2016-11-06
WeakMap#get can be undefined

gre:
in current defined types,
`Map#get` returns `V | void`
but
`WeakMap#get` returns `V`
.

It's inconsistent and this PR makes `WeakMap#get` to return `V | void` as well.


gabelevi:
Sorry for the delay! Nice find, you're right! Thanks for fixing this! I'll get it merged ASAP!

PR #2773, opened 2016-11-08
Added Gitter flow channel link

montogeek:
Added Gitter #flow channel link to Support page


gabelevi:
Sorry for the delay! We discussed this a little internally. We're a little worried about the Flow team being spread thin. We already have an IRC room, GitHub issues, a Stack Overflow tag, our internal channels, and have been playing with Slack. So we don't want to give the impression that we're monitoring a community when we aren't really.

That said, we totally support Flow users forming communities wherever they like! So maybe we could just change the wording to something like

```
You can also find a bunch of Flow users hanging out in the [#flow](https://gitter.im/facebook/flow) Gitter channel
```

montogeek:
Thanks @gabelevi :)

gabelevi:
Awesome! Looks good to me! Thanks!

PR #2779, opened 2016-11-10
Complete class declaration for FormData spec

acusti:
Reference: https://xhr.spec.whatwg.org/#interface-formdata

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -241,7 +241,11 @@ declare class DOMParser {
 }
 
 declare class FormData {
-    append(name: any, value: any, blobName?: string): void;
+    has(name: string): boolean;
+    get(name: string): ?string;
+    set(name: string, value: any, blobName?: string): void;
+    append(name: string, value: any, blobName?: string): void;
+    delete(name: string): void;
 }
 
 declare class MutationRecord {
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -242,9 +242,11 @@ declare class DOMParser {
 
 declare class FormData {
     has(name: string): boolean;
-    get(name: string): ?string;
-    set(name: string, value: any, blobName?: string): void;
-    append(name: string, value: any, blobName?: string): void;
+    get(name: string): ?(string | File);
+    set(name: string, value: string): void;
+    set(name: string, value: Blob, filename?: string): void;
+    append(name: string, value: string): void;
+    append(name: string, value: Blob, filename?: string): void;
     delete(name: string): void;
 }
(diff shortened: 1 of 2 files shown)


acusti:
I manually updated the line numbers in the expectations in `tests/fetch/fetch.exp` to match the new line numbers form the expanded FormData declaration. Those tests are now passing. Got an error from `_build/src/parser/test/run_hardcoded_tests.native src/parser/test/hardcoded_tests.js` however:

```
cd /home/travis/build/facebook/flow/src/parser//../..; \
	_build/src/parser/test/run_hardcoded_tests.native src/parser/test/hardcoded_tests.js
Fatal error: exception Not_found
```

Travis job: https://travis-ci.org/facebook/flow/jobs/174823776

Don’t know why that would be, but I’d be happy to take a look at it if someone could point me in the right direction.

acusti:
Also, if you’d prefer not to include the entire `FormData` interface as defined by the spec because it isn’t supported by all major browsers, we can close this PR.

lib/bom.js
```diff
@@ -241,7 +241,11 @@ declare class DOMParser {
 }
 
 declare class FormData {
-    append(name: any, value: any, blobName?: string): void;
+    has(name: string): boolean;
+    get(name: string): ?string;
```
gabelevi: Looking at the spec, I think the return type is `?(string | File)`

lib/bom.js
```diff
@@ -241,7 +241,11 @@ declare class DOMParser {
...
+    set(name: string, value: any, blobName?: string): void;
```
gabelevi: An overload would also work

```
set(name: string, value: string): void
set(name: string, value: Blob, filename?: string): void
```

lib/bom.js
```diff
@@ -241,7 +241,11 @@ declare class DOMParser {
...
+    append(name: string, value: any, blobName?: string): void;
```
gabelevi: ditto for append

gabelevi:
I'm cool including this! And sorry for the delay! I have some notes, but since this is so delayed I'll address them myself :)

gabelevi:
Well I feel silly. Went to rebase this branch and found out we already merged a PR with definitions for this (https://github.com/facebook/flow/pull/2471) 9 days ago. So closing this out :)

Thanks again and sorry for the trouble!

acusti:
Hah, that’s pretty funny! Glad to hear the definitions made it in, and thanks for the work and feedback.

PR #2782, opened 2016-11-10
Add indexed keys type to array definition

ForbesLindesay:
I'm attempting to create an interface for a subset of the array methods.  Specifically the read only methods.

Because the flow definition does not include this line, flow tells me that `Array<T>` doesn't match the interface:

```js
declare interface ReadOnlyArray<+T> {
  // ...other read only methods here...
  map<U>(
    callbackfn: (value: T, index: number, array: Array<T>) => U,
    thisArg?: mixed,
  ): Array<U>,

  length: number,

  [key: number]: T,
}
```

I believe adding it here will fix that.


vkurchatkin:
I suspect it wasn't there in the first place to support tuple functionality

ForbesLindesay:
That makes some sense.  Is there any way I can force flow to recognise that Array implements my interface?  Currently there isn't even anywhere good to add a `$FlowIgnore` comment.

ForbesLindesay:
The tests do seem to be passing though (on most platforms at least).  I would think we should have tests for the tuple types, and ideally tuple should be a separate type from array anyway.

ianobermiller:
The tuple tests are here: https://github.com/facebook/flow/tree/dd0603e9d8c9d5fb99a40f0b179a4d6a2b9e66b7/tests/tuples

None of them appear to cover indexing a tuple (oddly). Maybe add a couple test cases that check tuple indexing specifically? If they pass fine, then we can assume this change is safe.

samwgoldman:
Yeah, tuples are handled by a separate internal type. I think I've tried to do this before (for the same reason, to add a covariant Array interface) and ran into some issue but I forget what it was.

I'll import and check this out. Thanks, Forbes!

PR #2804, opened 2016-11-14
Type-preserving versions of `Object.entries`/`Object.values`

davidaurelio:
This adds signatures for `Object.entries` and `Object.values` that take a `{[key: string]: T}` and return `Array<[string, T]>` and `Array<T>`, respectively.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -35,6 +35,7 @@ declare class Object {
     static create(o: any, properties?: any): any; // compiler magic
     static defineProperties(o: any, properties: any): any;
     static defineProperty(o: any, p: any, attributes: any): any;
+    static entries<T>(object: {[key: string]: T}): Array<[string, T]>,
     static entries(object: any): Array<[string, mixed]>;
     static freeze<T>(o: T): T;
     static getOwnPropertyDescriptor(o: any, p: any): any;
@@ -49,6 +50,7 @@ declare class Object {
     static preventExtensions(o: any): any;
     static seal(o: any): any;
     static setPrototypeOf(o: any, proto: ?Object): bool;
+    static values<T>(object: {[key: string]: T}): Array<T>,
     static values(object: any): Array<mixed>;
     hasOwnProperty(prop: any): boolean;
     isPrototypeOf(o: any): boolean;


vkurchatkin:
This is unsafe in general:

```js
declare var x: { foo: string, bar: number }
var y: { foo: string } = x;

var a: { [k: string]: string } = y;

Object.values(a).forEach(v => v.toUpperCase())
```

davidaurelio:
True, but the original version is not useful for the common case. Any suggestions to make it better?

vkurchatkin:
> Any suggestions to make it better?

No, it's just something to consider. To be fair, the problem is with dictionary object, this just exposes it in a new way

davidaurelio:
Arguably, `{foo: string}` shouldn’t allowed to flow into `{[key: string]: string}`, whereas it is safe for `{|foo: string|}`.

vkurchatkin:
@davidaurelio Totally agree. I hope that it would be possible to make dictionaries safe in the future

conartist6:
My use case for this involves subtypes of string (a union of individual strings), would it be possible to not assume 'string' keys?

lib/core.js
```diff
@@ -35,6 +35,7 @@ declare class Object {
     static create(o: any, properties?: any): any; // compiler magic
     static defineProperties(o: any, properties: any): any;
     static defineProperty(o: any, p: any, attributes: any): any;
+    static entries<T>(object: {[key: string]: T}): Array<[string, T]>,
```
davidaurelio: @zertosh has two examples where this is broken: https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVxhgEICUDyA0gKIByqAJgKYDGMAhgE5Vh30DO7Y+ARgFa0ALmADeqMBPaD6ggJY0wVAHaDGsquwA8AQQA0OAHwAKOPyEAuUQG0A1lQCel7QF1L2AL4BKJ40b17mlZ6OM4GANyo7qgAbkxg9PQ+fgFWAOQ0qfqpFKmhYAC83GY0ggB0yqrq7EY1SgCuMDCW9Er23kUCJZ6eERhYAOL4+AAi6NRszKwMnB1CouKS0nIKFWoamtjGpp2CliK2DpZSakoA5q44Xkn+gceyZ-rYoRFRsYxgPDzXKemZYNm5AwFWYlcoqNbVWoNJrxVrtXg7boRIA

conartist6: Rebuttal: https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVxgDEDyAZPHAdQGUwBBEkgSQHEA5AWQFF6AVMLQo1AEwFMAxjACGAJ35hhIgM4ywOAEYArIQBcwAb1RhdMtSLUBLQWH4A7NWKP8ZAHnIAaMACEAfAAo4K9QC4tANoA1vwAnv7kALr+LgC+AJQRYmIioXYBTq6RbgDcqLGoAG7iYCLkckYA5uZgvprBYf4A5CJNkbVNik2xYAC8YB4e5gCuMDD+IuahiQo+gmrxecVipcyW1rZJKWkBTYJNzk28bW59s6rzAHQWVjYyg0Oj46VTM0oXC4vofEKiElKiOTndRaHR6AzGUw3Db2dxeOZqfz1ELhMD6azmSrRVwJLapdLooyY5wubJ5ArLMCKcoyKrmJG7fbtXxHbpnQYjMYTV4MvZtDrHBJ5TAuABKzHIAGkyEQABKsMCsNii6jMMiUEgAVRYZBwLgAUmASLLyAAFZhFEqKNa3TYUZL4xkHMCs7Jnd7qa7rO4PTnPSbTXlMgXdeKLIA

conartist6: I also just ran across this in our codebase:
```
declare class Seq<K,V> extends Iterable<K,V> {
...
<K:string,V>(object: {+[k:K]:V}): KeyedSeq<K,V>
```

gabelevi:
I think @samwgoldman might have plans here, though I might be misremembering. I thought that exact types were supposed to help with this kind of thing. Something like

```
declare class Object {
  ...
  static entries<T>(object: {| [key: any]: T |}): Array<[string, mixed]>;
  static entries(object: any): Array<[string, mixed]>;
  ...
}
```

Though I'm not sure if exact dictionaries work at the moment. That is, I don't think this currently works

```
({ foo: 'hello' }: {| [any]: string |});
```

Anyway, I don't think we'll be able to solve this with only a libdef update. So I'm going to close this PR, though feel free to continue the discussion!

zertosh:
These days, with `$Keys` and `$Values`, we can better type these, see #5663

PR #2806, opened 2016-11-14
Forward abnormal-return for `while(true) { return ; }`

jeffmo:
Before this change, the following would error stating that the function could implicitly return `undefined` (incompatible with the stated `string` return type):

```js
function f(): string {
  while (true) {
    return '';
  }
}
```

This change fixes the problem only for exact `while (true)` syntactic scenarios. There's a lot more work to be done to fix it for indirect scenarios (i.e. `let t = true; while (t)`), so we punt on fixing that for now.

### DIFF
--- a/newtests/while_true_return/_flowconfig
+++ b/newtests/while_true_return/_flowconfig
@@ -1,0 +1,8 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
+all=true
--- a/newtests/while_true_return/test.js
+++ b/newtests/while_true_return/test.js
@@ -1,0 +1,29 @@
+/* @flow */
+
+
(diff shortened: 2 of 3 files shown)


jeffmo:
<<Plan Changes>>

Looks like this fix also permits (i.e. doesn't error on) the following:

```js
function f(): string {
  while (true) {
  }
}
```

jeffmo:
<<Planning Changes>>

Looks like this fix also permits (i.e. doesn't error on) the following:

```js
function f(): string {
  while (true) {
    if (Math.random()) {
      return '';
    }
  }
}
```

gabelevi:
Closing this out for now, but feel free to re-open this if you pick it back up!

PR #2817, opened 2016-11-15
Fix the ocp-build build

gabelevi:
Stole this from @jeffmo since he's busy


gabelevi:
To test I ran `make all-ocp`

PR #2833, opened 2016-11-16
Add nullability to return of Element::getAttribute

disnet:
`Element::getAttribute` returns `null` when the attribute does not exist ([spec](https://dom.spec.whatwg.org/#dom-element-getattribute) and browsers agree).


michaelficarra:
@disnet The three functions that follow are also partial.

PR #2838, opened 2016-11-17
Add EventListenerOptions support

nkt:
This PR adds support for EventListenerOptions.
`passive`: https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md
`once` and others: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


gabelevi:
Ok, looks good to me! I made the names a little bit more precise, to reflect that these variables are serving a double purpose. I think it's better than adding more overloads, though, so I agree with your approach. Thanks for the contribution and sorry for the delay! I'll get this merged ASAP!

PR #2843, opened 2016-11-18
`Buffer#fill` returns `this`

davidaurelio:
The `fill` method of Buffer objects returns `this` rather than `void`.


mroch:
we really need to version the node defs. :/ this did return void in the past. I checked 3.0.0 and it returned `this`, and 0.10 and it returned void (https://github.com/nodejs/node/blob/e8eaaa772414e6193702f3933cbee919d6fcdc95/src/node_buffer.cc#L559). didn't check to see when it actually changed though.

PR #2863, opened 2016-11-21
[bom] Define MutationObserverInit

kuy:
`MutationObserverInit` is used as an argument named `options` in `MutationObserver#observe`.
The name is borrowed from here: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#observe()

I wanted to catch a runtime error "An invalid or illegal string was specified", but I don't know how to describe the condition written in [side note](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#MutationObserverInit).

> NOTE: At the very least, childList, attributes, or characterData must be set to true. Otherwise, "An invalid or illegal string was specified" error is thrown.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1222,7 +1222,890 @@ declare class CanvasRenderingContext2D {
   ellipse(x: number, y: number, radiusX: number, radiusY: number, rotation: number, startAngle: number, endAngle: number, anticlockwise?: boolean): void;
 }
 
-declare class WebGLRenderingContext {}
+
+// WebGL idl: https://www.khronos.org/registry/webgl/specs/latest/1.0/webgl.idl
+
+type WebGLContextAttributes = {
+  alpha: bool,
+  depth: bool,
+  stencil: bool,
+  antialias: bool,
+  premultipliedAlpha: bool,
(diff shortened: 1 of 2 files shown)


gabelevi:
> I wanted to catch a runtime error "An invalid or illegal string was specified", but I don't know how to describe the condition written in side note.

I suppose you could write a union, [like this](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgBiccYAvKmGAD5gDeYAhgFxgYBOArngDRgBGrfiRh5GAOz4BjISLHiwAX0o16TWXFES+gtl15gZAuRKUraDFsc3ydGrZMOsO3JQG50ACjpWXB3VCMMADOBkaBIXiKrMRwAJQe3lYRoXZ63NKsKVExJAmoSVlBqQJFkZnpOUR5iT5lJQHFYfVVsQlAA)

gabelevi:
Do you want me to merge this as is, or do you want to try and do the union approach?

kuy:
@gabelevi Thanks for replying and giving an example code of union type. I want to try the union approach.

kuy:
Oh... I did misoperation. I'm still working on this PR.

PR #2873, opened 2016-11-22
define_globals should not write to shared mem

gabelevi:
There are two main reasons why `define_globals` should not write to the shared memory

1. It is called once by the master process and once by each of the workers. It defines the local globals, but it does NOT initialize the shared memory
2. The memory isn't actually ready to be written yet on Windows.

`init_shared_globals` is the right place for this kind of code.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -735,6 +735,10 @@ static void memfd_reserve(char *mem, size_t sz) {
 
 #endif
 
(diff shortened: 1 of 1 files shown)


hack/heap/hh_shared.c
```diff
@@ -735,6 +735,10 @@ static void memfd_reserve(char *mem, size_t sz) {
 
 #endif
 
+// DON'T WRITE TO THE SHARED MEMORY IN THIS FUNCTION!!!
+// This function just calcuates where the memory is. It
```
mroch: tpyo

PR #2890, opened 2016-11-24
Add import type shorthand syntax

jeffmo:
Currently, importing both types and values from the same modules requires two statements with most of the same boilerplate in both statements:

```js
import {someValue} from "blah";
import type {someType} from "blah";
import typeof {someOtherValue} from "blah";
```

This gets tediously repetitive and adds a lot of boilerplate to JS files.

This PR implements a proposal from @leebyron to add an additional `type`/`typeof` shorthand to vanilla-`import` statements for individual named specifiers so that the above code can be written more concisely as:

```js
import {
  someValue,
  type someType,
  typeof someOtherValue,
} from "blah";
```

This PR does not remove `import type` or `import typeof` statements -- it only adds the shorthand to vanilla `import` statements.

Note also that I've banned usage of the shorthand with `import type` and `import typeof` statements since the need for shorthand in those is far less common and often combinatorially confusing:

```js
import type { type T } from "foo"; // <-- parse error
import typeof { type T } from "foo"; // <-- parse error
```

### DIFF
--- a/newtests/import_type_shorthand/_flowconfig
+++ b/newtests/import_type_shorthand/_flowconfig
@@ -1,0 +1,8 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
+all=true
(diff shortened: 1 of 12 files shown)


CodyReichert:
Nothing super constructive, but I'm really looking forward to this. Nice work!

Issue #1514, opened 2016-03-09
Node's Buffer is a Uint8Array


philikon:
Node's `Buffer` is a subclass of `Uint8Array` these days and there should be accepted where `Uint8Array` is specified:

```
> new Buffer(1).__proto__
Buffer { ... }
> new Buffer(1).__proto__.__proto__
Uint8Array {}
```

Moreover, `Buffer` instances have `Uint8Array` properties (such as `byteOffset`, etc.) and can be passed to the `Buffer` constructor.

philikon:
Is this as simple as changing `declare class Buffer` to `extend Uint8Array`? If so, happy to give that a shot.

PR #2897, opened 2016-11-26
Make node's Buffer class extend Uint8Array

aickin:
Node's `Buffer` class is now a `Uint8Array`, but the current flow builtins don't recognize that fact. This results in two problems:

1. There are a bunch of methods on `Buffer` that flow doesn't know about, like `forEach` and `reduce`.
2. Methods that take a `Uint8Array` can't take a `Buffer`, even though `(new Buffer(0)).__proto__.__proto__` returns `Uint8Array {}`.

I've tried to fix that in this PR, but it's a little hairy, because `Buffer` changes the signature of `Uint8Array.from`. From [the Node 6.9.1 documentation](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html#buffer_buffers_and_typedarray):

> The Buffer.from() and TypedArray.from() have different signatures and implementations. Specifically, the TypedArray variants accept a second argument that is a mapping function that is invoked on every element of the typed array

This PR defines `TypedArray.from` with the optional mapping function & this argument, which means that `Buffer` incorrectly inherits that function signature. Also, `Buffer.from(Array<number>)` has to be changed to `Buffer.from(Iterable<number>)` because the former is not compatible with `TypedArray.from(Iterable<number>)`. This means that `Buffer.from` will pass type checks incorrectly when passed an Iterable that is not an array. These effects are unfortunate, as they will allow some code to pass type checks incorrectly, but I don't see another way to do it. If there is a better way, please let me know.

Note also that there are a few fixes to function signatures that are just simple correctness fixes but are required to allow `Buffer` to extend `Uint8Array`: `TypedArray.entries`, `TypedArray.fill`, `TypedArray.findIndex`, `TypedArray.keys`, and `Buffer.slice`. I used the [MDN TypedArray page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) and [the Node 6.9.1 Buffer page](https://nodejs.org/dist/latest-v6.x/docs/api/buffer.html) as my references when trying to determine the true function signatures.

I believe this PR should fix issue #1514.

Thanks for all the work you do, and let me know what you think. Thanks!

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -626,7 +626,7 @@ type $ArrayBufferView = $TypedArray | DataView;
 // http://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%-intrinsic-object
 declare class $TypedArray {
     static BYTES_PER_ELEMENT: number;
-    static from(iterable: Iterable<number>): this;
+    static from(iterable: Iterable<number>, mapFn?: (element: number) => number, thisArg?: any): this;
     static of(...values: number[]): this;
 
     constructor(length: number): void;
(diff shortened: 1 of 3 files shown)


gabelevi:
Sweet, this looks good to me! Thanks for taking the time to clean this up! Will get this merged!

Issue #2544, opened 2016-09-29
crypto$Decipher#update type annotation incorrect


nicks:
According to
https://nodejs.org/api/crypto.html#crypto_decipher_update_data_input_encoding_output_encoding
the input_encoding should be `'latin1' | 'base64' | 'hex'`

for some reason, flow's definitions are `'utf8'| 'ascii' | 'binary'`, though it's not clear where that came from.
https://github.com/facebook/flow/blob/8e9755ed36d631679314110ea296683c71f269f0/lib/node.js#L340

glennreyes:
@nicks node v4 used different types for this API:
https://nodejs.org/dist/latest-v4.x/docs/api/crypto.html#crypto_class_decipher

I'll probably send a PR tackling this 👍

PR #2934, opened 2016-12-02
lib: fix typedef for `crypto$Decipher`.

iamchenxin:
[decipher.update api](https://nodejs.org/api/crypto.html#crypto_decipher_update_data_input_encoding_output_encoding)
The `input_encoding` should be `'latin1'| 'base64' | 'hex'`.
and `output_encoding` should be `'latin1' | 'ascii' | 'utf8'`.

Also modified all `binary` to `latin1`, for keeping synchronized with `Nodejs` docs.
The `binary` in `buffer$NonBufferEncoding` is retained, i'm not sure if there is someone need it.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -22,7 +22,7 @@ type buffer$NonBufferEncoding =
   'binary' | 'BINARY' |
   'base64' | 'BASE64' |
   'ucs2' | 'UCS2' | 'ucs-2' | 'UCS-2' |
-  'utf16le' | 'UTF16LE' | 'utf-16le' | 'UTF-16LE';
+  'utf16le' | 'UTF16LE' | 'utf-16le' | 'UTF-16LE' | 'latin1';
 type buffer$Encoding = buffer$NonBufferEncoding | 'buffer'
 type buffer$ToJSONRet = { type: string, data: Array<number> }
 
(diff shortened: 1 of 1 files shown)


mroch:
thanks! check out https://github.com/facebook/flow/issues/2544. the API changed and our libdefs are unversioned. for now, we should probably support the union of previously- and currently-valid values.

iamchenxin:
@mroch Maybe all typedefs  should be separated to flow-typed(with version). 
I used an `any` to cast `crypto$Decipher` today, and then cost hours to find a bug, that i typo `crypto.createDecipheriv` to `crypto.createCipheriv`. ~_~.

gabelevi:
Eventually we would like Flow to be aware of under which environments the code will run. Then Flow would grab the libdefs for those environments and merge them. However that's the future and for now we just try to provide builtin libdefs that work everywhere. So I think `'latin1' | 'binary'` might be the best thing to do for now

iamchenxin:
@mroch @gabelevi Make both of `'binary'` and  `'latin1'`` exist .
Seems, is `crypto$Decipher` really typo? [v4.x decipher.update](https://nodejs.org/dist/latest-v4.x/docs/api/crypto.html#crypto_decipher_update_data_input_encoding_output_encoding).
It is as the same param order as v6.x.

gabelevi:
Great! Thanks for making those changes! Looks good to me now! Importing...

Issue #2971, opened 2016-12-08
XMLHttpRequest: addEventListener event types


nhardy:
The class `XMLHttpRequest` does not currently specify the specific event types for events added with `.addEventListener()`, such as `ProgressEvent` for `'progress'` events.

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

PR #2972, opened 2016-12-08
Overload (add/remove)EventListener method of XMLHttpRequest

nhardy:
Relatively new to flow, but I believe this should do the trick fixing #2971 (I think).

Happy to make modifications as necessary.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -427,27 +427,34 @@ declare class XDomainRequest {
     }
 }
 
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks! If I recall correctly, we tried to import something like this before and it exposed a performance issue (with Flow itself). I will import this to run some performance tests we do internally and report back.

PR #2974, opened 2016-12-08
modify `mode?:  any` to `mode?: number`,  and add a new var `constants` for node 6.x.

iamchenxin:
The first commit  modified `mode?:  any` to `mode?: number` for both of `4.x` and `6.x`.

Second commit is for `6.x`. add a new var [`constants`](https://nodejs.org/dist/latest-v6.x/docs/api/fs.html#fs_fs_constants_1) for node 6.x. ( Seems the constansts of fs are collected to `constants` from 6.x). It will not make backward compatibility issues, since both of them are numbers. (`fs.F_OK == fs.constants.F_OK`).

Maybe in the future, when the node defs are separated by versions, we can use number literals to restrict the args.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -836,8 +836,8 @@ declare module "fs" {
   declare function watch(filename: string, options?: Object, listener?: (event: string, filename: string) => void): FSWatcher;
   declare function exists(path: string, callback?: (exists: boolean) => void): void;
   declare function existsSync(path: string): boolean;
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks! You're right that we need to figure out a solution for supporting versions for builtin libs. For now we're trying to support everything from every version, so this change seems correct to me.

Issue #2464, opened 2016-09-14
Document advanced Utility Types


gabro:
In the process of learning Flow and adopting it at work, I've often needed "advanced" operators that are completely missing from the documentation.

So far I relied on this blog post (now slightly outdated) http://sitr.us/2015/05/31/advanced-features-in-flow.html and code found in the Flow and React codebases.

I think I'd be nice to have this features explained in the official documentation.

The features I'd like to see documented are:
- [x] [$Keys<T>](https://flow.org/en/docs/types/utilities/#toc-keys)
- [x] [$Diff<A, B>](https://flow.org/en/docs/types/utilities/#toc-diff)
- [x] ~$Supertype<T>~ ([deprecated](https://flow.org/en/docs/linting/rule-reference/#toc-deprecated-utility))
- [x] ~$Subtype<T>~ ([deprecated](https://flow.org/en/docs/linting/rule-reference/#toc-deprecated-utility))
- [x] ~$Abstract<T>~ ([removed](https://github.com/facebook/flow/pull/4863))
- [x] [Class<T>](https://flow.org/en/docs/types/utilities/#toc-class)
- [x] [$PropertyType<T, x>](https://flow.org/en/docs/types/utilities/#toc-propertytype)
- [x] [\* (existential types)](https://flow.org/en/docs/types/utilities/#toc-existential-type)
- [x] [$ObjMap](https://flow.org/en/docs/types/utilities/#toc-objmap) / [$TupleMap](https://flow.org/en/docs/types/utilities/#toc-tuplemap)
- [x] [$ObjMapi](https://flow.org/en/docs/types/utilities/#toc-objmapi)
- [x] [$Shape](https://flow.org/en/docs/types/utilities/#toc-shape)
- [x] [$ReadOnlyArray](https://flow.org/en/docs/types/arrays/#toc-readonlyarray)
- [ ] $Compose
- [ ] $ComposeReverse
- [ ] $CharSet
- [x] [$NonMaybeType](https://flow.org/en/docs/types/utilities/#toc-nonmaybe)
- [x] [$Rest<A, B>](https://flow.org/en/docs/types/utilities/#toc-rest)

And maybe (they look like internal features though):
- [x] $Exports
- [x] ~$Pred~ ([experimental](https://github.com/facebook/flow/issues/2464#issuecomment-471115953))
- [x] ~$Refine~ ([experimental](https://github.com/facebook/flow/issues/2464#issuecomment-471115953))

Am I missing something?

[...]

PR #2983, opened 2016-12-08
[docs]: Documentation for Class<T>, $PropertyType<T, x>, $Exact<T>, and *

STRML:
Ref: #2464


website/docs/ref/utility-types.doc.js
```diff
@@ -10,10 +10,6 @@ next: react.html
 */
 
 /*
-# Utility Types
```
STRML: This was duplicated; the header is already present from the title on L5

STRML:
Is there an issue with this PR or something else that needs to be done to push it in?

PR #2987, opened 2016-12-09
Add type safe version of `process.nextTick` w/args

davidaurelio:
Adds a type safe version for calls to `process.nextTick` with a function with known parameter types, and passed arguments. Works for functions that take up to six arguments.

E.g.:

```js
function repeat(what: string, n: number): string {
  return Array(n + 1).join(what);
}

process.nextTick(repeat, 'foo', 3); // passes
process.nextTick((a, b) => {}, 1); // passes, unknown types
process.nextTick(repeat, 123, 2); // fails, `123` has the wrong type
```

The `...Array<void>` technique has been taken [from the flow docs](https://flowtype.org/docs/functions.html#too-many-arguments)

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1572,7 +1572,8 @@ declare class Process extends events$EventEmitter {
     heapTotal : number;
     heapUsed : number;
   };
-  nextTick(cb : Function) : void;
+  nextTick<A, B, C, D, E, F>(cb: (A, B, C, D, E, F) => mixed, A, B, C, D, E, F) : void;
+  nextTick(cb : Function, ...Array<void>) : void;
   pid : number;
   platform : string;
   release : {


davidaurelio:
Are you considering to add support for variadic generics? This would be so much better:

`nextTick<T, ...Ts>(cb: (T, ...Ts) => mixed, T, ...Ts): void`

and would make `...Array<void>` unnecessary for the simple version.

lib/node.js
```diff
@@ -1572,7 +1572,8 @@ declare class Process extends events$EventEmitter {
     heapTotal : number;
     heapUsed : number;
   };
-  nextTick(cb : Function) : void;
+  nextTick<A, B, C, D, E, F>(cb: (A, B, C, D, E, F) => mixed, A, B, C, D, E, F) : void;
+  nextTick(cb : Function, ...Array<void>) : void;
```
davidaurelio: ok, I will remove it. I was defensive and didn’t want to break anything.

vkurchatkin: I guess it will only work for arity > 6

vkurchatkin: but since you added `...Array<void>` it's not going to be matched anyway

davidaurelio: right, arity > 6 is a problem. If I leave the old generic version, then it will match all incorrect cases that fail for the generic version, and we don’t get any errors. 

Is there a way to resolve this?

samwgoldman: I'll fix in the import (should be `...Array<any>`)

PR #2994, opened 2016-12-10
Define MutationObserverInit v2

kuy:
`MutationObserverInit` is used as an argument named `options` in `MutationObserver#observe`.
The name was borrowed from here: https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#observe()

Previous PR: https://github.com/facebook/flow/pull/2863


samwgoldman:
Cool! Thanks for adding tests, too :)

PR #3000, opened 2016-12-12
Add emoji to server status messages

zertosh:
This change adds emoji to the various server status messages printed by flow. These messages are usually seen in square brackets with labels like "parsing" or "merging inference". The new behavior is configurable via the `emoji` option in `.flowconfig`, but will only be used in non-Windows platforms. It's disabled by default, so to opt-in you must add:

```
[options]
emoji=true
```

flow-emoji

Test plan (on a Mac):

* `make test`
* Started `flow` with an empty flowconfig and didn't see emoji.
* Started `flow` with a `.flowconfig` that had `emoji=true` under `[options]`, and saw emoji.
* Made a js change after the flow server had started, ran `flow` again, and saw the appropriate emoji given the config.
* Ran `flow | cat` with emoji enabled and didn't see any emoji.
* `make js` and `node -p 'require("./bin/flow.js").checkContent("-", "// @flow\n(1: string);")'` - this showed the correct error.

### DIFF
--- a/src/commands/commandConnect.ml
+++ b/src/commands/commandConnect.ml
@@ -60,24 +60,25 @@ let is_valid_line s =
 
 let msg_of_tail env tail_env =
   let line = Tail.last_line tail_env in
+  let use_emoji = env.emoji && Utils_js.can_emoji in
   if matches_re parsing_re line then
     Printf.sprintf "[%sparsing]"
-      (if env.emoji then (* Ghost *) "\xf0\x9f\x91\xbb  " else "")
+      (if use_emoji then (* Ghost *) "\xf0\x9f\x91\xbb  " else "")
   else if matches_re infer_re line then
     Printf.sprintf "[%slocal inference]"
-      (if env.emoji then (* Turtle *) "\xf0\x9f\x90\xa2  " else "")
+      (if use_emoji then (* Turtle *) "\xf0\x9f\x90\xa2  " else "")
   else if matches_re calc_deps_re line then
     Printf.sprintf "[%scalculating dependencies]"
-      (if env.emoji then (* Taco *) "\xf0\x9f\x8c\xae  " else "")
+      (if use_emoji then (* Taco *) "\xf0\x9f\x8c\xae  " else "")
   else if matches_re merging_re line then
     Printf.sprintf "[%smerging inference]"
-      (if env.emoji then (* Snail *) "\xf0\x9f\x90\x8c  " else "")
+      (if use_emoji then (* Snail *) "\xf0\x9f\x90\x8c  " else "")
   else if matches_re server_ready_re line then
     Printf.sprintf "[%sserver is ready]"
-      (if env.emoji then (* Unicorn Face *) "\xf0\x9f\xa6\x84  " else "")
+      (if use_emoji then (* Unicorn Face *) "\xf0\x9f\xa6\x84  " else "")
   else
     Printf.sprintf "[%sprocessing]"
-      (if env.emoji then (* Panda Face *) "\xf0\x9f\x90\xbc  " else "")
+      (if use_emoji then (* Panda Face *) "\xf0\x9f\x90\xbc  " else "")
 
 let arg name value arr = match value with
 | None -> arr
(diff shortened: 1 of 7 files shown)


samwgoldman:
Should we make emoji the default? ❤️ is yes, 🎉 if no.

src/commands/commandConnect.ml
```diff
@@ -57,20 +58,26 @@ let is_valid_line s =
   List.exists (fun re -> matches_re re s) re_list
 
 
-let msg_of_tail tail_env =
+let msg_of_tail env tail_env =
   let line = Tail.last_line tail_env in
   if matches_re parsing_re line then
-    "[parsing]"
+    Printf.sprintf "[%sparsing]"
+      (if env.emoji then (* Ghost *) "\xf0\x9f\x91\xbb  " else "")
```
adamjernst: book? magnifying glass? something indicative of parsing?

src/commands/commandConnect.ml
```diff
@@ -57,20 +58,26 @@ let is_valid_line s =
...
   else if matches_re infer_re line then
-    "[local inference]"
+    Printf.sprintf "[%slocal inference]"
+      (if env.emoji then (* Turtle *) "\xf0\x9f\x90\xa2  " else "")
```
adamjernst: "hmm" emoji?

src/commands/commandConnect.ml
```diff
@@ -57,20 +58,26 @@ let is_valid_line s =
...
   else if matches_re calc_deps_re line then
-    "[calculating dependencies]"
+    Printf.sprintf "[%scalculating dependencies]"
+      (if env.emoji then (* Taco *) "\xf0\x9f\x8c\xae  " else "")
```
adamjernst: something numeric for "calculating"?

src/commands/commandConnect.ml
```diff
@@ -57,20 +58,26 @@ let is_valid_line s =
...
   else if matches_re merging_re line then
-    "[merging inference]"
+    Printf.sprintf "[%smerging inference]"
+      (if env.emoji then (* Snail *) "\xf0\x9f\x90\x8c  " else "")
```
adamjernst: ⥷ or ⥺ ?

gabelevi:
Regarding the default - how widespread is emoji support? That would be my main concern with turning this on by default. @vkurchatkin - what concerns do you have?

vkurchatkin:
@gabelevi I think this change provides little to none benefit, but introduces a risk of breaking someone's setup

zertosh:
@Daniel15 what's emoji support like in Windows?

zertosh:
I've changed my mind and made the current behavior (no emoji) the default. So it's now opt-in. Also added a line in the docs about it.

Daniel15:
> what's emoji support like in Windows?

Not too good, unfortunately. They work in some configurations (for example, they work for me in Cmder), but they appear small and black-and-white. It seems like the new colourful Emoji can't be rendered in command-line apps yet. It seems like vanilla cmd.exe/conhost.exe don't support Emoji at all. See https://github.com/yarnpkg/yarn/issues/405

zertosh:
Thanks @Daniel15!

@samwgoldman after seeing yarnpkg/yarn/issues/405, I'll try to do something similar - I'll add a TTY check and exclude windows.

src/common/utils_js.ml
```diff
@@ -14,6 +14,12 @@ let spf = Printf.sprintf
 let print_endlinef fmt = Printf.ksprintf print_endline fmt
 let prerr_endlinef fmt = Printf.ksprintf prerr_endline fmt
 
+(* See https://github.com/yarnpkg/yarn/issues/405. *)
+let can_emoji =
```
zertosh: I wonder if it makes more sense to put this in `hack/utils/tty.ml`.

boris-marinov:
hm, this does not seem to work.

I specify
```
[options]
emoji=true
```
I get
``` 
Unsupported option specified! (emoji)
```

zertosh:
This was released in Flow 0.38.0 (https://github.com/facebook/flow/releases/tag/v0.38.0)

Issue #2450, opened 2016-09-13
Native String#match declaration incorrect


milesj:
The declaration here is incorrect: https://github.com/facebook/flow/blob/3a65d833a6a9b6ca02d006b1f41b2cc2030c7613/lib/core.js#L250

According to MDN, `match` returns an array of strings as well as an `index` and `input` property.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

alecperkins:
I just ran into this issue. It causes flow to fail at usage of the `match_instance.index` property:

```
return `${ source.substring(0, ext_match.index) }.css`;
                                         ^^^^^ property `index`. Property not found in
return `${ source.substring(0, ext_match.index) }.css`;
                               ^^^^^^^^^^^^^^^ Array
```

miklass:
Any progress? I've had to do some wacky "any" workarounds and the whole thing felt like a big hack.

milesj:
@miklass I was actually looking into this this week. However, it seems they don't review PRs often (I have another waiting).

rstuven:
In the meantime, I suggest you add the following to .flowconfig
```ini
[options]
suppress_comment=.*\\$FlowIssue
```

And then, just before the problematic line, add a comment like this:
```js
const rxa = ''.match(/./);
if (rxa) {
  // $FlowIssue: https://github.com/facebook/flow/issues/2450
  console.log(rxa.index);
}
```

milesj:
PR: https://github.com/facebook/flow/pull/3359

michaelhogg:
Since PR #3359 (9611080) was reverted in commit 3107fb3, I think this issue needs to be reopened.

Also see:

* PR #3007 – Fixes String#match and RegExp#exec declarations
* Duplicate issue #3554 – regex String .match flow type wrong

kesor:
@miklass @milesj I have a hack that does not involve sprinkling `any` everywhere, just correcting the signature locally using `flow-typed/node.js`, explained it here - https://stackoverflow.com/questions/48063016/how-to-extend-flow-bin-to-support-new-node-js-methods/48384325

PR #3007, opened 2016-12-13
Fixes String#match and RegExp#exec declarations. (#2450)

michelle:
The current declaration of String#match fails when trying to access the `index` and `input` properties of the return value.

Tests are failing because error messages are different due to new line numbers. I didn't want to update these by hand and I'm not sure how to update them otherwise (played around with `generate_exp.sh`).


michelle:
Hmm, it seems possible that intersection types don't play nicely with Arrays and Objects. After running tests, the only non-line-number-change diff is: 

```diff
-overload.js:7
-  7: var x1: number = "".match(0)[0];
-                      ^^^^^^^^^^^^^^ string. This type is incompatible with
-  7: var x1: number = "".match(0)[0];
-             ^^^^^^ number
-

-overload.js:8
-  8: var x2: number = "".match(/pattern/)[0];
-                      ^^^^^^^^^^^^^^^^^^^^^^ string. This type is incompatible with
-  8: var x2: number = "".match(/pattern/)[0];
-             ^^^^^^ number
-

-Found 12 errors
+Found 10 errors
```

avikchaudhuri:
I don't think this will work, see: https://flow.org/try/#0CYUwxgNghgTiAEA3W8AeAueBBGMoE8AeAZwBcYBLAOwHMA+eAMngG8AfeEVcqTMy2vDYBfANwAocQApUAOi49MVAK4BbAEYgYASlFA

Not sure if this is a bug that should be fixed separately in Flow, but this diff can't be merged in as is.

KevinGrandon:
> I don't think this will work, see: https://flow.org/try/#0CYUwxgNghgTiAEA3W8AeAueBBGMoE8AeAZwBcYBLAOwHMA+eAMngG8AfeEVcqTMy2vDYBfANwAocQApUAOi49MVAK4BbAEYgYASlFA
> 
> Not sure if this is a bug that should be fixed separately in Flow, but this diff can't be merged in as is.

@ avikchaudhuri I just went back and tried that try flow link, and it seems to be erroring as expected now. Perhaps something was fixed in flow in the last 1+ year that now allows this to work? Can we revisit this if possible now?

KevinGrandon:
Looks like the landscape has changed a bit. Using this as inspiration along with the new code in master I've opened: https://github.com/facebook/flow/pull/6790

Issue #1189, opened 2015-12-16
Allow `declare class Foo extends T` when `T` is a type


AgentME:
``` js
/* @flow */

import type Events from 'events';

declare class Foo extends Events.EventEmitter {
  bar(): number;
}

const f = new Foo();
f.on('bar', ()=>{});
```

In Flow 0.20, this gives an error:

```
test.js:5
  5: declare class Foo extends Events.EventEmitter {
                               ^^^^^^ Events. type referenced from value position
  3: import type Events from 'events';
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type Events


Found 1 error
```

Works fine with Flow 0.19.1.

avikchaudhuri:
Please try:

```
/* @flow */

import { EventEmitter } from 'events';

declare class Foo extends EventEmitter {
  bar(): number;
}

const f = new Foo();
f.on('bar', ()=>{});
```

As far as I can tell, the fact that it was working before was a bug.

AgentME:
`import type` has been very useful to me as often I don't want to import a module (so I can avoid its execution and side effects being taken immediately, or so I can avoid increasing the size of a javascript bundle) but the module is still involved in the type signatures. This worked correctly before. I can't see why it wouldn't be intended.

runn1ng:
This bug-slash-feature seems to extend *(no pun intended)* to definition file.

As in, in definition file, I cannot declare one type extending another type, because "type referenced in value position". Even when that makes zero sense in definition file.

See the immutable-js discussion that linked here

runn1ng:
As far as I can tell, you cannot now declare any kind of type hierarchy in the .js definition file, since it is always "type referenced in value position". It seems like a bug to me.

samwgoldman:
As @avikchaudhuri mentioned, it's an error to use a type as the argument to `extends`. This is part of JavaScript class syntax and isn't stripped away by Babel, so using a type there would cause a runtime error.

Let me know if there's something I missed here. Thanks!

AgentME:
I'm using `declare class`. The whole class declaration is stripped away by Babel, so there wouldn't be a runtime error.

samwgoldman:
Oh, I see. I've wanted this in the past, too. Thanks for clarifying!

joarwilk:
Any news on this? Fairly significant bug

calebmer:
It looks like this works now. Let me know if you still have issues 👍 

[...]

PR #3014, opened 2016-12-14
Mark defaultProps as covariant in React libs

### LINKED ISSUE #3008: Flow doesn't support getters for defaultProps in es6 classes
Given an example es6 style react class like:
```
/* @flow */

import React, { Component } from 'react';

type DefaultProps = {
    foo: number
}

class OnlyGetter extends React.Component {
    static get defaultProps(): DefaultProps {
        return {
            foo: 3
        }
    }
}
```

and `unsafe.enable_getters_and_setters=true` is specified under the `[options]` header of the `.flowconfig`

Flow incorrectly finds the following error, despite it being valid to react (in that react correctly makes use of the `defaultProps`):
```
test.js:5
  5: class ClassGetterTest extends Component {
           ^^^^^^^^^^^^^^^ ClassGetterTest. Covariant property `defaultProps` incompatible with invariant use in
  5: class ClassGetterTest extends Component {
                                   ^^^^^^^^^ statics of React$Component


Found 1 error
```

As far as I know, this is the only way to set `defaultProps` at runtime rather than load time.

From what I can tell, this is due to the fact that the flow type definition expecting `defaultProps` to be able to be set as well as gotten, even though afaik react really only ever gets the value (I could be wrong here?).

Adding a dummy setter seems to work around the problem and causes flow to succeed:
```
    static set defaultProps(props: DefaultProps) {
        // do nothing
    }
```

Unfortunately, it doesn't seem possible to solve this problem in the react definition by marking `defaultProps` as covariant (ie. `+defaultProps`) because this breaks valid situations like:
```
class LaterDefaultProps extends React.Component {
    static defaultProps: DefaultProps
}

LaterDefaultProps.defaultProps = {
    foo: 3
}
```

I can't find a way to express "invariant or covariant, but never contravariant" in the flow definition, but this may be able to be resolved if #1189 is by providing a union between a definition where `defaultProps` is invariant and a definition where `defaultProps` is covariant.

[There's a test case here](https://github.com/Novex/flow/tree/7226046335c5645e2da639b9054c5590ac94872c/tests/react_defaultProps) - making it pass should resolve this problem.

Novex:
This is so we don't have to include a dummy setter to satisfy Flow when using a getter to dynamically construct defaultProps

This resolves #3008, avoiding the `Covariant property `defaultProps` incompatible with invariant use in statics of React$Component` error and adds a test to ensure that future getters for defaultProps will typecheck properly.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -16,7 +16,7 @@
 declare class React$Component<DefaultProps, Props, State> {
   // fields
 
-  static defaultProps: $Abstract<DefaultProps>;
+  static +defaultProps: $Abstract<DefaultProps>;
   props: Props;
   state: $Abstract<State>;
 
--- a/tests/react_defaultProps/.flowconfig
+++ b/tests/react_defaultProps/.flowconfig
@@ -1,0 +1,2 @@
+[options]
+unsafe.enable_getters_and_setters=true
--- a/tests/react_defaultProps/react_defaultProps.exp
+++ b/tests/react_defaultProps/react_defaultProps.exp
@@ -1,0 +1,1 @@
+Found 0 errors
--- a/tests/react_defaultProps/test.js
+++ b/tests/react_defaultProps/test.js
@@ -1,0 +1,35 @@
+/* @flow */
+
+import React from 'react';
+
+type DefaultProps = {
+    foo: number
+}
+
+class OnlyGetter extends React.Component {
+    static get defaultProps(): DefaultProps {
+        return {
+            foo: 3
+        }
+    }
+}
+
+
+class GetterAndSetter extends React.Component {
+    static set defaultProps(props: DefaultProps) {
+        // do nothing
+    }
+
+    static get defaultProps(): DefaultProps {
+        return {
+            foo: 3
+        }
+    }
+}
+
+
+class StaticDefaultProps extends React.Component {
+    static defaultProps: DefaultProps = {
+        foo: 3
+    }
+}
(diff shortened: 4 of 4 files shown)


vkurchatkin:
The problem with this is that it's gonna break something like this:

```js
MyComponent.defaultProps = {};
```

Novex:
I *knew* I was missing an edge case!

I can't seem to find a way to mark a field as "covariant or invariant, but not contravariant" (ie. accepts anything but not only inputs).

Anyone wiser than me have any ideas here? Otherwise I'll close this and go back to the drawing board on #3008.

vkurchatkin:
Well, object with covariant property is a subtype of object with invariant property:

```js
declare function test(x: { +foo: string }): void;
declare var x: { foo: string };
test(x);
```

So, you can do this:

```js
class X {
 static +x: string;
}

class A extends X {
  static x: string;
}

A.x = '';
```

But it requires you redeclare inherited property as invariant every time.

Novex:
Thanks for the explanation - I think I understand what you're saying.

Unfortunately, the requirement to redeclare the inherited property as invariant when you want to set it after class definition is pretty much equivalent to the workaround of creating a dummy setter when you want to use a getter :(

I did stumble upon #1189, which if resolved would let us do something like define `React$Component` as a union type between a class with `defaultProps` as covariant and another class with `defaultProps` as invariant. But other than that I'm still stumped.

Novex:
I can't currently see a way to express this in the flow definition without breaking something, so I'm going to close this and come back to it (hopefully) in the future.

PR #3019, opened 2016-12-14
Declare HTMLFieldSetElement

josh:
Are there any docs or tests that need to be added or updated?

First flow PR 😆

##

CC: @tberman @mislav


mroch:
I expect the `tests/dom/` test will fail but travis is having issues again. run `./runtests.sh -r bin/flow dom` to re-record it, then check the diff to make sure it's just a bunch of line number changes.

josh:
@mroch ah, thanks for the tip. Committed changes to `tests/dom/dom.exp`.

josh:
Anything else you need me to fix with this patch?

Issue #2057, opened 2016-07-08
Use type from field of object type


jlongster:
If I have this object type:

``` js
type Person = {
  name: string
  age: number
}
```

There exists an undocumented `$Keys` feature for specifying a "key set" type (the key must be one of the keys of the object). It would be neat to also somehow enforce that a type must be a type of a specific field of the object, like

``` js
function get(person, field: $Keys<Person>) : $Value<Person(arg:field)> {
  return person[field];
}
```

I'm very new to flow, and I have no good proposal for the syntax. Writing this out, it is kind of difficult because you need to parameterize the field lookup with an argument value (given that it's a literal and flow can analyze it, just like it does with `$Keys`).

This came out of trying to type immutable.js. One can specify a type for `Record`:

``` js
export type Record<T: Object> = {
  get<A>(key: $Keys<T>): A;
  set<A>(key: $Keys<T>, value: A): Record<T>;
  setIn(keyPath: Array<any>, ...iterables: Array<any>): Record<T>;
  update<A>(key: $Keys<T>, updater: (value: A) => A): Record<T>;
  merge(values: $Shape<T>): Record<T>;
  mergeIn(keyPath: Array<any>, ...iterables: Array<any>): Record<T>;
  inspect(): string;
  toObject(): Object & T;
} & T;
```

And accesses like `foo.x` will be typed (because of the intersection type) but `foo.get("x")` will not be, even though Flow _does_ enforce that `x` is a valid key of the `foo` record.

jeffmo:
cc @bhosmer (per https://twitter.com/mroch/status/751496819327143936 )

jlongster:
Another issue with typing immutable.js (which is probably unrelated, but worth noting anyway) is `setIn` and friends. They take a path, `obj.setIn(["foo", "bar"], "value")` and it's impossible to type that, even though we have all the structure of `foo` and `bar` fields we need. I have no idea how that would even work, but it would be neat if it did.

vkurchatkin:
@jlongster here is an idea that might help: http://stackoverflow.com/questions/37952020/is-it-possible-to-wrap-a-flow-type-in-an-immutable-container. It's pretty verbose, but should work

vkurchatkin:
Syntax that makes sense could be like this:

```js
function get<K: $Keys<T>>(field: K) : $Value<T, K> {
  return person[field];
}
```

jlongster:
[...]

PR #2952, opened 2016-12-06
Allow all types as a key in `$PropertyType`.

pvolok:

PR #3023, opened 2016-12-15
Support computed properties for instances.

pvolok:
This is needed before #2952 can be merged.

Lookup rules:
1. If prop is a string literal (either computed or named), we look it up the prototype chain.
2. If class has an indexer, we add it to `try_ts_on_failure` of `LookupT` (only if there is no indexer queued yet, we distinguish by the reason `RObjectIndexer`).

We can't check indexer earlier because some classes like node's `Buffer` define indexer but has a parent class.

If this PR works, we can get rid of `$key` and `$value` props and use `dicttype` like in `ObjT`.

cc @samwgoldman

### DIFF
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -67,6 +67,7 @@ type reason_desc =
   | RObjectLit
   | RObjectType
   | RObjectClassName
+  | RObjectIndexer
   | RArray
   | RArrayLit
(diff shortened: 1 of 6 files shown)


STRML:
This unblocks #2952 and fixes #2057 after both are merged.

PR #3030, opened 2016-12-16
Add overload for setState to support function as argument

Alxswan:
Hi! I'm a junior developer and this is my first open source PR so please be gentle.

I have updated setState so that it can also take a function as per the React.js docs: https://facebook.github.io/react/docs/state-and-lifecycle.html

I'm not sure if I should add another test scenario to catch this change or if updating the test expectations as I have done is enough.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -22,8 +22,8 @@ declare class React$Component<DefaultProps, Props, State> {
 
   // action methods
 
-  // TODO: partialState could be a function as well
   setState(partialState: $Shape<State>, callback?: () => mixed): void;
+  setState(mapState: (prevState: State, props: Props ) => $Shape<State>, callback?: () => mixed): void;
 
   forceUpdate(callback?: () => void): void;
 
--- a/tests/new_react/new_react.exp
+++ b/tests/new_react/new_react.exp
@@ -45,6 +45,12 @@ classes.js:25
   9:   state: State;
               ^^^^^ object type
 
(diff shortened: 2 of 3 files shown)


CodyReichert:
I didn't even know that setState could do that. Nice job!

lib/react.js
```diff
@@ -22,8 +22,8 @@ declare class React$Component<DefaultProps, Props, State> {
 
   // action methods
 
-  // TODO: partialState could be a function as well
   setState(partialState: $Shape<State>, callback?: () => void): void;
+  setState(callback: (prevState: State, props: Props ) => $Shape<State>): void;
```
nmn:
```typescript
setState(callback: (prevState: State, props: Props ) => $Shape<State>, callback?: () => void): void;
```

apsavin:
@Alxswan Are you planning to finish this PR?

@nmn This PR actually fixes a bug. Can you add an appropriate label? 
If @Alxswan will not respond, I think I'll create another PR. 

By the way, why React (and some other libs) type definitions are here, not in https://github.com/flowtype/flow-typed ?

Alxswan:
@nmn @apsavin PR updated :)

goodmind:
This was fixed in https://github.com/facebook/flow/commit/95d6d11a88235a2106649aaa1afc134f0340c9da#diff-5ca8a047db3f6ee8d65a46bba4471236

Issue #1106, opened 2015-11-16
Switch from make to ocp-build?


mroch:
We are using ocp-build to support windows, so now we're maintaining both `build.ocp` and `Makefile`. Should we just switch to ocp-build? Having to install ocp-build makes it slightly harder to build flow, but if you have `opam` anyway, it's no big deal.

mroch:
cc @OCamlPro-Bozman @OCamlPro-Henry @gabelevi

OCamlPro-Henry:
@lefessan

lefessan:
I suppose that what you mean is using ocp-build just to build flow, and still use the Makefile for other targets. I think that such a use of ocp-build is safe. ocp-build is not as concise as Makefiles for generic targets, but it is quite simple and often much faster for OCaml targets. We can probably add some more comments in the `build.ocp` to make it easy to extend.

mroch:
yeah, I guess just to build flow, unless there's support for running generic commands but I don't see that in the user manual. I think we'd just keep the Makefile and have it run `ocp-build`, but that's better than the `ocamlbuild` stuff we're doing in the makefile now, which doesn't match how we build internally.

we use [Buck](https://buckbuild.com/) internally, which has files very similar in spirit to `build.ocp` (a bunch of targets with a list of sources and dependencies). I wonder if we could even generate the BUCK file from the build.ocp... that would solve a source of breakage that we run into when we add new files/directories and update the `BUCK` files but forget to update `Makefile` or `build.ocp`. I'll look more into what `build.ocp` supports to see if that's doable for us.

calebmer:
I’m going to close because of inactivity. Feel free to reopen if this is still a thing 😊

PR #3035, opened 2016-12-16
[RFC] Use ocp-build + watchman for oss build.

pvolok:
How about building with ocp-build by default? It's much faster and is the same build system as in travis.

Timings on my machine after touching `statement.ml`:
- `make fast-build` - 1.5s
- `make build-flow-debug` - 6.5s

Also, I think it would be convenient to have a watch command. So I added it via watchman.

If you are ok with this, I'll add docs and can remove building with ocamlbuild.

Related: #1106

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,4 +1,4 @@
 S hack/**
 S src/**
 
-B _build/**
+B _obuild/*
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,15 @@ OCP_BUILD_FILES=\
   ocp_build_flow.ocp\
   ocp_build_hack.ocp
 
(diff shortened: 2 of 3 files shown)


pvolok:
Added another idea: generate random build-id for debug builds.

My workflow looked like this:
1. Make changes
2. `make build-flow-debug`
3. `flow stop` in example directory
4. Test changes.

Would be cool to get rid of the step 3.

avikchaudhuri:
@mroch is working on related stuff.

nmote:
We have stopped using ocp-build altogether, so I think this PR is a non-starter at this point

PR #3037, opened 2016-12-16
[docs] Remove wrong spread operator in destructuring example.

atabel:


avikchaudhuri:
After the docs revamp, I can't find this example any more. cc @samwgoldman who might write some docs on destructuring.

PR #3046, opened 2016-12-19
Update handling of JSXText to trim and filter empty text

gabelevi:
The recent support for the `// @jsx Foo` pragma exposed that Flow wasn't properly desugaring JSX.  `JSXText` children are preprocessed and filtered before being passed to `React.createElement` or the custom function. Using babel as reference, this is what should happen

* Each line of the `JSXText` is trimmed of whitespace
* Empty lines are removed
* Tabs are replaces with spaces
* The lines are joined together with spaces.

Furthermore, empty `JSXText` children are thrown away and are not passed to the function. So

```
<Foo> {true} </Foo>
```

is correctly parsed as having 3 children (`" "`, `true`, and `" "`), but should desugar to `React.createElement(Foo, null, true)`;

### DIFF
--- a/hack/utils/string_utils.ml
+++ b/hack/utils/string_utils.ml
@@ -91,3 +91,44 @@ let rec is_not_lowercase str i j =
   if is_lowercase_char str.[i] then false
   else if i = j then true
   else is_not_lowercase str (i + 1) j
+
+(* String provides map and iter but not fold. It also is missing a char_list_of
+ * function. Oh well. You can use fold to simulate anything you need, I suppose
+ *)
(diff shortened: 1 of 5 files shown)


vjeux:
FYI, this would help with our js pretty printer for jsx nodes :)

gabelevi:
@vjeux - would it? This isn't a change to the parser.

pieterv:
Oh looks like it's standard for parsers to leave the white space in JSXText nodes (my bad i didn't look at the changes). cc @vjeux i think we will just need to copy this logic out unless we can share it with flow somehow.

PR #3071, opened 2016-12-22
Add Function.prototype.toString

disnet:
`toString` was missing from `Function.prototype`. This adds it.


mroch:
I think this is inherited from `Object.prototype`. was something complaining that it was missing?

disnet:
@mroch minimal test case:

```js
function F() {}
if (Function.prototype.toString !== F) {

}
```

errors with

```
3: if (Function.prototype.toString === F) {
                          ^ property `toString`. Property not found in
3: if (Function.prototype.toString === F) {
       ^ Function
```

mroch:
Hmm, this errors:

```
(Function.prototype.toString: empty);
```

([Try Flow](https://flowtype.org/try/#0BQMQrgdgxgLglgewgOgA4CcEywT1QU2WwGUZ04IBzALgAJ8BbVGHASgG4AoIA))

so I think in some cases we do know it's inherited. Something must be wrong with using inherited properties in a conditional, which goes through a slightly different codepath.

thejameskyle:
@mroch also pointed out that this no longer errors in master. Gonna close this.

mroch:
https://github.com/facebook/flow/commit/4b887d7972e5a7c597f4c337e8540fac6661eaed fixed this unintentionally.

Turns out that it was correctly falling back to looking at `Object.prototype`, but erroring because the `ObjProtoT ~> HasPropT` rule always said the prop was not found instead of allowing Object.prototype properties.

PR #3075, opened 2016-12-23
Fix syntax error in docs

baopham:


thejameskyle:
This is actually valid syntax for object types in Flow:

```js
type Foo = {
  prop1: 'value1';
  prop2: 'value2';
};

class Bar {
  prop: {
    prop1: 'value1';
    prop2: 'value2';
  };
}
```

[[Try](https://flowtype.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgBiccYAvGAN6phg4BOcOAjAFxgDkAbgIYwCueZhwDcNOoxwAmdtz6Cpo1AF8xqAMYweAZ21gAQj3pVxDJu2q1aZlrN4ChSqxKYzO9hU7CqVQA)]

mroch:
even though it's valid, I think it wouldn't hurt to be consistent.

https://flowtype.org/docs/syntax.html#object-types uses commas, https://flowtype.org/docs/objects.html uses semicolons. Facebook's internal style guide uses commas.

@baopham: if you want to make them all consistent, I think that'd be worth merging. I vote for making them all commas.

PR #3077, opened 2016-12-23
Improve lib def for WebSocket 'close', 'message' event handlers

philikon:
Handlers registered via `addEventListener('close' | 'message', ...)` now have their `event` argument spec'ed the same way that `onclose` and `onmessage` handlers have.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -368,6 +368,11 @@ declare class CloseEvent extends Event {
     wasClean: boolean;
 }
 
+type MessageEventHandler = (event: MessageEvent) => mixed;
+type CloseEventHandler = (event: CloseEvent) => mixed;
+type MessageEventListener = {handleEvent: MessageEventHandler} | MessageEventHandler;
+type CloseEventListener = {handleEvent: CloseEventHandler} | CloseEventHandler;
+
 declare class WebSocket extends EventTarget {
     static CONNECTING: 0;
     static OPEN: 1;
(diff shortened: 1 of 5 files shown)


philikon:
I get the `union_new` failure on master too, so I'm assuming this isn't due to my change. In fact, it seems like 56a82510efa774a9b079779fcfa4aa76abb6357d or 3e715557881eec7b8ce00edb017f63c9b1dc8b53 introduced this since that push is red and the revisions before it are green.

lib/bom.js
```diff
@@ -368,6 +368,11 @@ declare class CloseEvent extends Event {
     wasClean: boolean;
 }
 
+type MessageEventHandler = (event: MessageEvent) => mixed;
+type CloseEventHandler = (event: CloseEvent) => mixed;
+type MessageEventListener = {handleEvent: MessageEventHandler} | MessageEventHandler;
```
thejameskyle: I never knew about `addEventListener({ handleEvent(e) { ... } });`, we should probably add support for that everywhere.

gabelevi: I worry that overriding `addEventListener` is going to hit the same perf regression that https://github.com/facebook/flow/pull/2680 hit.

philikon: @gabelevi that's a bummer. I can give this a spin re: performance. Did any optimizations ever happen since #2680?

philikon: Indeed, running a random test (the `fetch` test) from the Flow test suite now takes >14 seconds rather than ~0.8 seconds. That's very unfortunate.

nmote:
I just tried out #2680 and unfortunately it hit the same issue that it originally did. I'm going to assume that this PR is similarly blocked

SamChou19815:

Issue #3108, opened 2016-12-30
String.prototype.split arity


EnoahNetzach:
`String.prototype.split` can have 0 arity ([**@see**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split)).

Flow throws an error, instead:
```
105:   return contents
              ^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
261:     split(separator: string | RegExp, limit?: number): Array<string>;
                          ^^^^^^^^^^^^^^^ union: string | RegExp. See lib: /tmp/flow/flowlib_413f687/core.js:261
```

The first argument should be `separator?: string | RegExp` in [core.js#L306](https://github.com/facebook/flow/blob/6d6c350ed492afcdfa161a8e672f836d5754b533/lib/core.js#L306).

mrkev:
This was solved

PR #3110, opened 2016-12-30
Remove the requirement of first argument in `String.prototype.split`

EnoahNetzach:
Solves #3108


jeffmo:
Thanks!

jeffmo:
Oh, actually it looks like the autocomplete test's `.exp` file just needs to be re-recorded.

If you just run the tests locally and then `cp tests/autocomplete/autocomplete.out tests/autocomplete/autocomplete.exp` that will re-record things for you, and then we can take this

EnoahNetzach:
@jeffmo I was just going to ask how to do it ;)

PR #3112, opened 2016-12-31
Add ParentNode and ChildNode pure interfaces to the appropriate types

tberman:
Some ParentNode methods were missing, and the ChildNode methods were missing from all 3 ChildNode types. whatwg spec docs for ChildNode here: https://dom.spec.whatwg.org/#interface-childnode


tberman:
Test changes look more than I would expect, but are outside of anything I touched and look similar to what I get off a test run on `master` so not quite sure what that means.

jeffmo:
Skimming the test failures -- these look normal/acceptable, they just need to be recorded. It seems most of them are just line number changes (or slight variations of the same previous message).

The way our older tests work is by comparing output, so if content in test files move around then line numbers change and the "expected" output changes and needs to be re-recorded.

If you can re-record the .exp file for `tests/dom/` so we can take this?

tberman:
@jeffmo how do you do that?

jeffmo:
Oh sorry:

If you run the tests then you'll be left with a `tests/dom/dom.out` file after the test fails. Just do `cp tests/dom/dom.out tests/dom/dom.exp` and then re-run tests to be sure things pass.

tberman:
@jeffmo Thanks! Seems to be working great now!

PR #3113, opened 2016-12-31
Add "dumpTypes" to flow.js

zertosh:
I'd like to add another tab to Try Flow with this information.

Test plan:

```
$ node -p 'JSON.stringify(require("./bin/flow.js").dumpTypes("-", `
// @flow
123;
class K {};
type T = {||};
const t: T = {}
`), null, 2)'

[
  {
    "type": "number",
    "reasons": [],
    "loc": {
      "source": "-",
      "type": "SourceFile",
      "start": {"line": 3, "column": 1, "offset": 10},
      "end": {"line": 3, "column": 3, "offset": 13}
    },
    "path": "-",
    "line": 3,
    "endline": 3,
    "start": 1,
    "end": 3
  },
  {
    "type": "[class: K]",
    "reasons": [],
    "loc": {
      "source": "-",
      "type": "SourceFile",
      "start": {"line": 4, "column": 1, "offset": 15},
      "end": {"line": 4, "column": 10, "offset": 25}
    },
    "path": "-",
    "line": 4,
    "endline": 4,
    "start": 1,
    "end": 10
  },
  {
    "type": "[type: {||}]",
    "reasons": [],
    "loc": {
      "source": "-",
      "type": "SourceFile",
      "start": {"line": 5, "column": 1, "offset": 27},
      "end": {"line": 5, "column": 14, "offset": 41}
    },
    "path": "-",
    "line": 5,
    "endline": 5,
    "start": 1,
    "end": 14
  },
  {
    "type": "{||}",
    "reasons": [
      {
        "desc": "T",
        "loc": {
          "source": "-",
          "type": "SourceFile",
          "start": {"line": 6, "column": 10, "offset": 51},
          "end": {"line": 6, "column": 10, "offset": 52}
        },
        "path": "-",
        "line": 6,
        "endline": 6,
        "start": 10,
        "end": 10
      }
    ],
    "loc": {
      "source": "-",
      "type": "SourceFile",
      "start": {"line": 6, "column": 7, "offset": 48},
      "end": {"line": 6, "column": 10, "offset": 52}
    },
    "path": "-",
    "line": 6,
    "endline": 6,
    "start": 7,
    "end": 10
  },
  {
    "type": "{}",
    "reasons": [],
    "loc": {
      "source": "-",
      "type": "SourceFile",
      "start": {"line": 6, "column": 14, "offset": 55},
      "end": {"line": 6, "column": 15, "offset": 57}
    },
    "path": "-",
    "line": 6,
    "endline": 6,
    "start": 14,
    "end": 15
  }
]
```

### DIFF
--- a/src/commands/dumpTypesCommand.ml
+++ b/src/commands/dumpTypesCommand.ml
@@ -63,7 +63,7 @@ let types_to_json types ~strip_root =
     in
     JSON_Object json_assoc
   ) in
-  (JSON_Array types_json)
+  JSON_Array types_json
 
 let handle_response types ~json ~pretty ~strip_root =
   if json
--- a/src/commands/dumpTypesCommand.ml
+++ b/src/commands/dumpTypesCommand.ml
@@ -40,32 +40,36 @@ let spec = {
   )
 }
 
+let types_to_json types ~strip_root =
+  let open Hh_json in
+  let open Reason in
+  let types_json = types |> List.map (fun (loc, _ctor, str, raw_t, reasons) ->
+    let json_assoc = (
+      ("type", JSON_String str) ::
+      ("reasons", JSON_Array (List.map (fun r ->
+        let r_loc = loc_of_reason r in
+        JSON_Object (
+          ("desc", JSON_String (string_of_desc (desc_of_reason r))) ::
+          ("loc", json_of_loc ~strip_root r_loc) ::
+          (Errors.deprecated_json_props_of_loc ~strip_root r_loc)
+        )
+      ) reasons)) ::
+      ("loc", json_of_loc ~strip_root loc) ::
+      (Errors.deprecated_json_props_of_loc ~strip_root loc)
+    ) in
+    let json_assoc = match raw_t with
+      | None -> json_assoc
+      | Some raw_t -> ("raw_type", JSON_String raw_t) :: json_assoc
+    in
+    JSON_Object json_assoc
+  ) in
+  (JSON_Array types_json)
+
 let handle_response types ~json ~pretty ~strip_root =
   if json
   then (
-    let open Hh_json in
-    let open Reason in
-    let types_json = types |> List.map (fun (loc, _ctor, str, raw_t, reasons) ->
-      let json_assoc = (
-        ("type", JSON_String str) ::
-        ("reasons", JSON_Array (List.map (fun r ->
-          let r_loc = loc_of_reason r in
-          JSON_Object (
-            ("desc", JSON_String (string_of_desc (desc_of_reason r))) ::
-            ("loc", json_of_loc ~strip_root r_loc) ::
-            (Errors.deprecated_json_props_of_loc ~strip_root r_loc)
-          )
-        ) reasons)) ::
-        ("loc", json_of_loc ~strip_root loc) ::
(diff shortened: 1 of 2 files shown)


zertosh:
^ I don't know OCaml, so this code prob sucks

jeffmo:
Love this idea!

We should probably factor out the actual dump-types command, though, and just use that logic directly. Here's where the command code sits:

https://github.com/facebook/flow/blob/master/src/commands/dumpTypesCommand.ml#L96

It looks like that just calls off to the server, though, so here's the hook in the server that runs that you probably want to re-use:

https://github.com/facebook/flow/blob/master/src/server/server.ml#L205

zertosh:
@jeffmo I'm not sure exactly what to factor out. The server already [calls](https://github.com/facebook/flow/blob/c2c3577/src/server/server.ml#L224) `Query_types.dump_types` like I'm doing here.

mroch:
we need to send this through a JSON printer instead of dumping js_of_ocaml's representation of ocaml's internal data structures

zertosh:
@mroch so instead of `Reason.string_of_loc` do this https://github.com/facebook/flow/blob/e21cb94a8bde71ea43ba529f33ddec3cd14cd870/src/commands/dumpTypesCommand.ml#L48-L68 ?

zertosh:
@jeffmo I factored out the "types to json" formatter in `DumpTypesCommand` and using that now.
@mroch I'm printing out the same thing that `flow dump-types` prints now.

src/commands/dumpTypesCommand.ml
```diff
@@ -40,32 +40,36 @@ let spec = {
...
+      ("loc", json_of_loc ~strip_root loc) ::
+      (Errors.deprecated_json_props_of_loc ~strip_root loc)
+    ) in
+    let json_assoc = match raw_t with
+      | None -> json_assoc
+      | Some raw_t -> ("raw_type", JSON_String raw_t) :: json_assoc
+    in
+    JSON_Object json_assoc
+  ) in
+  (JSON_Array types_json)
```
zertosh: n00b question: Do these parens matter?

jeffmo: No, they don't. Some people find parens like this help a little with clarity, but to each their own -- you decide

zertosh: 👍  don't like them. removed.

src/flow_dot_js.ml
```diff
@@ -224,6 +224,29 @@ let infer_type filename content line col =
...
+      in
+      let printer = Type_printer.string_of_t in
+      let raw_printer _ _ = None in
+      let types = Query_types.dump_types printer raw_printer cx in
+
+      let strip_root = None in
+      let types_json = DumpTypesCommand.types_to_json types ~strip_root in
+
+      js_of_json types_json
+    | _, _ -> failwith "parse error"
```
jeffmo: Any way to be more specific about the parse error here?

zertosh: That's the same messages used by `infer_type` when `parse_content` fails: https://github.com/zertosh/flow/blob/666003571ef634f186056891ddaf2287741906cd/src/flow_dot_js.ml#L225

PR #3117, opened 2017-01-01
lib: require.main property added

happylynx:
Summary:
Added property `main` of object (function) `require`. It allows to
access the main module. [Documentation][1].

[1]: https://nodejs.org/api/modules.html#modules_accessing_the_main_module


jeffmo:
Thanks!

PR #3118, opened 2017-01-02
Add statusMessage to http.IncomingRequest

robarnold:
This has existed for some time, per https://nodejs.org/api/http.html#http_message_statusmessage


gabelevi:
Looks good to me! Thanks for updating the lib!

PR #3127, opened 2017-01-03
Make a few DOM properties nullable and IDL classes iterable

loganfsmyth:
These DOM properties are definitely nullable, though in the average case they probably exist. I assume it's best for the type definitions to cover all the possibilities.

For the iterability, Chome added these in a few releases ago and the next release of core-js should include polyfill logic for them. More info in https://github.com/zloirock/core-js/pull/249


gabelevi:
I find the argument made in the OP of https://github.com/zloirock/core-js/pull/249 convincing. Definitely cool marking these as `Iterable`.

Other changes look good to me too! I'll get this merged! Thanks for the contribution!

OlegLustenko:
How to use this syntax now ?

```javascript
const body: HTMLElement = document.body;
// [flow] [flow] null (This type is incompatible with HTMLElement)
```

```javascript
const body: HTMLElement | null = document.body;
body.addEventListener('mousemove', event, false);
// [flow] [flow] call of method `addEventListener` (Method cannot be called on possibly null value)
```

Should I'm run this way now ?
```javascript
const body: HTMLElement | null = document.body;
```

From the documentation:
```javascript
/* @flow */
var o: ?string = null;
if (o == null) {
  o = 'hello';
}
print(o.length);
```

Do I need to perform a null check?
What If I'm sure my code will execute when the body isn't null?

loganfsmyth:
You'd have to perform a null check, and either handle that case specifically, or throw so Flow knows that the value is guaranteed to exist after that point.

PR #3150, opened 2017-01-07
Adding WebUSB API

runn1ng:
I have added WebUSB API definitions, based on the [draft](https://wicg.github.io/webusb/).

I am not sure if I even *want* this merged right now - the draft is still a draft, and it's supported only on Chrome/ium and only either via "trial mode" or via special flag in Chrome/ium settings.

But I am still making this PR - either for merging, or for tracking progress; as you wish. 

The definition is rewritten directly from the draft - partly by hand, partly by [webidl tools](https://www.npmjs.com/package/webidl-tools).

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -113,6 +113,7 @@ declare class Navigator mixins NavigatorCommon {
     vibrate?: (pattern: number|number[]) => bool;
     mozVibrate?: (pattern: number|number[]) => bool;
     webkitVibrate?: (pattern: number|number[]) => bool;
+    usb: USB;
 }
 
 declare var navigator: Navigator;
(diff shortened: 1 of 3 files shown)


runn1ng:
I cannot fix the travis issues now, because for some reason flow is not building on my PC, so I cannot repeat the tests (for fixing all those line shifting issues).

gabelevi:
Yeah, it seems a little early to merge. Putting stuff in the libdefs isn't exactly free at the moment, since every time Flow starts up it has to recheck the libdefs. We do plan on moving as much of the libdefs to flow-typed at some point.

Is that cool with you?

runn1ng:
I am not really sure *how*, technically, add this to a separate file, when this defines `navigator.usb` and `navigator` is defined in bom.js.

Also, I am not sure under which name this should be added to flow-typed :) you don't have lidefs there right now, right?

I am completely ok with you not merging this, the standard is draft after all.

gabelevi:
We don't currently have any way to extend an already defined class, other than redeclaring it. Ideally you'd be able to add your own libdef saying "Navigator also has this property" without completely redefining Navigator. It's been a todo for awhile :/

runn1ng:
Hm, it's a pity that there is no way to have this in flow right now :/ ok anyway

PR #3151, opened 2017-01-08
Socket.send has two optional arguments in the middle

sdwilsh:
This is likely the least-bad way to support this.

https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -545,10 +545,10 @@ type dgram$Socket = {
   ref(): void;
   send(
     buf: Buffer,
-    offset: number,
-    length: number,
-    port: number,
-    address: string,
+    offsetOrPprt: number,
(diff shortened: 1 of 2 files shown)


gabelevi:
@sdwilsh - why not use overloading? Given the docs specify

```
socket.send(msg, [offset, length,] port, address[, callback])
```

we could just write

```
send(
  msg: Buffer, 
  port: number, 
  address: string, 
  callback?: (err: ?Error, bytes: any) => mixed,
): void
send(
  msg: Buffer, 
  offset: number, 
  length: number, 
  port: number, 
  address: string, 
  callback?: (err: ?Error, bytes: any) => mixed,
): void
```

PR #3170, opened 2017-01-10
fix(bom): make readyStates static props of XMLHttpRequest

Delapouite:
Hi

I detected this issue with this false positive in my code:

```
  9:     if (this.readyState !== XMLHttpRequest.DONE) return
                                                ^^^^ property `DONE`. Property not found in
  9:     if (this.readyState !== XMLHttpRequest.DONE) return
                                 ^^^^^^^^^^^^^^ statics of XMLHttpRequest

```

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -493,14 +493,14 @@ declare class XMLHttpRequest extends EventTarget {
     getResponseHeader(header: string): string;
     msCachingEnabled(): boolean;
     overrideMimeType(mime: string): void;
-    LOADING: number;
-    DONE: number;
-    UNSENT: number;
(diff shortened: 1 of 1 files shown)


samwgoldman:
This PR doesn't do what it intends to do, which is add statics properties to the XMLHttpRequest class. To do that, you should add `static LOADING: number` and the like to the declare class body.

But upon further research, it looks like all the type defs for the XMLHttpRequest spec are pretty broken. (For example, I have no idea where the `statics` property came from, but there isn't even a `create` method in the spec, nor a `requestBody` property). I think we need to redo the entire thing, and that should happen in a new PR.

PR #3171, opened 2017-01-10
Define defaultProps directly in docs

ChaseBro:
Previously the doc recommended giving `defaultProps` a type and then doing the definition outside of the class definition. This is cleaner and more understandable. The type of `defaultProps` doesn't really matter since it should be static and should already be constrained by the type of `Props`.


samwgoldman:
Thanks for sending this PR. Sorry I let it sit unreviewed for so long. Since you shared this, we revamped the Web site. It turns out the [new docs](https://flow.org/en/docs/frameworks/react/) use this method already :)

PR #3184, opened 2017-01-13
[docs] Update quick reference for array syntax including shorthand

rmlewisuk:
**Summary**

Updates the [quick reference for arrays](https://flowtype.org/docs/quick-reference.html#arrays) to show the standard `Array<T>` syntax first in the example, and mention the shorthand syntax `T[]` explicitly.

Currently, the quick ref mentions the standard syntax but uses shorthand in the examples and only shows the standard syntax as an alternative. This struck me as something which could be confusing for anyone using the quick reference.

### DIFF
--- a/website/docs/ref/quickref.doc.js
+++ b/website/docs/ref/quickref.doc.js
@@ -53,16 +53,16 @@ next: syntax.html
   This type describes JavaScript array objects and the type of the elements
   contained within the array.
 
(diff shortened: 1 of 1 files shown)


samwgoldman:
Thanks for opening this PR. I'm sorry that I let this sit unreviewed for so long. We've since revamped the Web site and the [new docs on Arrays](https://flow.org/en/docs/types/arrays/) cover the material added in this patch. Thanks again :)

PR #3194, opened 2017-01-15
Getting Started page - Remove $DocIssue from line 36

newswim:
In the first code example, we haven't added any flow annotations, but there is an error being indicated. It seems like `$DocIssue` is responsible.

Note: I have not gone through the build steps with OCaml in order to test this..

---

flow_error


samwgoldman:
Thanks for opening this PR and I apologize for letting it sit unreviewed. Since we revamped the Web site, this issue no longer exists. Thanks again :)

Issue #2967, opened 2016-12-07
Autocomplete doesn't work with parametric types


gcanti:
Repro

```js
type Person<A> = { name: A };

const p: Person<string> = { name: 'Giulio' }

p. // <= no autocomplete
```

UnsungHero97:
Any update on when this will be fixed?

PR #3209, opened 2017-01-18
Fix autocomplete for generic type aliases

vkurchatkin:
Fixes https://github.com/facebook/flow/issues/2967

### DIFF
--- a/newtests/autocomplete/generic_alias.js
+++ b/newtests/autocomplete/generic_alias.js
@@ -1,0 +1,7 @@
+/**
+ *
+ */
+type Person<A> = { name: A };
+declare var obj: Person<string>;
+
+obj.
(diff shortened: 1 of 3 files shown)


UnsungHero97:
Any update on when this will be merged?

maoueh:
Sorry to disturb, could we get some attraction for merge here. It seems a fairly easy fix and it make it harder to play with RxJs for example when defining specific type aliases.

TrySound:
Ping guys

mvitousek:
It looks like sometime in the last year we fixed this issue by another approach, or at least the test you suggested is passing now (modulo line numbers). I'll go ahead and merge just the test from this PR, sorry we didn't get to it sooner. If this problem is actually persisting in a way that isn't captured by this test case, please let me know and I'll take another look.

PR #3216, opened 2017-01-19
Update Copyright in Footer

peduarte:
Updated Copyright from 2016 to 2017.

### DIFF
--- a/website/_includes/footer.html
+++ b/website/_includes/footer.html
@@ -1,6 +1,6 @@
 {% unless page.custom_footer %}
 <footer><div class='width'>
-    &copy; Copyright 2014 - 2016, Facebook Inc.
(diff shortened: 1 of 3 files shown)


davecardwell:
Could use ```{{ site.time | date: '%Y' }}``` instead of hard-coding the current year?

peduarte:
Agree that having it dynamic is better 👍 
Do you know if `{{ site.time | date: '%Y' }}` will work for sure or will I have to run the project locally to double check?

davecardwell:
@peduarte Yes, I just double-checked and that works fine. I found these files where it will need updating:
- website/_includes/footer.html
- website/_layouts/try.html
- website/index.doc.js

PR #3225, opened 2017-01-19
Buffer.from(): avoid clash between Iterable<number> and string case for first arg

philikon:
In some cases, Flow seems to eagerly choose the Iterable<number> case for the first argument to Buffer.from(), reporting the following error when Buffer.from(aString, encoding) is used:

```
inconsistent use of library definitions
274:     @@iterator(): Iterator<string>;
                                ^^^^^^ string. This type is incompatible with. See lib: /private/tmp/flow/flowlib_253af11f/core.js:274
111:   static from(value: Iterable<number>): this;
                                   ^^^^^^ number. See lib: /private/tmp/flow/flowlib_253af11f/node.js:111
```

Reordering the lib def for Buffer.from() fixes that problem for now, though the over-eagerness in choosing the Iterable<number> case should obviously be investigated too.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -108,10 +108,10 @@ declare class Buffer extends Uint8Array {
   static compare(buf1: Buffer, buf2: Buffer): number;
   static concat(list: Array<Buffer>, totalLength?: number): Buffer;
 
-  static from(value: Iterable<number>): this;
   static from(value: Buffer): Buffer;
   static from(value: string, encoding?: buffer$Encoding): Buffer;
   static from(value: ArrayBuffer, byteOffset?: number, length?: number): Buffer;
+  static from(value: Iterable<number>): this;
   static isBuffer(obj: any): boolean;
   static isEncoding(encoding: string): boolean;
 }


philikon:
Thanks to @gabelevi for helping diagnose this!

It'd be super helpful if we could uplift this to 0.38.x and release a new patch version.

gabelevi:
Ok, this looks good to me! I'm having a really hard time finding a repro, but still digging. I suppose a workaround in Flow v0.38.0 for anyone who hits this is to explicitly cast to string

    Buffer.from ((expression: string))

That should help Flow deal with whatever ambiguity is triggering this.

gabelevi:
Hey @philikon, since it's somewhat tricky to hit this bug and since I think it can be worked around somewhat easily with a string typecast, I'm going to avoid pushing 0.38.1 just for this. If I push a patch for another reason I'll piggyback this. Otherwise, this will be in 0.39.0 which is scheduled for February 1st.

philikon:
Sounds good, @gabelevi. Thanks!

philikon:
> `Buffer.from ((expression: string))`

Sadly that is not a workaround. Keep in mind that the trace shows that Flow totally understands that the first arg to `Buffer.from()` is a string. It's just simply not considering any other traces. Also, just for completeness sake, the expression it dislikes is `Buffer.from(aString, 'hex')`. Not sure if the second argument should matter here -- I feel like it should be an additional hint that the `from(value: string, encoding?: buffer$Encoding)` signature is desired...

gabelevi:
Hmm, I wish I had a repro. My best guess about the trouble with the intersection is that Flow realizes too late that the value passed to `Buffer.from` is a `string`. `Buffer.from ((expression: string))` should make that info available earlier and avoid the problem. If it doesn't work, maybe I'm wrong about the cause 😟

philikon:
Yeah, `Buffer.from((expression: string))` sadly doesn't fix the problem. :(

PR #3239, opened 2017-01-22
Fixed incorrect object destructuring

zerkms:
The code as it is implemented now (considering the stage-3 `Rest/Spread Properties` proposal) would type `o` as `{o: {z: boolean} }`

I think it's just a typo and it should be just `var {x, y, o}` in the destructuring declaration. Otherwise the typing + the following explanation should be fixed.


avikchaudhuri:
After the docs revamp, I don't know where this page is anymore. cc @samwgoldman who might write new destructuring docs.

PR #3244, opened 2017-01-22
Add a documentation note on troubleshooting slow inititalization

JeffRMoore:
I ran into this issue, where I needed to remove the cache after a scope decrease.  A friend who tried Flow for the first time this week complained about Flow hanging before he figured out to exclude
node_modules.  I also noticed a few "Hang" tickets when I was searching for how to clear the cache.  Hope this helps mitigate that.


JeffRMoore:
Oh.  

So I was experimenting with the all=true because my friend complained about the comment flag in his files (I was trying to convince him to use Flow). I created this Doc PR when I discovered that the cache clear helped on a scope down grade.

Doh.

You can't really ignore node_modules, can you?  Obviously, this PR should NOT be merged. 

Is it useful to mention that clearing the cache can help recover from a failed attempt to use --all?  If so, I'm happy to update the docs somewhere for that.

Otherwise, I'm just closing this.  Sorry for the noise.

PR #3254, opened 2017-01-23
Update ReactDOM.findDOMNode signature to match React

ianobermiller:
Update the `ReactDOM.findDOMNode` signature to match the type definitions React itself has for the function: https://github.com/facebook/react/blob/3438d503db2aba2b0e036912fa708cd3322310df/src/renderers/dom/shared/findDOMNode.js

Primarily allows `findDOMNode` to accept null, which is very common when working with refs. Also updates the return type of `findDOMNode` to be much more strict than `any`. This has the most potential to cause new (legitimate) errors.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -201,8 +201,8 @@ declare module react {
   ): React$Component<$DefaultPropsOf<Config>, $PropsOf<Config>, any>;
 
   declare function findDOMNode(
-    object: React$Component<any, any, any> | HTMLElement
-  ): any;
+    componentOrElement : Element | ?React$Component<any, any, any>,
+  ): null | Element | Text;
 
   declare function render<Config>(
     element: React$Element<Config>,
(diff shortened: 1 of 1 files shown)


thejameskyle:
Should it be this instead to be a bit more precise? I assume passing `null` or `void` to `findDOMNode` always results in `null`?

```js
declare function findDOMNode(componentOrElement: null | void): null;
declare function findDOMNode(
  componentOrElement: Element | React$Component<any, any, any>,
): null | Element | Text;
```

ianobermiller:
@thejameskyle seems reasonable, but will it really make a difference in practice?

skovhus:
As the function now can return Text, how would you recommend casting it to an Element? You would need this if I wanted to modify some Element properties than are not on Text. 

@ianobermiller

ianobermiller:
Since you need to null check anyway, do: `if (result instanceof
HTMLElement) {`. FWIW I ran into the same thing in my code ;)

Issue #2470, opened 2016-09-14
Doesn't check the type of an exception


langco:
This function will return a number, without flow complaining about it.

``` javascript
function foo(x: ? number): string {
try {
        throw x
    } catch (e) {
        return e // returns x, which is a number
    }
}
```

vkurchatkin:
I agree, `e` should be `mixed`  instead of `any`

vkurchatkin:
I think also annotating error should be possible:

```js
try {
 } catch (e: Error) { // mixed by default, but arbitrary annotation is allowed
 }
```

zertosh:
In flow 0.36.0, this is a syntax error:

```js
try {
} catch (e: Error) { // (e: Error) is a Syntax Error
}
```

vkurchatkin:
@zertosh just to clarify, my proposal is:

```js
try {
 } catch (e) {
   // e is mixed, safe behaviour
 }
```

```js
try {
 } catch (e: any) {
   // e is any, works exactly as it works today
 }
```

```js
try {
 } catch (e: Error) {
   // e is Error, works 99.99% percent of the time
 }
```

mroch:
`e: Error` is unsound because we don't check the `throw`s to make sure they're throwing `Error`.

allowing an annotation seems reasonable to me, but we'll have to flow `mixed` into it which would mean the only things you could actually annotate with are `mixed` or `any`. that's good because the default case will be `mixed`, but you can downcast to `any`.

you'd end up with errors like

```
} catch (e: Error) {
         ^ mixed. Some potential exception is not compatible with
} catch (e: Error) {
            ^^^^^ Error
```

vkurchatkin:
@mroch everything other then mixed is unsound, but it seems to be a good compromise to allow any annotation (i.e. not flowing mixed anything into it if it's present)

samwgoldman:
I think we should stick with `any` here and continue to forbid annotations unless:

1) We add support for effects, model exceptions as effects, add annotation support to cover libdefs, and actually model the data flow for exceptions

2) Find some way to indicate "untrusted" types and mark annotations here as untrusted. Any program location into which an `any` type flows would be untrusted, and this would be one such location.

3) Do what @mroch suggests and actually change from `any` to `mixed`. This wouldn't allow an explicit `Error` annotation, but you could add an explicit `any` annotation and cast to `Error`, or do a type test on the `mixed`.

vkurchatkin:
@samwgoldman 

[...]

PR #3259, opened 2017-01-24
Support annotations in catch clause

vkurchatkin:
This PR adds support for type annotations in `catch` clause, like this:

```js
try {
} catch (e: Error) {
}
```

By default `any` is used, so this change is fully backwards compatible.

Fixes: https://github.com/facebook/flow/issues/2470

### DIFF
--- a/src/parser/statement_parser.ml
+++ b/src/parser/statement_parser.ml
@@ -359,10 +359,11 @@ module Statement
         let start_loc = Peek.loc env in
         Expect.token env T_CATCH;
         Expect.token env T_LPAREN;
-        let id = Parse.identifier ~restricted_error:Error.StrictCatchVariable env in
-        let param = fst id, Pattern.Identifier {
-          Pattern.Identifier.name=id;
-                             typeAnnotation=None;
+        let loc, { Pattern.Identifier.name; typeAnnotation; _; } =
+          Parse.identifier_with_type env ~no_optional:false Error.StrictCatchVariable  in
+        let param = loc, Pattern.Identifier {
+          Pattern.Identifier.name=name;
+                             typeAnnotation;
                              optional=false;
         } in
         Expect.token env T_RPAREN;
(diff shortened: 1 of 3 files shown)


thejameskyle:
We should add tests for this, @vkurchatkin do you want to do that?

vkurchatkin:
@thejameskyle sure, will do

thejameskyle:
From @samwgoldman: (since he never responded here 😛)

> Because the catch clause binding's inflows are the result of non-local throws, there's no way to actually perform type checking here.
> 
> ```js
> function f() {
>   throw "a string";
> }
> 
> try {
>   f();
> } catch (e: Error) {
>   e.message; // `e` is a string
> }
> ```
> 
> We might be able to handle this with @avikchaudhuri's work on effects, but that hasn't landed.

vkurchatkin:
Well, using `Error` is not supposed to safe. But in general it will be safer then `any` since almost always it's `Error`

samwgoldman:
Oops, I responded on the original issue instead of the PR. I'm wary of adding unsoundness here, per my comment on the issue. I'll bring it up in our weekly meeting, in case mine is the minority opinion.

vkurchatkin:
> I'm wary of adding unsoundness here, per my comment on the issue.

There is no added unsoundness per se, it's` any` already. The only downside is that if you use annotation it won't show up in `flow cover`.

anmonteiro:
I wonder if this also fixes https://github.com/facebook/flow/issues/3874

vkurchatkin:
No, these issues are unrelated

LoganBarnett:
@samwgoldman Would it be acceptable to type the result as `mixed`? This lets us try to figure out what it is, and we can still achieve 100% type coverage without necessarily hiding it like we're doing [here](https://github.com/facebook/flow/issues/4120#issuecomment-345048942).

mrkev:
@lxe https://github.com/facebook/flow/issues/2470#issuecomment-281786393

Issue #2592, opened 2016-10-06
eventemitter libdefs should return subtypes


jgrund:
Consider a `Readable`, which extends from `events$EventEmitter`.

Given:

```
readable.on('end', fn) instanceof stream.Readable // true
```

The node.js libdef should reflect the same thing. Instead, the return value of `.on` is `events$EventEmitter` and not `Readable`. Output from `flow suggest`:

```
readable.on('end', fn): events$EventEmitter
```

PR #3273, opened 2017-01-26
Correct return types for events$EventEmitter .on, .once, etc

dy-dx:
Fixes https://github.com/facebook/flow/issues/2592


alexjhannan:
+1

PR #3282, opened 2017-01-27
[WIP] Flow IDE command

nmote:


src/commands/ideCommand.ml
```diff
@@ -0,0 +1,79 @@
...
+let main option_values root () =
+  let root = CommandUtils.guess_root root in
+  let ic, oc = connect option_values root in
+  ServerProt.cmd_to_channel oc ServerProt.CONNECT;
+  let stdin_fd = Unix.descr_of_in_channel stdin in
+  let ic_fd = Timeout.descr_of_in_channel ic in
+  let oc_fd = Unix.descr_of_out_channel oc in
+  while true do
+    let readable_fds, _, _ = Unix.select [stdin_fd; ic_fd] [] [] 1.0 in
+    match readable_fds with
```
samwgoldman: Might `readable_fds` include both `ic_fd` and `stdin_fd`?

```
if List.mem ic_fd readable_fds then handle_server_message ic_fd;
if List.mem stdin_fd readable_fds then handle_stdin_message oc_fd;
```

nmote: Yeah, but the one we don't deal with here will just come up again in the next select call. I don't see a reason to complicate it

PR #3305, opened 2017-01-30
Carry on even if there is no TERM environment variable set.

SamirTalwar:
There are a few continuous integration systems which act as STDOUT is a
TTY, but don't set `TERM`. On those occasions, we get the following
error message:

    Fatal error: exception Not_found

This, obviously, makes it pretty hard to figure out what the actual
problem is. An alternative implementation could throw a better error,
but it felt like defaulting to `"dumb"` was a reasonable solution here.

I couldn't see a reasonable place to put tests, so I omitted them. If there
is, please let me know.

### DIFF
--- a/src/common/utils_js.ml
+++ b/src/common/utils_js.ml
@@ -18,7 +18,7 @@ let prerr_endlinef fmt = Printf.ksprintf prerr_endline fmt
 let can_emoji =
   Sys.os_type <> "Win32" &&
   Unix.isatty Unix.stdout &&
-  Sys.getenv "TERM" <> "dumb"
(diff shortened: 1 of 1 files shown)


mroch:
nice catch, thanks!

there's a `Sys_utils.getenv_term ()` function that we could use here but i don't think it'd look cleaner than this.

SamirTalwar:
Thanks for merging this, @mroch!

Incidentally, debugging this would have been much easier if the release version of Flow was compiled with debugging symbols. It doesn't move noticeably slower with them switched on in my experience. Is there a reason we couldn't have them turned on all the time?

Issue #739, opened 2015-08-23
Improvements to frozen object types


samwgoldman:
1. Freeze any object type, not just object literals
   
   ``` javascript
   var a: { foo: string } = { foo: "foo" };
   Object.freeze(a);
   a.foo = "bar"; // error: a is frozen
   ```
2. Freezing an object should freeze aliases to same prop map
   
   ``` javascript
   var a = { foo: "foo" };
   var b = a;
   Object.freeze(a);
   b.foo = "bar"; // error: b is frozen, too
   ```
3. `$Frozen<T : Object>` type
   
   ``` javascript
   function test(o: $Frozen<{ foo: string }>) {
     o.foo = "bar"; // error: o is frozen
   }
   
   // OK
   test({ foo: "foo" });
   
   // OK
   type O = { foo: string };
   var o: O = { foo: "foo" }
   test(o);
   
   type F = $Frozen<O>;
   ({ foo: "foo" } : F); // OK, object literal
   (o : F); // error: o can mutate
   ```
4. Property-wise subtyping for frozen objects
   
   ``` javascript
   function test(o: $Frozen<{ foo: string | number }>) { }
   
   // OK, subtyping is safe
   var a: { foo: string } = { foo: "foo" };
   test(a);
   ```

mroch:
see also #18

goodmind:
What is difference between `$ReadOnly` and `$Frozen` ?

SamChou19815:
I think it's unlikely that we implement these improvements, especially since LTI that aims to kill the action at a distance behavior. A more practical way is to only allow literals in Object.freeze

Issue #18, opened 2014-11-18
Support for immutability checks


arthuredelstein:
When I write JavaScript code, I try to treat objects and arrays as immutable, and I prefer writing pure functions whenever possible. I find this makes code much easier to reason about and avoids lots of subtle bugs. (I learned about the practical benefits of immutable collections from Clojure.)

So it would be nice if Flow could have some sort of immutability checking. With a "const" annotation, the programmer could:
1. Annotate a variable to indicate it is not supposed to change after initialization
2. Annotate a function parameter (say an Object or an Array) to guarantee that it won't be modified by the function
3. Annotate a returned Object to indicate it is not supposed to be altered by the calling code. (This is important for closures or classes that might use the returned Object internally in the future.)

Additionally, a function could have a "pure" annotation to indicate that Flow should check that a function has no side effects. Of course, all functions called by a pure function would need to be checked for purity as well.

Preferably, immutability should apply to nested objects, such as `{ a : { b : 'c' }}`. (In JavaScript, unfortunately, the Object.freeze() function does not prevent nested objects from being modified.)

Flow looks extremely nice, BTW. Thank you for making it open source.

dgreensp:
I think the important thing here is the `const` annotation.  Persistent data structures are a whole different topic.

jussi-kalliokoski:
Idea: a contract of immutability for objects and arrays.

E.g.

```javascript
function foo <T> (x : Immutable<T>) : number {
  x.value = 1; // ERROR: `x` is Immutable<T>
  return x.value;
}
```

the contract is also considered broken if you call any function that doesn't implement the contract:

```javascript
function bar <T> (x : T) : number {
  x.value = 1;
  return x.value;
}

function foo <T> (x : Immutable<T>) : number {
  return bar(x); // ERROR: `x` is Immutable<T>, bar expects mutable T
}

function qoo <T> (x : Immutable<T>) : number {
  return x.value;
}

function doo <T> (x : Immutable<T>) : number {
  return qoo(x); // this is OK
}

function too <T> (x : T) : number {
  return x.value;
}

function hoo <T> (x : Immutable<T>) : number {
  return qoo(x); // this is OK too, b/c can be statically analysed
}
```

facebook-github-bot-2:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

angelf:
[...]

PR #3308, opened 2017-01-31
Add $ReadOnly<T> utility type

jonbretman:
This is in reference to https://github.com/facebook/flow/issues/739 and https://github.com/facebook/flow/issues/18 and adds the utility type `$ReadOnly` which can be used to indicate that a type is (or should be considered) read only.

For example if you define a type like so:

```js
type Person = $ReadOnly<{
    name: string
}>
```

All objects of type `Person` will now be considered read only.

```js
const person: Person  = {name: 'Sally'};
person.name = 'Helen'; // error
```

Full Disclosure: I am no OCaml expert and won't pretend I fully understand the implications of this change so am looking for feedback / reviews.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -1083,6 +1083,9 @@ module ResolvableTypeJob = struct
     | KeysT (_, t) ->
       collect_of_type ?log_unresolved cx reason acc t
 
+    | ReadOnlyT (_, t) ->
+      collect_of_type ?log_unresolved cx reason acc t
+
     | ShapeT (t) ->
       collect_of_type ?log_unresolved cx reason acc t
 
@@ -4236,6 +4239,18 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       let new_obj = ObjT (reason, {objtype with flags}) in
       rec_flow_t cx trace (new_obj, t)
 
+    (*********************************************************)
+    (* objects can be marked as read-only using $ReadOnly<T> *)
+    (*********************************************************)
+
+    | (ObjT (reason_o, objtype), UseT(_, ReadOnlyT (reason_op, t))) ->
+      let desc = RReadOnly (desc_of_reason reason_o) in
+      let reason = replace_reason_const desc reason_op in
+
+      let flags = {frozen = true; sealed = Sealed; exact = true;} in
+      let new_obj = ObjT (reason, {objtype with flags}) in
+      rec_flow_t cx trace (new_obj, t)
+
     (*******************************************)
     (* objects may have their fields looked up *)
     (*******************************************)
@@ -6166,6 +6181,10 @@ and subst cx ?(force=true) (map: Type.t SMap.t) t =
     let keys_t_ = subst cx ~force map keys_t in
     if keys_t_ == keys_t then t else KeysT (reason, keys_t_)
 
+  | ReadOnlyT (reason, readonly_t) ->
+    let readonly_t_ = subst cx ~force map readonly_t in
+    if readonly_t_ == readonly_t then t else ReadOnlyT (reason, readonly_t_)
+
   | SingletonNumT _
   | SingletonBoolT _
   | SingletonStrT _ -> t
@@ -6369,6 +6388,7 @@ and check_polarity cx ?trace polarity = function
   | ShapeT _
   | DiffT _
   | KeysT _
+  | ReadOnlyT _
   | ObjProtoT _
   | FunProtoT _
   | FunProtoApplyT _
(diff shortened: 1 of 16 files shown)


cloudkite:
@jonbretman does this work on arrays as well? Object.freeze works on arrays

jonbretman:
@cloudkite indeed, the tests cover this - https://github.com/facebook/flow/pull/3308/files#diff-9e6e392398531297e698f31266b1a0a3R11

jeffmo:
This seems to basically generalize the existing `Object.freeze` behavior into a utility type, which seems reasonable to me.

cc @mroch who built the original `Object.freeze` support in case he has any comments before we move forward

jonbretman:
@mroch any thoughts on this?

mroch:
Nice job! This is a really tempting feature, and the implementation is great, but first a bit of history...

so, we currently only support `Object.freeze({...})` (an object literal, not a variable) because it ensures that it's "fresh", meaning there are no other references to it. `Object.freeze` mutates the object's frozen bit, so any other references also become frozen when you do that.

```
function f() {
  "use strict";
  let x = { foo: 'foo' };
  let y = Object.freeze(x);
  console.log(y === x); // Object.freeze returns its argument, doesn't create a new frozen object
  y.foo = "bar"; // fails
  x.foo = "bar"; // also fails!
}
```

Object property refinements like `if (foo.bar === "some value")` are currently unsound, because Flow assumes that a property's value never changes. I added the basic `Object.freeze` support intending to follow up to fix this. Frozen objects' properties are essentially `const`, so refinements on frozen objects should be safe. (FB employees: see D2063440). For the props to be treated as consts, it's important that frozen objects can never be mutated by other references.

Here's an example... sorry in advance, I couldn't resist carrying on the "frozen person" theme from the OP. 

```
type Person = { name: string };
type FrozenPerson = $Frozen<Person>;
class JabbaTheHut {
  static capture(x: Person): $Frozen<Person> {
    return Object.freeze(x);
  }
}
let han: Person = { name: "Han Solo" };
let frozenHan = JabbaTheHut.capture(han);
// time passes, stuff happens, potentially including `han.name = "Chewbacca"`
if (frozenHan.name === "Han Solo") {
  // we would think this is true, but changing han.name also changes frozenHan.name
}
```

So, I think calling this `$Frozen` and flipping the `frozen` bit is problematic. But I wonder if we could call it `$ReadOnlyObject` and have it make all of the props covariant, instead. That would prevent it from being mutated, but not guarantee that it isn't mutated by someone else. cc @samwgoldman re: variance

jonbretman:
Ah I understand now why `Object.freeze` was only implemented for object literals, I didn't realise that it changes the object passed in - I had assumed (always bad) that it created a **new** frozen object.

I have no strong feelings on naming and based on your explanation `$ReadOnlyObject` sounds sensible. Alternatively just `$ReadOnly`?

I'll defer to your better judgement on how to move forward with this.

cloudkite:
I would advocate `$Readonly`, typescript has a similar modifier which they have named `Readonly` https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html

jonbretman:
@mroch @cloudkite @jeffmo Updated to be `$ReadOnly`.

jonbretman:
@mroch why does Travis hate me?

src/typing/flow_js.ml
```diff
@@ -4236,6 +4239,18 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
+    (*********************************************************)
+    (* objects can be marked as read-only using $ReadOnly<T> *)
+    (*********************************************************)
+
+    | (ObjT (reason_o, objtype), UseT(_, ReadOnlyT (reason_op, t))) ->
+      let desc = RReadOnly (desc_of_reason reason_o) in
+      let reason = replace_reason_const desc reason_op in
+
+      let flags = {frozen = true; sealed = Sealed; exact = true;} in
+      let new_obj = ObjT (reason, {objtype with flags}) in
```
jeffmo: You probably do also want to mark the all the [fields](https://github.com/facebook/flow/blob/master/src/typing/type.ml#L638) covariant here to reap the benefits of read-only-ness

Should also decide now: If you have an ObjT that inherits a `proto_t`, should this operation re-generate the entire proto chain as read-only as well? I suspect this behavior could get surprising if it did not freeze every accessible field...

jonbretman: What you are saying sounds sensible, but I'm not sure I know how to do that 😞 

At this stage I would really appreciate some help in moving this forward from someone with a bit more familiarity with the codebase.

jeffmo: That's fair. Also as @mroch mentioned, this remains unsound since it creates a new type representing the same object as a previous type that is more lenient (mutable) -- which means the new type's assertions that things can't change could be wrong if the previous one is mutated.

Anyway: I'm going to be out on leave for about a month starting soon, so it probably makes sense to hand this over to @mroch to decide how to move forward. Thanks a ton for starting the discussion!

rattrayalex:
What was the resolution here?

PR #3311, opened 2017-02-01
Add React context support to lib definitions

milesj:
I got annoyed with React context always being defined as `any`, so this is a quick PR in which context takes advantage of generics (the 4th argument). The PR takes the following assumptions into account:

* The typing of context is only necessary for the component itself, mainly the life-cycle events.
* Child context doesn't need to be typed as it only has 1 call site, `getChildContext()`. Perhaps this could use method generics?
* The `Config` in `ReactClass` doesn't require context for uniqueness.

Furthermore, I don't know OCaml, so I kind of guessed on how things work. I also avoided writing tests (besides fixing existing ones) in case this is a feature that is purposefully not supported.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -13,7 +13,7 @@
  * Base class of ES6 React classes, modeled as a polymorphic class whose type
  * parameters are DefaultProps, Props, State.
  */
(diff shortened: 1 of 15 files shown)


milesj:
Closing this as it's been almost a year, and a new context RFC is in the works.

Issue #1059, opened 2015-11-08
support destructuring, spread of all iterables, not just arrays


samwgoldman:
``` javascript
const [a, b, c] = new Set([1, 2, 3]);
```

With the above code, I expect there to be no errors, and each binding `a`, `b`, `c` should be `number`.

Today, I get the following errors:

```
test.js:2
 2: const [a, b, c] = new Set([1, 2, 3]);
           ^ element 0. Indexable signature not found in
Set. See: [LIB] core.js:376

test.js:2
 2: const [a, b, c] = new Set([1, 2, 3]);
              ^ element 1. Indexable signature not found in
Set. See: [LIB] core.js:376

test.js:2
 2: const [a, b, c] = new Set([1, 2, 3]);
                 ^ element 2. Indexable signature not found in
Set. See: [LIB] core.js:376
```

[relevant spec text](http://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-destructuringassignmentevaluation).

samwgoldman:
It should also be possible to spread arbitrary iterables, not just arrays:

```javascript
const set = new Set([1, 2, 3]);
console.log(...set); // 1, 2, 3
```

```
test.js:2
  2: const set = new Set([1, 2, 3]);
                 ^^^^^^^^^^^^^^^^^^ constructor call
  3: console.log(...set); // 1, 2, 3
                    ^^^ Set. This type is incompatible with
  3: console.log(...set); // 1, 2, 3
                    ^^^ spread operand
```

MoOx:
Meanwhile, any workarounds to ignore this issue for spread with set?

jchitel:
I'm a few months late, but a workaround is to wrap all iterables (set, map, generator, string, custom iterables) in [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) before spreading them.

jsdnxx:
Another example, but DOM NodeList should also be supported:
```js
[...document.querySelectorAll('img')].forEach(node => node.remove())
```

dendril:
Any idea about the roadmap for this improvement?

calebmer:
[...]

PR #3328, opened 2017-02-02
Set.prototype.size is a getter, not a data property

michaelficarra:


jeffmo:
It's true that it's a setter, but we don't have "non-experimental" support for getters/setters right now (since we can't reason about side effects yet).

Is there more context to this change? Does this affect how Flow analyzes uses of a `Set` in some way that's problematic?

michaelficarra:
@jeffmo Yeah, I was trying to add flow to [a recent project of mine](https://github.com/michaelficarra/samevalueset) which subclasses `Set`. I ran into a number of issues, including this one. It would not allow me to create a `size` getter because it was typed as a data property.

Among the other Flow bugs I ran into were [the complete lack of Iterable support](https://github.com/facebook/flow/issues/1059) and some `FunctionBindT` error trying to type

```js
const call = Function.prototype.call.bind(Function.prototype.call);
```

I eventually just gave up on Flow, which is a shame for such a small, modern, idiomatic module.

jeffmo:
Gotcha. Ok, so since getters/setters aren't really supported well in Flow, I think we'd either need to come up with a different solution here or find a way to map getters/setters in libdefs back to normal properties when getter/setter support isn't enabled.

Going to close this out for now, but I've opened an issue [here](https://github.com/facebook/flow/issues/3426) to help us find a solution

michaelficarra:
👍

samwgoldman:
> no __proto__ support in object initialisers

This is on my TODO list, FWIW.

michaelficarra:
Nice! Good to hear.

michaelficarra:
@jeffmo It appears Flow has support for getters/setters now. Can we re-open this PR?

PR #3359, opened 2017-02-09
Add new lib def for String#match() result

milesj:
For issue #2450.


samwgoldman:
Clever! Build failure is unrelated to this change. Thanks for this fix :)

gabelevi:
I'm afraid I had to revert this with 3107fb3394cce82abd3c964e992d1178da33fb2d. We have too many special cases for `Array` for extending it to work cleanly. In the mid-term, we plan to clean up those special cases, but for now I'm reverting.

It's possible that `type MatchResult = Array<string> & {index: number, input: string};` might work too. I tried it, but ran into a bug with refining intersection types, which meant you couldn't do

    var a: string = match && match[0];

However @mroch is attempting to fix the intersection bug, so this might be a good short term fix.

milesj:
Ah darn. Thanks for the info 👍

PR #3361, opened 2017-02-09
Tests: Add a test for bad intersection behaviour

JelteF:
The test that fails now should just work.


samwgoldman:
Unfortunately, we can't accept failing tests. Please open an issue with your repro case and we'll fix it. Thanks!

PR #3395, opened 2017-02-16
jimiguitar -> jimiGuitar

gabalafou:


mroch:
thanks! we ended up rewriting this whole page, so the typo went away. check out https://flow.org/en/docs/types/modules/

PR #3427, opened 2017-02-24
Show "type at position" in Try Flow

zertosh:
Adds cursor position and type at position in the footer:

short-type-at-pos

long-type-at-pos


gabelevi:
This is fantastic. Super cool, super useful. I can't see anything wrong with it and when I play with it locally it works great. I'm going to merge this immediately and we can iterate if we want to change anything.

Again, awesome feature! I'm going to use the heck out of it!

PR #3436, opened 2017-02-28
Fix typo: channelCoundMode -> channelCountMode on AudioNode

disnet:


mroch:
not sure if it was intentional but I like the term "typeo" in your branch name, to refer to a typo in a type :)

Issue #3448, opened 2017-03-02
Can't concat $ReadonlyArray


andy-ms:
```ts
function f(a: $ReadOnlyArray<string>, b: $ReadOnlyArray<string>): $ReadOnlyArray<string> {
  return a.concat(b);
}
```

This results in the following error:

```
src/a.js:2
  2:   return a.concat(b);
                       ^ read-only array type. This type is incompatible with
  1: function f(a: $ReadOnlyArray<string>, b: $ReadOnlyArray<string>): $ReadOnlyArray<string> {
                                                                                      ^^^^^^ string
```

PR #3459, opened 2017-03-04
Fixes type for $ReadOnlyArray

nmn:
fixes: #3448


mroch:
Can you rebase and re-record the type-at-pos test? Sorry that the Travis tests were broken for so long, they should be passing now.

PR #3490, opened 2017-03-08
Make `port` argument to `socket.bind` optional.

arthurschreiber:
According to https://nodejs.org/api/dgram.html#dgram_socket_bind_port_address_callback, the `port` argument to `socket.bind` of UPD Sockets is optional.


arthurschreiber:
Signed the CLA.

PR #3495, opened 2017-03-09
Add missing types for the buffer module

steelbrain:
Just like the title says

You can find these types documented in the [official docs](https://nodejs.org/api/buffer.html) or in my screenshot below

**Note:** I did not add types for `SlowBuffer` because according to the official docs, it's depreciated and no longer recommended for use.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -117,7 +117,10 @@ declare class Buffer extends Uint8Array {
 }
 
 declare module "buffer" {
+  declare var Buffer: Class<Buffer>;
+  declare var kMaxLength: number;
   declare var INSPECT_MAX_BYTES: number;
+  declare function transcode(source: Buffer, fromEnc: buffer$Encoding, toEnc: buffer$Encoding): Buffer;
 }
 
 type child_process$execOpts = {


lib/node.js
```diff
@@ -117,7 +117,10 @@ declare class Buffer extends Uint8Array {
 }
 
 declare module "buffer" {
+  declare var Buffer: Buffer;
```
mroch: this says that `require('buffer').Buffer` is an instance of a buffer, I think you'd want `...: Class<Buffer>`.

but since `Buffer` is a global, I don't think we need this one. I admit it's opinionated to say not to use `require('buffer').Buffer` but it seems to cause a minor headache for flow because it thinks the `source: Buffer` annotation below is a value since it shadows the name.

lib/node.js
```diff
@@ -117,7 +117,10 @@ declare class Buffer extends Uint8Array {
...
+  declare var kMaxLength: number;
   declare var INSPECT_MAX_BYTES: number;
+  declare function transcode(source: Buffer, fromEnc: buffer$Encoding, toEnc: buffer$Encoding);
```
mroch: was missing a `: Buffer` return type, I'll add it

lib/node.js
steelbrain: Taken care of. We actually need this one, mainly because having this doesn't really break anything existing because you'd have to manually import it to cause any interference. Secondly, some companies have usecases where they have to write their really focused module bundlers and those module bundlers don't add anything global like Buffer but support all requires, so they have to do `global.Buffer = require('buffer')` in their init script

lib/node.js
steelbrain: Taken care of

PR #3533, opened 2017-03-17
Fix the Windows build

gabelevi:
Forgot to update the Windows version of fsnotify


gabelevi:
Tested on Windows. The build was broken. Now it's fixed.

PR #3536, opened 2017-03-18
Add missing classes and functions to url module

LinusU:
This adds the two classes `URL` and `URLSearchParams` and the two functions `domainToASCII` and `domainToUnicode` that where missing from the `url` module.

Node docs for url module: https://nodejs.org/api/url.html

I have tested that this works in a local project that used these features


adriancooney:
Thanks for doing this, just ran into this problem. Be nice to get it merged!

mroch:
looks great, thanks!

LinusU:
Thanks for pulling 🎉

PR #3563, opened 2017-03-23
Extend Server.listen signature

mleg:
This is a tested valid signature, widely used, but Flow complains. Please, merge.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -935,6 +935,7 @@ declare class http$ServerResponse extends stream$Writable {
 declare module "http" {
   declare class Server extends net$Server {
     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server;
+    listen(port: number, callback?: Function): Server;
     listen(path: string, callback?: Function): Server;
     listen(handle: Object, callback?: Function): Server;
     close(callback?: (error: ?Error) => mixed): Server;


lib/node.js
```diff
@@ -935,6 +935,7 @@ declare class http$ServerResponse extends stream$Writable {
 declare module "http" {
   declare class Server extends net$Server {
     listen(port: number, hostname?: string, backlog?: number, callback?: Function): Server;
+    listen(port: number, callback?: Function): Server;
```
avikchaudhuri: Looks like this is redundant? If you pass just a `port: number` and a `callback?: Function`, the line above should be able to match it.

avikchaudhuri:
Redundant change as far as I can tell.

lib/node.js
mleg: Thanks for reply. Below example causes flow to complai with "Error:(23, 1) _Flow: call of method `listen`. Function cannot be called on any member of intersection type intersection_ ":
```js
server.listen(3000, function () {
  console.log('Server is listening on port 3000')
})
```
So the signature you suggested doesn't match.
If I add missing parameters -- no errors then:
```js
server.listen(3000, '', 0, function () {
  console.log('Server is listening on port 3000')
})
```
Flow version 0.43.1

PR #3581, opened 2017-03-27
Add docs about typing thunk actions to redux.md

jaredpalmer:
I picked this technique up from last year's f8 sample app source code.

https://github.com/fbsamples/f8app/blob/master/js/actions/types.js#L57

### DIFF
--- a/website/en/docs/frameworks/redux.md
+++ b/website/en/docs/frameworks/redux.md
@@ -124,6 +124,44 @@ function bar(value: boolean): BarAction {
 }
 ```
 
+##### Typing Redux thunk actions <a class="toc" id="toc-typing-redux-thunk-actions" href="#toc-typing-redux-thunk-actions"></a>
+
+In order to type your Redux [thunk actions](http://redux.js.org/docs/advanced/AsyncActions.html#async-action-creators),
+you'll add types for `ThunkAction` as a function `Dispatch`, and `GetState`. `GetState` is a function that returns an `Object`. `Dispatch` accepts a disjoint union of `Action`, `ThunkAction`, `PromiseAction` and `Array<Action>` and can return `any`.
+
+```js
+type Dispatch = (action: Action | ThunkAction | PromiseAction) => any;
+type GetState = () => Object;
+type ThunkAction = (dispatch: Dispatch, getState: GetState) => any;
+type PromiseAction = Promise<Action>;
+```
+
+Then to type a thunk action creator, add a return type of a `ThunkAction` to your action creator.
+
+```js
+// @flow
+type Action =
+  | { type: "FOO", foo: number }
+  | { type: "BAR", bar: boolean };
+
+type GetState = () => Object;
+type PromiseAction = Promise<Action>;
+type ThunkAction = (dispatch: Dispatch, getState: GetState) => any;
+type Dispatch = (action: Action | ThunkAction | PromiseAction | Array<Action>) => any;
+
+
+function foo(): ThunkAction {
+  return (dispatch, getState) => {
+    const baz = getState().baz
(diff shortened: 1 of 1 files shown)


samwgoldman:
Neat! I don't know redux myself, but trust anything that @frantic does :)

Thanks you for adding this useful information to the docs :)

website/en/docs/frameworks/redux.md
```diff
@@ -124,6 +124,44 @@ function bar(value: boolean): BarAction {
 }
 ```
 
+##### Typing Redux thunk actions <a class="toc" id="toc-typing-redux-thunk-actions" href="#toc-typing-redux-thunk-actions"></a>
+
+In order to type your Redux [thunk actions](http://redux.js.org/docs/advanced/AsyncActions.html#async-action-creators),
+you'll add types for `ThunkAction` as a function `Dispatch`, and `GetState`. `GetState` is a function that returns an `Object`. `Dispatch` accepts a disjoint union of `Action`, `ThunkAction`, `PromiseAction` and `Array<Action>` and can return `any`.
```
frantic: `PromiseAction` and `Array<Action>` require additional middleware to be installed.

website/en/docs/frameworks/redux.md
```diff
@@ -124,6 +124,44 @@ function bar(value: boolean): BarAction {
...
+
+```js
+type Dispatch = (action: Action | ThunkAction | PromiseAction) => any;
+type GetState = () => Object;
```
frantic: This `Object` should in theory be your `store` type, I'm not sure if Flow can infer its shape from reducers

website/en/docs/frameworks/redux.md
```diff
@@ -124,6 +124,44 @@ function bar(value: boolean): BarAction {
...
+type ThunkAction = (dispatch: Dispatch, getState: GetState) => any;
```
frantic: according to the [source code](https://github.com/gaearon/redux-thunk/blob/master/src/index.js) `ThunkAction` can take an `extraArgument` if it was configured with it

website/en/docs/frameworks/redux.md
```diff
@@ -124,6 +124,44 @@ function bar(value: boolean): BarAction {
...
+
+function foo(): ThunkAction {
+  return (dispatch, getState) => {
+    const baz = getState().baz
+    dispatch({ type: "BAR", bar: true })
+    doSomethingAsync(baz)
+      .then(value => {
+  	    dispatch({ type: "FOO", foo: value })
+      })
+	}
```
frantic: nit: indent

website/en/docs/frameworks/redux.md
jaredpalmer: Ahh i see that now.

website/en/docs/frameworks/redux.md
jaredpalmer: @frantic not sure either. I have always kept it as `Object` because "that's how f8 did it."

website/en/docs/frameworks/redux.md
jaredpalmer: So you prefer this

```
type ThunkAction = (dispatch: Dispatch, getState: GetState, extraArgument: Object) => any;
```

Or...something like this: (I often pass an axios instance)?

```
type ThunkExtraArg = { axios: Axios }
type ThunkAction = (dispatch: Dispatch, getState: GetState, extraArgument: ThunkExtraArg) => any;
```

frantic: technically the `extraArgument` could be anything, so `any` is more appropriate. But your second example is much better when working with real apps, so I'd recommend that.

Issue #2674, opened 2016-10-23
$ObjMapi, $ObjMap and $TupleMap allow invalid writes


vkurchatkin:
Example 1:

``` js
type A = $ObjMapi<{ FOO: null }, <K>(k:K) => K>
declare var a: A;

(a.FOO : 'FOO'); // ok
(a.FOO : 'BAR'); // error
a.FOO = 'BAR'; // no error
```

Example 2:

``` js
type A = $ObjMap<{ FOO: null }, <K>(k:K) => 'FOO'>
declare var a: A;

(a.FOO : 'FOO'); // ok
(a.FOO : 'BAR'); // error
a.FOO = 'BAR'; // no error
```

Example 3:

``` js
type A = $TupleMap<[mixed, mixed], <K>(k:K) => 'FOO'>
declare var a: A;

(a[0] : 'FOO'); // ok
(a[0]: 'BAR'); // error
a[0] = 'BAR'; // no error
```

See: https://github.com/gcanti/flow-runtime/issues/11

/cc @gcanti

gcanti:
This is pretty bad for libraries like [validated](https://github.com/andreypopp/validated) or [flow-runtime](https://github.com/gcanti/flow-runtime) which rely on `$ObjMap` in order to build and extract a static type from a runtime value representing a schema.

/cc @andreypopp

vkurchatkin:
This affects only setting properties and literal assignment:

```js
var a: A = { FOO: 'BAR' }; // no error
```

but:

```js
declare var b: { FOO: 'BAR' };
var a: A = b; // an error
```

and:

```js
type A = $ObjMap<{ FOO: null }, <K>(k:K) => 'FOO'>

declare var b: { FOO: number };
var a: A = b; // an error
```

but:

```js
declare var b: { FOO: string };
var a: A = b; // no error
```

samwgoldman:
You're right, this is a problem. The cause in the implementation is that we use the existing mechanism for function calls to perform the mapping. This is convenient, but inherits an unfortunate behavior:

```
declare function f(): string;
var x = f();
if (true) x = 0; // cond. assign to avoid definite assignment refinement
(x: void); // errors: number|string ~> void
```

Calling a function returns a new, open tvar with the function return type as a lower bound. This doesn't prevent more lower bounds from flowing in, but all inflowing lower bounds are checked against any uses, so it's not totally unsound.

[...]

PR #3589, opened 2017-03-28
Add documentation for $ObjMap

gabro:
This is long due.

I volunteered [some time ago](https://github.com/facebook/flow/issues/2464#issuecomment-274808633) to document advanced features, and @thejameskyle was so nice to even reach out and ask for help for the new website that saw the light yesterday.

Unfortunately "life happened" and I couldn't find the time to contribute as I thought.

Now that the new gorgeous docs are out, I feel really sorry for not having helped like I promised and I think the minimum I can do is to fill some of the remaining the gaps.

Here's the documentation for `$ObjMap`, a powerful advanced feature that many asked to explain.

Hope this helps!

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -249,3 +249,70 @@ const user2 = {name: 'John Wilkes Booth'};
 (user2: ExactUser);
 (user2: ExactUserShorthand);
 ```
+
+## `$ObjMap<T, F>` <a class="toc" id="toc-objmap" href="#toc-objmap"></a>
+`ObjMap<T, F>` is the type obtained by taking the type of the values of an object and mapping them with a type function.
+
+Let's see an example. Suppose you have a function called `run` that takes an object of thunks (functions in the form `() => A`) in input:
+
+```js
+// @flow
+function run<A, O: {[key: string]: () => A}>(o: O) {
+  return Object.keys.map(k => o[k]());
+}
+```
+
+The function's purpose is to run all the thunks and return an object made of values. What's the return type of this function?
+
+The keys are the same, but the values have a different type, namely the return type of each function. At a value level (the implementation of the function) we're essentially mapping over the object to produce new values for the keys. How to express this at a type level?
+
+This is where `ObjMap<T, F>` comes in handy.
+
+```js
+// @flow
+
+// let's write a typelevel function that takes a `() => V` and returns a `V` (its return type)
+type ExtractReturnType = <V>(() => V) => V 
+
+function run<A, O: {[key: string]: () => A}>(o: O): $ObjMap<O, ExtractReturnType> {
+  return Object.keys.map(k => o[k]());
+}
+```
+
+Let's try this out
+
+```js
+// @flow
+const o = {
+  a: () => true,
+  b: () => 'foo'
+};
+      
+(run(o).a: boolean); // Ok
+(run(o).b: string);  // Ok
+// $ExpectError
+(run(o).b: boolean); // Nope, b is a string
+// $ExpectError
+run(o).c;            // Nope, c was not in the original object
+```
+
+This is extremely useful for expressing the return type of functions that manipulate objects values. You could use a similar approach (for instance) to provide the return type of bluebird's [`Promise.props`](http://bluebirdjs.com/docs/api/promise.props.html) function, which is like `Promise.all` but takes an object as input.
+
+Here's a possible declaration of this function, which is very similar to our first example:
+
+```js
+// @flow
+declare function props<A, O: { [key: string]: A }>(promises: O): Promise<$ObjMap<O, typeof $await>>;
+```
+
+And use:
+
+```js
+// @flow
+const promises = { a: Promise.resolve(42) };
+props(promises).then(o => {
+  (o.a: 42); // Ok
+  // $ExpectError
+  (o.a: 43); // Error, flow knows it's 42
+});
+```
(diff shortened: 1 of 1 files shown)


vkurchatkin:
I think we should postpone this until $ObjMap is fixed

gabro:
Can you point to the specific issue?

There are some other features of flow that are bugged / crippled, but I'm not sure not documenting then is the way to go.

Maybe we can add a note in the doc, pointing the the specific issue so that people are aware and can subscribe to it if they wish?

vkurchatkin:
https://github.com/facebook/flow/issues/2674

> There are some other features of flow that are bugged / crippled, but I'm not sure not documenting then is the way to go.

I think it's fine to keep them private for the time being. Some features have issues, but are still useful. `$ObjMap` is so broken that it's really unsafe to use it

gabro:
Thanks. I understand your point.

Still, I think it's more beneficial to document the feature and clearly state what's broken, rather than leaving it undocumented, which doesn't prevent usage.

This is under the assumption that people are kind of used to undocumented features in Flow, so I don't think it signals that something shouldn't be used.

vkurchatkin:
We should draw the line somewhere. What about `$Facebookism` stuff? Clearly something should stay undocumented

gabro:
I agree a line should be drawn. My understanding is that `$ObjMap` is intended as a public API (although severely broken at the moment) and for this reason people are using it (and blogging about it) without the awareness of its current state.

Documentation with a clear warning linking to the bug report would make this better, not worse, in my opinion.

That being said, I'm far from suggesting that every `$Foo` looking type should be documented. Some internals are clearly internals, I just happen to think that `$ObjMap` is not an internal.

vkurchatkin:
I consider `$ObjMap` unfinished, but not private. It seems that it's not going to be fixed any time soon, sadly.

gabro:
>I consider $ObjMap unfinished, but not private. It seems that it's not going to be fixed any time soon, sadly.

Yet, this is not clear and I see people all around using it and basing libraries on it (@gcanti and @andreypopp come to mind).

I've added a warning to the documentation.

I think someone from Facebook should now chime in and decided what's the line to keep here. If the feature shouldn't be used, I'd prefer to see it removed from the public API, rather than left half-baked and undocumented.

samwgoldman:
Thanks for doing this! Experience has shown us that it doesn't work to omit features from the documentation just because there are bugs. Instead, I think we should try to document the features with the necessary disclaimers and provisos, as @gabro has done here.

mwalkerwells:
@gabro Unless I'm missing something obvious, some of the examples are not valid JS.
````js
// @flow

// let's write a typelevel function that takes a `() => V` and returns a `V` (its return type)
type ExtractReturnType = <V>(() => V) => V 

function run<A, O: {[key: string]: () => A}>(o: O): $ObjMap<O, ExtractReturnType> {
  return Object.keys.map(k => o[k]());
}
````
Not sure why this doesn't error either...but ``Object.keys.map`` is not a function.
````js
return Object.keys.map(k => o[k]());
````

gabro:
Woops, yes, it should be Object.keys(o), not sure why it doesn't error :(
Thanks, I'll submit a fix later today.

--
gabriele
~ buildo

gabro:
@mwalkerwells wow, they were more broken than I thought. Here's the fix https://github.com/facebook/flow/pull/3605 but it's troublesome that flow itself didn't catch that.

mwalkerwells:
Yeah, pretty surprising!

That's how I rewrote it also, but was surprised when it didn't error _again_ when not calling the function within the ``Object.assign`` like so:

````js
// @flow

// let's write a typelevel function that takes a `empty => V` and returns a `V` (its return type)
type ExtractReturnType = <V>(empty => V) => V 

function run<A, O: {[key: string]: empty => A}>(o: O): $ObjMap<O, ExtractReturnType> {
  // No Error       
  return Object.keys(o).reduce((acc, k) => Object.assign(acc, { [k]: o[k] }), {});
}

// Error: Type is still "empty => number" not "number"
const foo: number = run({ foo: () => 'foo' }).foo
````
````js
12: const foo: number = run({ foo: () => 'foo' }).foo
               ^ string. This type is incompatible with
12: const foo: number = run({ foo: () => 'foo' }).foo
               ^ number
````

https://flow.org/try/#0PTAEAEDMBsHsHcBQiSmgUwC4HIDOp4AnAS03VAENRMBPAB3QwDdHRIBXAOwGNNjZO1ABYVM1CgGt0+KgAN0AWzq1QAXgB8oAGqzKnACahCWdoU4zQsnaAAUpfMcynBtBgEpEr8gFEAHpkIKXgAlEzMAFXpyVVAAHi11G0VlGjVNLTc07VBkDh4+ASMuWIBBABpQAHkALlAAbwBtKRpa3ADiTgBzAF1a5JUNUBKAX0TYWsq3WoASSoAjACsAWQo6WMqKvwCgzFCnCKjNOsRQUFQAOVhQb0JCWEJTx9OTozDBeYX0XgA6ZtwbWBub7GfTsbjoGw2ILcCoSTKDD5fTDfCi4XDETqcKHcGH1UBNXqgWAE0DDNwVOpkgDciGGyFQNzuhFqkQYoGI+DaxGg0FAACJ+qlBpx2Ao5uhCHzQJxYGI+SKxRK+YhuAI2mxYONpaLxQ8YoQuDY6hqtTZ4ZpsJBNdhSUCrbAgA

PR #3595, opened 2017-03-28
use 'npm run build' for installation if npm is selected, update gem bundler…

danidewitt:
… website


samwgoldman:
We already show "npm run build" when npm is selected. I think this PR conflicted with @gabro's work in #3598

PR #3607, opened 2017-03-29
Redirect /docs/ to /en/docs/

### LINKED ISSUE #3602: Website: /docs should redirect to /en/docs
The old URL https://flowtype.org/docs/ should redirect to https://flowtype.org/en/docs/, so that links don't break.

gabro:
Fixes #3602

Not sure this is the way to go, and I can't really test it since it's at the S3 level, but maybe @samwgoldman can tell whether it's correct.

### DIFF
--- a/website/s3_website.yml
+++ b/website/s3_website.yml
@@ -13,6 +13,7 @@ cloudfront_wildcard_invalidation: true
 redirects:
   index.html: en/
   about/: en/docs/lang/
+  docs/: en/docs/
   docs/about-flow.html: en/docs/lang/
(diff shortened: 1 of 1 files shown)


gabelevi:
lgtm! Thanks!

mroch:
doesn't seem to work. neither does the /about/ above it... :/

`docs/index.html -> en/docs/` will probably work.

gabro:
Unfortunately this is hard to test. If you guys have a quicker way of testing it, please go ahead :)

PR #3610, opened 2017-03-29
Add `FocusEvent` type

fortes:
I noticed that the `FocusEvent` (used for `focus`, `blur`, `focusin`, and `focusout`) was missing. I've added it here.

Let me know if I've done something wrong here, this is my first PR for Flow. Thanks.


samwgoldman:
Thanks! Looks like this PR changed some line numbers in the expected output for tests, but I'll update the expectations internally before we sync the commit back out.

PR #3617, opened 2017-03-30
Add http.STATUS_CODES and http.METHODS to Node.js declarations

rolftimmermans:
https://nodejs.org/api/http.html#http_http_methods
https://nodejs.org/api/http.html#http_http_status_codes

Present since v0.11.8/ v0.1.22

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -980,6 +980,9 @@ declare module "http" {
     options: Object | string,
     callback?: (response: IncomingMessage) => void
   ): ClientRequest;
+
+  declare var METHODS: Array<string>;
+  declare var STATUS_CODES: {[key: number]: string};
 }
 
 declare module "https" {


lib/node.js
```diff
@@ -980,6 +980,9 @@ declare module "http" {
     options: Object | string,
     callback?: (response: IncomingMessage) => void
   ): ClientRequest;
+
+  declare var METHODS: Array<string>;
```
gajus: Any reason for not explicitly listing all possible values using ENUM construct?

nmote: Want to put up a PR?

gajus: I am out of capacity :-(

But it would be a simple copy-paste from:

* https://github.com/nodejs/node/blob/v7.8.0/lib/_http_server.js#L17-L80 and
* ... (looks like methods come from a native binding `http_parser`)

nmote: > I am out of capacity :-(

So are all of us :(

PR #3619, opened 2017-03-31
Docs Update: More Specific Union Types Are Supertypes

mwalkerwells:
Unless I'm missing something here, more specific union types are _supertypes_ of less specific union types. The opposite is true for intersection types: more specific intersection types are _subtypes_ of less specific intersection types.

// Less Specific

// A | B | C
//    A | B
//       A
//    A & B
// A & B & C

// More Specific

``A | B | C`` is _supertype_ to ``A & B & C``

### DIFF
--- a/website/en/docs/lang/subtypes.md
+++ b/website/en/docs/lang/subtypes.md
@@ -13,17 +13,13 @@ all the possible values for both types and figure out if the other has a
 _subset_ of the values.
 
 For example, if we had a `TypeA` which described the numbers 1 through 3, and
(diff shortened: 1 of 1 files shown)


mwalkerwells:
Still a bit of a n00b, but I recently recapped my understanding here: https://medium.com/@mwalkerwells/disjoint-union-intersection-relationships-with-flow-600b0cde9b32

samwgoldman:
Oh brilliant. This just came up in #3619. Thanks for finding and fixing :D

PR #3620, opened 2017-03-31
Subtypes of unions of functions

mwalkerwells:
Adding a new example to the "Subtypes of complex types" section. Hopefully people find this helpful!

### DIFF
--- a/website/en/docs/lang/subtypes.md
+++ b/website/en/docs/lang/subtypes.md
@@ -117,56 +117,38 @@ can decide if we have a subtype or not.
 
 #### Subtypes of functions <a class="toc" id="toc-subtypes-of-functions" href="#toc-subtypes-of-functions"></a>
 
-Flow compares two functions by comparing its inputs and outputs. If all the
-inputs and outputs are a subset of the other function, then it is a subtype.
+Flow compares two functions by comparing its inputs and outputs. Subtypes have
+less specific inputs and more specific outputs. Less specific types are supertypes,
+or contravariant, and more specific types are subtypes, or covariant.
 
 ```js
-type Func1 = (1 | 2)     => "A" | "B";
-type Func2 = (1 | 2 | 3) => "A" | "B" | "C";
-```
+type Func1 = string => ?boolean;
+type Func2 = ?string => boolean;
+````
+> ```Func2``` is a subtype of ```Func1``` since ```?string``` (_an optional ```string```
+> or ```string | null```_) is a less specific type of ```string``` and ```boolean``` is
+> a more specific type of ```?boolean```.
 
-This also applies to the number of parameters in the functions. If one function
-contains a subset of the parameters of the other, then the other is a subtype.
-
+Let's take a look at a more practical example:
 ```js
-// @flow
-type Func1 = (number) => void;
-type Func2 = (number, string) => void;
+function needsWelcome(handler: (string) => ?boolean): void {
+  const didWelcome = handler("hello world!");
+  if (didWelcome) {
+    console.log(`didWelcome: ${didWelcome.toString()}`);
+  }
+}
 
-let func1: Func1 = (a: number) => {};
-let func2: Func2 = func1;
+function welcome(value: ?string): boolean {
+  if (value) {
+    console.log(value);
+    return true;
+  }
+  return false;
+}
+
+needsWelcome(welcome); // No error!
 ```
+The function ```needsHandler``` needs a function that can handle at least a ```string``` as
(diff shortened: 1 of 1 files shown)


website/en/docs/lang/subtypes.md
```diff
@@ -136,3 +136,37 @@ type Func2 = (number, string) => void;
 let func1: Func1 = (a: number) => {};
 let func2: Func2 = func1;
 ```
+
+#### Subtypes of unions of functions <a class="toc" id="toc-subtypes-of-unions-of-functions" href="#toc-subtypes-of-unions-of-functions"></a>
+
+A function is a subtype of another when the input is contravariant (_less specific type / supertype_) and the output is covariant (_more specific type / subtype_).
```
samwgoldman: The wording here makes me imagine that some functions don't have contravariant inputs and covariant outputs, but all functions have this behavior.

It's more precise to say that function types are contravariant with respect to their parameter types and covariant with respect to their return type. That is, contra- and covariance are properties of the function type.

However, I think what you have is easier to understand. I would just remove the variance stuff (at least to start with):

"A function is a subtype of another when its argument types are less specific (supertype) or its return type is more specific (subtype)."

You can then go on to introduce the technical jargon: "In technical terms, we say that function types are contravariant with respect to their parameter types and covariant with respect to their return type."

website/en/docs/lang/subtypes.md
```diff
@@ -136,3 +136,37 @@ type Func2 = (number, string) => void;
...
+
+```js
+// @flow
+declare var foo1: (string => ?string) => void
+declare var bar1: ?string => string
+
+foo1(bar1) // Works!
```
samwgoldman: This example gets straight to the point, but I think a concrete example would be better at building intuition.

A lot of javascript developers are familiar with callback functions, so how about examples using that?

```
function f(callback: (name: string) => void) {
  callback("Flow");
}

function g(name: ?string): void {
  let name = name || "World";
  console.log("Hello, " + name);
}

f(g); // `?string => void` is a subtype of `string => void`
```

```
function f(callback: () => ?string) {
  var name = callback() || "World";
  console.log("Hello, " + name);
}

function g(): string {
  return "Flow";
}

f(g); // `() => string` is a subtype of `() => ?string`
```

website/en/docs/lang/subtypes.md
```diff
@@ -136,3 +136,37 @@ type Func2 = (number, string) => void;
...
+```
+So what would be a subtype of a union of functions? Well we know that a subtype of ``A | B`` woudld be either ``A`` or ``B``, but we are looking for the subtype of ``A => B | C => D``. Let's walk through this...
```
samwgoldman: To be honest, I think this is a fun exercise for a blog post, but I'm worried that this example is too abstract and might discourage folks from grasping the concepts if introduced too early.

website/en/docs/lang/subtypes.md
```diff
@@ -136,3 +136,37 @@ type Func2 = (number, string) => void;
...
+foo2(baz2); // Works!
+```
+But is there anything else we can pass into ``foo2``? Yes! 
+```js
+// @flow
+
+declare var bar3: (A | C) => (B & D)
+foo2(bar3);
+```
+A subtype of a union of functions is a union (``|``) of the inputs and a intersection (``&``) of the outputs. While not covered here, the inverse is true for a intersection of functions: a subtype of a intersection of functions (``A => B & C => D``) is a intersection (``&``) of the inputs and a union (``|``) of the outputs.
```
samwgoldman: This *is* a very cool derivation, though!

mwalkerwells:
> I'm not sure about including the (A|C => B&D) <: (A => B) | (C => D) example in the docs, because I'm not sure how helpful it is for most people. Very cool stuff, though!

Yeah, it definitely would not have been helpful to me when I was first learning, but it was a really helpful realization when I was writing a library that heavily relied on passing functions as values & expanding the type via ```(...rest: Array<Fn>)```.

I'd love to see an "Advanced" section that this sort of thing could be a part of. For now, I'll drop it. :)

website/en/docs/lang/subtypes.md
mwalkerwells: Yeah callbacks are pretty familiar, but they generally lack any return type. :( It makes it harder to illustrate that subtypes are depended on both input and output.

website/en/docs/lang/subtypes.md
mwalkerwells: Thanks!

PR #3632, opened 2017-04-01
Point to the latest WebStorm version

lll000111:
The current WebStorm editor documentation page only points to WebStorm 2016.3. I added an additional link for WebStorm 2017.1.


samwgoldman:
Thanks!

PR #3641, opened 2017-04-03
Add v8 module definitions

### LINKED ISSUE #2832: `required module not found` for Node's 'v8'
```js
  9: const v8 = require('v8');
                ^^^^^^^^^^^^^ v8. Required module not found
```

It's definitely there: https://nodejs.org/api/v8.html

R1ZZU:
Closes #2832

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1538,6 +1538,29 @@ declare module "assert" {
   }
 }
 
+type HeapStatistics = {
+  total_heap_size: number,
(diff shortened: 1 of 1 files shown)


gabelevi:
Cross checked with [the docs](https://nodejs.org/api/v8.html#v8_v8_getheapstatistics) and this looks good to me! Thanks for the contribution!

PR #3646, opened 2017-04-03
Adding missing Google Analytics

ericnakagawa:
The new design was missing GA tracker.

### DIFF
--- a/website/_layouts/default.html
+++ b/website/_layouts/default.html
@@ -44,6 +44,15 @@
         }
       });
     </script>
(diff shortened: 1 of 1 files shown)


mroch:
[the docs](https://support.google.com/analytics/answer/1008080?hl=en) say to put it just before the `</head>`. I think that's where we had it before. not sure it really matters...

ericnakagawa:
Ok @mroch. Moved to before </head>

PR #3650, opened 2017-04-03
change return types for some Geolocation callback functions;

thurt:
success & error are callback functions, so it seems like it would be an error for the user to try and return a value from them. That is why I voided them.

Thoughts?


vkurchatkin:
It's a matter of preference. Theoretically if we know that result is ignored, it can be `mixed` or `any`.

thurt:
interesting, i hadn't thought of it like that.  so `void` is actually too strict.

PR #3680, opened 2017-04-07
Add missing property "valid" to ValidityState

gniw:
W3C, MDN or MSDN saids ValidityState has "valid" property, and  almost browsers have implemented this.


gabelevi:
Thanks for the contribution! This looks good...I'll get it merged!

PR #3697, opened 2017-04-10
update Geolocation types

### LINKED ISSUE #3665: flow coverage issue with Geolocation
i'm not able to get flow coverage on `navigator.geolocation.getCurrentPosition`

I have reproduced using the simplest possible example that won't generate a flow error (ref: https://github.com/facebook/flow/blob/v0.42.0/lib/bom.js#L521)

screen shot 2017-04-05 at 11 06 53 am

thurt:
fixes #3665

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -523,13 +523,13 @@ declare class Geolocation {
         success: (position: Position) => any,
         error?: (error: PositionError) => any,
         options?: PositionOptions
-    ) => any;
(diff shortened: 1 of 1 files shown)


thurt:
refs: 
https://dev.w3.org/geo/api/spec-source.html#geolocation_interface
https://developer.mozilla.org/en-US/docs/Web/API/Geolocation

Issue #3703, opened 2017-04-11
Missing DOM HTML elements 


mislav:
These HTML element classes are defined in Google Chrome but missing from `dom.js` definitions:

- [x] HTMLAreaElement
- [x] HTMLBodyElement
- [ ] HTMLContentElement
- [x] HTMLDataListElement
- [x] HTMLDetailsElement
- [x] HTMLDialogElement
- [ ] HTMLDirectoryElement
- [x] HTMLFieldSetElement
- [ ] HTMLFontElement
- [ ] HTMLFrameElement
- [ ] HTMLFrameSetElement
- [x] HTMLHeadElement
- [x] HTMLHtmlElement
- [x] HTMLLegendElement
- [x] HTMLMapElement
- [ ] HTMLMarqueeElement
- [x] HTMLMeterElement
- [x] HTMLModElement
- [x] HTMLObjectElement
- [ ] HTMLOutputElement
- [x] HTMLParamElement
- [x] HTMLPictureElement
- [x] HTMLProgressElement
- [x] HTMLQuoteElement
- [ ] HTMLShadowElement
- [x] HTMLTableColElement
- [x] HTMLTitleElement
- [x] HTMLTrackElement

This list was generated with:

```js
Object.getOwnPropertyNames(window).filter(n => /^HTML/.test(n)).join("\n")
```

ghodir:
@mislav I implemented the HTMLObjectElement class. Should I open a pull request?

mislav:
@ghodir Always feel free to open a pull request!

Keep in mind that adding some tests for this could be useful. See existing `tests/dom/*.js` files. Re-run tests with `make && ./runtests.sh bin/flow dom`. Record new test expectations with `./runtests.sh -r bin/flow dom`.

mroch:
https://github.com/facebook/flow/commit/7b971fbd7b0ec8c66613459eac6cfd31a0237b26 added a bunch, I updated the checklist above

mislav:
@mroch Fantastic; thank you

romeovs:
`HTMLSummaryElement` is not on this list! Should I open a new issue or can we add it to this one?

mislav:
@romeovs The specs suggests that `<summary>` does not have its own HTMLSummaryElement interface and that it instead uses HTMLElement https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element

Similarly, Google Chrome does not seem to have HTMLSummaryElement defined either.

[...]

PR #3704, opened 2017-04-11
Add DOM `fieldset`, `legend`, `details` elements

mislav:
Ref. https://github.com/facebook/flow/issues/3703


nmote:
Could you provide a link to where these changes are specified?

Also, CI is still failing, could you sort that out?

mislav:
* `<details>` https://html.spec.whatwg.org/multipage/forms.html#the-details-element
* `<fieldset>` https://www.w3.org/TR/html5/forms.html#the-fieldset-element
* `<legend>` https://www.w3.org/TR/html5/forms.html#the-legend-element

I've pushed a commit that re-records expected test output for "dom". I must have recorded the previous test output with an outdated version of `bin/flow`. This should fix CI.

nmote:
Looks good to me, thanks!

PR #3721, opened 2017-04-13
setup-yarn.md: propose note about empty .flowconfig

piperchester:


nmote:
The next page (https://flow.org/en/docs/usage/) instructs you to run `flow init` which will create a `.flowconfig`. Because of this, I don't think this change is necessary.

Perhaps it would be worthwhile to note that you need to go to the next part of the guide in order to actually use Flow, but if you'd like to do that I think it should be on the page regardless of whether you've selected yarn or npm.

nmote:
I was just talking with @samwgoldman about #3726 and he thinks it would be fine to have some redundant information here. If you could update your PR to suggest running `flow init`, consistent with #3726, I'll merge.

Thanks for your patience!

piperchester:
I can do that, thanks @nmote !

PR #3724, opened 2017-04-13
Add Buffer/EventEmitter export vars

### LINKED ISSUE #3723: Can't import {Buffer} from 'buffer', same with EventEmitter/events
Flow (tested with 43 and 42) rejects the following line, claiming that `buffer` does not have an export named `Buffer`:

```
import {Buffer} from 'buffer';
```

It allows `import Buffer from 'buffer'`, except that this doesn't work: 'buffer' doesn't export a `default` value, so that throws when compiling ES6 imports to require statements e.g. with Babel. 

The same is true for `import {EventEmitter} from 'events'`, which works but is rejected by Flow.

The use case I'm trying to solve is explicitly Buffer/EventEmitter so that I can produce node builds (using the actual node modules) and browser builds (rewriting those imports to 'buffer/' and 'events/') that require browser polyfills for them.

josephsavona:
Fixes #3723: the node "buffer" module exports a `Buffer` property, and "events" exports an `EventEmitter` property.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -118,6 +118,7 @@ declare class Buffer extends Uint8Array {
 
 declare module "buffer" {
   declare var kMaxLength: number;
+  declare export var Buffer: typeof Buffer;
   declare var INSPECT_MAX_BYTES: number;
   declare function transcode(source: Buffer, fromEnc: buffer$Encoding, toEnc: buffer$Encoding): Buffer;
 }
@@ -675,6 +676,7 @@ declare module "events" {
     static EventEmitter: typeof EventEmitter;
   }
 
+  declare export var EventEmitter: typeof EventEmitter;
   declare var exports: typeof EventEmitter;
 }
 


lib/node.js
```diff
@@ -118,6 +118,7 @@ declare class Buffer extends Uint8Array {
 
 declare module "buffer" {
   declare var kMaxLength: number;
+  declare export var Buffer: typeof Buffer;
```
avikchaudhuri: This looks OK.

lib/node.js
```diff
@@ -675,6 +676,7 @@ declare module "events" {
     static EventEmitter: typeof EventEmitter;
   }
 
+  declare export var EventEmitter: typeof EventEmitter;
```
avikchaudhuri: No idea why this is needed, given that `EventEmitter` is already declared above. @jeffmo please have a look.

lib/node.js
avikchaudhuri: Although without testing I don't know whether the type will refer to the `Buffer` class in the outer scope or just create a vacuous reference to the `var` just declared. cc @samwgoldman

lib/node.js
josephsavona: Yeah, this is weird. Without it I get:

```
21: import {EventEmitter} from 'events'; // rewritten for browsers
             ^^^^^^^^^^^^ Named import from module `events`. This module only has a default export. Did you mean `import EventEmitter from ...`?
```

lib/node.js
mroch: i deleted this from https://github.com/facebook/flow/pull/3495 for the same reason

josephsavona: Any suggestions for alternatives? I tested this and it appears to work correctly.

mroch: what does your test look like? it should be something that generates an (expected) error so you can be sure that it's not just acting like `any`/unbound

josephsavona: @mroch I tried the following:

```
const {Buffer} = require('buffer');

function readUInt8(buffer: Buffer, offset: number): number {
  return buffer.readUInt8(buffer, 'zero');
}
```

Which Flow (with this PR applied) correctly rejects with:

```
return buffer.readUInt8(buffer, 'zero');
^^^ string. This type is incompatible with the expected param type of
100:   writeUInt8(value: number, offset: number, noAssert?: boolean): number;
^^^^^^ number. See lib: flow/node.js:100
```

josephsavona: sooo... cool?

nmote:
Based on internal discussions, it seems that Joe has abandoned this PR

josephsavona:
Yup, abandoned due to lack of consensus about the right way to express this.

CMCDragonkai:
Did this pr not work?

PR #3611, opened 2017-03-29
update bundler link, add further instructions for npm setup

danidewitt:

PR #3726, opened 2017-04-13
update bundler link, add more instructions for setting up npm

danidewitt:


danidewitt:
i was having trouble with squashing commits in https://github.com/facebook/flow/pull/3611, so i opened this PR, hopefully it is cleaner

Issue #3729, opened 2017-04-14
JSON.stringify may also be undefined


xjamundx:
```
> JSON.stringify(undefined)
undefined     
```

But in flow I can do this:
```
var y: string = JSON.stringify(undefined)
```
Which will make `y` as `undefined` which is too bad.

deecewan:
As an aside, this seems like weird behaviour, right?

```
> JSON.stringify(undefined)    
< undefined
> JSON.stringify()    
< undefined
> JSON.stringify(null) 
< "null"
```

I'd probably expect it to be an empty string, I guess.

goodmind:
Fixed by https://github.com/facebook/flow/pull/7447

PR #3736, opened 2017-04-14
Add method type definition for JSON.stringify with undefined input fi…

aackerman:
…xes #3729


vkurchatkin:
This approach doesn't seems useful. I think the best thing to do is to disallow to call `stringify` with `undefined`.

aackerman:
@vkurchatkin the existing typedef allows `any` as input is there a way to specifically disallow `void`?

vkurchatkin:
@aackerman no, you should use union of everything that is allowed

aackerman:
JSON.stringify really allows anything AFAIK, but I think undefined is a special case. It doesn't really seem tenable to me to create a union to disallow undefined.

Is there some other way to approach this other than a typedef? Are there special cases for any other existing built-ins where certain types of input are disallowed?

vkurchatkin:
>  It doesn't really seem tenable to me to create a union to disallow undefined.

You'll have to do it either way. The only question is whether to add an overload that allows `undefined` or not.

aackerman:
@vkurchatkin I'm unclear on how I would proceed to do that, maybe it would be best if we close this and leave it for another person to take on.

nmote:
You could try `$NonMaybeType<mixed> | null` as the argument type for the ordinary `stringify`.

aackerman:
Thanks @nmote I wasn't aware of `$NonMaybeType`, it's not documented yet. I'll try to get my PR updated with that sometime tonight.

nmote:
I still think we should have the overload that allows `undefined`

nmote:
Thanks. Unfortunately, when I was testing something similar it seemed that due to a bug in Flow, you had to put the `void` one first in order to achieve the desired effect. Could you add some tests, or explain the manual tests you performed, and reorder if necessary?

aackerman:
@nmote I would be happy to add some tests. I'm unclear on where typedef related tests are located. Can you point me in the right direction?

vkurchatkin:
It looks like `$NonMaybeType<mixed>` doesn't actually work

nmote:
@aackerman Looks like https://github.com/facebook/flow/tree/master/tests/core_tests is the right place.

@vkurchatkin I noticed that too, and brought it up with the team. However in this case it does appear to work in conjunction with overloading, although it is possible it has no effect and Flow simply selects the first case if both may apply. However, as longs as the tests pass I think it's fine to commit.

aackerman:
I'll look into creating the tests today, if `$NonMaybeType<mixed>` doesn't work because of bugs this would probably be held up until that is fixed.

vkurchatkin:
> However in this case it does appear to work in conjunction with overloading

Not really: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVATApgYxgQwCcsw98BncsAKQGUB5AOTAG9UwOxyAXfbgSxxgADkXJYAFNywAPbgC4u3QvwB2AcwA0YYgDd+urIQD8iiQGssAT0X5VV7bvwwArllv2AlGAC8APjA7K08PKwBudk4ePkElFQ1+KCsJSM4wJ1d3MAASRjhVAFl8KwAjLAAVK2EsAB4AW34ZLAwAgB8wVRcYGE1UzmJhAhwjUzBjCQtrRR54rXTnN1Dvf0CvMHaAQUJCYpqgv160jnJRYdGZtXV1jpc6ssI+kLjLiLTogSELhKSUo-nMxS6OD8DCHI4DIYjRTjSY2Z4aRwLLJBZYBFHXLY7Kx7ewHPrHU5Yc7KS7XTp3IyPQHAjARAC+6GwZGI80IgUUDSatMwuAILKcbJK1JBEVQEjoTAAdF91IlkvgnjLPGEwMBgFwABZwLoYMBlVZgIyEOAPcUMRjSknfZIlRVW9TK1Xq8hanV6kh2Q3bE1AA

Flow just selects first overload all the time

aackerman:
The test I created is supposed to fail and require the output to be regenerated but it doesn't. Are there any open bugs around the behavior of `$NonMaybeType<mixed>`?

nmote:
Not sure, but as I noted above if you put the `void` case first it might give the expected results. As long as the tests pass I'm fine with this change, even though `$NonMaybeType` is broken.

aackerman:
Rearranging the cases worked. I've added an expression to `core_tests/json.js` that correctly identifies that `JSON.stringify` with `undefined` input does not return a string. The test outputs will have to be regenerated to account for line differences and I'm not sure how to do that.

vkurchatkin:
@aackerman `$NonMaybeType<mixed>` simply does nothing, there is no point in using it

nmote:
@aackerman  To re-record the tests just run `make test` and then copy the `testName.out` file to `testName.exp` (after, of course, scanning the diff to make sure that nothing is out of place).

@vkurchatkin You're right, but I'll merge it either way. At the very least it is documenting intent, and the tests will enforce that it doesn't bitrot, so I'm happy to allow it in.

vkurchatkin:
@nmote I don't get it, sorry. It's totally possible to implement this correctly, so there is no reason to merge incorrect implementation

nmote:
If it works as expected and there are tests enforcing the correct behavior is it really an incorrect implementation? I just don't have a strong opinion either way. If you feel strongly I would happily look at a PR that does this the way you want, or you can lobby @aackerman to implement it that way.

vkurchatkin:
> If it works as expected and there are tests enforcing the correct behavior is it really an incorrect implementation?

`$NonMaybeType<mixed> | null` is basically the same as `mixed`, so it works incorrectly when argument is `mixed`

aackerman:
Looks like the tests are failing for real reasons, I'll try to look into it today.

aackerman:
@nmote any more changes necessary for this?

I'm unclear on what changes would be satisfactory for @vkurchatkin and it might be better that he make his own follow on PR with the changes he would like to see.

nmote:
This looks good. Sorry it took so long for me to follow up, I somehow didn't see a notification. I'll import and re-record the tests myself since it's my fault they are out of date.

Agreed, @vkurchatkin feel free to CC me on a follow-up PR and I'll make sure to look at it quickly.

nmote:
Unfortunately, while I was playing around with this PR I found that it does not work as desired with `any`:

```js
(JSON.stringify((5: any)): string);
```

This results in the following error:

```
  3: (JSON.stringify((5: any)): string);
      ^^^^^^^^^^^^^^^^^^^^^^^^ undefined. This type is incompatible with
  3: (JSON.stringify((5: any)): string);
                                ^^^^^^ string
```

Adding `any` should never add type errors, since it's meant to circumvent the type checker. Unfortunately, even though this isn't actually your fault, this means we can't merge this PR right now.

gabelevi:
@nmote - are you sure that's a dealbreaker? I think I have a workaround: adding another overload using the `empty` type. Nothing other than `any` can Flow to the `empty` type, so it should work.

[try flow example](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwFIGUDyA5eA3gFDynzIAuUFAlmORTDQHYDmNAZgJ4AUJZpAG5QIAVxAAueCAC2ABwpcANPwFw50MCBgB+KTp48A1iC5TKTNkvjCxk+FGZcAlPAC8APgdPXAH3gAgjAwUFwAPI5cHioCpMhyUFp6DJas8P7MojIARtqq8M7mjCysANxE+ZTUdCklnLz5QiLiUoIA9jTAMbHqmtrJBsamRanWti3eLu5ekX6BwaERTtGN5AlJIyXp8Jk5eQKFNh3A5ZVUtPQWddx8sTbN9gAkuG3MALKhuQAqXHIgYTIaAAPEDALwZUQQCDdNQgDSJfr6QwmMy1Kz3OxSWbTSZzIIhcKRFZ3eII5JXNjbXa5GD5Q4UspEAC+FR4WDwADoGfVDABGABMAGYsT56cU2M5ymycLgueL2DdRMxQBwWKDDu1OpKiNLOdybgAiAAWIChbQNYtSkqAA)

nmote:
Interesting, if there's a hack that will keep the current tests passing plus the `any` example I gave, I'll still merge.

gabelevi:
BTW, nice catch @nmote! I totally would have missed the `any` case!

vkurchatkin:
This could be avoid but simply making `string` overload first

tsrc/test/assertions/stderr.js
```diff
@@ -6,7 +6,8 @@ import type {AssertionLocation, ErrorAssertion} from './assertionTypes';
 
 function formatIfJSON(actual: string) {
   try {
-    return JSON.stringify(JSON.parse(actual), null, 2);
+    let parsed: mixed = JSON.parse(actual)
+    return JSON.stringify(parsed, null, 2);
```
vkurchatkin: This only works because current implementation is incorrect.

gabelevi:
Hmm, I'm actually not sure if `$NonMaybeType<mixed>` works as you'd expect.

    (undefined: $NonMaybeType<mixed>);

seems to be no error. Honestly, it might just be worth making the [third overload take any](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwFIGUDyA5eA3gFDynzIAuUFAlmORTDQHYDmNAZgJ4AUJZpAG5QIAVxAAueCAC2ABwpcANPwFw50MCBgB+KTp48A1iC5TKTNkvjCxk+FGZcAlPAC8APgdPXAH3gAgjAwUFwAPI5cHioCpMhyUFp6DJas8P7MojIARtqq8M7mjCysANxE+ZTUdCklnLz5QiLiUoIA9jTAMbHqmtrJBsamRanWti3eLu5ekX6BwaERTtGN5AlJIyXp8Jk5eQKFNh3A5ZVUtPQWddx8sTbN9pHdaiAaif36hiZmtVb3dlJZtNJnMgiFwpEVnd4u9klc2NtdrkYPlDvCykQAL4VHhYPAAOnR9UMAFZAT40cU2M5yricLhCVT2DdRMxQBwWCBgId2p0aUQ6QSiTcAEQACxAEAgbRFlNSNKAA)

vkurchatkin:
> Hmm, I'm actually not sure if $NonMaybeType<mixed> works as you'd expect.

It works simply as `mixed`

> Honestly, it might just be worth making the third overload take any

Won't it just match first overload anyway?

gabelevi:
>> Hmm, I'm actually not sure if $NonMaybeType works as you'd expect.

> It works simply as mixed

Right, I thought maybe the intention was `mixed`, but without `null` and `void`.

> Won't it just match first overload anyway?

`any` will match the first overload no matter what. As I understand it we want

1. `undefined` to return `undefined`
2. anything other than `undefined` to return `string`
3. `any` to return...well, now that I think about it, it should probably return `any`. Or maybe `string & void`?

So maybe something [like this](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwFIGUDyA5eA3gFDynzIAuUFAlmORTDQHYDmNAZgJ4AUJZpAG5QIAVxAAueCAC2ABwpcANPwFw50MCBgB+KTp48A1iC5TKTNkvjCxk+FGZcAlPAC8APgdPXAH3gAgjAwUFwAPI5cHioCpMhyUFp6DJas8P7MojIARtqq8M7mjCxpAGQ2APY0wADcRPmU1HQpJZy8+UIi4lKCVcAxseqa2skGxqZFqda23d4u7l6RfoHBoRFO0R3kCUmTJenwmTl5AoWV1XUNVLT0Fq3cfLE2XfYyNAAeIP1bQ4kj+oYTGYWlZnnYpEsFnNlkEQuFIpsnvE-sk7mwDkdcjB8mc0aw6gBfeo8LB4AB0eLahgArBCfLjimxnHUSThcBTGewHjxRMxQBwWF86S4zr1qsyiKzyZSHrz+YLgKK+hKpeyZbwAEQACxAEAgFQ1DNSzKAA)?

vkurchatkin:
> undefined to return undefined

This is not particularly useful on its own

> any to return...well, now that I think about it, it should probably return any

Or maybe just string?

Like [this](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwFIGUDyA5eA3gFDynzIAuUFAlmORTDQHYDmNAZgJ4AUJZpAG5QIAVxAAueM1EQI8AD4MmbRdNEBbAEYgYarQHsDEEFGZrsWgFbgKAGn4C4AB2hhdAfikeePANYgXFKUKqx28MJikvBmXACU8AC8AHwxzPFqAIIwMFBcADyxyQ4CpMjOUO5eyiysajLauo7wccGMtQDcRM2U1HQ1bJy8zUIi4lIaNAAeIMAlpS5unt6+AUEDYRFj0bEJKWkZStm5BUXzAuWVINUhtfWaOjDNrRtqggY0wF0AvkRAA)

gabelevi:
> This is not particularly useful on its own

I'm not sure if I understand what you're saying. Since `JSON.stringify(undefined)` returns `undefined`, it's good to express this in the type. But certainly, it should be part of an overload.
The problem with having the `mixed` case return `string | void` is that 

    (JSON.stringify(undefined): void); // This errors, but shouldn't

There's also this principle to which we loosely try to adhere where adding `any` shouldn't cause errors. So

    (JSON.stringify((undefined: any)): void); // This shouldn't error

And yeah, I very much doubt that people literally write `JSON.stringify(undefined)` very often. Nevertheless, it's good to be able to accurately type it!

vkurchatkin:
> And yeah, I very much doubt that people literally write JSON.stringify(undefined) very often. 

That's what I was trying to say.

nmote:
This hasn't seen any recent activity and the discussion hinged on some Flow bugs that may or may not still exist. Anyone, feel free to reopen or open a new PR if you would like to continue work on this.

PR #3737, opened 2017-04-14
Fix typedefs of static properties on DOM Range class fixes #3734

### LINKED ISSUE #3734: Range comparison constants like START_TO_START should be statics, not instance attributes
[The DOM spec defines `START_TO_START`](https://dom.spec.whatwg.org/#dom-range-start_to_start) as a constant export from the `Range` interface: `Range.START_TO_START`.

However, [Flow defines `START_TO_START`](https://github.com/facebook/flow/blob/v0.44.0/lib/dom.js#L956) as an instance attribute rather than a static of `Range`: `document.createRange().START_TO_START`.

The expression `Range.START_TO_START` therefore fails type checking, despite being correct, and the expression `document.createRange().START_TO_START` passes type checking, despite depending on behavior not included in the spec.

Flow's built-in DOM type definitions should be updated to make these fields static.

This also applies for `Range.START_TO_END`, `Range.END_TO_START`, and `Range.END_TO_END`, in addition to `Range.START_TO_START`.

aackerman:


nmote:
Looks right. I'll merge once CI is green (currently pending)

PR #3738, opened 2017-04-15
Remove accidental line from generics docs

steveluscher:


steveluscher:
Look a few lines up to understand why this doesn't belong here.

PR #3752, opened 2017-04-16
Fix recursive type definitions

tberman:
HTMLBRElement was incorrectly a subclass of HTMLHRElement and HTMLHRElement was incorrectly subclassing itself.


gabelevi:
Thanks for fixing! I'm cherry-picking this into v0.44.1 and will have this out shortly!

tberman:
Thank you!

gabelevi:
@tberman Ok, v0.44.1 has been deployed to flow-bin and homebrew! Thanks for the report and the fix!

PR #3756, opened 2017-04-17
Fix the travis deploy

gabelevi:
`make build-flow-with-ocp` would build Flow with ocp-build, but wouldn't do the `objcopy` step which would copy the flowlibs into the binary. Since it was performed last, it was overriding the `bin/flow` binary that we had previously tested. Then, on tagged commits, we would publish the bad binary.

This makes two main changes:

1. Use `make all-ocp`, which does the full ocp-build build.
2. Move the ocp-build build before we test, so that whatever is in `bin/flow` is properly tested.


gabelevi:
Let's see if the CI systems like this...

PR #3760, opened 2017-04-18
Better definition for `process.nextTick`

davidaurelio:
Newer versions of Buck sometimes have trouble deciding the correct definition of `process.nextTick`. On the other hand, we can now use a single generic type and spread it as function parameters.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1729,8 +1729,7 @@ declare class Process extends events$EventEmitter {
     heapTotal : number;
     heapUsed : number;
   };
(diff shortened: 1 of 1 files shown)


zertosh:
Nice!

How about adding a test like this:

```
process.nextTick(
  (a: string, b: number, c: boolean) => {},
  'z',
              1,
                         true
);

process.nextTick(
  (a: string, b: number, c: boolean) => {},
  0, // Error: number ~> string
              1,
                         null // Error: null ~> boolean
);

process.nextTick(
  (a: string, b: number, c: boolean) => {},
  'z',   
              'y', // Error: string ~> number
                         true
);

PR #3776, opened 2017-04-20
Fix naming of optional property in example

floriangosse:


mroch:
oops, merged https://github.com/facebook/flow/pull/3985 first. sorry about that!

PR #3779, opened 2017-04-20
Update interfaces.md: fixed typo

DanArthurGallagher:
I changed the spelling of the example interface: the word is spelt "serializable" rather than "serializeable".
(The "z" is already hard so removing the "e" along with the general rule doesn't change the pronunciation: it should be dropped.)


gabelevi:
Thanks for cleaning this up! We'll get it merged!

PR #3792, opened 2017-04-22
Use correct number of arguments in getting-started demo

polybuildr:


polybuildr:
I'm not sure if there's a reason for using the incorrect number of arguments. Even though JS allows argument length mismatches, I'm not sure it makes sense to someone reading the getting started tutorial.

gabelevi:
Oh wow, yeah, that makes no sense. Thanks for fixing this for us! Really appreciate it!

PR #3801, opened 2017-04-24
Add TRACE method type

alcaeus:
TRACE is a valid HTTP method and should be in the type declaration.

Technically, the [method specification](https://fetch.spec.whatwg.org/#methods) claims "There are no restrictions on methods", so technically any string should be allowed as method.


gabelevi:
Looks good to me! Thanks for the contribution! We'll get it merged right away!

PR #3807, opened 2017-04-24
Added OGP images.

DanArthurGallagher:
I added two [Open Graph Images](http://ogp.me#array) of the sizes recommended by the [Facebook Developer documentation](https://developers.facebook.com/docs/sharing/best-practices#images) using the SVG icon from the website's header and with the background colour of the header.
I added these to the default layout so they should be crawled on any page.


gabelevi:
Oh awesome! Thanks for fixing this! I'll double check this with @mroch and get this merged!

gabelevi:
@DanArthurGallagher - I'm only passingly familiar with the open graph tags. Why provide two images of different sizes? I know it's supported, but it wasn't clear to me from the best practices page how two sizes will behave differently than just one large size.

DanArthurGallagher:
@gabelevi I think it's in the same spirit as other responsive images: not downloading too large an image on a lower-resolution screen. I think, say, Skype on a mid-range Android phone might use it, or Microsoft Teams on a lower-resolution monitor. Facebook itself might display it through `sizes`, at a guess, or some other responsive image implementation.

gabelevi:
Without really knowing what I'm talking about, wouldn't the Facebook website crawler cache the images locally? And I'm guessing that they'd create smaller versions of the cached images, rather than sending the huge image each time. I'm also unable to find an example of another website that uses multiple `og:image` tags to provide two different sizes of the same image.

So I'm wondering whether we could just provide the single, larger image.

DanArthurGallagher:
@gabelevi I think it might come down to your discretion: I put in both to follow the guidelines I linked to, but if you'd prefer only the one I will remove the smaller and commit it up.

DanArthurGallagher:
@gabelevi I was working in Open Graph images for another site and read the docs again - I think I've misread it at first. It doesn't seem to be suggesting that using multiple images is recommended.

I've removed the smaller image and the link to it in [my latest commit](https://github.com/facebook/flow/pull/3807/commits/3197e64d49fed5cdc002480f3baa1e5a8054bf87).

gabelevi:
Thanks for updating! I think you're right - only the larger image should be necessary. I'm tweaking it a little bit to add some padding, but otherwise this looks good to me!

PR #3818, opened 2017-04-27
fixed a typo in build command via npm

gogeta95:


gabelevi:
Thanks for the contribution, but I don't think that's a typo. I think that might be necessary to show that everything following `--` is an argument to then `babel` script and not `yarn`. `--` is a pretty standard way of saying "there are no more flags after this".

PR #3825, opened 2017-04-28
Update README with instructions on how to build flow.js

matiasleidemer:
Add instructions on how to build `flow.js`.

While trying to build the flow website (which I'm still trying) I notice that it expects the `flow.js` file inside the `bin` folder. 

This PR adds instructions on how to achieve that.


gabelevi:
Looks good to me! Thanks for documenting this!

PR #3830, opened 2017-04-28
Add `make all-homebrew`

gabelevi:
This upstreams the homebrew build that @ilovezfs suggested in
https://github.com/Homebrew/homebrew-core/pull/12970. It's basically just
`make all` which assumes you have `opam` installed but no opam packages.

When I deploy v0.46.0 to homebrew, I'll update the install step to just be
`make all-homebrew`


ilovezfs:
:tada: Thanks @gabelevi

PR #3834, opened 2017-04-28
Print out usage guidelines when non-numeric line & column are given to type-at-pos

inlinestyle:
Hello!
I noticed that the error feedback when using `type-at-pos` incorrectly could be a little more consistent.

In the current build, passing a filename and line but no column results in the following output:
```bash
$ ./bin/flow type-at-pos some/file 2
Unhandled exception: Failure("int_of_string")
```
This is because with two arguments `type-at-pos` expects to receive line and column, and for the filename to be given via stdin. Most filenames can't be parsed as ints, hence the `Failure`.

The same error occurs if three args are passed, but the second two are non-numeric, e.g.
```bash
$ ./bin/flow type-at-pos some/file 2 a
Unhandled exception: Failure("int_of_string")
```

This change just makes `type-at-pos` bail out if the line & column are unparseable with the same behavior used for the zero arguments case.

Please let me know if there's anything I can do to make this better!

### DIFF
--- a/src/commands/typeAtPosCommand.ml
+++ b/src/commands/typeAtPosCommand.ml
@@ -41,18 +41,27 @@ let spec = {
   )
 }
 
+let exit () =
+    CommandSpec.usage spec;
+    FlowExitStatus.(exit Commandline_usage_error)
+
+let parse_line_and_column line column =
+    try (int_of_string line), (int_of_string column)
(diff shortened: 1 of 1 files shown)


gabelevi:
This is awesome and looks good to me! Thanks for fixing this up!

PR #3840, opened 2017-04-29
Fixes array destructuring with defaults pattern parsing. Fixes #1985

### LINKED ISSUE #1985: Flow doesn't seem to understand default values in array destructuring
From the Firefox console:

``` js
const [a, b = {}, c = {}] = [1, {a: 1}];
a
1
b
Object { a: 1 }
c
Object {  }
```

https://tryflow.org/?code=Y29uc3QgW2EsIGIgPSB7fSwgYyA9IHt9XSA9IFsxLCB7YTogMX1dOw==
in flow:

```
assignment of identifier `c`. Could not resolve name
```

mikaelbr:
There's an error in the AST parsing for default values in array pattern parsing, where as the second destructured pattern whould be sequence expression in stead of array of assignment expressions.

It seems to me that it should be parsed as an assignment instead of expression, as `1, y = 1` is a valid expression.

This PR fixes this issue and would hopefully fix the issue in prettier (https://github.com/prettier/prettier/issues/1457) as well.

I don't know if there is anything more to test then the hardcoded tests? Let me know if there's anything else you need from me.

### DIFF
--- a/src/parser/pattern_parser.ml
+++ b/src/parser/pattern_parser.ml
@@ -196,7 +196,7 @@ module Pattern
         let pattern = match Peek.token env with
           | T_ASSIGN ->
             Expect.token env T_ASSIGN;
-            let default = Parse.expression env in
+            let default = Parse.assignment env in
             let loc = Loc.btwn (fst pattern) (fst default) in
             loc, Pattern.(Assignment Assignment.({
               left = pattern;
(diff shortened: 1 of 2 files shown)


vkurchatkin:
See https://github.com/facebook/flow/pull/3835

mikaelbr:
Ah. It didn't exist yesterday when I started working on it. Glad to see it being fixed

gabelevi:
Thanks for taking a swing, @mikaelbr! Sorry about the conflict!

mikaelbr:
No worries. Happy to help and learn more about how flow does the parsing. Thanks for commenting.

PR #3843, opened 2017-04-29
Fix various lib def function parameter types

zertosh:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString#Parameters
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString#Parameters
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#Parameters

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap#Parameters
* https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent#Values


gabelevi:
lgtm! Merging!

PR #3844, opened 2017-04-30
Changes dynamic import syntax to work without assignments

mikaelbr:
Looks to me that the parsing only looks at the `import` only supports static imports and not dynamic imports, as it's matching the token and parsing it as a import statement, even if the next token is a parenthesis. But when it is dynamic it should be treated as an expression.

This matches up with how babylon6 ast parser does it: http://astexplorer.net/#/gist/4dcc5f44e952f664c4e9f4c9098b864c/3db751c05b8b6ce7dd0f813f913c0c79efd7b120

This breaks with calling `import()` without `.then()`, but as promises are eager does really observing the result make a difference? I couldn't find any information on whether or not this is not supported. I know `import` isn't strictly a function, but a keyword with parenthesis, so maybe this changes the semantics. If this is the case we need to see if there is an identifier property as a port of the expression.

Looks like @deecewan added the initial testcase. Maybe you have some inside into if `import('dsa')` is valid or not?

### DIFF
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -180,8 +180,6 @@ module rec Parse : PARSER = struct
     (* If we see an else then it's definitely an error, but we can probably
      * assume that this is a malformed if statement that is missing the if *)
     | T_ELSE -> _if env
-    (* Handle dynamic imports *)
-    | T_IMPORT when Peek.token ~i:1 env = T_LPAREN -> expression env
     (* There are a bunch of tokens that aren't the start of any valid
      * statement. We list them here in order to skip over them, rather than
      * getting stuck *)
@@ -201,7 +199,6 @@ module rec Parse : PARSER = struct
     | T_EXTENDS
     | T_STATIC
     | T_EXPORT (* TODO *)
-    | T_IMPORT (* Not dynamic import *)
     | T_ELLIPSIS ->
         error_unexpected env;
         Eat.token env;
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -153,6 +153,7 @@ module rec Parse : PARSER = struct
     | _ when Peek.is_class env -> class_declaration env decorators
     | T_INTERFACE -> interface env
     | T_DECLARE -> declare env
+    | T_IMPORT when Peek.token ~i:1 env = T_LPAREN -> expression env
     | T_TYPE -> type_alias env
     | _ -> statement env)
 
--- a/src/parser/parser_flow.ml
+++ b/src/parser/parser_flow.ml
@@ -153,7 +153,6 @@ module rec Parse : PARSER = struct
     | _ when Peek.is_class env -> class_declaration env decorators
     | T_INTERFACE -> interface env
     | T_DECLARE -> declare env
-    | T_IMPORT when Peek.token ~i:1 env = T_LPAREN -> expression env
     | T_TYPE -> type_alias env
     | _ -> statement env)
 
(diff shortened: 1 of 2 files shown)


vkurchatkin:
There is no reason for `import('dsa')` to be invalid. Could you also add a test where `import` is not top-level, i.e. inside of a function or a block. I'm not sure, but I think your patch doesn't cover this.

mikaelbr:
I'll check it out. Thanks for the feedback 👍

mikaelbr:
There. Took me a while to navigate the code base but in the end it was pretty obvious. Added a test for import inside block-statement.

src/parser/test/hardcoded_tests.js
```diff
@@ -3018,17 +3018,60 @@ module.exports = {
...
+      // Invalid syntax tests
       'const a = import("asdf", nope);': {
         'errors': {
           '0.message': 'Unexpected token ,',
         }
+      },
+      // Should not be detected as dynamic import.
+      '{ import "asdf"; }': {
+        'errors': {
+          '0.message': 'Unexpected token import',
```
vkurchatkin: Shouldn't it be `Unexpected token "`?

src/parser/parser_flow.ml
```diff
@@ -150,6 +153,7 @@ module rec Parse : PARSER = struct
     | _ when Peek.is_class env -> class_declaration env decorators
     | T_INTERFACE -> interface env
     | T_DECLARE -> declare env
+    | T_IMPORT when Peek.token ~i:1 env = T_LPAREN -> expression env
```
vkurchatkin: I think it should be handled in `statement`, it's probably just enough to remove `T_IMPORT` there

deecewan:
hmm.  I believe I took these test cases straight from the espreema implementation.  I was just as confused as to why it couldn't/shouldn't work.

mikaelbr:
I moved the import statement management to the statement function. I thought about just letting every T_IMPORT case be handled as an expression, but I was thinking `Unexpected token import` would be better error to get than `Unexpected string`.

If you look at things like babylon the error will be `'import' and 'export' may only appear at the top level`.

Let me know what you think and I'll fix it, but to me `Unexpected token import` seems better?

vkurchatkin:
> I was thinking Unexpected token import would be better error to get than Unexpected string.

IMO that's misleading

mikaelbr:
Now it is changed to be handled as expression always.

jeffmo:
Thanks for fixing this, I should've added a test for the expression-statement form.

I'm not at a computer so it's hard for me to review, but for whoever does I wanted to just mention that these new tests look correct to me.

jeffmo:
> > I was thinking Unexpected token import would be better error to get than Unexpected string.
> IMO that's misleading

It might be nice to have a more specific/descriptive error message, but the import token is indeed the one that is unexpected.

gabelevi:
This looks good to me, especially now that @jeffmo has signed off on the tests. Thanks for fixing this!

mikaelbr:
@jeffmo 
> It might be nice to have a more specific/descriptive error message, but the import token is indeed the one that is unexpected.

To be clear, the change pulled in now is with the error `Unexpected string` not `Unexpected token import`. I can change that in a new PR if you want. 💃

gabelevi:
> To be clear, the change pulled in now is with the error Unexpected string not Unexpected token import. I can change that in a new PR if you want.

I think it's ok. If we get reports of it being confusing, we can fix. The "unexpected token" parse error is pretty common in the flow parser at the moment, though.

Issue #3786, opened 2017-04-21
Refinement of maybe properties only lasts one line


zandaleph:
The following code works as expected:

```
function consumer(val: string) {}

function example1(foo: {bar: ?string}) {
  if (foo.bar) {
    consumer(foo.bar);
  }
}
```

However, the next example uses the refined `foo.bar` twice, and the second use is unexpectedly an error:

```
function example2(foo: {bar: ?string}) {
  if (foo.bar) {
    consumer(foo.bar);  // works
    consumer(foo.bar);  // error!
  }
}
```

Worse, any function use before accessing `foo.bar` seems to invalidate the refinement:

```
function doNothing() {}

function example4(foo: {bar: ?string}) {
  if (foo.bar) {
    doNothing();
    consumer(foo.bar);  // error!
  }
}
```

As I wrote up this issue I realized that in some frustrating bozo-code universes, `consumer(foo.bar)` or `doNothing()` could modify `foo` to remove `bar`, thus invalidating the refinement.  However, when I first stumbled across it I was sure it was a bug.  Could you add something about this surprising behavior to the Maybe types docs at https://flow.org/en/docs/types/maybe/ ?

FWIW, capturing the value in a local variable solves both the Flow errors and the pathological case Flow is worried about:

```
function workingExample(foo: {bar: ?string}) {
  if (foo.bar) {
    var localBar = foo.bar;
    doNothing();
    consumer(localBar);  // works
  }
}
```

AugustinLF:
Flow is pessimistic about refinements, and considers that any function call might make your previous assertion invalid. Therefore, a function call invalidates refinements. See the [Refinement Invalidations](https://flow.org/en/docs/lang/refinements/) section of the docs.

zandaleph:
I spent an hour figuring this our instead of finding this on the last page of the docs.  Please add a link to Refinement Invalidations from the Maybe Type docs.

mwalkerwells:
@AugustinLF Click "Edit this page" on https://flow.org/en/docs/lang/refinements/ and submit a PR. 😃

gkz:
This is explained here: https://flow.org/en/docs/lang/refinements/#toc-refinement-invalidations

PR #3851, opened 2017-05-01
[doc] Link to refinement invalidation from maybe

AugustinLF:
Some people are confused by the refinement invalidation (see #3786), directing them to the refinement section from the maybe types section might help.

### DIFF
--- a/website/en/docs/types/maybe.md
+++ b/website/en/docs/types/maybe.md
@@ -62,3 +62,5 @@ function acceptsMaybeNumber(value: ?number) {
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


AugustinLF:
I'm not sure that the build errors are related to my PR, is it a problem?

avikchaudhuri:
The build errors are probably unrelated. But...
could you please delete the phrases "Do not hesitate to" and "you should not forget that"?

I think the docs read better without such phrases. If you think it's useful, you can use "Note that" to emphasize a point.

AugustinLF:
@avikchaudhuri Done! I trust you on that! I amended the commit.

PR #3865, opened 2017-05-03
Simplifying & Clarifying Module Docs

mwalkerwells:
The examples shown in "Importing and exporting types" are inconsistent with the examples shown in the other sections. The exports from "exports.js" do not correspond with the imports from "imports.js", thus confusing any reader reasonably assuming correlation. In addition, the examples are duplicative of the first section.

This diff merely consolidates the first two sections and maintains a coupling of "exports.js" to "imports.js" to help the reader understand the relationship.

### DIFF
--- a/website/en/docs/types/modules.md
+++ b/website/en/docs/types/modules.md
@@ -1,13 +1,14 @@
 ---
 layout: guide
 ---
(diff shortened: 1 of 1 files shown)


mwalkerwells:
Looks like the continuous integration system is broken...

PR #3903, opened 2017-05-08
[Blog] Nit: use correct its for posession (it's being 'it is')

wbinnssmith:


gabelevi:
Whoops! Thanks for reading and fixing!

PR #3911, opened 2017-05-09
fix docs typo: defaultProps -> props

dntrkv:


samwgoldman:
Thanks! This isn't a typo. The example here is showing that you can statically annotate the type of defaultProps, then assign the property later.

Issue #3914, opened 2017-05-10
Enforce use of function return


jacobp100:
Swift and Rust both have the behaviour that if you define a function that returns a value, you must use the value.

```js
function return5(): number { return 5 }
return5() // Error
const five = return5() // No error
```

In both of these languages, you can disregard the return value by assigning to a special variable, `_`. In JS, we could use the `void` keyword.

```js
void return5() // No error
```

Would this be a good addition?

And if so, what would be the best approach? Would something like this work,

```diff
  and funcalltype = {
    call_this_t: t;
    call_args_tlist: call_arg list;
    call_tout: t;
    call_closure_t: int;
    call_strict_arity: bool;
+   call_return_used: bool;
  }
```

And set `call_return_used` to true whenever the `CallExpression` is a direct child of `ExpressionStatement`. Then finally match `CallT(_ { call_return_used: true })` to add errors?

borela:
[You can set the expected returned type](https://flow.org/en/docs/types/functions/#function-returns-a-classtoc-idtoc-function-returns-hreftoc-function-returnsa) and flow will warn you about these mistakes.

jacobp100:
@borela, I think you misunderstood my post. Let me know if you need anything clearing up!

borela:
Yeah, my bad. There are some cases where I would want to force the user of the API to not ignore the returned value and receive at least a warning.

jacobp100:
I've not had issues with Swift forcing me to use a return value, even if it is just throwing it away. Would this not be a good default?

borela:
It's javascript default behavior, many libraries rely on it and it would suddenly show errors where none existed.

jacobp100:
Can you point to any code that makes extensive deliberate use of this?

> it would suddenly show errors where none existed before

Isn't that the point of flow? 😂

borela:
> Can you point to any code that makes extensive deliberate use of this?

All functions have a implicit "return undefined" and many programmers [take advantage of that](https://github.com/expressjs/express/blob/master/lib/router/index.js#L210) to save 1 line of code, this is more common than I would like.

> Isn't that the point of flow? :joy:

It would show errors on valid code.

jacobp100:
> many programmers take advantage of that to save 1 line of code

[...]

PR #3921, opened 2017-05-11
[Not ready] Begin implementing enforcement of use of function return value

jacobp100:
Following on from #3914, I thought I'd take a stab at it.

My heuristic for whether a function call is not used is if it is the direct child of an expression statement.

This does, however, mean that `f() || null`, while unused, is not flagged up—but I think a linter should pick that up.

The thing I'm least sure on is that this heuristic is derived from the AST, not the types themselves.

It's not ready yet, there's a few bugs, I just wanted feedback on the approach before continuing! Any pointers would be useful too!


vkurchatkin:
> My heuristic for whether a function call is not used is if it is the direct child of an expression statement.

Something like this seems to be a better fit for linter, like ESLint

jacobp100:
You wouldn't be able to determine whether the function returns a non-void value with a linter

samwgoldman:
It's awesome that you're digging into the type system and making changes! Unfortunately, I don't think this change has a good chance of being merged. You're certainly free to fork and hack on Flow as much as you'd like, though!

One suggestion: what you're catching here is a specific case of a more general (and simpler) rule. In `statement.ml` where we handle `ExpressionStatement` AST nodes, we ignore the result. Instead, you can flow the result into `VoidT`.

PR #3931, opened 2017-05-13
Add Node type to React definition

thejameskyle:
This is useful to model `PropTypes.node`

cc @sebmarkbage

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -231,7 +231,16 @@ declare module react {
   declare var PureComponent: typeof React$PureComponent;
   declare var Element: typeof React$Element;
   
-  declare type Node = string | number | React$Element<any> | Array<Node>;
+  declare type Empty = null | void | boolean;
+  declare type Text = string | number;
+
+  declare type Node =
+    | React$Element<any>
+    | Text
+    | Fragment;
+
+  declare type Fragment = Iterable<Node>;
+  declare type NodeList = Empty | Node;
 }
 
 // TODO Delete this once https://github.com/facebook/react/pull/3031 lands
--- a/lib/react.js
+++ b/lib/react.js
@@ -231,7 +231,7 @@ declare module react {
   declare var PureComponent: typeof React$PureComponent;
   declare var Element: typeof React$Element;
   
-  declare type Node = string | number | React$Element<any> | Array<React$Node>;
+  declare type Node = string | number | React$Element<any> | Array<Node>;
 }
 
 // TODO Delete this once https://github.com/facebook/react/pull/3031 lands
--- a/lib/react.js
+++ b/lib/react.js
@@ -230,6 +230,8 @@ declare module react {
   declare var Component: typeof React$Component;
   declare var PureComponent: typeof React$PureComponent;
   declare var Element: typeof React$Element;
+  
+  declare type Node = string | number | React$Element<any> | Array<React$Node>;
 }
 
 // TODO Delete this once https://github.com/facebook/react/pull/3031 lands


lib/react.js
```diff
@@ -230,6 +230,8 @@ declare module react {
   declare var Component: typeof React$Component;
   declare var PureComponent: typeof React$PureComponent;
   declare var Element: typeof React$Element;
+  
+  declare type Node = string | number | React$Element<any> | Array<Node>;
```
sebmarkbage: This can have boolean/void/null in it as well. It can also be an Iterable of Node.

https://github.com/facebook/react/blob/master/src/renderers/shared/fiber/isomorphic/ReactTypes.js#L18-L24

Although I was thinking that `Node` should not allow boolean/undefined/void those but a fragment should be able to contain them.

thejameskyle: Should there be two types then? `Node` and `Fragment` which are defined as:

```js
type Node = string | number | React$Element<any> | Array<Node>;
type Fragment = void | null | string | number | boolean | React$Element<any> | Array<Fragment>;
```

thejameskyle: Actually I can probably just copy all those type in here, with the exception of some of those Fiber types:

```js
export type Text = string | number;
export type Empty = null | void | boolean;

export type Fragment = Empty | Iterable<Node>;

export type Node =
  | React$Element<any>
  | Text
  | Fragment;

export type NodeList = Empty | Node;
```

sebmarkbage: Except Fragment should just be `Iterable<Node>`. Otherwise it makes `Node` allowed to be empty which I think is not usually what you want.

Although since these nodes can end up rendering to whatever, it's not super helpful for types anyway.

thejameskyle: done

samwgoldman:
Would you mind adding a test showing how this would be used? You can also just add an example in a gist / Try Flow and I will add it as a test.

thejameskyle:
```js
class Product extends React.Component {
  // prop-types
  static propTypes = {
    title: PropTypes.node.isRequired,
    description: PropTypes.node.isRequired,
  };

  // flow props
  props: {
    title: ReactNode,
    description: ReactNode,
  };

  render() {
    return (
      <div className="product">
        <div className="product-title">{this.props.title}</div>
        <div className="product-description">{this.props.description}</div>
      </div>
    );
  }
}

let p1 = <Product title="Foo" description="Lorem ipsum"/>
let p2 = <Product title="Bar" description={<p>Lorem ipsum</p>}/>
let p3 = <Product title="Baz" description={"Lorem/nIpsum".split('\n').map((line, index) => <p key={index}>{line}</p>)}/>
```

[[Try Flow](https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocgEyTQBsUi4YBPMJRYmAUXFbgC8cAHYBXOnTgAfOADcIwKtLgAjCBDqphFGvUYdW7TuhgAVJAA94QgM4wowYQHNlYkCqRQK1WgyaGOZBMAOQgaQXI4ZSCMABJuTRAkYRgAHhRhFgA+SOiucytcmRiYADEoFCcklO9dPwM2QK5yyurrOABJGE8UFU1UktCaLJ1ffWZG4wwhpAAZYDtBKZ4+FjyQsKRvAHptuG4LFHBNOABVGyQALko9Gxs4AAUcKlEMOEtu4Sp7koA6AGFcJBhMl4ABvXJ2FAwYBoOBgHBgUyNe5CCFRKIwmCaS7LX5PCBIlG-YSbX4LZAAR1EwCIVAANLkojQbGgHGAYRBhLi-gSiewbCSyRSkNTaUgGbkAL7eKIIwk2XHojFYnHLGaMjFwFls4Ac4BcnlcDXS2VwIhfTwACgAlHBlVEiDBRFBhHArUyMakqMBZHBbjZgkckAIAETyl4YUM5LVa72+-0MO5BpJhiOvGAAWlVSGjYJgAAsFr95WBBTmpaltj7ZDHY1F436AymQ+HnhnMzr2ZzhHnC8XS4Ku3qe5Xq7663Hx7XPTaKFEpeRF+RNPAwABGJapAmR+A5sOldSh7VIVndg3CMOzaBIPB6myiECh7Y5VfwgBMW53GeYwGxrYAIUYY9h31LkBDBVIwCya8iDvMtHyraCpRfFckDXABmL92zefdQyAgAvEDT11MDLzBUNYNvbZhA6BCn1+GwwDoP8rXwAAdYR8BtX4QBQMArStFiQXpOBHBoCw7QELI4CguAAGskBYCDxMsKUsjBYSkDHaCbRQnIgA)]

samwgoldman:
Where does `ReactNodeList` fit in? I'm particularly curious about a) the inclusion of `boolean` and b) whether there is some kind of `.isRequired` logic that would filter out the `null` and `void` bits.

thejameskyle:
That seems more like an internal React type, but it would represent all the types of values that React is able to render. It's third parameter in `React.createElement(node, props, children)`

samwgoldman:
Cool, thanks! I'll pick this up on Monday

rbalicki2:
@thejameskyle right now, React$Children is of type: any in react.js. Is the proper type for Children Iterable<Node>?

thejameskyle:
Should be the `NodeList` type here, not sure if that will delay this PR though cause it will add new type errors

samwgoldman:
We didn't end up taking this PR, but we do now support this type in lib/react.js.

PR #3937, opened 2017-05-15
Feature/node https agent

bwlt:
Flow now supports https agent extra params.
See https://nodejs.org/dist/latest-v7.x/docs/api/https.html#https_https_request_options_callback


nmote:
Sorry this never got looked at. We've since added the `https.Agent` typedef, though this PR was more thorough about the options. Since part of this has already been landed separately, I'm going to close, but feel free to reopen if you would like to get the options in.

PR #3944, opened 2017-05-15
Fix ocp-build parser test

mroch:
I broke the parser test runner in ocp-build.

This fixes the .ocp file, makes Appveyor actually run the test instead of just building it, and makes Travis build and run the test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ LINKER_FLAGS=$(BYTECODE_LINKER_FLAGS) $(SECTCREATE)
 all: $(FLOWLIB) build-flow copy-flow-files
 all-ocp: build-flow-with-ocp copy-flow-files-ocp
 
-all-homebrew: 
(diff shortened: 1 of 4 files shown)


resources/appveyor/build.sh
```diff
@@ -20,4 +20,4 @@ opam depext -u flowtype-ci
 opam install flowtype-ci --deps-only
 opam install camlp4
 make all-ocp
-make build-parser-test-with-ocp
+make test-parser-ocp
```
gabelevi: Isn't it a little weird to run tests during the build step?

It looks like appveyor has a literal test step: https://github.com/facebook/flow/blob/master/appveyor.yml#L29

And don't you need to update that script, which is trying to run `  .\_obuild\flow-parser-hardcoded-test\flow-parser-hardcoded-test.asm.exe`?

PR #3955, opened 2017-05-17
Hash.update accepts encoding for buffers

davidaurelio:
Allows to call `crypto.Hash.update` with `Buffer, 'utf8' | 'ascii' | 'latin1' | 'binary')`

While this might seem incorrect, node.js just [ignores this parameter for buffers](https://github.com/nodejs/node/blob/master/src/node_crypto.cc#L3905-L3914).

The same is true for `Cipher`, `Hmac`, `Sign`, and `Verify`

This change allows calling code to work with `string | Buffer` unions without having to refine types.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -357,17 +357,17 @@ declare class crypto$Cipher extends stream$Duplex {
   setAutoPadding(auto_padding?: boolean): crypto$Cipher;
   update(
     data: string,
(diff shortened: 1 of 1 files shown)


davidaurelio:
I didn’t re-record the tests yet, as that changes more than just the expectations for crypto…

PR #3988, opened 2017-05-20
[RFC] Improve core types for (Async) Iterable/Iterator/Generator

leebyron:
* Changes `$Iterator`'s `@@iterator` method to return `this`.
* An `$Iterator` is an `$Iterable`
* Add optional `return` and `throw` methods to `$Iterator` (http://www.ecma-international.org/ecma-262/7.0/#table-54)
* A `$Generator` is an `$Iterator`
* Fix mistaken function name to `$asyncIterate` to match style of `$iterate`.


leebyron:
#3990

Issue #1163, opened 2015-12-06
Can't get iterator from iterable object


ikokostya:
When I try to get `iterator` from `iterable` object using `Symbol.iterator`

```ts
/* @flow */

'use strict';

const a = [1, 2, 3];
for (const value of a[Symbol.iterator]()) {
    console.log(value);
}
```

`flow check` produces error:

```
test.js:6
  6: for (const value of a[Symbol.iterator]()) {
                         ^^^^^^^^^^^^^^^^^^ access of computed property/element. Element cannot be accessed with
  6: for (const value of a[Symbol.iterator]()) {
                           ^^^^^^^^^^^^^^^ string

```

I have the same problem with `Iterable` interface:

```ts
/* @flow */

'use strict';

const iterable: Iterable<number> = [1, 2, 3];
const iter = iterable[Symbol.iterator];
```

```
$ flow check
test.js:6
  6: const iter = iterable[Symbol.iterator];
                  ^^^^^^^^^^^^^^^^^^^^^^^^^ access of computed property/element. Indexable signature not found in
  6: const iter = iterable[Symbol.iterator];
                  ^^^^^^^^ $Iterable
```

```
$ flow version
Flow, a static type checker for JavaScript, version 0.19.1
```

runn1ng:
+1

This simple code does not work

```
function iterator<X>(x: Iterable<X>): Iterator<X> {
  return x[Symbol.iterator];
}

```

There seems to be `@@iterator` in flow/core.js but seems like a hack.

iamchenxin:
Are there some process for this?
Is there some standard way to write iterator protocol in Flow?

reneloperena:
Getting the same error on 0.32

drborges:
Any news on this?

zacharygolba:
[...]

PR #3990, opened 2017-05-21
[RFC] Improve core types for (Async) Iterable/Iterator/Generator

leebyron:
* Changes `@@iterator` method to return more specific type in Generators.
* An `$Iterator` is an `$Iterable`
* Add optional `return` and `throw` methods to `$Iterator` (http://www.ecma-international.org/ecma-262/7.0/#table-54)
* A `$Generator` is an `$Iterator`
* Fix mistaken function name to `$asyncIterate` to match style of `$iterate`.


goodmind:
@nmote do you think this is ok to merge?

lib/core.js
```diff
@@ -547,20 +549,22 @@ type IteratorResult<Yield,Return> =
...
 interface $Iterable<+Yield,+Return,-Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
 }
 type Iterable<+T> = $Iterable<T,void,void>;
 
-interface Generator<+Yield,+Return,-Next> {
+interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
+    +return?: <R>(value: R) => IteratorResult<Yield,R|Return>;
```
mwiencek: A `return` definition would be great. :)

I'd think it should just be `() => IteratorResult<Yield, Return>` (same for `$AsyncIterator` below) as long as that doesn't break passing a generator where an iterator is expected.

If you're hand-writing a `return` method to handle abrupt completions, you normally wouldn't accept any value because none are passed on `IteratorClose`. See http://www.ecma-international.org/ecma-262/7.0/#sec-iteratorclose

lib/core.js
```diff
@@ -547,20 +549,22 @@ type IteratorResult<Yield,Return> =
...
-type Iterator<+T> = $Iterator<T,void,void>;
-
 interface $Iterable<+Yield,+Return,-Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
 }
 type Iterable<+T> = $Iterable<T,void,void>;
 
-interface Generator<+Yield,+Return,-Next> {
+interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
```
mwiencek: I wish Flow didn't require an `@@iterator` method on Iterators, even though they typically have one that returns `this` by convention. (But I'm not sure how to use this without hacks in any case: #1163.)

nmote: #6075 has some relevant discussion

goodmind:
@nmote anything preventing merging this?

Beanow:
Happy comment anniversary @goodmind!
goodmind commented on Mar 10, 2019

I'm also finding lots of issues with async generators, but don't have the OCaml chops to help here.
Would love to see improvements like these :+1:

jedwards1211:
TIL that there's a convention for `AsyncIterator.return` being able to accept a promise:

> If the argument value is used in the typical manner, then if it is a rejected promise, a promise rejected with the same reason should be returned; if it is a fulfilled promise, then its fulfillment value should be used as the value property of the returned promise's IteratorResult object fulfillment value. However, these requirements are also not enforced.

The TS type for `AsyncIterator.return`, consequently, is

```
return?(value?: TReturn | PromiseLike<TReturn>): Promise<IteratorResult<T, TReturn>>;
```

Issue #2048, opened 2016-07-06
Static methods on an interface


appsforartists:
Say I have an interface I want a class to define, and one of the methods/properties is static.  What's the right way to define that?

I tried this:

``` typescript
export type PlanAndTargetT = {
  plan:PlanT,
  target:mixed
};

export interface PerformerI {
  static canHandle(planAndTarget:PlanAndTargetT):boolean;
  someProperty:string;
}
```

and Flow didn't know how to parse it.  This appears to work, but it seems hacky-at-best:

``` typescript
interface PerformerInstancesI {
  someProperty:string;
}

type PerformerStaticsT = {
  canHandle(planAndTarget:PlanAndTargetT):boolean;
}

export type PerformerI = PerformerInstancesI&PerformerStaticsT;
```

nmn:
ProTip: use ` ```typescript ` at the starting of your code block to get syntax highlighting.

vkurchatkin:
Static methods on interface doesn't make much sense. If you want your class to have some static methods, then you need to create a separate type containing those methods.

popham:
Static methods in ES6 are perfectly analogous to ordinary methods, unlike in Java or C++. Further, they're accessible within ordinary methods through the `constructor` property. It therefore makes perfect sense to support them on interfaces. Submitting pull request.

PR #3994, opened 2017-05-21
Typecheck interface statics

### LINKED ISSUE #803: Should `interface` reject static interface members?
I just tried to use  a static member function within an interface, and I got an unexpected identifier error:

``` js
/* @flow */

interface A {
  static someStatic(): void;
} // inter.js:4:10,10: Unexpected identifier
```

Is relaxing `Type._object env` to `Type._object ~allow_static:true env` ([src/parser/parser_flow.ml](https://github.com/facebook/flow/blob/25e6368959740e845cb962fe2843c503c7d6c4cc/src/parser/parser_flow.ml#L2885)) fraught with peril?  (I expect the override semantics of `class` to get applied, so I don't see any problem.)

popham:
Interfaces admit statics in their bodies, but no typechecking occurs. This adds statics support to interfaces.

Closes #803, #1704, and #2048.


src/typing/flow_js.ml
```diff
@@ -4883,22 +4889,60 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
+        | DefT (reason, StrT _) ->
+            let string = get_builtin_type cx ~trace reason "String" in
+            lookup_static cx trace reason string
+        | DefT (reason, BoolT _) ->
+            let boolean = get_builtin_type cx ~trace reason "Boolean" in
+            lookup_static cx trace reason boolean
+        | DefT (reason, NumT _) ->
+            let number = get_builtin_type cx ~trace reason "Number" in
+            lookup_static cx trace reason number
+        | _ -> assert false
```
popham: I'm prefer failing loudly on unhandled edge cases, but this may be a little too loud, no? Is there some extant error that invites the user to open an issue?

tests/structural_subtyping/structural_subtyping.exp
```diff
@@ -76,5 +76,37 @@ optional.js:12
...
+                   ^^^^^^ number. This type is incompatible with
+ 10:   static bar: string; //wrong type
+                   ^^^^^^ string
+
+statics.js:10
+ 10:   static bar: string; //wrong type
+                   ^^^^^^ string. This type is incompatible with
+  5:   static bar: number;
+                   ^^^^^^ number
+
```
popham: The preceding pair of errors are triggered twice, once by [statics.js#L13](https://github.com/facebook/flow/pull/3994/files#diff-368dcd24f729b3ce6926597abd693cd0R13) and once by [statics.js#L14](https://github.com/facebook/flow/pull/3994/files#diff-368dcd24f729b3ce6926597abd693cd0R14). The problem already existed and continues to exist.

mike-marcacci:
I see that this isn't merged, but have these features already landed? Using `0.49.1` I am able to declare statice on an interface (although they aren't being stripped by the latest `babel-plugin-transform-flow-strip-types`).

mike-marcacci:
OK, so [I see it has partial support](https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWA7ALgUwE6QIYDGWoAyrALZYCSmuBxoA3oqKANQDOlNanG+NMQCyWDAAtYAEwBcoABQBKUAF4AfKH450AcwDcrTQIwpCmnqWOnRE6XKWqNW3QYC+yQtHydOoAMKwODhYhBigKBQADtBYVJi+5FS02HhEJCxs3El8AkJYNpJSCsoZbKDBGACuOGigAOQSKL5NoILhOYKMVLZSdQZs7myG-PgmZllYlqPWYoXFzIZsFdW1DeItLfhG02bdhX2G7u6Int6+tISBwaHhUTFxGAk8yfRpC5nPHXkF0vOlS2IVvVGs1fG10CM8qA9tIDgNEMdLjlQABBOR+Lw+AA8iV4KQYWA0Kn8VxCGAMKIAdBMpmMflIlAYTrBkQAhdGYzg4z74tJE0AXUmhAys6kWKyEemMjws-itOS4l6pRjEtBYeAkoJk6X4SlIrSVUKBMVUWkzHrS5nIgBGCp5rxVoDVGsFWtC0utetlGBwhowxppEqligMQA), in that the `implements` keyword doesn't actually check statics, even though variables of the interface type recognize the static.

I'm going to open a bug for the current behavior.

bdbergeron:
Would really love to see this merged in. Any status update?

samwgoldman:
Tim, I know you were a proponent of static support on interfaces, so I'm sorry that we're going in a different direction. Thanks for putting the work in here.

popham:
I've been disappointed before :-). Thanks for the ack.

magicmark:
@samwgoldman @popham  what would be the recommended alternative here? How can I tell flow a class with some static methods should match a certain shape?

(Worth a note on this [in the docs](https://flow.org/en/docs/types/interfaces/)?) Thanks so much!

popham:
@magicmark You can put a typecast after the class definition. For instance,

```js
interface I {
  m(): void;
}

interface Statics {
  staticMethod1(): void;
}

class A implements I {
  static staticMethod1(): void;
  m(): void {}
}
(A: Statics);
```

I've abandoned the `new` operator outside of static methods, yielding lots of code like the following:

```js
interface Ctor<INSTANCE> {
  create(/*...*/): INSTANCE;
  staticMethod1(/*...*/): void;
  //...
}

interface I {
  method1(/*...*/): void;
  //...
}

class A implements I {
  constructor(/*...*/) {/*...*/}
  static create(/*...*/): A {
    return new A(/*...*/);
  }
  static staticMethod1(/*...*/): void;
  method1(/*...*/): void;
  //...
}
(A: Ctor<I>);

function f(K: Ctor<I>): void {}
```

While abandoning `new` outside of class methods may sound/look like a nasty imposition, Flow doesn't handle constructors soundly. For instance, Flow doesn't complain about the following:

```js
class A {
  constructor(s: string) {}
}

class B extends A {
  constructor(n: number) {}
}

function f(K: Class<A>): A {
  return new K("asdf");
}
const b = f(B);
```

(I'm pretty sure that this unsoundness can be sidestepped by (1) using `new` only within static methods and (2) avoiding `new this(/*...*/)` within static methods. I tend to use `new this(/*...*/)` within static methods while subtyping any constructor overloads by eye.)

hellsan631:
I've found a way to do this now, using the ```Class<T>``` type, along with classes instead of interfaces.

[Here is an example](https://flow.org/try/#0C4TwDgpgBAsg9gOwM7AgJwHIEMC20C8AUFFAOQDmaEOANgJYKlQA+xZOArjjiE6yRTg0ujANyFCoSFADCcOGgAmDLKgAKWOmij4oAbQRcARugA0UQzhNoAuuMnhoAZQ4BjVxCRIdUI-JoQWAj2rjRYXrCIKOhQAN5sKKp0rlAA1Ai4EABckciomJniJInAyWlG4RAAEmA5ltZFUCVlqarAWK4A1jkAFK7ySkg5cgrKGeqaaACUOgB8UGpocDh0SBAAPC7unkiz4gC+EqHh3jBcPFAQAB6oCIqnUflxCe2lKRl4PqSc3LyNzSkKmsaj4AIwABnB-1eZXCIAQKTaHU6fQG92GaJUEy0Uxyi2Wqw2Ww8Xnm8RIJCowA4aAQUH6oyQenBNig8wZgz0oLsbEOh0IxwiAHEhCJnsUYe9Ml9yKKcGIXklAZUQbpSAAmSGkA5HMLCqi0BiXG4QO4PPIxckWaVqyjUegKiVK3wqsA+ACsUMVbygSK6qMZGNGWIgGhxOWJO3FFKpNLpHPuzNZ7LRTO5jT5upOuWiaAAYh1gAoQNGAfSqKoID0Ptkc-lsHhcbI9Uh1vALWgyWxigB3OjAVwACyg1cyMytFPplXY514WW7k6gsdpsFnjUnrmngmE8tI88XUAAkMu6SKd8EFyRN2syHbDYx94vjxBqSuhQaHeuKYoIAAzLAcDQwCPgewCDksPYWBAkEAKJoEsaA9KQ7a5lAGBwMAUB5nAHB3KQUxfocJB8kAA)

Lets say I want to create an interface, ```MonsterInterface```. I can declare this as a class instead, create statics, and make them readonly.

```ts
type MonsterName =
  'gremlin' |
  'mummy' |
  'golumn';

type CoordinatePair = [number, number];
type Success = boolean;

class MonsterInterface {
  static +name: MonsterName;
  static +baseHp: number;
  static +attack: (coords: CoordinatePair) => Promise<Success>;
}
```

Whenever I want to create a class that implements this, say our mummy class. I would just extend our interface.

```ts
class Mummy extends MonsterInterface {
  static name = 'mummy';
  static baseHp = 100;
  static async attack(coords: CoordinatePair): Promise<Success> {
    return await (coords[0] > coords[1]);
  }
}
```

And If I want to ever use this type for type hinting, using, lets say, a factory. I can do something like this:

```ts
class MonsterFactory {
  static create(name: MonsterName): Class<Monster> {
    switch (name) {
      case 'mummy':
        return Mummy;
      default:
        throw new Error('Monster Not Found');
    }
  }
}
```

Notice the ```Class<T>``` type. It functions a lot like ```$Shape<T>``` where anything returned from this class needs to match the shape of Monster, otherwise it will throw errors.

oonsamyi:
@hellsan631 If you specify:
```js
class Mummy extends Monster {
  static name = 'mummy';
}
```
It is possible to call undeclared methods:
```js
  Mummy.attack([1,3]) // no flow error, runtime error
```

PR #4012, opened 2017-05-24
Change JSON.parse return type from to `mixed`

spudly:
The return type of JSON.parse should be mixed, because it can return just about anything and still be valid (especially if a `reviver` function is provided).

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -533,7 +533,7 @@ declare class URIError extends Error {
 }
 
 declare class JSON {
-    static parse(text: string, reviver?: (key: any, value: any) => any): any;
+    static parse(text: string, reviver?: (key: any, value: any) => any): mixed;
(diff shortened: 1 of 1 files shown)


nmote:
Thanks for putting up the PR. Unfortunately this is not a clear-cut issue. I think that people understand that using `JSON.parse` can give them back just about anything. Forcing them to do runtime checks that Flow can understand before using their data may be more trouble than it's worth. In practice, I bet that people would just cast through `any`, so they wouldn't end up with any more safety.

It would be an invasive change for our users, without an adequately compelling benefit. Personally I would rather have the stricter definition in your PR, but we have to be careful rolling things like this out and carefully balance our soundness goals with our desire to minimize disruption. So, for now we would like to leave this typedef as-is.

PR #4013, opened 2017-05-24
Fix minor goof on ArrayBuffer

popham:
It's `ArrayBuffer[Symbol.species]`, not `someBuffer[Symbol.species]`.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -673,7 +673,7 @@ declare class ArrayBuffer {
     constructor(byteLength: number): void;
     byteLength: number;
     slice(begin: number, end?: number): this;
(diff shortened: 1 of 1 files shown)


nmote:
Could you link some docs that support this change?

popham:
In a node shell, try

```js
ArrayBuffer[Symbol.species] === ArrayBuffer
```

If that's inadequate, then check in the ES6 spec under 24.1.3 (Properties of the ArrayBuffer Constructor) for 24.1.3.3 (get ArrayBuffer [ @@species ]). It lists a single step: "Return the `this` value." Since it's attached to the constructor, the `this` value is the constructor itself.

Issue #4019, opened 2017-05-25
Native node module method not recognised


Mottoweb:
```
import cluster from 'cluster'
cluster.on // property not found in cluster
```

Looking thru similar issues tells me it should have been fixed.
Any thoughts?

PR #4020, opened 2017-05-25
[PR] Native node module method not recognised #4019

Mottoweb:


nmote:
Could you link docs that support this change?

Mottoweb:
Could not find docs related exactly to listener method, but here is official stuff.
https://nodejs.org/api/cluster.html#cluster_class_worker

nmote:
Looks good. You'll have to update the tests with the new line numbers, then I'll merge.

vkurchatkin:
If it's not in the docs, then it's not a public API. That said, this is definitely a public API, but it's the wrong way to add it. Instead cluster should inherit from `events$EventEmitter`

nmote:
@vkurchatkin It actually is mentioned in the link there so I think we're good to go. But you make a good point that it should instead inherit.

Mottoweb:
```declare class events$EventEmitter {```
```  // deprecated```

Should i just do 
```
declare function on(event: string, listener: Function);
```
Inside cluster module declaration?

PR #4030, opened 2017-05-26
Make readyState constants static

barbu110:
The XMLHttpRequest.readyState constant values should be static.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -513,14 +513,15 @@ declare class XMLHttpRequest extends EventTarget {
     getResponseHeader(header: string): string;
     msCachingEnabled(): boolean;
     overrideMimeType(mime: string): void;
-    LOADING: number;
-    DONE: number;
-    UNSENT: number;
-    OPENED: number;
-    HEADERS_RECEIVED: number;
 
     statics: {
         create(): XMLHttpRequest;
+      
+        LOADING: number;
+        DONE: number;
+        UNSENT: number;
+        OPENED: number;
+        HEADERS_RECEIVED: number;
     }
 }
 
--- a/tests/fetch/fetch.exp
+++ b/tests/fetch/fetch.exp
@@ -1,106 +1,106 @@
 fetch.js:12
  12: const b: Promise<string> = fetch(myRequest); // incorrect
                       ^^^^^^ string. This type is incompatible with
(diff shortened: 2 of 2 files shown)


nmote:
This looks right but could you link the docs that support this change?

barbu110:
@nmote, yes of course!

First of all, there is the Google Chrome source code [file](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/core/xmlhttprequest/XMLHttpRequest.idl) which describes the `XMLHttpRequest` interface like this:

```idl
[
    ActiveScriptWrappable,
    Constructor,
    ConstructorCallWith=ScriptState,
    DependentLifetime,
    Exposed=(Window,DedicatedWorker,SharedWorker),
] interface XMLHttpRequest : XMLHttpRequestEventTarget {
    // event handler
    attribute EventHandler onreadystatechange;

    // states
    const unsigned short UNSENT = 0;
    const unsigned short OPENED = 1;
    const unsigned short HEADERS_RECEIVED = 2;
    const unsigned short LOADING = 3;
    const unsigned short DONE = 4;
    // ...
};
 ```

And according to the [standard](https://www.w3.org/TR/WebIDL-1/#idl-constants) 

> Constants are not associated with particular instances of the interface on which they appear. It is language binding specific whether constants are exposed on instances. **NOTE** The ECMAScript language binding does however allow constants to be accessed through objects implementing the IDL interfaces on which the constants are declared. For example, with the following IDL: `(truncated)` the constant value can be accessed in ECMAScript either as `A.rambaldi` or `instanceOfA.rambaldi`.

So this should probably describe the status both static and not-static. (But, IMHO, using them as statics is the correct approach).

barbu110:
What do you you think?

nmote:
This looks good, but the tests need to be updated since line numbers changed. I'll merge once you do that.

barbu110:
Could you guide me please? I am not familiar with the directory structure. At a glance, I couldn't find the tests responsible for what I've changed. 

Thank you.

philikon:
Can we make this not an either-or thing please? Every single browser and XHR polyfill exposes the state constants as both instance *and* static members, and even the spec says that *either* is ok. Why does Flow get to tell what to do here?

PR #4042, opened 2017-05-27
add getContext("experimental-webgl") typecheck

gre:
as IE still seems to needs that, we should keep doing code like: 

```js
const gl = canvas.getContext("webgl") || canvas.getContext("experimental-webgl");
```

and assume that experimental-webgl is same API as webgl.

with today's flow, the inferring won't work and you gets a weird union of types (that probably is the 'fallback' `RenderingContext` union )

screen shot 2017-05-27 at 18 54 10

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -2449,6 +2449,8 @@ declare class HTMLCanvasElement extends HTMLElement {
   height: number;
   getContext(contextId: "2d", ...args: any): ?CanvasRenderingContext2D;
   getContext(contextId: "webgl", contextAttributes?: $Shape<WebGLContextAttributes>): ?WebGLRenderingContext;
+  // IE currently only supports "experimental-webgl"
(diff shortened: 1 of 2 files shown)


nmote:
This sounds plausible, but could you add a comment explaining when you need to use this (sounds like just for IE?). Also, it looks like you'll have to re-record tests since some line numbers changed.

gre:
I've added a comment, however i'm currently having trouble building the project so i'm not sure how to regenerate the files.

gre:
oh ok, i had to do `eval $(opam config env)`  – hold on, i'll try to generate

gre:
done (but BTW travis have some issues today https://www.traviscistatus.com/ )

nmote:
Looks good, I'll land once CI finishes. Thanks for the heads up about Travis.

PR #4079, opened 2017-06-01
Add missing parameters to localeCompare definition

yanndinendal:
Declare the `locales` and `options` optional parameter of the `localeCompare` string method.

References:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#options_argument

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -285,7 +285,7 @@ declare class String {
     indexOf(searchString: string, position?: number): number;
     lastIndexOf(searchString: string, position?: number): number;
     link(href: string): string;
-    localeCompare(that: string): number;
+    localeCompare(that: string, locales?: string | Array<string>, options?: Object): number;
     match(regexp: string | RegExp): ?Array<string>;
     normalize(format?: string): string;
     padEnd(targetLength: number, padString?: string): string;


yanndinendal:
Thanks! :)

yanndinendal:
@mroch : Hi, looking at the git history, looks like this was part of the 0.48 release, but I don't get the fix after updating to flow-bin 0.48. Others in my time don't either.

``` js
src/components/OrganizationMenu.js:119
119:             a.name.localeCompare(b.name, this.props.locale))
                                              ^^^^^^^^^^^^^^^^^ unused function argument
  288:     localeCompare(that: string): number;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type expects no more than 1 argument. See lib: /tmp/flow/flowlib_1254e9eb/core.js:288
```

Looking at the file used by flow, the fix is indeed not included:

```sh
$ flow version
Flow, a static type checker for JavaScript, version 0.48.0

$ vi /tmp/flow/flowlib_1254e9eb/core.js +288
```
```js
localeCompare(that: string): number;
```

Any idea what happened?

PR #4081, opened 2017-06-01
Swap 'soundness' and 'completeness'; tweak grammar

wcerfgba:
The definitions of soundness and completeness were the wrong way round, see:
http://logan.tw/posts/2014/11/12/soundness-and-completeness-of-the-type-system/
https://stackoverflow.com/questions/21437015/soundness-and-completeness-of-systems

and other resources.

I took the liberty of tweaking a couple of bits of grammar/wording while I was here too.

On a separate note, the sentence "Unit/Integration testing can be an extreme form of this approach." is thrown in without any elaboration. Although I can see how you could make this argument / draw this analogy, I think it could do with expansion/justification. Happy to write a separate change for this if people agree.

Thanks. :)

### DIFF
--- a/website/en/docs/lang/types-and-expressions.md
+++ b/website/en/docs/lang/types-and-expressions.md
@@ -69,30 +69,31 @@ of values. But still Flow checks _every_ possible value. In this way Flow is
 checking too many things or _over-approximating_ what will be valid code.
 
 By checking every possible value, Flow might catch errors that will not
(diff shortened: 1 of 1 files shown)


vkurchatkin:
Original text seems to be correct

wcerfgba:
@vkurchatkin Apologies, I was completely incorrect. Closing.

PR #4096, opened 2017-06-04
Fix typo on HTMLElement.dir

### LINKED ISSUE #2864: Element direction value rtl seems to be wrongly reported
I've just started debugging an existing code base that uses HTML5 Canvas elements with flow.

One of the error flow reports regards these lines:

```js
const newCanvas = document.createElement('canvas');
newCanvas.dir = 'rtl';
```

The specific error is: `This type is incompatible with string enum`
And my flow IDE integration links to a dom.js definition file quoted here:

```
declare class HTMLElement extends Element {
  ...
  dataset: {[key:string]: string};
  dir: 'ltr' | 'trl' | 'auto';
```

But looking at the [spec](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dir) it seems that the second string should be `rtl` rather than `trl`.

I don't have much experience with flow, so if this issue is incorrect please correct me. :)

Flow version: 0.35.0
OS Platform: OSX El Capitan - 10.11.6

yuchi:
Original implementation had `trl` instead of `rtl` as one of the possible values of the `dir` attribute.

Please see [MDN’s definition of `dir`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir) and its possible values, which are indeed `ltr`, **`rtl`** and `auto`.

Fixes #2864

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1194,7 +1194,7 @@ declare class HTMLElement extends Element {
   contentEditable: string;
   contextMenu: ?HTMLMenuElement;
   dataset: {[key:string]: string};
-  dir: 'ltr' | 'trl' | 'auto';
+  dir: 'ltr' | 'rtl' | 'auto';
   draggable: bool;
   dropzone: any;
   hidden: boolean;


yuchi:
Thanks @mroch !

Issue #3674, opened 2017-04-06
Octal and binary numbers > 2^63-1 crash the parser


mroch:
we're using `Int64.of_string` to parse octal and binary numbers (e.g. `0o1234` or `0b10101`). `Int64.of_string` throws when called with a number larger than `Int64.max_int` (2^63-1).

Instead, I think we can just use `string_of_float`. It seems to support `0b` and `0o`.

PR #4117, opened 2017-06-07
Fix large octal and binary literals crashing the parser by building o…

fletcherw:
Fix large octal and binary literals crashing the parser by building on the hexadecimal parser added in commit e3fda214. This pull request fixes issue #3674. I did not use float_of_string, as suggested, because it appears that float_of_string only supports hexadecimal and decimal literals.


fletcherw:
Accidentally closed

fletcherw:
Am having trouble reproducing travis test failures on my machine, will keep looking them.

fletcherw:
I've figured out the source of the travis failures (overflow bug when the mantissa got big), but I'm wondering what the expected behavior should be with regards to rounding. Is it acceptable if, for example, the hex literal 0x35237678ab3e246cd (decimal: 61264549901378930381) is rounded to 61264549901378930000 in Node and Firefox's js engines, but 6.1264549901400000000 in flow? Should I remove the tests that test for such specific behavior in the parser and instead just make sure that flow does not crash when run on a file with those large numbers?

PR #4128, opened 2017-06-08
Fix typo in website / docs

Aaronik:
Change incorrect usage of "it's" to the grammatically correct "its"


Aaronik:
I did actually sign the CLA. I just did a few seconds ago, maybe I jumped the gun with this.

PR #4158, opened 2017-06-13
Disable all warnings for homebrew & opam release builds

gabelevi:
This should avoid issues like https://github.com/Homebrew/homebrew-core/pull/14545 where our CI is happy with the build (since we use 4.03.0) but it hits a warning in 4.04.0 when built with homebrew.

Test Plan: 
```
opam switch 4.04.0
eval `opam config env`
make clean
make # Errored
make clean
FLOW_RELEASE=1 make # Worked
make all-homebrew # Worked
opam pin add flowtype .
```

Previously `make all-homebrew` would error with 4.04.0

I could never get `opam install flowtype` to error, but I can at least confirm that this doesn't make it worse.


gabelevi:
Just realized that the same warning is going to break opam too...let's see if I can fix that as well

ilovezfs:
@gabelevi thanks for looking into this!

avikchaudhuri:
@gabelevi This is the opam build PR: https://github.com/ocaml/opam-repository/pull/9491

gabelevi:
Well, it's good to know opam doesn't error, despite the warning. However, I distinctly remember it erroring on warnings before, hence https://github.com/facebook/flow/pull/2237

PR #4176, opened 2017-06-15
Add util.promisify

yuffiy:
Add util.promisify for NodeJs v8.0.0

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1500,6 +1500,7 @@ declare module "util" {
   declare function isError(object: any): boolean;
   declare function inherits(constructor: Function, superConstructor: Function): void;
   declare function deprecate(f: Function, string: string): Function;
+  declare function promisify(f: Function): Function;
 }
 
 type vm$ScriptOptions = {


nanot1m:
I think it would be better to specify more explicit type:
```js
type CallbackFn1<T1, T2> = (T1, (*, T2) => void) => void
type CallbackFn2<T1, T2, T3> = (T1, T2, (*, T3) => void) => void
type CallbackFn3<T1, T2, T3, T4> = (T1, T2, T3, (*, T4) => void) => void
/* ... */
declare function promisify<T1, T2>(CallbackFn1<T1, T2>): T1 => Promise<T2>
declare function promisify<T1, T2, T3>(CallbackFn2<T1, T2, T3>): (T1, T2) => Promise<T3>
declare function promisify<T1, T2, T3, T4>(CallbackFn3<T1, T2, T3, T4>): (T1, T2, T3) => Promise<T4>
/* ... */
```

stouf:
I totally agree with @nanot1m . I would even go further and say that I would rather not have a type definition for `util.promisify()` at all. Having a wrong type definition for promisified functions could lead to errors since one would probably use those promisified functions thinking Flow will make sure they are being called correctly, which could not be the case.

That being said, I have to admit I have no idea how to type `util.promisify()` so the number of arguments is not limited 😥

dckc:
> I think it would be better to specify more explicit type:

FWIW, I've been using this approach:

```js
/*
 * Adapt callback-style API using Promises.
 *
 * Instead of obj.method(...arg, callback),
 * use asPromise(cb => obj.method(...arg, cb)) and get a promise.
 *
 * @param calling - a function of the form (cb) => o.m(..., cb)
 * @return A promise for the result passed to cb
 */
function asPromise/*:: <T>*/(calling /*: Callback<T> => mixed*/) /*: Promise<T> */{
  function executor(resolve, reject) {
    const callback = (err, result) => {
      if (err) {
        reject(err);
        return;
      }
      if (typeof result === 'undefined') { throw new TypeError('must give err or result'); }
      resolve(result);
    };

    calling(callback);
  }

  return new Promise(executor);
}
/*::
type Callback<T> = (error: Error | null, response?: T) => void;
*/
```

I keep copying and pasting it between projects; I'm not sure what to do with it. I thought about contributing it to one of the popular promisify / as-promise packages, but in researching them I discovered node's `util.promisify`, which seems to subsume those packages (though perhaps not in a browser context...).

PR #4201, opened 2017-06-18
fix minor error in flow react docs example

RobertBorg:


RobertBorg:
looks like i just didn't understand the docs fully, might resubmit with changes to the phrasing.

PR #4206, opened 2017-06-19
XMLHttpRequest: make state consts available on both constructor and instances

philikon:
As per spec: https://www.w3.org/TR/WebIDL-1/#idl-constants

See also #4030 which changed Flow's behavior from one arbitrary choice to another, while in fact the spec says that both are valid, so I think Flow should accept both.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -514,10 +514,15 @@ declare class XMLHttpRequest extends EventTarget {
     getResponseHeader(header: string): string;
     msCachingEnabled(): boolean;
     overrideMimeType(mime: string): void;
+    LOADING: number;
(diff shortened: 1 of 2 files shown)


philikon:
ping?

philikon:
ping?

gabelevi:
Yeah, I kind of agree with you. That PR kind of bugged me too. Let's get this merged.

gabelevi:
By the way, sorry for the delay. This was my responsibility to merge, but instead I went on holiday. Sorry :(

philikon:
@gabelevi hah no problem, people gotta do holidays

PR #4209, opened 2017-06-20
Node HTTPS Agent: Add the Agent class to the https module.

AdrienFromToulouse:
## What is implemented:

Add the Agent class to the https module.

Flow version 0.48.0 does not support it.
screen shot 2017-06-20 at 12 10 35

## How is it implemented:

By reporting 
```js
declare class Agent extends http$Agent {}
```
into `module "https"`


AdrienFromToulouse:
Could someone help me to found out why my change impact the `test` on `Crypto` and `process`? 
What should I update?
https://travis-ci.org/facebook/flow/jobs/244787793

Haroenv:
I think you need to record the tests' changes @AdrienFromToulouse

AdrienFromToulouse:
@Haroenv thanks for you reply, however I get the same fails on the master branch itself, when running locally... so I don't think it is because of my changes.

AdrienFromToulouse:
@Haroenv thx for your reply!

Indeed you were right :) not as obvious as `Jest` tho.
Would be nice to have a small mention of it in the README under the test section.

`-r re-record failing tests to update expected output`

Will update the PR then.

Cheers!

AdrienFromToulouse:
Anyone to merge it?

gabelevi:
Sorry for the delay! This looks good to me! Thanks for the fix!

PR #4217, opened 2017-06-21
Add IntersectionObserver declarations

pascalduez:
Hi,

add `IntersectionObserver, IntersectionObserverEntry` declarations.

Docs from:
https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API  
https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp  
https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -170,6 +170,10 @@ declare class PluginArray {
     [key: number | string]: Plugin;
 }
 
+// https://www.w3.org/TR/hr-time-2/#dom-domhighrestimestamp
+// https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp
+declare type DOMHighResTimeStamp = number;
+
 // https://www.w3.org/TR/navigation-timing-2/
 declare class PerformanceTiming {
     connectEnd: number;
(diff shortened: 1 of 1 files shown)


billschaller:
@pascalduez Thanks!

lib/bom.js
```diff
@@ -362,6 +366,55 @@ declare class MutationObserver {
...
+declare type IntersectionObserverOptions = {
+    root?: Node | null,
+    rootMargin?: string,
+    threshold?: number | Array<number>,
+};
+
+declare class IntersectionObserver {
+    constructor(
+      callback: IntersectionObserverCallback,
+      options: IntersectionObserverOptions
```
julienw: The second argument is actually optional, see https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver

lib/bom.js
```diff
@@ -362,6 +366,55 @@ declare class MutationObserver {
...
+    top: number,
+    right: number,
+    bottom: number,
+    left: number,
+    constructor(x: number, y: number, width: number, height: number): void,
+    fromRect(rectangle?: { x?: number, y?: number, width?: number, height?: number }): DOMRect,
+}
+
+declare class DOMRect extends DOMRectReadOnly {
+    constructor(x: number, y: number, width: number, height: number): void,
```
julienw: These are all optionals too, according to https://drafts.fxtf.org/geometry/#DOMRect

pascalduez: Thanks @julienw, I will give it another look.

lib/bom.js
Hypnosphi: There is an open PR fixing this: https://github.com/facebook/flow/pull/4664

Issue #420, opened 2015-04-26
Why is the `&` operator called an intersection?


popham:
I had to check out the tests to figure out what the intersection operator does.  My intuition says that an intersection of two types should yield a type that is of equal or greater strength than both operands.  The `&` operator smells more like a sum to me.  I guess it doesn't matter, but a `{x:number}&{y:number}` example in the docs would have saved me some aggravation.

It's probably too late, but I would expect an intersection operator to provide `(string | number) \cap (string | SomeClass) = string` and `{x: number | string, y: string} \cap {x: string, y: number} = {x: string}`.  I could use this intersection operator as a workaround (`type Null = ?string \cap ?number`), but it could be useless in practice.  I mention the alternative in case it justifies deprecating the `&` symbol in favor of a `+` symbol.  Then an intersection (`\cap` or something else) could take the `&` symbol someday.

avikchaudhuri:
@popham `&` indeed means intersection in the sense you describe, but our support for intersection types right now is pretty hacky. I'm working on making them better.

To elaborate on your intuition: if you think of a type as describing a set of values, an intersection type describes an intersection of sets of values. The type `{ x: number }` represents the set of objects that have at least property `x: number` (could have more, via structural subtyping), and similarly, `{ y: number}` the set of objects having at least `y: number`. So their intersection is the set of objects that have at least `x: number` and `y: number`, which is described by the type `{ x: number, y: number }`. This is a common source of mindfuck, but makes sense when you realize that "adding more properties" actually makes an object type more restrictive (i.e., "smaller").

popham:
Ah.  Thanks for the explanation.  I parse your explanation to the following:

```
{x: number} => {js objects s.t. there exists a numeric property x}
{y: number} => {js objects s.t. there exists a numeric property y}
{x: number} & {y: number} =>
    {js objects s.t. there exists both numeric properties x and y} = {x: number, y: number}
```

strangemonad:
I know this issue is old but I was just running through the math in my head to convince myself of the same thing so I thought I’d provide an alternative explanation while it’s fresh in my head since others might search old bugs for explanations. (I’ll try and come up with a nicer explanation for the docs).

The set of values interpretation is correct but didn’t help my brain grasp why the intersecion type has all properties (my initially wrong assupmtion was that shouldn’t it be a restriction of the properties). 

The alternative, order-theoretic, interpretation is to think of the subtyping relation that must hold. Assuming from above:
```
type A = {x: number}
[...]

PR #4223, opened 2017-06-22
[docs] Correct description of subtype relation

joelochlann:
Having read this carefully, I think there's a typo - since it's `ObjectB` that contains all the keys of `ObjectA`, `ObjectB` is the subtype and `ObjectA` is the supertype.

Of course, being new to Flow, it's also possible I misunderstood something so please correct me if I'm wrong.

(And apologies for the branch name - I edited this quickly in the web UI without downloading the repo, and as far as I can tell there's [still no support](https://github.com/isaacs/github/issues/228) for supplying a custom branch name when doing it that way)

### DIFF
--- a/website/en/docs/lang/subtypes.md
+++ b/website/en/docs/lang/subtypes.md
@@ -82,7 +82,7 @@ the keys of another object, then it may be a subtype.
 
 For example, if we had an `ObjectA` which contained the key `foo`, and an
 `ObjectB` which contained the keys `foo` and `bar`. Then it's possible that
-`ObjectA` is a subtype of `ObjectB`.
+`ObjectB` is a subtype of `ObjectA`.
 
 ```js
 // @flow


popham:
Sorry, dude, the original's correct. Given `a = { foo: string }` and `b = { foo: string, bar: number }`, `b` can be substituted for `a`, not vice versa. "Can be subsitituted for" implies "is subtype of" comes from https://en.wikipedia.org/wiki/Liskov_substitution_principle.

Reminds me of a goof of mine: https://github.com/facebook/flow/issues/420 :).

joelochlann:
Thanks for having look at this Tim. Do you have any idea why I got your
comment via email, but can't see it on the PR on GitHub?

popham:
I deleted it after I realized that I was mistaken. Apologies for the ping.

joelochlann:
Is this OK to merge then?

gabelevi:
You're absolutely right! Thanks for finding and fixing this mistake! And sorry for the delay in merging!

PR #4230, opened 2017-06-23
Remove forEach method from NodeList

gustavpursche:
According to MDN [`forEach()` isn't in any Spec](https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach). It's only implemented by some Vendors, to be nice to us developers. Even though there is a wide adoption, this shouldn't be in Flow.

At the guardian we are trying to [align the flow behavoir with polyfill.io](https://github.com/Financial-Times/polyfill-service/issues/1233) so that all developers can rely on the type definitions being eventually polyfilled.

Opinions?

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -479,7 +479,6 @@ declare class NodeList<T> {
   item(index: number): T;
   [index: number]: T;
 
-  forEach(callbackfn: (value: T, index: number, list: NodeList<T>) => any, thisArg?: any): void;
   entries(): Iterator<[number, T]>;
   keys(): Iterator<number>;
   values(): Iterator<T>;


gustavpursche:
Sorry for the test failures. I can't see how they relate to the actual change. Do you have a hint how to fix them?

popham:
Note the different line numbers in the error messages. You removed a line and that broke all of the tests whose error messages reference lines after your alteration.

I would add a blank line at your removal point and run the tests to remove all of the line numbering noise so that you can see what tests your changes have broken. Once you understand the implications of your changes and have repaired the tests, then the `runtests.sh` shell program has a `-r` flag for updating the tests' `.exp` files (execute `.runtests.sh --help` for specifics).

gustavpursche:
@popham Thanks for the help! I checked the tests with the empty line instead and it seems no tests are affected ✨. Regenerated the `.exp` file for `dom`.

gabelevi:
Hmm, this is a tricky situation. According to the linked spec, there is a candidate recommendation spec which gives all iterables in the webidl spec, including `NodeList`, a `forEach` method. When new things show up in the spec, we could in theory make them nullable, so that people have to make sure that they exist before using them. In practice, though, we don't.

So I'm leaning against taking this PR, since this feature is in the candidate spec.

What do you think?

mislav:
NodeList is [defined as "iterable"](https://dom.spec.whatwg.org/#interface-nodelist). Per https://heycam.github.io/webidl/#es-forEach (emphasis mine):

> If the interface has any of the following:
>
> - an **iterable declaration**
> - a maplike declaration
> - a setlike declaration
>
> then **a property named “forEach” must exist** with attributes { [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true } and whose value is a function object.

Issue #4204, opened 2017-06-19
What is the best way to express an inner class?


Ailrun:
I'm working with [openlayers](http://openlayers.org/) typing.
There is following inner class in code of _openlayers_.

``` javascript
// ol.Collection class
ol.Collection = function (/* ... parameters ... */) {
  // ... class construction codes ...
}
// ... some methods

// ol.Collection.Error class
ol.Collection.Error = function (/* ... parameters ... */) {
  // ... class construction codes ...
}
```

I try some typescript-like, C++-like, ... methods, (inner module, static inner class, ....) but no one works.
For example,

``` javascript
declare module 'ol' {
  declare module 'Collection' {
    declare class Error {
      // ....
    }
  }
}
```

is rejected by flow.

Is there any clean choice for this case? If it is, could you give me some direction for typing this one?

popham:
`Error` is a static member of class `Collection`, yeah? Try

```js
declare module 'ol' {
  declare class _Error { /*...*/ }
  declare class Collection {
    static Error: _Error;
    constructor( /*...*/ ): void;
    // ... some methods
  }
}
```

Ailrun:
@popham
At that case, does flow allow to use `new` keyword on `ol.Collection.Error`?

popham:
Yup. Just beware that many pre-ES6 libraries don't implement ES6's class behavior. In particular I'm thinking of the prototype chaining on static methods that ES6 provides with its classes (if your library "extends" `Collection`, then the library may not propagate the `Error` static member to the extension classes, while the ES6 semantics of Flow will expose the `Error` static member on extension classes).

popham:
Whoops. I just fixed a goof in my earlier code by wrapping a `Class<.>` around `_Error` at the static `Error` field's annotation.

Ailrun:
@popham Thx!

popham:
Sure. By the way, the give-away for pre-ES6 libraries that do implement ES6 semantics is `Subclass.__proto__ = Superclass` for a `Subclass` that extends `Superclass` (in addition to the normal prototype chaining).

Ailrun:
@popham What I notice is they (_openlayers_) use `Collection` as a namespace, not just a class,
I mean,

`ol.Collection` class extends (of course they use their own implementation, not ES6 class) `ol.Object` class, and `ol.Object` also has `ol.Object.Error`.

[...]

PR #4243, opened 2017-06-25
Expose pre-ES6 inheritance through `inherits` on `declare class...`

popham:
See https://github.com/facebook/flow/issues/4204#issuecomment-309883477 for the problem. See https://github.com/facebook/flow/issues/4204#issuecomment-310235076 for the hackish workaround that relies on a bug (in my opinion).

### DIFF
--- a/src/parser/ast.ml
+++ b/src/parser/ast.ml
@@ -369,6 +369,7 @@ and Statement : sig
   module Interface : sig
     type t = {
       id: Identifier.t;
+      es6: bool;
       typeParameters: Type.ParameterDeclaration.t option;
       body: Loc.t * Type.Object.t;
       extends: (Loc.t * Type.Generic.t) list;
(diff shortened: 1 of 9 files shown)


popham:
I don't know the policy on updating parser tests. Assuming y'all are open to pulling these changes, should I create a separate PR to `ast-types` and widen this PR to update the tests and bump the version on the `ast-types` dependence?

gabelevi:
Hey @popham, sorry for the delayed response. I don't think we'd like to take a non-standard feature like `inherits`. It's a little too close to `extends` and only really seems useful in a few cases.

I wonder, could the code in question use interfaces instead? And if you're just using `declare class`, there is a `mixins` keyword that [looks like this](https://github.com/facebook/flow/blob/dd0603e9d8c9d5fb99a40f0b179a4d6a2b9e66b7/tests/traits/test.js) and let's you copy a definition in from one declared class into another.

popham:
I would quibble over the "useful" word choice. It's *necessary* in only a few cases, but I suspect that there's a bunch of flow declarations out there that slightly mis-model the associated libraries. Given pre-ES6 inheritance where statics exist on the base class, the derived class mistakenly exposes the base's statics. This feature would certainly find common use in such cases.

The issue referenced in my initiating comment needs a feature like this because two static field names collide on a base and a derived class without subtyping. That issue's OP is kinda up a creek without something like this PR. Is there a possibility for a compromise with an ugly name instead of `inherits` maybe?

Issue #4265, opened 2017-06-28
A tuple allows `length` setting which can mutate it


popham:
Judging from the tuple docs' "You cannot use `Array.prototype` methods that mutate the tuple, only ones that do not," I suspect that `length` setting should be forbidden. The following passes without error (see the note from the spec's 22.1.4.1):

```js
declare a: [number];
a.length = 0;
(a: [number]); //ng, but Flow passes
```

popham:
Tuple uses $ReadOnlyArray, so #4266 fixes this (see https://github.com/facebook/flow/blob/49ca0a17cedca8f8b3c056fab9548480302bbc42/src/typing/flow_js.ml#L5519-L5523)

Issue #4246, opened 2017-06-26
Array#includes check with enum


deecewan:
Hey,

So, I'm pretty sure that this will have a fairly logical solution, but I've been trawling through bugs and stack overflow and, I can't find a) a solution, and more importantly b) an explanantion.

I'm trying to validate that an item is part of an `enum`.

```js
type StringEnum = 'first_option' | 'second option' | 'third option'

const options: Array<StringEnum> = ['first_option', 'second option', 'third option']

function checkInEnum(item: string): boolean {
     if (options.includes(item)) {
        console.log('it\'s a valid option')
        return true;
     }
     return false;
}

checkInEnum('some item')
checkInEnum('first_option')
```

However, in the `includes` call, flow gets mad because `item` is not a valid subset of `StringEnum`. Which I (probably) know, and I'm trying to check for.

Is there a nice way to fix this?

[Flow Try](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgDKGATgJYB2A5gKIUCuAtmALxgDkUZJAzhgPpwcGMnArswAHw488AYzEATMEJFiJ09hgAW3ZatHj0CinxXDDPAFxgAgiRIBDLAB5i5anSYA+VmADanNx8ghbqADQy8krmauKRWrok+mHiALroUPQUcnFgctryANYAkhRejAAUZBh4jDZ8HlQAlDYARnBwMHiOFGAA3qhgw8NkUGCVBmI8AHSUcjD0ing81bWMzc0DQyO7JjxdeDPwVJXsNQA67DxgjmAAbo4wZClx7M07u8MkeBj0JH1SPQ8ABuT7DAC+4LAPz+ALAUCesjBUNQBWKZQqZwOjAINTq7zRhTkpXKDCqQV4Aim4maQA)

niieani:
Use: `checkInEnum(item: StringEnum)` instead. This way you'll catch incorrect usage of `checkInEnum` build-time, rather than in runtime. This kinda makes the function not necessary. 

Why would you want to do the check otherwise? Are you receiving the value from a backend?

deecewan:
that was actually a terrible example, sorry.

a better function signature would be:

```js
function normalizeValue(item: string) {
  if (options.includes(item)) {
    return item;
  }
  return item.split(' ').join(',');
}
```

[...]

PR #4266, opened 2017-06-28
Covariant length on ReadOnlyArray<.>

popham:
$ReadOnlyArray shouldn't allow `length` setting because it can mutate the array.


vkurchatkin:
I personally would prefer to have `length` covariant even on mutable arrays

popham:
@vkurchatkin Sure, although I recall reading something about Flow striving toward "JavaScript as it is." Constraining `length` on `Array` seems more in line with "JavaScript as it ought to be." I personally agree with you because I prefer false negatives to false positives (https://github.com/facebook/flow/issues/4246#issuecomment-311188651 was another instance of "JavaScript as it ought to be" in tension with "JavaScript as it is").

Can I get a "shall" directive from the Flow hive at FB? I was too lazy to look it up on the spec, but I noticed that the Node interpreter admits `length` assignments on strings by ignoring them:

```js
const a = "asdf"; //a.length===4
a.length = 1; //a.length===4
a.length = 7; //a.length===4
```

If `length` assignment on `Array` is forbidden, then it seems reasonable that it should be forbidden on `String` also.

vkurchatkin:
`length` mutation on arrays is definetely observable and might even be considered useful. On strings it should do nothing, so that is a no-brainier

popham:
Agreed. The PR is good as-is.

nmote:
Looks reasonable to me. I'd like to see some tests to ensure that (a) `length` mutations are in fact banned on `$ReadOnlyArray`s, and that (b) `length` mutations are still allowed on `Array`s.

popham:
Sure. This closes #4265 BTW.

PR #4291, opened 2017-06-30
feat: add lib defs for the node repl module

zacharygolba:
Adds lib defs for the [repl](https://nodejs.org/dist/latest-v6.x/docs/api/repl.html) module.


nmote:
Looks good, could you update the expected test output so the tests pass? Should just be able to run `make test` then copy `node_tests.exp` to `node_tests.out` (of course, after verifying that the changes look reasonable).

zacharygolba:
😃 Thanks! I just updated `node_tests.exp`. The changes look reasonable. Only line numbers have been modified.

PR #4293, opened 2017-07-01
Extend cluster from events$EventEmitter

### LINKED ISSUE #4019: Native node module method not recognised

Corei13:
https://github.com/nodejs/node/blob/master/lib/internal/cluster/master.js#L11
Fixes #4019

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -327,23 +327,27 @@ type cluster$setupMasterOpts = {
 }
 
 declare module "cluster" {
-  declare var isMaster: boolean;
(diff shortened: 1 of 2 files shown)


nmote:
Looks good, could you just update the expected test output? Looks like it should be just the `node_tests` one, the other failures in CI look spurious.

Corei13:
@nmote Updated tests

PR #4322, opened 2017-07-05
Fixed group of different possible types example

Andarist:
boolean type cannot be converted automatically to string by concatenation ('' + true)


mroch:
thanks!

#4321 fixed this by removing the `boolean` from the union, just an hour before, so going to close this. but if you think changing from + to String makes the example cleaner, feel free to rebase and reopen!

PR #4334, opened 2017-07-07
[docs] Fix symbol declaration typo

roya3000:
According to ECMAScript 2015 Language Specification symbols do not support the syntax "new Symbol()".


chyzwar:
well this is still invalid, there is no constructor for symbols
Symbol("foo") return primitive data type of symbol.
Object(Symbol("foo")) would return symbol wrapped in object

https://developer.mozilla.org/pl/docs/Web/JavaScript/Reference/Global_Objects/Symbol

Issue #368, opened 2015-04-07
Correct way to type React events


gcazaciuc:
I have the following code:

 handleChange: function(event:SyntheticEvent): void {
         var el = event.target;
         this.props.setNewLanguage(el.value);
    }

Can you please advise on the correct way of typing this function so that it doesn't throw a type error accessing el.value.

I've found that using an union type works var el:HTMLInputElement|EventTarget but I'm not convinced that this is the correct way to type it. 
Typing it as only var el:HTMLInputElement = event.target throws a type error.

samwgoldman:
You could cast `event.target` to be a `HTMLInputElement` or you could do a [dynamic type test](http://flowtype.org/docs/dynamic-type-tests.html), also called a type refinement to see if `event.target` is-a `HTMLInputElement` and handle the else clause as an error.

gcazaciuc:
Thank you very much for the suggestions. Using dynamic type test makes sense to me but  I'm not quite clear  on how to do the cast( and haven't found a suggestion in the docs).

samwgoldman:
Turns out I was wrong about casting (there's a blog post describing it). Dynamic type test is the way to go, IMO.

samwgoldman:
@gcazaciuc are you still having this issue?

gcazaciuc:
It was resolved by checking if the event.target is instanceof HTMLElement before accesing data on it. It seems like flow is smart enough to detect than inside that if statement event.target must be of type HTMLElement. Thanks for your support( and looking forward for let/const in flow :)  )

rosskevin:
Relates to #218 

I'd hate to have to sprinkle this dynamic type test code into every event handler:

```javascript
function myEventHandler(e: Event) {
  // We only know that e.target is an EventTarget
  e.target.value;
  if (e.target instanceof HTMLInputElement) {
    // Now we know it's an <input />, with a `value` property.
    businessLogic(e.target.value);
  } else {
    // error handling
  }
}
```

Is there a way I can define my own `type` that extends `Event` and sets a static expectation of `HTMLInputElement`?

rosskevin:
Here is my current dynamic polymorphic method hack to shorten up usage:

```javascript
// events.js
export function target<E> (e:Event, t:E):E {
  if (e.target instanceof E) {
    // Now we know it's an <input />, with a `value` property.
    return e.target
  } else {
    throw new Error(`Expected Event target to be ${E} but found ${typeof e.target}`)
  }
}
```

Called like:
```javascript
[...]

PR #4344, opened 2017-07-10
Use SyntheticEvent in React event docs.

damian:
This is more readable and accurate(i.e. Flow doesn't give a type error) as demonstrated [here](https://github.com/facebook/flow/issues/368#issuecomment-289219832).


mrkev:
This is no longer necessary per the new docs: https://flow.org/en/docs/react/events/

PR #4392, opened 2017-07-15
Simplification of example

cPlayIt:
- Easier to skim over examples if variable names are self-describing. 
- May be a good convention to make the documentation more readable. 
- Distracting for the example to use 'method' for a function name when it's not a method, here it would be more easier to read if 'func' or 'f' was used.


jbrown215:
Totally agree! Thanks for the change

PR #4393, opened 2017-07-15
Example code update

cPlayIt:
- Easier to skim over examples if variable names are self-describing. 
- May be a good convention to make the documentation more readable. 
- Distracting for the example to use 'method' for a function name when it's not a method, here it would be more easier to read if 'func' or 'f' was used.

------------------------------

Still learning and wasn't sure exatcly what the bottom 2 line of code was trying to show. I believe this may be along the lines. Is my usage of subtype correct in the comments?

### DIFF
--- a/website/en/docs/lang/subtypes.md
+++ b/website/en/docs/lang/subtypes.md
@@ -117,8 +117,24 @@ Flow compares two functions by comparing its inputs and outputs. If all the
 inputs and outputs are a subset of the other function, then it is a subtype.
 
 ```js
-type Func1 = (1 | 2)     => "A" | "B";
-type Func2 = (1 | 2 | 3) => "A" | "B" | "C";
+// @flow
+type LessSpecificReturn = (1 | 2 | 3) => "A" | "B" | "C";
+type MoreSpecificReturn = (1 | 2 | 3 ) => "A" | "B";
+
+type LessSpecificParam = (1 | 2 | 3 ) => "A" | "B" | "C";
+type MoreSpecificParam = (1 | 2 ) => "A" | "B" | "C";
+
+var func1: MoreSpecificReturn = (num) => 'B';
+var func2: LessSpecificReturn = func1; // Less specific return is OK! (Subtype)
+
+/* More specific return is not a valid Subtype
+   .. otherwise you may return a type that cannot be handled */
+
+var func3: LessSpecificParam = (num) => "B";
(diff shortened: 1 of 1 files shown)


website/en/docs/lang/subtypes.md
```diff
@@ -117,8 +117,24 @@ Flow compares two functions by comparing its inputs and outputs. If all the
...
+type MoreSpecificParam = (1 | 2 ) => "A" | "B" | "C";
+
+var func1: MoreSpecificReturn = (num) => 'B';
+var func2: LessSpecificReturn = func1; // Less specific return is OK! (Subtype)
+
+/* More specific return is not a valid Subtype
+   .. otherwise you may return a type that cannot be handled */
+
+var func3: LessSpecificParam = (num) => "B";
+var func4: MoreSpecificParam = func3; // More specific param is OK! (Subtype)
```
jbrown215: // func3 is a subtype of MoreSpecificParam since func3 has more general params than MoreSpecificParam and the same return type

jbrown215: I think it might be easier to understand if we talk about the subtype always being "more" of something. So the params of the subtype are more general and the return type is more specific.

website/en/docs/lang/subtypes.md
```diff
@@ -117,8 +117,24 @@ Flow compares two functions by comparing its inputs and outputs. If all the
...
-type Func2 = (1 | 2 | 3) => "A" | "B" | "C";
+// @flow
+type LessSpecificReturn = (1 | 2 | 3) => "A" | "B" | "C";
+type MoreSpecificReturn = (1 | 2 | 3 ) => "A" | "B";
+
+type LessSpecificParam = (1 | 2 | 3 ) => "A" | "B" | "C";
+type MoreSpecificParam = (1 | 2 ) => "A" | "B" | "C";
+
+var func1: MoreSpecificReturn = (num) => 'B';
+var func2: LessSpecificReturn = func1; // Less specific return is OK! (Subtype)
```
jbrown215: // MoreSpecificReturn is a subtype of LessSpecificReturn since the params are the same and the return type is a subtype of LessSpecificReturn

jbrown215:
The examples were updated in a way that better explains what's going on in https://github.com/facebook/flow/pull/4635

Closing this for now. If you feel they could be further clarified, feel free to rebase and update the examples :)

PR #4416, opened 2017-07-19
Node.js: add rawHeaders to http$IncomingMessage

sstur:
Add support for rawHeaders as per Node docs:

https://nodejs.org/api/http.html#http_message_rawheaders


sstur:
The build is failing but I don't think that's related to my changes.

PR #4420, opened 2017-07-19
Add @flow mention in module

jdeniau:
`@flow` is mandatory in `export.js` files otherwise flow just do not report errors.


pronebird:
It's implied that `// @flow` should follow each file processed by flow, this is mentioned in Getting Started section of README.md

PR #4433, opened 2017-07-21
Add FreeBSD support

Parakleta:
This set of changes makes it possible to compile `flow` on FreeBSD.

It does require `libinotify` to be install separately, but otherwise seems fine.  I'm not sure how to make the local `libinotify` a dependency through the `opam` build system.

Additionally, the current install instructions in `README.md` specify that `opam init --comp 4.03.0` is sufficient to change versions but on FreeBSD it is necessary to use `opam switch 4.03.0` since the `init` subcommand will not replace the system version (currently 4.02.3 on FreeBSD) which is automatically installed as a dependency of `opam`.

### DIFF
--- a/hack/heap/hh_shared.c
+++ b/hack/heap/hh_shared.c
@@ -162,7 +162,6 @@
     #endif
   #endif
 
-  #define MEMFD_CREATE 1
   #include <asm/unistd.h>
 
   /* Originally this function would call uname(), parse the linux
@@ -547,12 +546,6 @@ void memfd_init(char *shm_dir, size_t shared_mem_size, uint64_t minimum_avail) {
 
 static int memfd = -1;
 
-static void raise_failed_anonymous_memfd_init() {
-  static value *exn = NULL;
-  if (!exn) exn = caml_named_value("failed_anonymous_memfd_init");
-  caml_raise_constant(*exn);
-}
-
 static void raise_less_than_minimum_available(uint64_t avail) {
   value arg;
   static value *exn = NULL;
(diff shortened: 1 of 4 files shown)


Parakleta:
I don't think the appveyor failure is mine?

samwgoldman:
@gabelevi this is not really my area of expertise. Can you take a look?

hack/utils/sysinfo.c
```diff
@@ -13,11 +13,9 @@
 #include <caml/mlvalues.h>
 
 #include <assert.h>
-#ifndef _WIN32
-#ifndef __APPLE__
+#ifdef __linux__
```
gabelevi: LGTM, since the functions in this file are no-ops for non `__linux__`

hack/third-party/inotify/inotify_stubs.c
```diff
@@ -28,7 +28,9 @@
 #include <caml/callback.h>
 #include <caml/unixsupport.h>
 
+#ifdef __linux__
 #include <features.h>
```
gabelevi: This looks like dead code. Looking at https://github.com/facebook/hhvm/commit/d50dd15bc34c60a4bcc02dc6bd95fdf32effc54b (from ages ago), it was doing something like

```
#if __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 4
```

so I'll delete this include when merging.

hack/heap/hh_shared.c
```diff
@@ -176,6 +176,10 @@
   }
 #endif
 
+#ifndef MAP_NORESERVE
+  #define MAP_NORESERVE 0
```
gabelevi: Was wondering what this was doing, and found https://reviews.llvm.org/D6953 which explained that FreeBSD 11.x removes this flag, and it was a no-op before that. Makes sense. I'll add a comment.

hack/heap/hh_shared.c
```diff
@@ -148,7 +148,7 @@
  * with the MAP_ANONYMOUS flag. The memfd_create() system call first
  * appeared in Linux 3.17.
  ****************************************************************************/
-#if !defined __APPLE__ && !defined _WIN32
+#ifdef __linux__
```
gabelevi: If you look below where `MEMFD_CREATE` is checked, you'll see the non-apple case will fall through to a pretty terrible create-a-file-on-tmpfs code. FreeBSD seems to have [`shm_open` and `shm_unlink`](https://www.freebsd.org/cgi/man.cgi?query=shm_open&sektion=2&manpath=FreeBSD+8.4-RELEASE) so I wonder if they can share the apple implementation, which also uses those APIs.

But this is fine for now. Falling through to a terrible implementation is better than not working at all!

gabelevi:
Oh and thanks for this PR! Really appreciate it!

hack/heap/hh_shared.c
Parakleta: Well those APIs are just posix, so it's fine.  FreeBSD has an extension flag `SHM_ANON` which makes an anonymous map automatically but I don't think it's worth diverging for that.  I'll update the patch to make use of the posix variant.  In general your system detection flags should probably be along the lines of:

```
#ifdef _WIN32
  /* Windows */
#else
  /* Unix-ish (POSIX&Linux) */
  #ifdef __linux__
    /* Linux-isms */
  #else
    /* POSIX */
    #if defined(__APPLE__)
    #elif defined(__FreeBSD__)
    #elif defined(/* other posix variant */)
    #end
  #end
#end
```

Parakleta: OK, new version is up.

Parakleta: Reading back on the penultimate comment it may not have been clear, but what I was getting at was preferring `#ifndef __linux__` over `#ifdef __APPLE__` as a test, since generally the BSDs are close enough that you can cover them all together in the `#ifndef __linux__` branch.

Parakleta:
There was an additional patch added later that didn't make the merge, will this be handled separately, or should I make a new PR, or is it not wanted (it is only really a small clean-up and minor optimisation)?

PR #4446, opened 2017-07-23
HTMLMediaElement#play() return Promise

umanoda:
# Problem

`HTMLMediaElement#play()` return Promise object on browser.

ref. https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play

but the following code will result in an type error

```js
function playMedia(media: HTMLMediaElement): Promise<any> {
  return media.play();
}
```
```
Error: eample.js:5
  5:   return media.play();
              ^^^^^^^^^^^^ undefined. This type is incompatible with the expected return type of
  4: function playMedia(media: HTMLMediaElement): Promise<any> {
                                                  ^^^^^^^^^^^^ Promise
```

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -2679,7 +2679,7 @@ declare class HTMLMediaElement extends HTMLElement {
   ended: boolean;
   autoplay: boolean;
   loop: boolean;
-  play(): void;
+  play(): Promise<void>;
   pause(): void;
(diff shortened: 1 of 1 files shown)


pronebird:
`HTMLMediaElement#play` does not return any result with `Promise`, therefore should be `Promise<void>`

PR #4478, opened 2017-07-28
Turn on strict call arity everywhere by default.

### LINKED ISSUE #4324: Strict function call arity check not working
Hi, it seems that strict function call arity check introduced in version [`0.47.0`](https://flow.org/blog/2017/05/07/Strict-Function-Call-Arity/) doesn't work with later versions (`>= 0.48.0`).

Example code:
```
function add(a: number, b: number) {
  return a + b;
}

add(10, 15, 20);
```
It doesn't produce any errors with version different than `0.47.0`.

Try flow example:
https://flow.org/try/#0PTAEAEDMBsHsHcBQjIFcB2BjALgS1uqAIYAmJAFEQFyjqoC2ARgKYBOANKIzXU2wJSgA3olChWzbKlaEioANRcA3IgC+yUhQCMABk5aArJwBMO-kqA

Any ideas?

stevemolitor:
Fixes #4324. 'Try flow' (flow.org/try) has it off, which is confusing.


samwgoldman:
thank you!

Issue #3714, opened 2017-04-12
node "fs" lib definitions: optional parameter "mode" is not optional to node (due to variable nr. of params?)


lll000111:
Test code:

```javascipt
// @flow
'use strict';

const fs = require('fs');

fs.open('filename', 'r', (err: ?ErrnoError, fd) => {
    // do stuff
});

fs.open('filename', 'r', 0o666, (err: ?ErrnoError, fd) => {
    // do stuff
});
```

The second `fs.open` works, but the first one without (a superfluous) `0o666`  mode parameter leads to this error:

```
test.js:6
  6: fs.open('filename', 'r', (err: ?ErrnoError, fd) => {
                              ^ arrow function. This type is incompatible with the expected param type of
777:   declare function open(path: string | Buffer, flags: string | number, mode?: number, callback?: (err: ?ErrnoError, fd: number) => void): void;
                                                                                   ^^^^^^ number. See lib: ../../../AppData/Local/Temp/flow/flowlib_3a866f4f/node.js:777


Found 1 error
```

I don't quite know how to fix the definitions, after all, `mode` is defined as optional already but it doesn't help. It seems Flow has a problem here when the number of parameters and with it their meaning is variable. Is this even expressible?

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

Issue #3081, opened 2016-12-25
Wrong positive in fs.writeSync(fd, string[, position[, encoding]])


doberkofler:
I do not understand why flow report an error in `fs.writeSync(fd, string[, position[, encoding]])` given the [definition](https://nodejs.org/api/fs.html#fs_fs_writesync_fd_string_position_encoding)  in the nodejs documentation.

Example:
```
// @flow

const fs = require('fs');

let fd = fs.openSync('file.txt', 'w');
fs.writeSync(fd, 'hello world');
fs.closeSync(fd);
```

Error:
```
index.js:6
  6: fs.writeSync(fd, 'hello world');
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `writeSync`
  6: fs.writeSync(fd, 'hello world');
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
777:   declare var writeSync: (fd: number, buffer: Buffer, offset: number, length: number, position?: number) => number
                                                                   ^^^^^^ number. See lib: C:\temp\flow\flowlib_15ce7844\node.js:777

index.js:6
  6: fs.writeSync(fd, 'hello world');
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `writeSync`
  6: fs.writeSync(fd, 'hello world');
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ undefined (too few arguments, expected default/rest parameters). This type is incompatible with
777:   declare var writeSync: (fd: number, buffer: Buffer, offset: number, length: number, position?: number) => number
                                                                                   ^^^^^^ number. See lib: C:\temp\flow\flowlib_15ce7844\node.js:777

index.js:6
  6: fs.writeSync(fd, 'hello world');
                      ^^^^^^^^^^^^^ string. This type is incompatible with the expected param type of
777:   declare var writeSync: (fd: number, buffer: Buffer, offset: number, length: number, position?: number) => number
                                                   ^^^^^^ Buffer. See lib: C:\temp\flow\flowlib_15ce7844\node.js:777
```

Issue #2902, opened 2016-11-27
fs.write and fs.writeSync have incorrect signature


dignifiedquire:
The signatures defined in https://github.com/facebook/flow/blob/master/lib/node.js#L712-L715 are not correct according to https://nodejs.org/docs/latest/api/fs.html#fs_fs_write_fd_buffer_offset_length_position_callback

Both `offset` and `length` are optional parameters, but not defined as such

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

PR #4493, opened 2017-07-31
Fix library definitions for some node `fs` functions

### LINKED ISSUE #3992: Incorrect types for fs.write and fs.writeSync
Should be overloads instead of unions

adanoff:
Fix definitions for:
- write(Sync)
- readFile(Sync)
- readdir(Sync)
- readFile(Sync)
- appendFile(Sync)
- open
- close

I believe this closes #3992, #3714, #3081, and #2902.

**Note:** Some of these changes are breaking, as they change some `callback` arguments from optional to required (as indicated in the node documentation).e

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -777,11 +777,32 @@ declare module "fs" {
   declare function mkdirSync(path: string, mode?: number): void;
   declare function mkdtemp(prefix: string, callback: (err: ?ErrnoError, folderPath: string) => void): void;
   declare function mkdtempSync(prefix: string): string;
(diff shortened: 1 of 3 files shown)


adanoff:
It seems like some of the callbacks are required though. For example, calling `fs.readFile` without a callback results in a deprecation warning from node. I suppose there's an argument to be had about whether that truly means it's required, but I think it's fair to say a callback should definitely be provided for that function (otherwise, you can't get the data!). Do you think it would be best to have some functions require the callback and others not?

PR #4496, opened 2017-07-31
Make String.replace callback args less unassuming

atk:
The callback in `String.prototype.replace(regex, callback)` will always receive the following arguments:

* 0: full match as string,
* 1-n: single matches from parantheses inside the regex (no match will return empty string)
* n+1: start position of full match as number
* n+2: full string

So `(substring: string, ...args: Array<any>)` is too unassuming.

In the future, some logic that was able to detect parantheses in RegExp and computed the callback arguments accordingly would be desirable.


nmote:
Looks like the CLA was never signed, feel free to reopen if this changes.

PR #4521, opened 2017-08-03
Fix link to flowlint-comments

TiddoLangerak:
The current link is incorrect and ends up on a 404. This change should fix that.


ballardrog:
Whoops. Thanks for the fix!

PR #4523, opened 2017-08-03
fix(docs): Link to IDE Integration page

nodkz:


ballardrog:
Whoops. Thanks for the fix!

PR #4527, opened 2017-08-03
Provide more helpful values for Object.entries() and Object.values()

jcready:


jcready:
I'm not familiar with how the tests work, and I honestly cannot figure out what I broke in the tests from Travis-CI output. If anyone can point me in the right direction I'd be more than happy to adjust any test expectations to get these tests passing.

vkurchatkin:
Unfortunately, this is very unsound

jcready:
Why is that?

vkurchatkin:
Here is a simple example: 

```js
function test(val: { foo: string }) {
  Object.values(val).forEach(v => {
  	v.toLowerCase();
  });
}

test({ foo: 'x', bar: 123 });
```

This going to pass typechecking, but fail at runtime

jcready:
Oh, blerg. That is truly unfortunate.

samwgoldman:
We're slowly getting Flow to a place where we can exploit exact objects to provide more specific types for own- and width-sensitive APIs, like Object.values, object spread properties, etc. Rest assured that this is my primary focus right now.

Issue #4543, opened 2017-08-04
Raise error when an undefined property is referenced on a type.


dmueller39:
Recently I was doing some refactoring of code that was covered by flow. The following type:

```
type Route = {
  name: ?string,
};
```

Was changed to this:

```
type Route = {
  routeName: ? string
}
```

My expectation was that it would ensure that the property `name` was not being referenced in the code. Things like this gave an error:

```
const route = { routeName: 'foo' };

// $ExpectError
const theName = route.name;
```

However this did not:
```
const route = { routeName: 'foo' };

// $ExpectError
if (route.name != null) {
  // $ExpectError
  const theName = route.name;
}
```

What would be ideal is a  lint rule that ensures that raises an error whenever a undefined member is referenced on a type.

I tested this out on try flow: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgBCAhgE5gC8YA3mFHHAFxgDOGpAlgHYDmYAvgG50AYzhc2YAEZlKNOg0YByYkoHDUUAK5cRGDuIVwAFAA9GJUgEpmbTrxqowYDlDBmAdDIBeYAIRUXFowMGAAZGFg2PhwbqZexL4UVEp23DxKVo7OzqR4GFqkXGDxPsLO-E5geQVFYABEACZ4UMTBGKzs6fXClUA

@calebmer

PR #4547, opened 2017-08-06
Added unknown_property_in_branch_test lint rule

dmueller39:
Rule fires when an unknown property is tested on a typed object. Resoves https://github.com/facebook/flow/issues/4543

### DIFF
--- a/src/common/lints/lintSettings.ml
+++ b/src/common/lints/lintSettings.ml
@@ -16,7 +16,7 @@ type sketchy_null_kind =
 type lint_kind =
   | SketchyNull of sketchy_null_kind
   | UntypedTypeImport
-  | UnknownProperty
+  | UnknownPropertyInBranchTest
 
 let string_of_sketchy_null_kind = function
   | SketchyBool -> "sketchy-null-bool"
@@ -27,7 +27,7 @@ let string_of_sketchy_null_kind = function
 let string_of_kind = function
   | SketchyNull kind -> string_of_sketchy_null_kind kind
   | UntypedTypeImport -> "untyped-type-import"
-  | UnknownProperty -> "unknown-property"
+  | UnknownPropertyInBranchTest -> "unknown-property-in-branch-test"
 
 let kinds_of_string = function
   | "sketchy-null" -> Some [
@@ -41,7 +41,7 @@ let kinds_of_string = function
   | "sketchy-null-number" -> Some [SketchyNull SketchyNumber]
   | "sketchy-null-mixed" -> Some [SketchyNull SketchyMixed]
   | "untyped-type-import" -> Some [UntypedTypeImport]
-  | "unknown-property" -> Some [UnknownProperty]
+  | "unknown-property-in-branch-test" -> Some [UnknownPropertyInBranchTest]
   | _ -> None
 
 type lint_state =
--- a/src/common/lints/lintSettings.ml
+++ b/src/common/lints/lintSettings.ml
@@ -16,6 +16,7 @@ type sketchy_null_kind =
 type lint_kind =
   | SketchyNull of sketchy_null_kind
   | UntypedTypeImport
+  | UnknownProperty
 
 let string_of_sketchy_null_kind = function
   | SketchyBool -> "sketchy-null-bool"
@@ -26,6 +27,7 @@ let string_of_sketchy_null_kind = function
 let string_of_kind = function
   | SketchyNull kind -> string_of_sketchy_null_kind kind
   | UntypedTypeImport -> "untyped-type-import"
+  | UnknownProperty -> "unknown-property"
 
 let kinds_of_string = function
   | "sketchy-null" -> Some [
@@ -39,6 +41,7 @@ let kinds_of_string = function
   | "sketchy-null-number" -> Some [SketchyNull SketchyNumber]
   | "sketchy-null-mixed" -> Some [SketchyNull SketchyMixed]
   | "untyped-type-import" -> Some [UntypedTypeImport]
+  | "unknown-property" -> Some [UnknownProperty]
   | _ -> None
 
 type lint_state =
--- a/src/typing/flow_error.ml
+++ b/src/typing/flow_error.ml
@@ -165,7 +165,7 @@ type error_message =
       null_loc: Loc.t;
       falsy_loc: Loc.t;
     }
-  | ENonstrictLookupFailed of (reason * reason) * reason * string option
+  | ENonstrictLookupFailed of reason * reason * string option
 
 and binding_error =
   | ENameAlreadyBound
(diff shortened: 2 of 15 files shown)


src/common/lints/lintSettings.ml
```diff
@@ -16,6 +16,7 @@ type sketchy_null_kind =
 type lint_kind =
   | SketchyNull of sketchy_null_kind
   | UntypedTypeImport
+  | UnknownProperty
```
calebmer: We need a better name then `UnknownProperty` since [we do error for unknown properties](https://flow.org/try/#0MYewdgzgLgBCBGArAXDA3gMxCVYCuAtvAKYBOAvjALzpY4wAsATOQNwBQ7CiAdPAJZgA5qxgB6MTACipUiFKd+GGAApufQUICU6ShJgA5ELACGYGGTkKgA). We don’t error specifically in branch tests. (Such as `if ()` or `while ()`.) How about `UnknownPropertyInBranchTest` and `unknown-property-in-branch-test`?

src/typing/flow_error.ml
```diff
@@ -165,6 +165,7 @@ type error_message =
       null_loc: Loc.t;
       falsy_loc: Loc.t;
     }
+  | EUnknownProperty of Loc.t * string
```
calebmer: Could we rename this error to `NonstrictLookupFailed`? This would be symmetrical with our error for strict lookup failures: [`EStrictLookupFailed`](https://github.com/facebook/flow/blob/dab35ec2fe38a83f2e6a25a5ba8aa082a26ef902/src/typing/flow_js.ml#L5471-L5476) (you might also want to look at [the message for `EStrictLookupFailed`](https://github.com/facebook/flow/blob/dab35ec2fe38a83f2e6a25a5ba8aa082a26ef902/src/typing/flow_error.ml#L773-L791)).

calebmer: It would probably be a good idea to copy the implementation for `EStrictLookupFailed` for `ENonstrictLookupFailed`. We’ll just output an `UnknownPropertyInBranchTest` lint.

src/typing/flow_error.ml
```diff
@@ -1413,3 +1415,9 @@ let rec error_of_msg ~trace_reasons ~op ~source_file =
         null_loc, ["Potentially null/undefined value."];
         falsy_loc, [spf "%s value." value_str]
       ]]
+  | EUnknownProperty (loc, name) ->
+    mk_error
+      ~kind:(LintError LintSettings.UnknownProperty)
```
calebmer: Try temporarily making this an error and running our test suite. I’d imagine that this error would show up in some unexpected places.

src/typing/flow_js.ml
```diff
@@ -5555,10 +5555,10 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
 
     (* LookupT is a non-strict lookup *)
-    | (ObjProtoT _ |
-       FunProtoT _ |
-       DefT (_, MixedT (Mixed_truthy | Mixed_non_maybe))),
-      LookupT (_, NonstrictReturning t_opt, [], _, _) ->
+    | (ObjProtoT reason |
+       FunProtoT reason |
+       DefT (reason, MixedT (Mixed_truthy | Mixed_non_maybe))),
+      LookupT (reason_op, NonstrictReturning t_opt, [], _, _) ->
```
calebmer: We will probably need to add a second parameter to `NonstrictReturning` or another variant to that data structure that lets us know we are coming from `TestPropT`. We use `NonstrictReturning` in a few other places that it would not make sense to add a lint warning.

src/typing/flow_error.ml
dmueller39: The case that has me nervous is the refinement of a mixed type: https://github.com/facebook/flow/blob/master/tests/refinements/cond_prop.js#L23

Everything else is more or less expected.

My sense is that many developers are used to this pattern, and it should be excluded from the rule. What do you think the way forward should be?

src/typing/flow_js.ml
dmueller39: I can do that, but I would argue that the pattern below https://github.com/facebook/flow/pull/4547/files/a11ccd0d4b85a1f59ea163a6203bd42c03e4f693#diff-f14393048a5695c25d5b2c97f6eca481R5576 of `RProperty Some name, RObjectType` it properly. Basically saying if `obj.prop` is non strict, trigger the error. The exercise above about making this a global error bore this out.

src/common/lints/lintSettings.ml
dmueller39: 👍

src/typing/flow_js.ml
dmueller39: Given the intent of scoping this to tests I agree.

dmueller39:
@calebmer I spent some time with this, and it was getting close. I ran out of time for the moment, but can get back to it relatively soon. In the mean time, if you could take a brief look at my progress that would be appreciated. I added a test case for refining a mixed type (because it should either definitely fail, or definitely not fail depending on your feedback). Additionally I am getting a Not_found exception, I believe in here: https://github.com/facebook/flow/pull/4547/files#diff-c35e29b70fc92e53566835b00fd10ad0R1419

If you see anything obvious I would appreciate it. Otherwise I can continue to investigate later.

dmueller39:
At the moment I am blocked due to a Not_found exception, happening somewhere in the error suppression phase. If someone could either take a look OR give guidance on the best method on how to attach a debugger (tutorial guide or something like that) I can get unblocked. Until then I'll put this on hold.

ballardrog:
A little while back we had some issues with Not_found in error suppression code due to some assumptions of LintSettingsMap being violated. [This diff](https://github.com/facebook/flow/commit/ac035a61d0d72e07fcbdeccaad3f166e656b3099) fixes the problem and [this one](https://github.com/facebook/flow/commit/f01db0df2ba79b43f7cec041d91520c75c81387b) gives a better error message if the assumption is violated. Try rebasing on top of those and see if that fixes it.

dmueller39:
@ballardrog @calebmer Thank you! It looks like the lreason location was being included with the error, which has a location of `"<NONE>": (0, 0) to (0, 1)` A "<None>" source does not seem to be in the lint settings map at the moment. Is that an issue? I'm not 100% certain. For the moment, I am formatting my LintError similar to how the other lint errors are being formatted, with an explicit location and a message.

I am still looking for guidance on handling mixed types in a conditional statement. My guess is that this lint rule should allow for that, but I would like one of the core team members to second that before I go forward.

ballardrog:
We have locationless reasons (the kind that you ran into) for a couple of internal operations that aren't supposed to be surfaced in an error. Errors with locationless reasons break several things when we go to filter or print errors, including looking up the severity that should be associated with a lint. I looked at the code you wrote for the linting part of things, and it looks good to me.

I'm of the opinion that refinements on mixed types shouldn't trigger this lint, otherwise it would be too noisy. I'd like one of the team members who have been around longer to weigh in on this, though.

yeknomerif:
Any updates on this?

dmueller39:
Recently @calebmer mentioned that yes indeed we want the refinement of a mixed type to be a valid case, so a little investigation needs to be done on how to add that. afaict that's the big outstanding issue.

PR #4610, opened 2017-08-16
Disable coersion of addition of number and string.

### LINKED ISSUE #2303: Question: Prevent coercing unsafe values?
Is there a way to make flow fail upon coersion of unsafe values, such as something like this:

```js
function unsafeCoersion(num: number) {
  return number * 'a string'; // <- Make flow throw error here
}
```

weakish:
Now flow will report type error on code like `1 + '%'`.

resolve #2303

---

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -6114,8 +6114,8 @@ and needs_resolution = function
  * is that we would error on both l and r, saying neither is compatible with
  * `string`.
  *
- * We are less permissive than the spec when it comes to string coersion:
- * only numbers can be coerced, to allow things like `num + '%'`.
+ * Unlike the spec, we do not allow string coersion:
+ * for example, numbers cannot be coerced, so `num + '%'` is invalid.
  *
  * TODO: handle symbols (which raise a TypeError, so should be banned)
  *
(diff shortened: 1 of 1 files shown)


nmote:
This seems like a good idea to me, but it will likely lead to a tough rollout on large codebases. I'm not sure if there would be consensus among the team.

If you are still interested, could you rebase and make sure that tests are passing? If you do, I'll bring this up with other team members to see what they think.

Thanks for the contribution, and sorry for the long wait.

TrySound:
Or maybe we should go with ocaml way and leave this unchanged? More safety and unexpected string conversions which are common problem in javascript.

I work with JS for 5 years and still not sure what is the result of `'5' + 5` :)

nmote:
@TrySound, this PR *disables* implicit coercions. I think we are in agreement here.

TrySound:
Ah, sorry. Misread. I really like it then! Thanks

weakish:
> could you rebase and make sure that tests are passing?

Rebased.

nmote:
I started a discussion with others on the team about this. While we agree that this is a good idea in principle, we are concerned about the rollout pain (this causes 16,000 new errors in one of our repositories) and the frustration that it could cause users. Of course, the improved safety is a worthwhile goal but we don't think the tradeoff is worth it at this time.

We are interested in making this a Flow lint rule, though. That way it could be configured as a warning or error (or left off altogether) on a per-project basis. I'm not very familiar with how lint rules are implemented but I could spend some time on it if you are interested in reimplementing this as a lint rule.

Feel free to CC me in an issue if you would like implementation guidance.

weakish:
> We are interested in making this a Flow lint rule, though.

Great. Should this be part of the [sketchy-number] rule or  a seperated rule?

[sketchy-number]: https://flow.org/en/docs/linting/rule-reference/#toc-sketchy-number

nmote:
@weakish, sorry, your response slipped through the cracks. This should be a separate lint rule. Could we move any future design discussion to an issue? Feel free to CC me when you open it.

PR #4635, opened 2017-08-17
[Docs] Explain function subtyping and variance

asolove:
I'm not an expert at type theory, but trying out the previous function subtyping example seemed to convince me that it was not correct because it explained that inputs and outputs should both be subtypes, whereas inputs should actually be a super type.

I'm also not an expert at technical writing, so please let me know (or just edit) if my tone and style don't fit with the rest of the docs.

I tried to add a light explanation of function subtyping that might help for simple cases, while still pointing to the Variance page for a more detailed discussion. 

(Note that, in reading the Variance page, I notice that it mostly focuses on variance as applied to OO class hierarchies. I think there might be room for a simpler explanation of function input/output types. There is also probably a need for some examples of just functions with complex cases: arguments that are callbacks, whose arguments are in output position and therefore covariant, arguments that are written to, and therefore invariant, etc.)

### DIFF
--- a/website/en/docs/lang/subtypes.md
+++ b/website/en/docs/lang/subtypes.md
@@ -113,22 +113,32 @@ recursively until we can decide if we have a subtype or not.
 
 #### Subtypes of functions <a class="toc" id="toc-subtypes-of-functions" href="#toc-subtypes-of-functions"></a>
 
-Flow compares two functions by comparing its inputs and outputs. If all the
-inputs and outputs are a subset of the other function, then it is a subtype.
+The subtyping of functions is a bit trickier. So far we've used the rule that one
+type is a subtype of another if it contains all the possible values of the parent.
+It's not clear how to apply this rule to function types.
+
+For functions, we'll adopt a different way of looking at that rule: for a function
+type to be a subtype of another, it should be safe to use in all the places where
+its parent can be used.
+
+Let's say we have a function type:
 
 ```js
-type Func1 = (1 | 2)     => "A" | "B";
-type Func2 = (1 | 2 | 3) => "A" | "B" | "C";
+type Func1 = (1 | 2) => "A" | "B";
 ```
 
-This also applies to the number of parameters in the functions. If one function
-contains a subset of the parameters of the other, then the other is a subtype.
+For another function type to be a subtype of this, it must be usable in any code that
+relies on the `Func1` type. That code expects to be allowed to send `1` or `2` as
+an argument and receive back `"A"` or `"B"` as the result. For another function type
+to be a subtype, it must:
 
-```js
-// @flow
-type Func1 = (number) => void;
-type Func2 = (number, string) => void;
+- Accept _at least_ `1` and `2` as inputs, though it may also accept other values (or additional parameters)
+- Return _at most_ `"A" | "B"`, though it may only return one or the other.
 
-let func1: Func1 = (a: number) => {};
-let func2: Func2 = func1;
-```
+In general, for one function type to be a subtype of another, its inputs must be a supertype
+of the parent's, but its output must be a subtype of the parent's. (The distinction between
+input and output becomes even trickier if some of the function's arguments are themselves functions
+or are written to.)
+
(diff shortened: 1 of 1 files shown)


website/en/docs/lang/subtypes.md
```diff
@@ -113,22 +113,32 @@ recursively until we can decide if we have a subtype or not.
...
-type Func1 = (number) => void;
-type Func2 = (number, string) => void;
+- Accept _at least_ `1` and `2` as inputs, though it may also accept other values (or additional parameters)
+- Return _at most_ `"A" | "B"`, though it may only return one or the other.
 
-let func1: Func1 = (a: number) => {};
-let func2: Func2 = func1;
-```
+In general, for one function type to be a subtype of another, its inputs must be a supertype
+of the parent's, but its output must be a subtype of the parent's. (The distinction between
```
jbrown215: I don't think you need to mention this extradistinction here, but you are definitely correct!

website/en/docs/lang/subtypes.md
```diff
@@ -113,22 +113,32 @@ recursively until we can decide if we have a subtype or not.
...
+input and output becomes even trickier if some of the function's arguments are themselves functions
+or are written to.)
+
+This subtyping rules for inputs and outputs are called "variance" and are the subject of the next
```
jbrown215: "The decision of which direction to apply the subtyping rule on inputs and outputs is governed by variance, which is the topic of the next section."

website/en/docs/lang/subtypes.md
```diff
@@ -113,22 +113,32 @@ recursively until we can decide if we have a subtype or not.
...
 
 ```js
-type Func1 = (1 | 2)     => "A" | "B";
-type Func2 = (1 | 2 | 3) => "A" | "B" | "C";
+type Func1 = (1 | 2) => "A" | "B";
 ```
 
-This also applies to the number of parameters in the functions. If one function
-contains a subset of the parameters of the other, then the other is a subtype.
+For another function type to be a subtype of this, it must be usable in any code that
```
jbrown215: Introduce another function, Func2, and then describe the relationship as you do. Better to be explicit :)

website/en/docs/lang/subtypes.md
```diff
@@ -113,22 +113,32 @@ recursively until we can decide if we have a subtype or not.
 
 #### Subtypes of functions <a class="toc" id="toc-subtypes-of-functions" href="#toc-subtypes-of-functions"></a>
 
-Flow compares two functions by comparing its inputs and outputs. If all the
-inputs and outputs are a subset of the other function, then it is a subtype.
+The subtyping of functions is a bit trickier. So far we've used the rule that one
```
jbrown215: Subtyping rules for functions are a bit more complicated. So far, we've seen that `A` is a subtype of `B` if `B` contains all the possible values in `A`. For functions, it's not clear how this relationship would apply. To simplify things, you can think of a function type `A` being a subtype of a function type `B` if functions of type `A` can be used wherever a function of type `B` is expected.

asolove:
@jbrown215 thanks for your feedback! I completely agree with all of it. I've updated the text to reflect your suggestions and add some more concrete cases.

website/en/docs/lang/subtypes.md
```diff
@@ -113,22 +113,34 @@ recursively until we can decide if we have a subtype or not.
...
+- `f3` can accept all the argument values that `FuncType` does, and only returns
+values that `FuncType` does, so its type is a subtype of `FuncType`.
 
-```js
-// @flow
-type Func1 = (number) => void;
-type Func2 = (number, string) => void;
+In general, the function subtyping rule is this: A function type `B` is a subtype
+of a function type `A` only if `B`'s inputs are a superset of `A`'s, and `B`'s outputs
+are a subset of `B`'s. The subtype must accept _at least_ the same inputs as its parent,
```
jbrown215: small typo: `B`'s outputs are a subset of `A`'s (instead of `B`'s)

jbrown215:
Thanks again for these changes. I think your examples make the docs way clearer!

PR #4652, opened 2017-08-18
Fix exported type of React.StatelessFunctionalComponent

### LINKED ISSUE #4650: React.StatelessFunctionalComponent<Props> type arguments errors
Using the new `React.StatelessFunctionalComponent<Props>` type, I tried to declare a component like this:

```js
type FooProps = {| size?: 1 | 2 | 3 |};
const Foo: React.StatelessFunctionalComponent<FooProps> = props => /* component definition */
```
but this resulted in the following errors:

```
9: const Foo: React.StatelessFunctionalComponent<FooProps> = props =>
              ^ type application of StatelessFunctionalComponent. Too few type arguments. Expected at least 2
[LIB] static/v0.53.1/flowlib/react.js:204:   declare export type StatelessFunctionalComponent<P, C> =
                                                                                              ^ See type parameters of definition here
9: const Foo: React.StatelessFunctionalComponent<FooProps> = props =>
              ^ type application of polymorphic type: type `React$StatelessFunctionalComponent`. Too many type arguments. Expected at most 1
[LIB] static/v0.53.1/flowlib/react.js:124: declare type React$StatelessFunctionalComponent<Props> = {
                                                                                           ^ See type parameters of definition here
```

https://flow.org/try/#0PTAEAEDMBsHsHcBQiCWBbADrATgF1AFSgCGAzqAEoCmxAxvpNrGqAETY32sDcytsAO1L5aoALyVOuAHS0OxXFQCi0KmioDcvRLgCeGKqABisWAAUmGchIDeAH0ShQpFAC8qAfgBcoAIyg7UAAmANAAZkQ7AF9tfiF8E1gfajoZAGVcBSpVUlIjAFcBehRBYmgAYWYsAQ1cAB5Ei1grAD5xUAxLaxbHUFoAClYAExQAN1YAGlAbXqdhXVUfGacV0AALKhQAczXcH07m0mkXdwDA3wnZlfgUIdw1-a7jt0M7c6uo3qiASm0qAA8sHhQEMqJBiPloAlTLwgA

Is declaring components like this not supported, or am I missing something?

asolove:
Resolves #4650.

I don't know if the `C` type variable was intended to model something else, but it was not used and seemingly causing a lot of problems. Removing it made things work as (I, at least) expected.

I also noticed that the tests previously did not contain any explicit usages of `StatelessFunctionalComponent`, which is presumably how this wasn't noticed. So I added a new test using that type, which passes only after this commit.

(I've also added the contextTypes as an optional value on a stateless functional component following the pattern currently used for full Components. That could be committed separately, but I thought it was worth adding while I was in there...)

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -123,8 +123,8 @@ declare class LegacyReactComponent<Props, State>
  */
 declare type React$StatelessFunctionalComponent<Props> = {
   (props: Props, context: any): React$Node,
-
   propTypes?: $Subtype<{[_: $Keys<Props>]: any}>,
+  contextTypes?: any
 };
 
 /**
@@ -201,8 +201,8 @@ declare module react {
 
   declare export var Component: typeof React$Component;
   declare export var PureComponent: typeof React$PureComponent;
-  declare export type StatelessFunctionalComponent<P, C> =
-    React$StatelessFunctionalComponent<P, C>;
+  declare export type StatelessFunctionalComponent<P> =
+    React$StatelessFunctionalComponent<P>;
   declare export type ComponentType<P> = React$ComponentType<P>;
   declare export type ElementType = React$ElementType;
   declare export type Element<C> = React$Element<C>;
--- a/tests/react_functional/test.js
+++ b/tests/react_functional/test.js
@@ -1,4 +1,4 @@
-import React from "react";
+import * as React from "react";
 
 function F(props: { foo: string }) {}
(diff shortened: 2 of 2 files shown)


Peeja:
I'm not sure I follow. Why was `C` removed? Doesn't this mean we can't type the context for stateless components? I'd really like to be able to use `C`.

PR #4664, opened 2017-08-21
Fix IntersectionObserver constructor definition

apostolos:
The options argument in IntersectionObserver's constructor should be optional:
https://wicg.github.io/IntersectionObserver/#intersection-observer-interface


lvdang:
This is definitely needed. Got code that doesn't use that options. Therefore having the "?" options? is great. Would like to have this merged

Hypnosphi:
Is there anything that blocks merging this? Can I help in any way?

PR #4675, opened 2017-08-22
Remove redundant `react-native` text in lieu of just `react`

peterpme:


asolove:
Just a note that this is the opposite of [another recent PR](https://github.com/facebook/flow/pull/4653). I'm not sure which way is right, but it's either one or the other.

peterpme:
you're right! Closing this one out. Thanks :)

PR #4687, opened 2017-08-23
Update EventTarget definition to extend HTMLElement

RajaBellebon:
```js
document.body.addEventListener('click', event => {
  console.log(event.target instanceof HTMLElement); // true
});
```

> Note: Hi this is @thejameskyle

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -154,7 +154,7 @@ type EventListenerOptionsOrUseCapture = boolean | {
   passive?: boolean
 };
 
-declare class EventTarget {
+declare class EventTarget extends HTMLElement {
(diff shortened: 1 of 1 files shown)


thejameskyle:
What a great PR, this guy makes the best PRs, I can tell you that.

jcready:
I don't believe this is valid since the target could be an SVGElement.

asolove:
Not just SVG, this could be a ton of things. In Node, that includes lots of objects that can be `EventTarget`s, like `FileReader`.

TrySound:
Maybe better to use polymorphic type like `SyntheticEvent`? It will be more consistent then.

samwgoldman:
As noted, this isn't correct. And it's also not correct to make Event polymorphic on the target type, due to bubbling.

thejameskyle:
Last we talked about this, `currentTarget` could be made polymorphic though, right?

PR #4713, opened 2017-08-25
Add Recoverable and context to repl

SimenB:
Never done this before...

Jest uses these.
image

https://github.com/facebook/jest/blob/f509856258499bdfa4b1bb43016652b9acd86b81/packages/jest-repl/src/cli/repl.js#L36
https://github.com/facebook/jest/blob/f509856258499bdfa4b1bb43016652b9acd86b81/packages/jest-repl/src/cli/repl.js#L79

https://nodejs.org/api/repl.html

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1899,6 +1899,10 @@ declare module 'repl' {
     replMode?: $SymbolReplModeMagic | $SymbolReplModeSloppy | $SymbolReplModeStrict;
     breakEvalOnSigint?: boolean;
   }): REPLServer;
+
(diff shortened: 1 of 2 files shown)


SimenB:
Guessing we're seeing errors because of #4291. /cc @zacharygolba

zacharygolba:
Looks like I forgot to add the definition for [`repl.Recoverable`](https://nodejs.org/dist/latest-v8.x/docs/api/repl.html#repl_recoverable_errors)!

PR #4722, opened 2017-08-25
feat: add indexer property to string lib def

### LINKED ISSUE #4716: String lib definition is not a subtype of indexed collections
Currently the [`String` lib def](/facebook/flow/blob/master/lib/core.js#L290) does not implement the following interface _([example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAdgFwKYBOUAhgMZ5gCSWAJngB540DCcMMepOGcWAPABUAfGAb5aAZyr4CxAEYdBIgN6owYANoBrPAE8AXGCwBXALZzCAXUMCANGrAcsAcxwALQyfOF7AX3RQxlhcPFhgUBgEEjhKABQAbsQwxniG1HSMLGwcIbxKAJQ2YKpgBHg4xgRhicl4GgAMlqj+qKS80WDEnmYWBGAAvOGR0bEaAIy2YABMkwDMlvmt7ThgcobRBNjOA0NROLEA5PjRB4tAA))_:

```javascript
interface IndexedCollection<T> extends Iterable<T> {
  [key: number]: T,
  length: number,
}
```

However, (internally?) strings are treated as such _([example](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAKA5AFwKYDOOWA2gAwC6AXGEQE4CWAdgOYCUqQA))_. A workaround could be achieved with a union but it is lossy in comparison to the `IndexedCollection` interface. You still have to refine `string | T` and a lot of the benefits of parametric polymorphism are nullified.

```javascript
// 😃
function forEach1<T>(fn: T => void, value: IndexedCollection<T>): void {
  for (let i = 0; i < value.length; i++) {
    fn(value[i])
  }
}

// 😐
function forEach2<T>(fn: (string | T) => void, value: string | IndexedCollection<T>): void {
  for (let i = 0; i < value.length; i++) {
    fn(value[i])
  }
}
```

**tldr;**

Is it sound to add a `[key: number]: string` property to the lib def for `String`? If so, I would be happy to create a PR.

zacharygolba:
Closes #4716

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -323,6 +323,7 @@ declare class String {
     valueOf(): string;
     toString(): string;
     length: number;
+    [key: number]: string;
     static (value:any):string;
     static fromCharCode(...codes: Array<number>): string;
     static fromCodePoint(...codes: Array<number>): string;
--- a/tests/core_tests/core_tests.exp
+++ b/tests/core_tests/core_tests.exp
@@ -3,16 +3,16 @@ Error: map.js:23
                  ^^^^^^^^^^^^^^^^^^^^^ constructor call
  23:     let x = new Map(['foo', 123]); // error
                           ^^^^^ string. This type is incompatible with
-545:     constructor(iterable: ?Iterable<[K, V]>): void;
-                                         ^^^^^^ tuple type. See lib: <BUILTINS>/core.js:545
+546:     constructor(iterable: ?Iterable<[K, V]>): void;
+                                         ^^^^^^ tuple type. See lib: <BUILTINS>/core.js:546
 
 Error: map.js:23
  23:     let x = new Map(['foo', 123]); // error
                  ^^^^^^^^^^^^^^^^^^^^^ constructor call
  23:     let x = new Map(['foo', 123]); // error
                                  ^^^ number. This type is incompatible with
-545:     constructor(iterable: ?Iterable<[K, V]>): void;
-                                         ^^^^^^ tuple type. See lib: <BUILTINS>/core.js:545
+546:     constructor(iterable: ?Iterable<[K, V]>): void;
+                                         ^^^^^^ tuple type. See lib: <BUILTINS>/core.js:546
 
 Error: map.js:24
  24:     let y: Map<number, string> = new Map([['foo', 123]]); // error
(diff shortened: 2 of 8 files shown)


lib/core.js
```diff
@@ -323,6 +323,7 @@ declare class String {
     valueOf(): string;
     toString(): string;
     length: number;
+    [key: number]: string;
```
jcready: Should this instead be `+[key: number]: string | void`?

I mean, it's definitely covariant, but should the `| void` be added?

```js
let a = ''
a[10].toString()
```

zacharygolba: That was also my gut reaction but I chose to omit the `| void` for compatibility with the [`Array` indexer property](https://github.com/facebook/flow/blob/master/lib/core.js#L277).

None the less, I agree with you. This definitely feels like a bug in the `Array` lib def _([example](https://flow.org/try/#0PTAEAEDMBsHsHcBQiDGsB2BnALqAhgE4F4CeAXKAIJGkA86ArgLYBGApgQHygC8oA2gEYANKABMogMwBdVBhygAbnmgM2FZara98NEvwC0g2UA))_.

jcready: I feel like you should copy [the definition for $ReadOnlyArray](https://github.com/facebook/flow/blob/master/lib/core.js#L237-L238) and make both the `length` and indexer properties covariant.

zacharygolba: That makes sense semantically although I don't think that makes the `string` indexer property any more or less sound.

**Rational**

Writing into index `[0]` on `Array<*>` with a `number` and then index `[1]` with a `string` produces `Array<number>` and then `Array<number | string>` respectively. The assignment to the indexer property will change the type of it's value (`T`). In this case covariance is beneficial because it guarantees the value of `T` does not change at "compile time". 

On the other hand, `string` is monomorphic and immutable at the language level so we do not need to worry about the type of the indexer property value from changing. In this case, making the indexer property covariant only makes the type less flexible.

**tldr;**

In my opinion, keeping the indexer property bivariant provides greater flexibility without sacrificing soundness. 

In any case, [I could be missing something](https://github.com/facebook/flow/pull/4713#issuecomment-324923277). I am happy to amend this PR if the bivariant indexer property is unsound. 😃

asolove: Now that I think about it, I'm wondering if this isn't part of string because of Unicode issues. Subscripting directly into a string isn't safe because you may get part of a multi-byte Unicode character. If you want a character array you are supposed to get the string's iterator, which will let you iterate over the Unicode characters.

zacharygolba: Ahhh that's a great point. Funny enough, you just made me find a bug in an iterator library I'm working on! Still, that feels better suited lint warning than a type error.

For example you may want to write the bytes of a string into some arbitrary destination such as Uint8Array.

asolove: Hmm, yeah, but in some sense the right return type for string subscript isn't string, but rather a code point or something like that...

zacharygolba: We need more types!

zacharygolba: https://doc.rust-lang.org/std/primitive.char.html

zacharygolba: Hmm after giving it some thought I still feel like the return type for string subscript is string. 
`'🍕'[0].constructor` is `String`. To me, from a type perspective that is safe. You can call `.toUpperCase()` without a run-time error. Your 100% right that it is potentially incorrect but I feel like that would be a logic error at the application level (at least in JavaScript).

nmote:
I spent some time looking into this, and I believe this change is correct. I don't think we should make it return `string | void`, even though that is technically correct, since I think consistency with other APIs is very important. I think it should mirror what is already in place for `Array` and `string` (the indexer on `string` is implemented [inside Flow](https://github.com/facebook/flow/blob/0af3f3a5d6b6c5f92a0df1f3d1f83b2fab57b755/src/typing/flow_js.ml#L4859-L4861) and also returns `string`).

I tend to agree that the indexer should be left as invariant in this case, though I'm not 100% sure. It's not a runtime error to write to a string property, but we might still want it to be a type error, since doing so still indicates at the very least a misunderstanding of the API.

Anyway, I'll import this and clean up the test output. Thanks for the PR, and sorry for letting it sit for so long.

PR #4735, opened 2017-08-29
Correct intersection type link in hoc doc

SamWyld:
The link within the hoc doc leads to a 404 page as the link is incorrect, I corrected the link from '../../types/intersection/' to '../../types/intersections/'


nmote:
Good catch, thanks for the fix!

PR #4741, opened 2017-08-30
Change the message when flow is rechecking (#3970)

### LINKED ISSUE #3970: 'this should not take long'
Please remove the 'this should not take long' remark while running flow. First of all 'long' is a very subjective property and for me it does take very long (subjectively measured of course :-)). It very often takes more than a minute, which is not weird because it completely depends on the size of the project.

VincentCordobes:
Fix #3970

### DIFF
--- a/src/commands/commandConnect.ml
+++ b/src/commands/commandConnect.ml
@@ -263,7 +263,7 @@ let rec connect env retries start_time tail_env =
         FlowExitStatus.(exit ~msg Server_initializing)
       end
   | Result.Error CCS.Server_rechecking ->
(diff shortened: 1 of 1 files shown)


nmote:
You're right that Flow is slower than we would like in some cases, but rewording the message doesn't change that. As such I don't think it's worth spending resources on this. There's always the risk that other tools rely on strings that Flow outputs, and even though that risk is small in this case there's no concrete upside to the change.

nmote:
Looks like a number of other people disagreed with me -- 46b585269bbdc5e8258c0686515955844419fd1e landed on Friday. Thanks for sparking the conversation!

PR #4747, opened 2017-08-30
Added in Express HTTP Improvements

### LINKED ISSUE #4500: Improve Node HTTP type definition
I would like to propose an improvement of HTTP request's [method](https://github.com/facebook/flow/blob/master/lib/node.js#L928).

```
declare type express$RequestMethod = 'GET' | 'POST' | 'PUT' | 'DELETE' | 'HEAD' | 'PATCH' | 'CONNECT' | 'OPTIONS' | 'TRACE'
```

jdhorwitz:
Fixes #4500.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1975,3 +1975,5 @@ declare var __dirname: string;
 
 declare function setImmediate(callback: ((...args: Array<any>) => mixed), ...args: Array<any>): Object;
 declare function clearImmediate(immediateObject: any): Object;
+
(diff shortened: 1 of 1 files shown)


asolove:
This defines the type, but you also probably want to use it in appropriate places where HTTP methods are represented in the node typing. [Here](https://github.com/facebook/flow/blob/master/lib/node.js#L1058) is one example. Possibly [here](https://github.com/facebook/flow/blob/master/lib/node.js#L1122) too.

PR #4762, opened 2017-08-31
Add MediaQueryList and related types

### LINKED ISSUE #4470: Add MediaQueryList type definitions
Flow seems not to recognise these standard interfaces, which are [widely supported](http://caniuse.com/#search=matchMedia) in browsers:

- [MediaQueryList](https://developer.mozilla.org/en/docs/Web/API/MediaQueryList)
- [MediaQueryListEvent](https://developer.mozilla.org/en/docs/Web/API/MediaQueryListEvent)
- [MediaQueryListListener](https://developer.mozilla.org/en/docs/Web/API/MediaQueryListListener)

I'm currently using this in a libdef in the root of my project:

```js
declare type MediaQueryListEvent = {
  matches: boolean,
  media: string,
};

declare type MediaQueryListListener = MediaQueryListEvent => void;

declare type MediaQueryList = {
  matches: boolean,
  media: string,
  addListener: MediaQueryListListener => void,
  removeListener: MediaQueryListListener => void,
};
```

Happy to PR if someone can guide me – should it go in the [BOM](https://github.com/facebook/flow/edit/master/lib/bom.js) file? Is it correct? Should I also add a declaration for [window.matchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia), and how?

callumlocke:
Please let me know if this is the wrong way to do it. I'm still learning about libdefs.

Closes #4470.


lib/bom.js
```diff
@@ -1324,3 +1324,17 @@ declare class VRDisplayEvent extends Event {
 
   constructor(type: string, eventInitDict: VRDisplayEventInit): VRDisplayEvent;
 }
+
+declare type MediaQueryListEvent = {
+  matches: boolean,
+  media: string,
+};
+
+declare type MediaQueryListListener = MediaQueryListEvent => void;
```
nmote: > A MediaQueryListListener is a callback function that gets invoked with a single argument, the MediaQueryList for which the evaluated result changed.

From https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryListListener

And

> [MediaQueryListEvent] instances are available as the event object on a function referenced by a MediaQueryList.onchange property or MediaQueryList.addEvent() call.

From https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryListEvent

So I think the callback passed to `addListener` and `removeListener` should just take a `MediaQueryList` -- `onchange` or `addEvent` are the only things that involve the `MediaQueryListEvent` type. Does that seem right?

nmote:
I think the test failure is spurious.

lib/bom.js
callumlocke: But the [addListener](https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener) docs specifically say the callback receives a `MediaQueryListEvent` (emphasis mine):

> A function or function reference representing the callback function you want to run when the media query status changes. In the original implementation, the callback was a non-standard MediaQueryListListener object. In the new implementation the standard event mechanism is used, the callback is a standard function, and **the event object is a MediaQueryListEvent**, which inherits from Event.

I think the docs for `MediaQueryListEvent` itself are just a bit unclear/incomplete.

nmote:
Thanks for the clarification -- I'll merge assuming tests pass.

PR #4777, opened 2017-09-01
Render typeinfo as text rather than HTML

Gozala:
https://flow.org/try/ prints type information of item under the cursor but information was rendered by updating `innerHTML` which leads to incorrect rendering as `Array<string>` would render Array string and `string` element. Using `.textContent`  instead will render text `Array<string>` instead.


nmote:
Nice catch!

PR #4778, opened 2017-09-01
Change definition for node cluster module workers

magicmark:
The following diff:

```js
    if (cluster.isMaster) {
        Object.values(cluster.workers).forEach(worker => {
            worker.on('message', msg => {
                if (msg.dumpMasterMemory) {
                    dumpMemory();
                }
            });
        });
    }
```

Produces the following error: (flow 0.53.1)

```
                 v----------------------------
 39:             worker.on('message', msg => {
 40:                 if (msg.dumpMasterMemory) {
 41:                     dumpMemory();
 42:                 }
 43:             });
                 -^ call of method `on`. Method cannot be called on
 39:             worker.on('message', msg => {
                 ^^^^^^ mixed
````

Looking at the def, it shows `workers` as `Object`: https://github.com/facebook/flow/blob/v0.53.1/lib/node.js#L343

The node docs also shows only `Object`, but describes the type of the object:
https://nodejs.org/api/cluster.html#cluster_cluster_workers

Would updating the flow def as such be appropriate?

Thanks!

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -340,7 +340,9 @@ declare module "cluster" {
       gid: number;
     };
     worker: cluster$Worker;
-    workers: Object;
+    workers: {
+        [id: number]: cluster$Worker;
+    };
 
     disconnect(callback?: () => void): void;
     fork(env?: Object): cluster$Worker;


nmote:
Yes, this is a reasonable change. The docs support your change. You'll just need to update the tests since this changes line numbers (`make test` then copy the appropriate `.out` file to `.exp` after verifying that it looks reasonable).

Unfortunately, I don't think this change will solve your problem, since `Object.values` returns `Array<mixed>`.

PR #4788, opened 2017-09-03
Property descriptors

swansontec:
The core type library was using `any` as the type for `Object.getOwnPropertyDescriptor` and friends. This pull request fixes that by defining a proper `Object$PropertyDescriptor` type, and applying it to all the relevant functions. It also adds the ES2017 'Object.getOwnPropertyDescriptors' method for good measure.

This pull request affects a gazillion files, since it changes line numbers in core.js. Sorry about that.


swansontec:
Sorry for the churn on this; I made some mistakes in the original commits, which were not in line with the ES spec.

nmote:
Could you link supporting docs?

bakkot:
It looks like this was done in #5067, other than `getOwnPropertyDescriptors`.

swansontec:
Cool! That's what we needed.

PR #4792, opened 2017-09-03
Support named imports properly with flow-upgrade codemod

### LINKED ISSUE #4786: flow-upgrade: Default import with named imports can be invalid and break
Hi! We're in the process of upgrading from 0.52 to 0.53. Thank you SO MUCH for creating the codemods to ease the upgrade!

I came across one case that causes quite a lot of flow errors in our codebase after running the codemod. I have a strong feeling that we could fix these with the upgrade codemods themselves.

We've got a few components that both import `React` from the default export as well as import `Component` as a named export. Like so:

```js
/* @flow */
import React, { Component } from 'react';

export default class SomeComponent extends Component {
  props: {
    children?: React$Element<*>
  };
  // ...
}
```

Currently the codemod outputs code that looks like this:

```js
/* @flow */
import React, { Component } from 'react';

export default class SomeComponent extends Component<{
  children?: React.Element<any>
}> {
  // ...
}
```

This causes type checks to fail, as we now have `import React from 'react';` instead of `import * as React from 'react';`. Here's an example error we've got:

```
Error: app/assets/javascripts/common/components/ContentHider.js:25
 25: export const createContentHider = (HiderComponent: React.ComponentType<*>) => {
                                                        ^^^^^^^^^^^^^^^^^^^^^^ ComponentType. Property not found in
                              v-
233:   declare export default {|
234:     +DOM: typeof DOM,
235:     +PropTypes: typeof PropTypes,
...:
247:   |};
       -^ object type. See lib: /private/tmp/flow/flowlib_3e2ef16f/react.js:233
```

And this is because the original source has this code:

```js
/* @flow */
import React, { Component } from 'react';
// ...
export const createContentHider = (HiderComponent: React.ComponentType<*>) => {
```

I can see that the existing snapshots also have this very same case with probably incorrect implementation:

https://github.com/facebook/flow/blob/bafcbde69953a0ae2d34dea5fadfdb03517b8a42/packages/flow-upgrade/src/upgrades/0.53.0/ReactUtilityTypes/__tests__/__snapshots__/codemod-test.js.snap#L31-L43

Specifically, here the `React.Element` won't be discovered by flow correctly, as the default export of `react` package does not have the `Element` property.

---

There are likely many ways to fix this case. I've outlined some suggestions below and would appreciate your input on their viability and what you'd prefer to see.

As a reminder, here's the input to these suggestions:

```js
/* @flow */
import React, { Component } from 'react';

export default class SomeComponent extends Component {
  props: {
    children?: React$Element<*>
  };
  // ...
}
```

## A. Remove the named imports and fetch them directly from `React.X`

```js
/* @flow */

export default class SomeComponent extends React.Component<{
  children?: React.Element<any>
}> {
  // ...
}
```

```diff
-import React, { Component } from 'react';
+import * as React from 'react';
```

```diff
-export default class SomeComponent extends Component<{
+export default class SomeComponent extends React.Component<{
```

# B. Add explicit type imports for new types

```js
/* @flow */
import React, { Component } from 'react';
import type { Element as ReactElement } from 'react';

export default class SomeComponent extends Component<{
  children?: ReactElement<any>
}> {
  // ...
}
```

* Add new line: `import type { Element as ReactElement } from 'react';`
* Output `ReactElement<any>` instead of `React.Element<any>`

This would naturally look ahead to what `React.X` helper types are used and then import those specifically.

We don't have to prefix the imported types with `React` but I think it would make the new code more clear.

# C. Import new types together with the existing import

This is practically the same as suggestion B but with only one `import` line:

```js
/* @flow */
import React, {
  Component,
  type Element as ReactElement
} from 'react';

export default class SomeComponent extends Component<{
  children?: ReactElement<any>
}> {
  // ...
}
```

* Add `type Element as ReactElement` to the list of named imports from `react` package
* * Output `ReactElement<any>` instead of `React.Element<any>`

---

I'd be open to taking a crack at modifying these codemods to support whatever we'd like to see. I've played around with codemods and ASTs before but haven't really used them to more complex scenarios. I'd appreciate any implementation suggestions :relaxed:

Thanks for your time! I'm looking forward to analyzing the new flow errors we now have. I might even discover new cases which the codemod could handle if just implemented. Do you think it's a good idea for me to open issues like these if I discover related cases?

valscion:
This PR fixes #4786 by removing named imports and changing the affected identifiers to use `React.X` directly. Basically, this is the solution A specified in the linked issue.

With this fix in place, flow can now find the `React.Element` and other helper types as is documented.

### DIFF
--- a/packages/flow-upgrade/src/upgrades/0.53.0/ReactUtilityTypes/__tests__/__snapshots__/codemod-test.js.snap
+++ b/packages/flow-upgrade/src/upgrades/0.53.0/ReactUtilityTypes/__tests__/__snapshots__/codemod-test.js.snap
@@ -29,7 +29,7 @@ const expression2 = () =>
 `;
 
 exports[`HiddenInSuperClassTypeArgs.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 class Buz {}
 class Foo extends Bar<React.Element<any>> {}
@@ -215,7 +215,7 @@ function create(): React.ComponentType<any> {}
 `;
 
 exports[`ReactClass.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 function create(): React.ComponentType<any> {}
 "
@@ -331,7 +331,7 @@ exports[`ReactComponent.js 2`] = `
 `;
 
 exports[`ReactComponent.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 (any: React.Component<Props, State>);
 (any: React.Component<Props, State>);
@@ -625,7 +625,7 @@ class N extends React.Component {
 `;
 
 exports[`ReactComponentRenderReturn.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 class A extends React.Component {
   render(): React.Node {}
@@ -1435,7 +1435,7 @@ exports[`ReactElement.js 2`] = `
 `;
 
 exports[`ReactElement.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 (any: React.Element<React.ComponentType<Config>>);
 (any: React.Element<React.ComponentType<Props>>);
@@ -1905,7 +1905,7 @@ class X {
 `;
 
 exports[`ReactNodeWithTypeArgs.js 3`] = `
-"import React, {PropTypes} from 'react';
+"import * as React from 'react';
 
 class X {
   render(): ?React.Node {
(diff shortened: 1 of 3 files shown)


valscion:
I'm on mobile and can't see why appveyor is failing. Should be something minor though -- jest passes for me locally. Feel free to do any necessary changes here if you know what's up with that :)

valscion:
The appveyor failure doesn't seem to have got anything to do with this PR. I closed and reopened to trigger a new build that should pass.

valscion:
A friendly ping @calebmer, if you could check this out :)

valscion:
Let me know if you'd like be to rebase this pull request or do something here still ☺️. I'm not sure if this would still be useful to people but I still believe this would be the correct fix here

Issue #1194, opened 2015-12-17
Union subset still not working when it's a subset of a sentinel type


zgao:
```es6
/* @flow */

type Subset = 'a' | 'b';

type A = {
  sentinel: 'a';
};

type B = {
  sentinel: 'b';
};

type C = {
  sentinel: 'c';
  contents: string;
};

type Fullset = A | B | C;

function foo(bar: Subset): Fullset {
  return ({
    sentinel: bar
  });
}
```

This fails to typecheck, but it should. Here's the error which shouldn't exist:

```
test.js:21
 21:   return ({
               ^ object literal. This type is incompatible with
 20: function foo(bar: Subset): Fullset {
                                ^^^^^^^ union: A | B | C
```

pasha-mf:
I believe it's the same issue as https://github.com/facebook/flow/issues/1014. Of course, it would be nice to hear from the developers on whether they plan to fix this in a foreseeable future...

avikchaudhuri:
@zgao This is working as intended. Unfortunately without an assurance that `sentinel` is read-only, we cannot conclude that `{ sentinel: 'a' | 'b' }` is the same as `{ sentinel: 'a' } | { sentinel: 'b' }`.

```
function gen(sentinel: 'a' | 'b') {
  const obj = { sentinel };
  sub(obj, () => { 
    obj.sentinel = 'b';
  });
}

gen('a');

function sub(obj: { sentinel: 'a' } | { sentinel: 'b' }, callback: () => void) {
  if (obj.sentinel === 'a') fix(obj, callback);
}

function fix(obj: { sentinel: 'a' }, callback: () => void): 'a' {
  callback();
  (obj.sentinel: 'a');
}
```

PR #4806, opened 2017-09-04
Adding DefaultProps Typing Back

mwalkerwells:
After doing a pretty large [Styled Components](https://github.com/styled-components/styled-components) refactor (PR [#1194](https://github.com/flowtype/flow-typed/pull/1194/)), which includes updates for Flow v0.53+, I started started to view HOCs differently.

For the following example, let's consider a ``React$StatelessFunctionalComponent `` supertype for simplicity.

```js
declare type SimpleComponent<Props> = {
  (Props): React$Node,
}
```
A HOC is a pure function that takes a callable object as input and returns a callable object as output. With the current set of library definitions, a HOC would be typed as:

```js
declare type Foo = { foo: 'foo' }
declare var HOC:
  <Props: Foo>(SimpleComponent<Props>) => SimpleComponent<$Diff<Props, Foo>>
```
This is effectively a transformation from ``{ $call: Foo => React$Node }`` to ``{ $call: void => React$Node }``.

This is a lossy type morphism that throws away information related to ``Foo``.

I'd like to _politely_ 😀 disagree with...
```js
  // We don't add a type for `defaultProps` so that its type may be entirely
  // inferred when we diff the type for `defaultProps` with `Props`. Otherwise
  // the user would need to define a type (which would be redundant) to override
  // the type we provide here in the base class.
```
https://github.com/facebook/flow/blob/master/lib/react.js#L81-L84

Is there any reason ``DefaultProps`` has to be a supertype of ``Props``?

---

Let's look at a new example & consider modeling ``DefaultProps`` orthogonally to ``Props`` for a moment...
```js
declare type Component<Props, DefaultProps> = {
  (Props): React$Node,
  defaultProps: DefaultProps,
}
```
```js
declare type Foo = { foo: 'foo' }
declare var HOC:
  <Props: Foo, DefaultProps: {}>(Component<Props, DefaultProps>) =>
    Component<$Diff<Props, Foo>, DefaultProps & Foo>
```

This is effectively a transformation from ``{ $call: Foo => React$Node, defaultProps: {} }`` to ``{ $call: void => React$Node, defaultProps: Foo }``.

This is now a lossless type morphism that retains all type information related to ``Foo``.

---

Contrived Example:

```js
declare type DB = { database: * }

declare var needsComponentWithDatabaseConnectionDependency: <Props: DB>(
   | Component<Props, *>
   | Component<*, Props>
) => void

declare var providesDatabaseConnection: <Props: DB>(SimpleComponent<Props>) => SimpleComponent<$Diff<Props, DB>>

declare var NeedsDBComponent: SimpleComponent<DB>

var TestComponent = providesDatabaseConnection(NeedsDBComponent)
// <TestComponent /> // <— valid

needsComponentWithDatabaseConnectionDependency(TestComponent) // ERROR
```

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -26,9 +26,9 @@ declare type React$Node =
  * Base class of ES6 React classes, modeled as a polymorphic class whose main
  * type parameters are Props and State.
  */
-declare class React$Component<Props, State = void> {
+declare class React$Component<Props, DefaultProps = *, State = void> {
   // fields
-
+  static defaultProps: DefaultProps;
   props: Props;
   state: State;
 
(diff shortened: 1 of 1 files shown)


calebmer:
I’m not entirely sure I understand why you want this 😣

Removing the `DefaultProps` type is the right thing to do for product developers. Is there anything stopping you from creating your own type for `React.StatelessFunctionalComponent` or `React.Component` that has a slot for `DefaultProps`? These types could extend `React.Component` for instance.

nmote:
Closing this due to a lack of a response. I'm not the right person to review but if you'd still like to move forward with this PR, feel free to reopen and maybe @jbrown215 could review.

PR #4837, opened 2017-09-08
[DOM] Support URL.createObjectURL from MediaStream

ovr:


nmote:
According to [the docs](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL), calling this method with a `MediaStream` is deprecated. Could you make a case that we should add it anyway?

ovr:
@nmote Yeap, will be better to use srcObject, but the biggest problem is that React didn't support `srcObject` property for `video`/`audio` tag

https://github.com/facebook/react/search?q=srcObject&type=Issues&utf8=%E2%9C%93

image

Yet another moment, possible we can add this method to Flow, but mark it as `deprecated`?

Thank

mvitousek:
It looks like this behavior, which was already deprecated in 2017, was removed from browsers in 2018. Unless I'm mistaken (in which case feel free to reopen, and I'll run this PR past other folks as well), this obviates this PR. Thanks for the submission though and I'm sorry we didn't fully take care of this one way or another sooner.

PR #4858, opened 2017-09-12
ServiceWorkerMessageEvent extends ExtendableEvent

keyiiiii:
See https://w3c.github.io/ServiceWorker/#service-worker-postmessage 
and https://github.com/GoogleChrome/samples/blob/da20dccd06edfb5008bdf04a02950c1b7969198d/service-worker/post-message/service-worker.js#L138-L140

### DIFF
--- a/lib/serviceworkers.js
+++ b/lib/serviceworkers.js
@@ -125,7 +125,7 @@ declare class ServiceWorkerContainer extends EventTarget {
   onmessage?: EventHandler,
 }
 
-declare class ServiceWorkerMessageEvent {
+declare class ServiceWorkerMessageEvent extends ExtendableEvent {
   data: any,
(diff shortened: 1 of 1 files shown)


calebmer:
It looks like the `ServiceWorkerMessageEvent` is deprecated and extends `Event` but not `ExtendableEvent`: https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerMessageEvent

It looks like we should be adding a type for `ExtendableMessageEvent` instead and having that extend `ExtendableEvent` 😊: https://developer.mozilla.org/en-US/docs/Web/API/ExtendableMessageEvent

Could you change the types to this?

keyiiiii:
Thank you review for me.

Could you review that again?

PR #4869, opened 2017-09-13
Add type definitions for Navigator.share

neeharv:
Ref: https://wicg.github.io/web-share


neeharv:
Woah, I have no idea why tons of tests failed. Could anyone help me understand why (seemingly unrelated) tests have bonked for such a simple type definition update?

asolove:
Looks like all that changed is the line numbers because you added content in the middle of the file. You can run locally and update the numbers and then should be ok.

calebmer:
Thank you for this 😊

It’s not too hard for me to update the tests so I’ll do it 👍

MDN link: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share

PR #4886, opened 2017-09-15
Allow React lifecycle methods to return mixed

### LINKED ISSUE #1803: Allow react components' lifecycle methods to be async
I tried to declare `componentWillMount` in a component as async but got a flow error: `Promise This type is incompatible with undefined`. I found out that it's caused by this definition:

``` js
componentWillMount(): void;
```

in https://github.com/facebook/flow/blob/master/lib/react.js

Shouldn't it be possible for components' lifecycle methods which return void to return a Promise instead?

thejameskyle:
Resolves #1803

This will allow people to write stuff like:

```js
class Component extends React.Component {
  async componentDidMount() {
    // ...
  }
}
```

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -43,12 +43,12 @@ declare class React$Component<Props, State = void> {
 
   constructor(props?: Props, context?: any): void;
   render(): React$Node;
-  componentWillMount(): void;
-  componentDidMount(): void;
+  componentWillMount(): mixed;
+  componentDidMount(): mixed;
   componentWillReceiveProps(
     nextProps: Props,
     nextContext: any,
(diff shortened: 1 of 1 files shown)


calebmer:
Thank you!

capaj:
@thejameskyle wouldn't it be better as `void | Promise<void>` instead of `mixed`?

IMHO it could be confusing for new react users if they see returns from these methods. We should encourage users not to return anything. Flow should warn about returns IMO.

thejameskyle:
It doesn't really seem like the responsibility of Flow to enforce something that does not make a difference at runtime

PR #4887, opened 2017-09-15
console.profile* methods both receive an optional string

iamdustan:
According to MDN at least:

https://developer.mozilla.org/en-US/docs/Web/API/Console/profile
https://developer.mozilla.org/en-US/docs/Web/API/Console/profileEnd


calebmer:
Thank you!

Issue #869, opened 2015-09-28
Don't typecheck or load files under node_modules/ unless they're imported by flow-typed files


AgentME:
Most of the files under node_modules aren't the responsibility of my project, errors in them are probably due to them being written for a different Flow version, and the only reason to parse them would be if my project has flow-typed files referring to types in them.

It seems like a common issue is that Flow takes longer to start up than it ought to because it reads all of the javascript files under node_modules/ when the user isn't interested in typechecking most of them (#863). Tons of files under node_modules/ are sub-dependencies of other non-Flow-typed dependencies. People can manually ignore individual dependencies to speed things up ([example](https://github.com/hammerlab/pileup.js/pull/269/files)), but this loses its effectiveness with the newly-released npm 3 because now dependencies aren't usually nested. (Browserify's many dependencies are now placed directly under node_modules/ instead of node_modules/browserify/node_modules/, so ignoring `.*node_modules/browserify.*` is no longer nearly as effective.)

It makes sense to check the flow types of files under node_modules/ if they're imported by flow-typed files in the project itself, and maybe the types of the files those ones include if it's necessary to fully figure out their types, but I think most/all users would prefer that files under node_modules/ would not be even loaded for typechecking unless they're imported by a flow-typed file in the project proper. Maybe this could be a (default-on?) setting in `.flowconfig`.

GGAlanSmithee:
Also #719 and #835

samwgoldman:
This seems like a good idea to me, but it also sounds really hard, given how things currently work.

danvk:
Also #863 for an easy way to make this problem more user-visible.

jamiter:
When I test my own code with Flow I'm not interested at all in type errors in the node_modules directory. I can't change anything about it anyway. But I do like to be able to import a module without having to declare the module interface for **every** NPM module I use. This is suggested in #719. I find it hard to accept this as a workaround... That means a lot of manual work for anybody using Flow.

Shouldn't node_modules be treated as a library in some way? As in: do not type check them or report on errors in them, but do find them and use them when imported.

> I think most/all users would prefer that files under node_modules/ would not be even loaded for typechecking unless they're imported by a flow-typed file

I agree, but do you expect Flow to show you errors in those imported files? Personally I only want errors from my own code. Flow should be smart enough to understand that code in `node_modules` isn't something you have control over.

samwgoldman:
[...]

PR #4916, opened 2017-09-18
Add silence support.

LegNeato:
With this patch there is a new section in `.flowconfig`:

```ini
[silence]
/path/to/file.js
/path/to/dir
```

Paths in `[silence]` will still be typechecked, but any
errors will be silenced as if there was a `suppress_comment` comment string
on every line where there are errors.

I called it `silence` to make it stand out from `ignore` and `suppress_comment` as
it currently silences all errors, including lints.

I believe this fixes https://github.com/facebook/flow/issues/869, though I admit I didn't read
the whole issue.

(If this is accepted I'll add docs in a follow-up).


LegNeato:
Note that I have never written OCaml before...let me know if there is anything I can do better.

mruzekw:
I'm curious to see how this will be received by the flow team.

All for suppressing flow errors internal to dependencies, while also checking against the API.

LegNeato:
Test failure is legit...apparently `make test` doesn't run `make js`. I'l fix it tonight.

LegNeato:
Open questions:

* Should we accept regexes instead of paths?
* Should we throw a warning when a file is silenced but there are no problems in the file to silence?
* Should we to silence `node_modules` by default?

mruzekw:
>Should we accept regexes instead of paths?

I would follow whatever [ignore] does.

>Should we throw a warning when a file is silenced but there are no problems in the file to silence?

I agree, silent means silent.

>Should we to silence node_modules by default?

Not at first. Let people get used to [silence]. Then down the road it can be a default.

rcknight:
Silencing node_modules will fix the spam which has been putting me off from using flow - Looking forward to the merge, thanks @LegNeato

LegNeato:
👉 @gabelevi

zoontek:
No ETA on this? 😕

alexmbarton:
@samwgoldman is this PR of interest to you?

jlaustill:
I CAN'T use flow until this is merged, my company/boss simply won't allow it.  This is THE feature that will make flow TAKE OFF.  Anything I can do to help test/improve this feature please reach out to me.  I want this!

TrySound:
@jlaustill There's `[untyped]` section for this case. You can see reference in thread.

jlaustill:
@TrySound I didn't notice that, thanks for pointing it out!  I'll check it out.  This has been frustrating me for quite awhile, and have not come across untyped.

jlaustill:
Has anyone looked at this in the last week?

icaroharry:
+1 for this

TrySound:
I don't think broken types should implicitly works and hide errors. It's like knowing there's a problem with your health but don't care about it. And one day you will die. Who cares?

TrySound:
Don't shoot in your feet, guys.

Palisand:
@TrySound That depends on your definition of "broken". If your project contains a dependency that is using older versions of flow (which is a common scenario), your project's flow version might consider that dependency's typings to be broken. However, they are still perfectly valid for the flow version used by the dependency and can be safely silenced.

TrySound:
`[untyped]` safely treats it as `any` without hiding dangerous errors

jlaustill:
@TrySound [untyped] would also ignore any types that WERE working, why would you EVER want to do that???

I just need to silence errors in node_modules that I have ZERO control over.  If flow can get types and help me with the rest of the library I want that, hence [silence].  Give me the help, don't break my build because of errors which I can not control though...

haggholm:
@TrySound What I (and I think many others) really want is for Flow to verify our use of 3rd party APIs, but not to make us responsible for 3rd party internals. `[untyped]` and `[ignore]` fails to do the former. Doing nothing does the latter. `[silence]` hits the sweet spot, insofar as it is possible.

vincenzovitale:
+1000 for merging this.

jlaustill:
@TrySound your thumbs down tells me that you really don't care about other peoples needs and only your own.  How is that helpful in the open source community?

TrySound:
I just think that hiding errors is nonsense which will lead community to really bad feel about flow correctness.

jlaustill:
@TrySound you are literally arguing that ignoring typing is better than silencing an error.  Silencing an error still allows flow to do its job, it merely silences SPECIFIC CHOSEN errors that we can NOT control.  Your suggestion goes a step further and completely ignores the typing.  How does one make the community feel worse than the other?  In both cases you are ignoring an error, you are merely going about it differently.  Your chosen path goes WAY further than merely silencing and in effect renders flow useless.  Want the community to have a negative view, render your library useless...

jlaustill:
Why is this taking so long?  Someone already did the work, the community has been very patient.  I'm still not using flow because of this missing necessity, but I'd really like to!

jlaustill:
How many weeks will I keep checking back for this before I give up on flow and just recommend typescript to my team?  I'm starting to wonder.

TrySound:
@jlaustill This doubtful change is not critical for using flow.

jlaustill:
@TrySound yes it is, so I will just go to typescript

TrySound:
Good luck

vith:
@TrySound Why would you prefer to disable type checking the public APIs of dependencies, when they only have type issues internally? Is that an unfair reading of your position? I just can't understand it.

Wouldn't using `[silence]` be equivalent to what happens when typings for a library are maintained externally?

This is getting very repetitive at this point, but I don't know what else to say: there is value in knowing that I've misused an API regardless of whether their flow annotations internally are correct. It seems ideal that we'd get errors initially, report them upstream, and silence them locally so that for example CI can pass in the meantime.

icaroharry:
Honestly, I'd like to hear from other devs. I still didn't get if @TrySound is a troll or just very closed-minded

Palisand:
I think other devs are just tired of their requests for silence support falling on deaf ears. This PR is 7 months old and the issue it solves (#869) was first reported 3 years ago.

icaroharry:
@Palisand I mean other devs responsible for Flow

StribMaxMarti:
I too support this feature, I've tried to add flow to my react native project a couple times, but the inability to hide all the flow issues in dependencies made it nearly impossible to see / manage real issues in my code.

timLoewel:
We currently have 371 flow errors in our react-native project.
361 of which are in the node_modules folder.
But 10 are currently in our code. And no one in our team can be bothered to check her flow errors before checking in, because they are so hard to see among the many errors we do not control.

Due to this missing feature flow is now just a warning system in our editors, but not part of the build infrastructure, as it breaks anyway.

Please accept the pull request! So that we can finally move on to type checked code in javascript.

TrySound:
@timLoewel You probably should treat react-native as untyped since it's are broken. This will allow you to run flow on ci.

STRML:
What's frustrating about that strategy @TrySound is that the externally facing types are not broken. They are simply self-referentially broken. Things like component prop types should still be preserved. `[silence]` would fix this very nicely.

timLoewel:
Exactly!
I would have to throw out half of my type checks, just to not see errors/warnings, I am not interested in.
@TrySound  why do you keep repeating the same argument, that we could use the current configuration options to silence the errors? Do you not listen? 
The answer is always the same: We want types! We love types! That is why we use flow. 

I was surprised by this, when I started using flow, because the suggested _silence_ solution is the one I was expecting to be the default. This is how many other code verifiers work.

And why do you say, that this was "not critical for using flow"?
If flow can not be used in a build pipeline, because it fails every build is that not critical? 
So many users in this thread say, that it is critical for them. Is that not enough proof that it is critical?

This is not about the objective truth. It is about making the tool work for the community.

yleflour:
> How many weeks will I keep checking back for this before I give up on flow and just recommend typescript to my team? I'm starting to wonder. @jlaustill 

The transition is already starting for our teams. New projects are getting Typescript support from now on. Our decision is motivated by 2 reasons:

- node_modules which can't be silenced
- Too many breaking changes (which makes silencing node_modules all the more important)

I've been watching this PR for > 6 months and the discussion seems to be going nowhere. I just don't understand why there are backlashes against this feature. A lot of people seem to need it and it wouldn't introduce breaking changes as it's a new config option.

apsavin:
@mrkev Could you please take a look on this PR?

mrkev:
Woah ok, this is a popular and meaty one 😅

I can't really do much here since there's merge conflicts. The diff looks scary because it's so much code, but relatively straightforward. If someone can fix it I can't promise I can get it merged, but I think it's a good addition and will vouch for it 👍

LegNeato:
I can rebase. Should be able to get to it a bit later today.

LegNeato:
@mrkev Rebased.

jlaustill:
@LegNeato amazing!  If this gets merged I will start testing it asap.

mrkev:
Does anyone have a quick test repo that would show why this change is necessary? It'd help.

From what I read `node_modules` is a big issue, but external dependencies from here by default get imported as `any` and display no errors. Are people trying to force flow to infer types inside `node_modules`?

LegNeato:
FWIW I posted a bunch of real world use-cases in the task that I hit in a project.

zeorin:
@mrkev As I understand it, files in `node_modules` that _aren't_ Flow files don't get type-checked and by default get imported as `any`, but files in `node_modules` that _are_ Flow files do get checked and are imported as whatever type Flow determines them to be.

It's possible that such a dependency might be written against a version of flow which did not raise any errors, but that the project which is consuming the dependency is using a different version of Flow that does raise errors in the dependency's code. Another possibility is that the dependency used an error suppression comment that is not commonly used, and thus not likely to be set up in any consuming project's `.flowconfig`; causing known and supressed errors to be unsupressed.

JSON files are still a problem, if they're malformed they will cause a flow error. Since they cannot have pragmas, they're automatically checked. There are some projects that include intentionally malformed JSON files in their tests (in order to ensure specific handling of such files).

In all of these cases, we will get errors that we cannot fix. These errors add noise and negatively impact the developer experience. We can eliminate them by ignoring the problematic files in `.flowconfig`, but this is sub-optimal, as it causes the ignored file to not be type-checked at all, leading to an implicit `any` import, or having to create a libdef.

`[silence]` fixes this by suppressing only the error notification, not the type inference.

I'm on mobile right now and can't create a sample repo, but the `config-chain` package is one that includes a malformed JSON file. It is depended on by (old?) versions of `npm`, which are depended on by other packages; one example of which is `gitbook-cli`. Simply including that package as a `devDependency` in a Flow project should demonstrate the problem.

jlaustill:
@mrkev I just made one quick
git clone https://github.com/jlaustill/flow-needs-silence.git
cd flow-needs-silence
yarn install
flow

You will see one error from immutable inside node_modules.  If node_modules was silenced, problem solved :)

LegNeato:
Here are the exact real-world scenarios that caused me to write this patch:

-------

1. **Different flow configuration from third-party libs**
   Your flow version doesn't match what a lib in `node_modules` is using for their project. Their types may depend on updated typedefs that ship with flow, causing errors for you but showing clean for them.

   Similarly, when there are suppressed errors using `suppress_comment` in a lib in `node_modules` and it doesn't match YOUR `suppress_comment` values you will see errors where they will not.

-------

2. **Fixed flow bugs causing errors**
   I updated to the latest version of `flow`, which fixed a flow bug. One of my dependencies has a `$FlowFixMe` to workaround the bug. I now get an _"Error suppressing comment. Unused suppression error"_ because the bug has been fixed and the suppression is not necessary with my `flow` version.
  
   The dependency (https://github.com/callemall/material-ui) doesn't get an error with their `flow` version and can't remove that suppression line if they want to support older versions of `flow` that other users may have installed.

-------

3. **Built-in typedef and version hell**
 - Started with no flow errors or warnings.
 - I updated `material-ui` to get new library features.
 - The updated `material-ui` was using `flow` >= 0.53.0, so its React types didn't match the ones my version of `flow` (0.51.0) understood. **I now had a bunch of type errors due to React type mismatches**.
 - I upgraded my version of `flow` to >= 0.53.0. The React errors went away as expected, but NEW errors in a subdep of some library I am using popped up due to changed `flow` behavior (example: https://github.com/cssinjs/jss/pull/580)
 
   So now I have to either fork the whole dep chain to fix the issue, wait for the various libraries to fix, release, and update all their deps, or use `yarn`'s overrides.

mrkev:
Update: busy week this week + short week next week = slow process

mrkev:
@LegNeato A team member involved in this just came back from PTO this week, so this was sort of blocked-- trying to get this reviewed this week though!

jlaustill:
@mrkev Thanks for taking the time with @LegNeato 's much-needed work!  I look forward to adding flow back to my projects and giving this a go.

mrkev:
Ok, this seems to be good, albeit one change. Instead of calling it "silenced files" we'd want to call them "declaration files", in line with [Hack's decl mode](https://docs.hhvm.com/hack/typechecker/modes#decl-mode). There currently is no `@flow decl`, but it's an idea that been played around with, and were it to materialize, it'd be good to have the terms align. 😄

LegNeato:
@mrkev Rebased and changed naming.

Wasn't sure grammatically to go with `[declaration]` (matching `include` and `exclude`) or `[declarations]` (matching `libs` and `lints`). Ended up with `[declarations]`, as it isn't a verb like `include` or `exclude`.

Also, I noticed the command line option for `[libs]` is `--lib`, so I made the command line option for `[declarations]` be `--declaration`. Let me know if you'd prefer it to be plural.

Finally, I didn't squash so you can see the changes. I assume you'll do so when you merge. Let me know if you'd rather have them squashed into one commit here in the PR.

Thanks again for pushing this through!

mrkev:
Oh nice, that's what you call quick turnaround 😄 Alright re-importing-- the squash should be no issue.

LegNeato:
Failure on AppVeyor looks unrelated:

```
[00:24:59] [—] FAIL:  lsp\queries                       (10/10 tests. 36 steps passed. 1 steps failed)
```

mrkev:
Wouldn't suprise me, AppVeyor has been flaky lately.

mrkev:
I got a thumbs up, but it's generally bad practice to merge things late on a Friday. I'll rebase and attempt to land next week 👍

STRML:
Awesome!

FYI for those of you at Facebook: this was merged (https://github.com/facebook/flow/commit/433581b51b44f345e7ffa9ba8b78487da245d7dc) with references to `[silence]`, but it was changed to `[declarations]` by the time it was merged, so be aware of that in the next changelog :+1:.

vladshcherbin:
Finally I can give Flow another chance.

@LegNeato thank you!

lll000111:
Aaaaand the Changelog is wrong, it says [silence] (so does the commit message of https://github.com/facebook/flow/commit/433581b51b44f345e7ffa9ba8b78487da245d7dc - but the commited code clearly shows [declarations])

https://github.com/facebook/flow/issues/6572

kylebebak:
The changelog is wrong, but it doesn't matter, because `[declarations]`, although it's now a supported config section, doesn't actually work.

I'm running flow *0.76.0*. 16 files in node modules don't pass the flow check. One of these files is `/Users/kylebebak/Code/Work/919/web/node_modules/npm/node_modules/config-chain/test/broken.json`.

None of the following patterns silences flow errors in this file (or any of the other 15 files, for that matter):

- `/Users/kylebebak/Code/Work/919/web/node_modules/npm/node_modules/config-chain/test/broken\.json`
- `<PROJECT_ROOT>/node_modules/.*`
- `.*/node_modules/.*

So either the docs are wrong, or the feature was shipped but it's broken.

gabrielenosso:
I confirm that it seems it doesn't work.

LegNeato:
I'm digging into this right now, thanks for the reports!

LegNeato:
Ok, I think this is a different "issue". This PR (as intended) takes a __valid parsed file__ and suppresses any _type_ errors. It doesn't suppress any _parse_ errors on __invalid files__ (this is what the existing `[ignore]` section is for).

So, for the case @kylebebak  mentions above `[declarations]` wouldn't work for `config-chain/test/broken.json` because it isn't a valid file with type errors, it is invalid json and thus can't be parsed by flow. You need to use `[ignore]` for those types of issues.

One may think this is bad UX ("why not have an overall "hide any errors in this file?"), but I think it is correct and is the least surprising for your project. They are two fundamentally different types of issues that may affect your project in different ways. There is a big difference between a parse error (_flow can't make sense of this file and doesn't know what to do so it is bailing and no types are used_) and a type error (_flow can make sense of this file and can use types and found an issue_). Depending on the particular project merging those two cases into one override could hide correctness issues and cause bugs in production for you. But, the solution is simple...use `[ignore]` for the former and `[declarations]` for the latter.

In all my tests with _type_ errors (including @jlaustill's https://github.com/jlaustill/flow-needs-silence.git) `[declarations]` works as intended. Please open an issue and tag me if you find a case where that doesn't hold.

@kylebebak @gabrielenosso please confirm you are trying to set to `[declarations]` something that isn't able to be parsed rather than something that is parsed and has type errors.  Please open an issue and tag me if you find a case where type errors are not suppressed.

kylebebak:
@LegNeato 
You're absolutely right. I ignored the offending `node_modules` (some were tests, others were written in TypeScript) and now everything passes.

It might be a good idea to include the first paragraph of your comment, which explains things very nicely, as an addendum to the documentation for `declarations`.

Thanks for getting back to me, and sorry for the false alarm!

yamov:
@LegNeato first of all, thanks a lot for this great feature ❤️ 

However, it doesn't seem to work for me and I hope it is just me doing something wrong. 

These are the steps that illustrate the issue: 
1. Create a new RN 0.55.4 project `react-native init FlowDeclarations --version 0.55.4`
2. Install Flow
3. Run flow
At this point in the new RN project there are 15 Flow errors ( all inside `node_modules/react-native/Libraries/`).
4. Add the libraries with Flow errors to .flowconfig's [declarations] section
5. Run Flow again.

**Actual result**: The above mentioned Flow errors still stay and are not suppressed. More than that, there are new ones (now there are 47 errors)

**Expected result**: 15 Flow errors  from the RN Libraries are suppressed, no new errors are added.

I have set up a repo for this use case, [you can have a look here](https://github.com/yamov/FlowDeclarations/commits/master).

Am I doing it wrong or it indeed does not work?

LegNeato:
@yamov thanks for the repo! I can reproduce the behavior and am looking into it.

ntomallen:
Yeah, I'm seeing the same problem as @yamov. They all seem to be type errors coming from `react-native-fbsdk` and `react-native`, but the other third-party packages don't seem to be throwing any errors, which is good!

LegNeato:
Still digging in, I don't really know OCaml so not sure where to make the fix. Rather than pollute this PR, I filed a new issue for those that want to keep track of the `[declarations]` fix for flow builtin types:

https://github.com/facebook/flow/issues/6631

nodkz:
I dislike `[declarations]` name for this option. It does not explain what it really means.

Small poll for option naming:
https://github.com/facebook/flow/issues/6631#issuecomment-411285015

jedwards1211:
> It doesn't suppress any parse errors on invalid files (that is what the existing [ignore] section is for).
Groaaaaaaaaan, so what we need now is
```
[just STFU about]
.*\.json
```

jedwards1211:
@nodkz that's to mimic the fact that probably no one would assume that `[ignore]` means "act like this file is missing" :trollface: 

I agree it's a terrible name.  It's basically saying **"these files are type declarations.  Unlike all of my other type declaration files, which are intended for type checking, I'm explicitly declaring these declarations because obviously that means you shouldn't check these declarations"**

mrkev:
As stated above, it's meant to align with Hack's `decl` mode. From [the docs](https://docs.hhvm.com/hack/typechecker/modes):

> Decl mode code is not typechecked. However, the signatures of functions, classes, etc in decl mode are extracted, and are used by the typechecker when checking other code. Decl mode is most useful when converting over code written in PHP: although the body of that code might have type errors that you don't want to deal with right away, it's still beneficial to make the signature of that code, or at the very least its mere existence, visible to other code.

That's also the idea behind surfacing parse errors: you can't extract signatures if you can't understand the code. It is purposefully different from `[ignore]`.

vladshcherbin:
@mrkev `decl` is pure shit title no one understands. 

[This poll](https://github.com/facebook/flow/issues/6631#issuecomment-411285015) gives you understanding what developers expect it to be called. Changing `silence` naming was a bad idea. Please stop acting like it's fine just because Hack language named it so.

STRML:
I'll say it more gently, but I agree with @vladshcherbin. `[declarations]` sounds an awful lot like `[libs]` and does not meaningfully describe what it does. It also is confusing given Flow syntax like `declare class` or `declare module`. 

`[silence]`, while also not perfect due to the existence of `[ignore]` (which treats the module as completely missing), at least better describes what it does. I'd prefer something even more descriptive like `[ignoreErrors]`.

TrySound:
`[flow_fix_me]: *`

jedwards1211:
@mrkev yeah I know, I already saw that, but how is Hack in any way relevant to this?  (not to mention PHP, for goodness sake...)
It seems to me like you guys are so used to its terminology that you have a hard time seeing how confusing this will be to non-Hack users (i.e. most of us, probably).

jedwards1211:
@mrkev it's also worth mentioning that **all .js.flow files can be thought of as "declaration files"** even if we want to see errors from them.  "flow-typed declarations" is also a commonly-used term.  And "type declarations" can even live in ordinary .js files.

lll000111:
@mrkev Weighing whether yet another comment is justified, I think it _is_ possibly okay to repeat what I wrote elsewhere, because I pointed out that the naming violates a principle that every programmer is familiar with: information hiding.

In #6572 I wrote:

> By the way, when this is documented some day I hope there will also be an explanation for the name "declarations". The previous choice "silence" was more or less obvious to anyone, "declarations" requires internal knowledge about "decl mode" that most Flow users are completely unaware of (I never heard about it before reading through the PR discussion). I think it's unfortunate that a naming **that exposes internals** was chosen, I think one should aim to name things meant for external consumption based on the outside view, not based on what's (well hidden) inside.

mrkev:
Unfortunately I didn't make the call on this one. Flow was born largely out of the Hack team (and in fact still a lot of code is shared) so to me it's not surprising they wanted to align things this way. IMO `[silence]` is a little removed from from the functionality this provides because it doesn't silence all errors: parse errors will still be as loud as always. Discussion about this is perhaps best had in a separate issue though 👍

jedwards1211:
@mrkev well I think most people here want it to silence parse errors as well

mrkev:
That's what `[ignore]` is for.

TrySound:
And `[untyped]`. There  is nothing to silence with parse errors. $FlowFixMe does work in this case too.

lll000111:
@mrkev `ignore` ignores the whole thing. What I think people want is that *internal* issues are not reported. I cannot fix them. I still want to know about issues of how I use the API!

lll000111:
@mrkev `ignore` ignores the whole thing. What I think people want is that *internal* issues are not reported. I cannot fix them. I still want to know about issues of how I use the API! I DO care about he boundary layer. I do NOT care about what goes on behind it (I have no influence over it, not my package).

jedwards1211:
@jamesisaac it's not so much that tons of libraries have this problem, but rather that things like this which are stupidly cumbersome get more and more annoying over time.  There's no 100% ideal solution I can copy into each of my projects.

`<PROJECT_ROOT>/node_modules/[a-z0-9\-]+/(lib|src|test)/.*\.json` wouldn't catch files in `node_modules/foo/node_modules/bar/test/bad.json`, so the character range should at least include `/`.  Even this is not a foolproof solution, because should there be file with `@flow` that tries to `require` one of those ignored `.json` files will cause a Flow error.  Probably not the case in any of my dependencies, but I am agitating for a 100% foolproof way to suppress the malformed JSON errors once and for all.

AFAIK, Flow parses files under `node_modules` whether they're referenced or not.

ntomallen:
The reality is that (it seems) a lot of us want type processing in node modules but NO errors shown in any node modules. We want errors to be shown in our own code when we missuse a module, but we don't want to see that any node modules did something wrong internally. We want to be able to maintain code that is error free and see that on a PR there are no flow errors, but we can't do that when node modules reports hundreds of errors for dependencies I can't control. Many of those errors for us are from Facebook modules, such as the react native fbsdk or react native itself. I still haven't found an easy way to accomplish this. Should be as easy as "If error originates from $glob, do not report or display"

PR #4922, opened 2017-09-20
Recommend using the new Atom IDE packages

jamiebuilds:


mrkev:
Looks like Atom IDE this is in the docs now ¯\\_(ツ)_/¯

PR #4928, opened 2017-09-20
Add DOMStringMap

### LINKED ISSUE #4926: Could not resolve name 'DOMStringMap'
What's the best way to resolve this one?
It's [in the standard API](https://developer.mozilla.org/en-US/docs/Web/API/DOMStringMap) so...

index.js:
```js
const rootElement = document.getElementsByClassName('root-element')[0];
ReactDOM.render(<App dataset={rootElement.dataset}/>, rootElement);
```

App.jsx:
```js
const App = ({dataset}: { dataset: DOMStringMap }) => (
  /* ... */
);
```

Flow error:
```bash
Error: src/App.jsx:19
 19: const App = ({dataset}: { dataset: DOMStringMap }) => (
                                        ^^^^^^^^^^^^ identifier `DOMStringMap`. Could not resolve name
```

jcready:
[`DOMStringMap`](https://developer.mozilla.org/en-US/docs/Web/API/DOMStringMap)
[`HTMLElement.dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset)

Resolves #4926

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -91,6 +91,10 @@ declare class DataTransferItem {
 
 /* DOM */
 
+declare class DOMStringMap {
+  [key:string]: string;
+}
+
 declare class DOMError {
   name: string;
 }
@@ -1489,7 +1493,7 @@ declare class HTMLElement extends Element {
   className: string;
   contentEditable: string;
   contextMenu: ?HTMLMenuElement;
-  dataset: {[key:string]: string};
+  dataset: DOMStringMap;
   dir: 'ltr' | 'rtl' | 'auto';
   draggable: bool;
   dropzone: any;


lib/dom.js
```diff
@@ -1194,7 +1198,7 @@ declare class HTMLElement extends Element {
   className: string;
   contentEditable: string;
   contextMenu: ?HTMLMenuElement;
-  dataset: {[key:string]: string};
+  dataset: DOMStringMap;
```
MichaelDeBoey: Isn't the dataset optional?
Or at least the keys are optional I think?

MichaelDeBoey:
@jcready Any idea on fixing the CI errors?

MichaelDeBoey:
@jcready Any news on this one? 🙂

lib/dom.js
jcready: The object type `{[key:string]: string}` has only an indexer defined. An indexer just means that **if** there are keys aside from the ones already defined, the keys and values must match those of the indexer. So all of the following inhabit the type `{[key:string]: string}`: `{}`, `{ foo: 'bar' }`, etc. Where as this would not inhabit the type: `{ 1: 'bar' }`, or `{ foo: 1 }`.

MichaelDeBoey: Yeah I figured that out already, but thanks for making it clear to me once again 🙂

MichaelDeBoey:
@asolove Maybe you can help @jcready and me here?

MichaelDeBoey:
Can somebody help @jcready and me here please?

CC/ @asolove @samwgoldman @mroch @nmote

mrkev:
Sorry this fell off the radar. If you rebase this real quick I'll be able to merge it-- right now our internal tool is complaining cause it hasn't been rebased in a while so @facebook-github-bot can't import it.

jcready:
@mrkev done.

PR #4951, opened 2017-09-22
Normalize `;` vs `,` in interface separators

vjeux:
The interface docs was inconsistent, it uses `;` and `,` as separators interchangeably. This PR normalizes the entire document to use `,`, which is what prettier does and what I've seen most used in flow codebases.


vjeux:
cc @calebmer

vjeux:
I just updated it to go the other way, prefer `;` for interfaces.

PR #4956, opened 2017-09-23
allow async functions for react lifecycle hooks CWM, CDM and CWUN

capaj:
currently a an `async` hook is a flow error. Yes you can work around by defining a extra method and calling it from the hook, but it's cumbersome.


vkurchatkin:
See: https://github.com/facebook/flow/pull/4886

PR #4957, opened 2017-09-23
Add "external" property to `process.memoryUsage`

Couto:
According to the documentation of node.js the `process.memoryUsage()` returns a property `external`. [More info](https://nodejs.org/dist/latest-v4.x/docs/api/process.html#process_process_memoryusage)

This property, however, was missing from the nodejs type declarations, causing type errors when used.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1936,7 +1936,7 @@ declare class Process extends events$EventEmitter {
     rss : number;
     heapTotal : number;
     heapUsed : number;
-    external: number;
+    external : number;
   };
   nextTick: <T>(cb: (...T) => mixed, ...T) => void;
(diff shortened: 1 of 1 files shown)


Couto:
Tests seem to be failing according to Travis.

I can't figure out why (it's my first time looking at flow's code). If someone could point me to the right direction in order to fix the broken test, I would really appreciate :)

mroch:
Some of our tests include line numbers, so adding a line means you need to update all of the subsequent line numbers. Annoying, but easy to do: `./runtests.sh -r bin/flow` (the `-r` "records" the updated tests).

I did that while importing this, so it's good to go now.

Thanks for the fix!

PR #4968, opened 2017-09-26
Docs: Fix react events example

k15a:
This Counter example doesn't make sense otherwise :)


mroch:
whoops, https://github.com/facebook/flow/pull/4627 already did this

PR #4973, opened 2017-09-26
Add test-utils to ReactDOM definitions

wbinnssmith:
react-addons-test-utils was deprecated with React 16, as most of its functionality was merged into `react-dom/test-utils`. The shallow renderer is now a separate package that can be typed on `flow-typed`, but these other definitions likely belong here as they're now part of react-dom.

This adds lib definitions to react-dom/test-utils, mostly borrowed from [the flow-typed definitions for the now-deprecated react-addons-test-utils](https://github.com/flowtype/flow-typed/blob/0bdbd3d3e8537541baf762d93e2b7c25fc38a024/definitions/npm/react-addons-test-utils_v15.x.x/flow_v0.15.x-/react-addons-test-utils_v15.x.x.js).

I haven't interacted directly with this API much, so a quick sanity check on the tests for this would be much appreciated :)

### DIFF
--- a/lib/react-dom.js
+++ b/lib/react-dom.js
@@ -49,6 +49,26 @@ declare module 'react-dom/server' {
   declare function renderToStaticMarkup(element: React$Node): string;
 }
 
+declare module 'react-dom/test-utils' {
+  declare var Simulate: {
+    [eventName: string]: (element: Element, eventData?: Object) => void;
+  };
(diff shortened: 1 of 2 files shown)


wbinnssmith:
👋 @calebmer 

(for some reason I can't assign a reviewer to this)

PR #5001, opened 2017-09-29
Closes #4999

### LINKED ISSUE #4999: property `destroy`. Property not found in ReadStream
# Bug
NodeJS' ReadStream supports the method `.destroy` since v8.0.0. So this shouldn't be marked as an error.
Reference from NodeJS docs: https://nodejs.org/api/stream.html#stream_readable_destroy_error

Example which throws flow error as per v0.55.0:

```js
readStream.on('data', chunk => {
  // ...
  if (found) {
    readStream.destroy();
    readStream.emit('end', found);
  }
});
```

jkomyno:
Closes issue #4999 added destroy method as per official NodeJS documentation.
https://nodejs.org/api/stream.html#stream_readable_destroy_error

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1407,7 +1407,9 @@ declare class stream$Stream extends events$EventEmitter {}
 
 type readableStreamOptions = { highWaterMark? : number, encoding? : ?string, objectMode? : boolean };
 declare class stream$Readable extends stream$Stream {
+  _destroy(error: Error, callback: () => void): void;
   constructor(options?: readableStreamOptions): void;
+  destroy(error?: Error) => void;
   setEncoding(encoding : string): stream$Readable;
   isPaused(): boolean;
   pause(): stream$Readable;


jkomyno:
I'd love to add tests, but I can't seem to wrap my head around the way the test suites are set up in this project. I'm on Windows with GIT bash, `make test` doesn't work at all, while something like `./runtests.sh bin/flow node | grep -v 'SKIP'` fails with the following stack. Note that it failed for me even before adding my edit.

```
[✗] ERRORED: config_module_name_rewrite_node
[✗] ERRORED: config_module_system_node_resolve_dirname
[✗] ERRORED: declaration_files_incremental_node
flow start exited code 127
[✗] ERRORED: declaration_files_node
[✗] ERRORED: lazy_node_modules
[✗] ERRORED: node_haste
[✗] ERRORED: node_modules_with_symlinks
[✗] ERRORED: node_modules_without_json
[✗] ERRORED: parse_error_node
[✗] ERRORED: unchecked_node_module_vs_lib
flow start exited code 127

Passed: 0, Failed: 0, Skipped: 370, Errored: 10

```

lib/node.js
```diff
@@ -1407,7 +1407,9 @@ declare class stream$Stream extends events$EventEmitter {}
 
 type readableStreamOptions = { highWaterMark? : number, encoding? : ?string, objectMode? : boolean };
 declare class stream$Readable extends stream$Stream {
+  _destroy(error: Error, callback: () => void): void;
```
mrkev: The docs mention 

> callback <Function> A callback function that takes an optional error argument.

Could you add this optional error argument to the callback?

DullReferenceException:
@jkomyno: are you able to make the requested changes? If not, are you ok with someone forking your fix?

lib/node.js
jkomyno: Sure thing. It should become the following, right?

  `_destroy(error: Error, callback: (error? : Error) => void): void;`

DullReferenceException: `error` is also optional. Maybe the best option is `_destroy(error: ?Error, callback?: (error?: ?Error) => void): void;`

jkomyno:
@DullReferenceException Hi, sorry but I didn't see the message from @mrkev.
I have proposed my fix as a comment after changes have been requested, but I can't seem to edit it. I must have deleted my patch branch, since a year passed and I haven't used Flow since then.
Is it ok for you if I create a new fork and PR, and reference it below as a comment?

lib/node.js
jkomyno: This is how they documented that type in `@types/node`.

types_node

That implies that neither `writable._destroy();` or `writable._destroy(null);` are valid in Typescript at the moment, as the two parameters are required.
In fact, a quick test reveals that they are right. If I try to compile the following Typescript source, I get an error.

```typescript
import * as stream from 'stream';
const writable = new stream.Writable();
// @ts-ignore
writable._destroy(); // this throws
// writable._destroy(null, () => { /* */ }); (this wouldn't throw)
```

```
_stream_writable.js:673
  cb(err);
  ^

TypeError: cb is not a function
    at Writable._destroy (_stream_writable.js:673:3)
    at Object.<anonymous> (C:\Work\test\index.js:4:10)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:191:16)
    at bootstrap_node.js:612:3
```

DullReferenceException: I suspect the TypeScript definition is wrong, then. The [documentation for stream](https://nodejs.org/dist/latest-v8.x/docs/api/stream.html#stream_writable_destroy_err_callback) states that `error` is a _possible_ `Error`:

jkomyno: @DullReferenceException With my example I wanted to demonstrate that `@types/node` isn't wrong, and that my implementation is based on top of it.

My typescript example compiled to javascript with `tsc` gives the following:

```
const stream = __importStar(require("stream"));
const writable = new stream.Writable();
// writable._destroy(null, () => { /* */ });
// @ts-ignore
writable._destroy();
//# sourceMappingURL=index.js.map
```

The `// @ts-ignore` directive simply forced Typescript to compile even when if fails to type check.
The error is thrown from the NodeJS core itself. The documentation states that the `err` parameter is a *possible* `Error`, meaning that it can also be `null`. But you can't omit the two params, since there isn't any check that tests wether `cb` (the callback with an optional error) is `null` / `undefined` or not.

jkomyno: Yes, you are correct. New PR incoming. The first error can in fact accept 3 alternatives:
1) an `Error` instance
2) `null`
3) `undefined`

I just didn't remember the exact syntax in Flow (i.e. inverting the `:` and the `?` symbols).
Thanks for the refresh.

jkomyno:
PR #6935 opened. Closing this @DullReferenceException

PR #5004, opened 2017-10-01
Define zlib.constants for nodejs

### LINKED ISSUE #4803: [defs] zlib.constants is not defined
Node moved zlib's constants from under the zlib module to under zlib.constants e.g. `zlib.Z_BEST_COMPRESSION` is now `zlib.constants.Z_BEST_COMPRESSION`:

>Note: Previously, the constants were available directly from require('zlib'), for instance zlib.Z_NO_FLUSH. Accessing the constants directly from the module is currently still possible but should be considered deprecated.

https://nodejs.org/api/zlib.html#zlib_constants

Flow's defs are here: https://github.com/facebook/flow/blob/master/lib/node.js#L1730. Probably could pull the fields out into a constants type and then intersect it into the zlib type for backwards compatibility.

iboss-ptk:
resolve  #4803

Declare zlib.constants while retaining existing constants var at the root of the module for backward compatibility.


iboss-ptk:
`make test` has no error on my local machine. Could anybody suggest how to figure out why travis is failing? @asolove @mroch @gabelevi

PR #5017, opened 2017-10-03
[DOCS] Improve Type System docs

DSouzaM:
I made some improvements to the Type System docs while reading through them. Things improved/changed:

- general readability
- description of structural typing (I [don't think it's used in Haskell](https://www.google.ca/search?q=haskell+structural+typing), and in OCaml/Elm it seems to be [just used for record types](https://en.wikipedia.org/wiki/Structural_type_system#Description). Correct me if I've made a mistake)
- ~~description of soundness vs completeness (I think their meanings were reversed)~~

### DIFF
--- a/website/en/docs/lang/types-and-expressions.md
+++ b/website/en/docs/lang/types-and-expressions.md
@@ -69,30 +69,30 @@ of values. But still Flow checks _every_ possible value. In this way Flow is
 checking too many things or _over-approximating_ what will be valid code.
 
 By checking every possible value, Flow might catch errors that will not
-actually occur when the code is run. Flow does this in order to be _"complete"_.
+actually occur when the code is run. Flow does this in order to be _"sound"_.
 
-In type systems, ***completeness*** is the ability for a type checker to catch
+In type systems, ***soundness*** is the ability for a type checker to catch
 every single error that _might_ happen at runtime. This comes at the cost of
 sometimes catching errors that will not actually happen at runtime.
 
-On the flip-side, ***soundness*** is the ability for a type checker to only
+On the flip-side, ***completeness*** is the ability for a type checker to only
 ever catch errors that _would_ happen at runtime. This comes at the cost of
 sometimes missing errors that will happen at runtime.
 
-In an ideal world, every type checker would be both complete _and_ sound so
+In an ideal world, every type checker would be both sound _and_ complete so
 that it catches _every_ error that _will_ happen at runtime.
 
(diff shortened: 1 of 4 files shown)


website/en/docs/lang/types-and-expressions.md
```diff
@@ -69,30 +69,30 @@ of values. But still Flow checks _every_ possible value. In this way Flow is
...
-actually occur when the code is run. Flow does this in order to be _"sound"_.
+actually occur when the code is run. Flow does this in order to be _"complete"_.
 
-In type systems, ***soundness*** is the ability for a type checker to catch
+In type systems, ***completeness*** is the ability for a type checker to catch
 every single error that _might_ happen at runtime. This comes at the cost of
 sometimes catching errors that will not actually happen at runtime.
 
-On the flip-side, ***completeness*** is the ability for a type checker to only
+On the flip-side, ***soundness*** is the ability for a type checker to only
```
rattrayalex: Hmm, really? This is a pretty big change to my understanding.

DSouzaM: To my knowledge, a sound system will only ever make correct decisions (an unsound system might make incorrect decisions, e.g. letting program X pass type-checking despite having a type error). A complete system makes it possible, for any correct program, to show that it is correct (an incomplete system might not be sufficiently powerful to prove that program X is valid even if it is).

I'm drawing my definitions from [first order logic](https://math.stackexchange.com/a/105576), but maybe I'm misunderstanding it in this context.

asolove: I too think this was correct in the original. Soundness means it does not allow any incorrect program, but may also not allow programs that would be fine at run time. Completeness means that any correct program is allowed, although possibly some incorrect ones are as well.

DSouzaM: Okay, I follow now. Fixed.

clemmy: @asolove Your explanation seems correct to me. The part that makes the documentation confusing IMO is

> On the flip-side, completeness is the ability for a type checker to only ever catch errors that __would happen__ at runtime. This comes at the cost of sometimes missing errors that __will happen__ at runtime.

If there is a way to phrase this better, it'll be a lot more clear.

SamChou19815:
I will close this since it has not been updated for years, and we just pushed out a new website so the edited files are no longer there.

PR #5027, opened 2017-10-04
Add support for Crypto

### LINKED ISSUE #5019: DOM missing defs for crypto
There is node crypto, but no browser crypto in the defs.

MichaelDeBoey:
Closes #5019

Ref: https://developer.mozilla.org/en-US/docs/Web/API/Crypto

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -4,6 +4,116 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
+/* Crypto */
+
+// This is a helper type to simplify the specification, it isn't an interface
+// and there are no objects implementing it.
+// https://developer.mozilla.org/en-US/docs/Web/API/BufferSource
+type BufferSource = ArrayBuffer | $ArrayBufferView;
+
+type DOMString = string;
+
+declare interface Crypto extends RandomSource {
+  +subtle: SubtleCrypto;
+}
+
+type CryptoKey$Type =
+  | 'secret'
+  | 'public'
+  | 'private';
+type CryptoKey$Usages =
+  | 'encrypt'
+  | 'decrypt'
+  | 'sign'
+  | 'verify'
+  | 'deriveKey'
+  | 'deriveBits'
+  | 'wrapKey'
+  | 'unwrapKey';
+declare interface CryptoKey {
+  +algorithm: Object;
+  +extractable: Boolean;
+  +type: CryptoKey$Type;
+  +usages: CryptoKey$Usages[];
+}
+
+// TODO: CryptoKeyPair
+declare interface CryptoKeyPair {
+}
+
+// TODO: JSONWebKey
+type JSONWebKey = {};
+
+type RandomSource = {
+    getRandomValues<T: $TypedArray>(typedArray: T): T;
+};
+
+type SubtleCrypto$Algorithm =
+  | {name: 'AES-CBC', iv: BufferSource}
+  | {name: 'AES-CTR', counter: BufferSource, length: number}
+  | {name: 'AES-GCM', iv: BufferSource, additionalData?: BufferSource, tagLength?: number}
+  | {name: 'RSA-OAEP', label?: string};
+type SubtleCrypto$DerivedKeyAlgorithm = {
+  name: DOMString,
+  length?: 128 | 192 | 256,
+};
+// TODO: fix types for parameters
+type SubtleCrypto$DeriveKeyAlgorithm =
+  | {name: 'ECDH', public: any}
+  | {name: 'DH', public: any}
+  | {name: 'PBKDF2', salt: BufferSource, iterations: number, hash: DOMString}
+  | {name: 'HKDF-CTR', hash: any, label: any, context: any};
+type SubtleCrypto$GenerateKeyAlgo =
+  | 'AES-CBC'
+  | 'AES-CTR'
+  | 'AES-GCM'
+  | 'RSA-OAEP'
+  | 'AES-KW'
+  | 'HMAC'
+  | 'RSASSA-PKCS1-v1_5'
+  | 'ECDSA'
+  | 'ECDH'
+  | 'DH';
+type SubtleCrypto$HashAlgo =
+  | 'SHA-1'
+  | 'SHA-256'
+  | 'SHA-384'
+  | 'SHA-512';
+type SubtleCrypto$KeyFormat =
+  | 'raw'
+  | 'pkcs8'
+  | 'spki'
+  | 'jwk';
+type SubtleCrypto$SignAlgo =
+  | 'HMAC'
+  | 'RSASSA-PKCS1-v1_5'
+  | 'ECDSA';
+type SubtleCrypto$WrapAlgo =
+  | 'AES-CBC'
+  | 'AES-CTR'
+  | 'AES-GCM'
+  | 'RSA-OAEP'
+  | 'AES-KW';
+declare interface SubtleCrypto {
+  decrypt(algorithm: SubtleCrypto$Algorithm, key: CryptoKey, data: BufferSource): Promise<ArrayBuffer>;
+  // TODO: fix type for 'algorithm' param
+  deriveBits(algorithm: any, baseKey: CryptoKey, length: number): Promise<ArrayBuffer>;
+  deriveKey(algorithm: SubtleCrypto$DeriveKeyAlgorithm, masterKey: CryptoKey, derivedKeyAlgorithm: DOMString | SubtleCrypto$DerivedKeyAlgorithm, extractable: Boolean, keyUsages: CryptoKey$Usages[]): Promise<CryptoKey | CryptoKeyPair>;
+  digest(algo: SubtleCrypto$HashAlgo, buffer: BufferSource): Promise<ArrayBuffer>;
+  encrypt(algorithm: SubtleCrypto$Algorithm, key: CryptoKey, data: BufferSource): Promise<ArrayBuffer>;
+  exportKey(format: SubtleCrypto$KeyFormat, key: CryptoKey): Promise<CryptoKey>;
+  generateKey(algo: SubtleCrypto$GenerateKeyAlgo, extractable: Boolean, keyUsages: CryptoKey$Usages[]): Promise<CryptoKey | CryptoKeyPair>;
+  importKey(format: SubtleCrypto$KeyFormat, keyData: ArrayBuffer | JSONWebKey, algo: SubtleCrypto$GenerateKeyAlgo, extractable: Boolean, usages: CryptoKey$Usages[]): Promise<CryptoKey>;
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -4,6 +4,117 @@
...
+  | 'AES-CTR'
+  | 'AES-GCM'
+  | 'RSA-OAEP'
+  | 'AES-KW';
+declare interface SubtleCrypto {
+  decrypt(algorithm: SubtleCrypto$Algorithm, key: CryptoKey, data: BufferSource): Promise<string>;
+  // TODO: fix type for 'algorithm' param
+  deriveBits(algorithm: any, baseKey: CryptoKey, length: number): Promise<ArrayBuffer>;
+  deriveKey(algorithm: SubtleCrypto$DeriveKeyAlgorithm, masterKey: CryptoKey, derivedKeyAlgorithm: DOMString | SubtleCrypto$DerivedKeyAlgorithm, extractable: Boolean, keyUsages: CryptoKey$Usages[]): Promise<CryptoKey | CryptoKeyPair>;
+  digest(algo: SubtleCrypto$HashAlgo, buffer: BufferSource): Promise<string>;
```
qm3ster: This definitely returns something `ArrayBuffer`-like and not a string.

lib/dom.js
```diff
@@ -4,6 +4,117 @@
...
+  encrypt(algorithm: SubtleCrypto$Algorithm, key: CryptoKey, data: BufferSource): Promise<ArrayBuffer>;
+  exportKey(format: SubtleCrypto$KeyFormat, key: CryptoKey): Promise<CryptoKey>;
+  generateKey(algo: SubtleCrypto$GenerateKeyAlgo, extractable: Boolean, keyUsages: CryptoKey$Usages[]): Promise<CryptoKey | CryptoKeyPair>;
+  importKey(format: SubtleCrypto$KeyFormat, keyData: ArrayBuffer | JSONWebKey, algo: SubtleCrypto$GenerateKeyAlgo, extractable: Boolean, usages: CryptoKey$Usages[]): Promise<CryptoKey>;
+  sign(algo: SubtleCrypto$SignAlgo, key: CryptoKey, text2sign: BufferSource): Promise<BufferSource>;
```
qm3ster: How did you choose to return a `BufferSource`-implementing type rather than just an `ArrayBuffer`? both **MDN** and **w3.org** itself fail to specify the content of the promise, at the time of writing. Could it depend on the algo?(unlikely)

lib/dom.js
```diff
@@ -4,6 +4,117 @@
...
+  | 'RSASSA-PKCS1-v1_5'
+  | 'ECDSA';
+type SubtleCrypto$WrapAlgo =
+  | 'AES-CBC'
+  | 'AES-CTR'
+  | 'AES-GCM'
+  | 'RSA-OAEP'
+  | 'AES-KW';
+declare interface SubtleCrypto {
+  decrypt(algorithm: SubtleCrypto$Algorithm, key: CryptoKey, data: BufferSource): Promise<string>;
```
qm3ster: Also definitely `Promise<ArrayBuffer>`, the example uses TextDecoder and I looked up type in Chrome and FF debugger.
```js
(async function test(){

const ptUtf8 = new TextEncoder().encode('my plaintext')

const pwUtf8 = new TextEncoder().encode('my password')
const pwHash = await crypto.subtle.digest('SHA-256', pwUtf8)

const iv = crypto.getRandomValues(new Uint8Array(12))
const alg = { name: 'AES-GCM', iv: iv }
const key = await crypto.subtle.importKey('raw', pwHash, alg, false, ['encrypt'])

const ctBuffer = await crypto.subtle.encrypt(alg, key, ptUtf8)

const key1 = await crypto.subtle.importKey('raw', pwHash, alg, false, ['decrypt'])

const ptBuffer = await crypto.subtle.decrypt(alg, key1, ctBuffer)

const plaintext = new TextDecoder().decode(ptBuffer)
debugger
return plaintext
}())
```

lib/dom.js
```diff
@@ -4,6 +4,117 @@
...
+
+// This is a helper type to simplify the specification, it isn't an interface
+// and there are no objects implementing it.
+// https://developer.mozilla.org/en-US/docs/Web/API/BufferSource
+type BufferSource = ArrayBuffer | $ArrayBufferView;
+
+type DOMString = string;
+
+declare interface Crypto extends RandomSource {
+  subtle: SubtleCrypto; // readonly
```
qm3ster: My head is currently in TypeScript land, but I am almost certain [this is the way to make read-only props on an interface](https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only).

lib/dom.js
MichaelDeBoey: Because the input (`text2sign `) is of type `BufferSource`, it seemed logical the `Promise` would be something of `BufferSource` as well

MichaelDeBoey: Because the input (`text2sign `) is of type `BufferSource`, it seemed logical the `Promise` would be something of `BufferSource` as well

lib/dom.js
MichaelDeBoey: This is was not defined in any spec, but it should be a buffer indeed 🙂
`BufferSource` is more appropriate I think, since the input (`buffer`) is of type `BufferSource`?

lib/dom.js
MichaelDeBoey: `BufferSource` is more appropriate I think, since the input (`data `) is of type `BufferSource`?

lib/dom.js
```diff
@@ -4,6 +4,117 @@
...
+// TODO: CryptoKeyPair
+declare interface CryptoKeyPair {
+
+}
+
+// TODO: JSONWebKey
+type JSONWebKey = {};
+
+type RandomSource =  {
+  getRandomValues(typedArray: $TypedArray): void;
```
qm3ster: The [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/RandomSource/getRandomValues) uses mutating the array, but [w3.org](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues) officially specifies the array must also be returned.
MDN also depends on that feature in [the `encrypt` and `decrypt` examples](https://github.com/facebook/flow/pull/5027/commits/d56ffe786f3509999c304605e36e96979f04a3bd#r145700797).

TL;DR: I imagine it should go a little something like `getRandomValues<T: $TypedArray>(typedArray: T): T;`

lib/dom.js
MichaelDeBoey: That's indeed the way to make read-only props on an interface, but I did it this way, because in other parts of the file it was also done like this and I wasn't sure what to do with them so I stuck to the current way

lib/dom.js
qm3ster: It doesn't preserve the type in any way though? It's impossible to guess what was used from the ciphertext, so it's a safe bet it will always be an `ArrayBuffer`? :thinking:
It literally operates on the raw bytes.

lib/dom.js
MichaelDeBoey: I just followed the [MDN spec](https://developer.mozilla.org/en-US/docs/Web/API/RandomSource/getRandomValues), but if [w3.org](https://www.w3.org/TR/WebCryptoAPI/#Crypto-method-getRandomValues) specifies it with a return-variable and MDN uses it in there examples, this one should defo have one too. 🙂

lib/dom.js
qm3ster: Regarding all methods:
Saying it is a `BufferSource` won't break anything, but I don't see why any vendor would return an array representation of a buffer with a particular bitness and not a naked buffer.
I suspect `ArrayBuffer` is more correct for the returns.
`BufferSource` is indeed the type for all inputs, even `decrypt`.

MichaelDeBoey: OK will make them `ArrayBuffer` for now 🙂
We can always change them to `BufferSource` if necessary 🙂

MichaelDeBoey:
@qm3ster I addressed your comments 🙂

MichaelDeBoey:
@qm3ster I addressed your comments 🙂

The CI tests will still fail, 'cause of the line-numbers.
Don't know what's the best way to make them pass?

lib/dom.js
```diff
@@ -4,6 +4,116 @@
...
+
+// TODO: CryptoKeyPair
+declare interface CryptoKeyPair {
+}
+
+// TODO: JSONWebKey
+type JSONWebKey = {};
+
+type RandomSource = {
+    getRandomValues(typedArray: $TypedArray): $TypedArray;
```
qm3ster: This should be a generic, because it is always guaranteed to return the same member of $TypedArray, since it just returns the same instance.

qm3ster: Is there any reason `getRandomValues<T: $TypedArray>(typedArray: T): T;` didn't work?

qm3ster:
How naughty to do an interactive rebase on a publically accessible branch :rofl:

MichaelDeBoey:
@qm3ster I know, but it makes the commit cleaner, which would help when merging 🙂

MichaelDeBoey:
@qm3ster The CI tests will still fail, 'cause of the line-numbers.
Don't know what's the best way to make them pass?

qm3ster:
@MichaelDeBoey that it do, but it hid the old CI reference so I couldn't see why it fails + it disconnected my review comments from the code so I had to go line hunting.
I think it's best to do that at the end, before the merge.
Anyway, I am happy with the result and will show myself out.

MichaelDeBoey:
@asolove Maybe you can point me in the right direction on what's the best way to make the failed CI tests ('cause of the line-numbers) pass?

qm3ster:
Well just to make it pass as is, overwriting `tests/dom/dom.exp` would be enough.

However, this brings to my attention: 
does this need any tests?
or are they only for preventing regressions on real bugfixes?

MichaelDeBoey:
@qm3ster Indeed changing the line-numbers in `tests/dom/dom.exp` would fix the tests, but it looks very cumbersome to do so every time we add/update something in `lib/dom.js` (or other files, which will cause failures in other tests), so that's why I was asking for a better solution.

qm3ster:
Disabling line numbers would make it easier to see the change in the diff when test results change, and avoid failing changes to the lib when tests and test results are unchanged. But that is still a hack.
The everchanging libs should not be used as an integration test. There should be errors expected inside each test with the javascript API. The error printer should not be involved.

MichaelDeBoey:
Can somebody point me in the right direction on what's the best way to make the failed CI tests ('cause of the line-numbers) pass please?

CC: @asolove @samwgoldman @mroch @nmote

qm3ster:
Maybe @asolove is kill 🤔

MichaelDeBoey:
@nmote Can take a look at it later today if that's OK for you guys?

MichaelDeBoey:
@qm3ster @nmote I created #7714,  since I deleted the repo from my account

PR #5052, opened 2017-10-07
Update type definition of `module`

SimenB:
What I really wanted was `module.paths`, ended up tweaking more though.

I _think_ it's correct, but I might be wrong.

https://nodejs.org/api/modules.html

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -783,19 +783,21 @@ declare class Proxy<T> {
 declare var global: any;
 
 declare var module: {
-    exports: any;
-    require(id: string): any;
+    exports: typeof exports;
+    require(id: string): typeof exports;
     id: string;
-    filename: string;
+    filename: ?string;
     loaded: boolean;
-    parent: any;
-    children: Array<any>;
+    paths: Array<string>;
+    parent?: typeof module;
+    children: Array<typeof module>;
 };
 declare var require: {
-  (id: string): any;
+  (id: string): typeof exports;
   resolve: (id: string) => string;
   cache: any;
-  main: typeof module;
+  main?: typeof module;
+  extensions: {[string]: Function}
 };
 declare var exports: any;
 


lib/core.js
```diff
@@ -783,19 +783,21 @@ declare class Proxy<T> {
 declare var global: any;
 
 declare var module: {
-    exports: any;
-    require(id: string): any;
+    exports: typeof exports;
+    require(id: string): typeof exports;
     id: string;
-    filename: string;
+    filename: ?string;
```
SimenB:
```sh-session
$ node
> module.filename
null
```

lib/core.js
```diff
@@ -783,19 +783,21 @@ declare class Proxy<T> {
...
     loaded: boolean;
-    parent: any;
-    children: Array<any>;
+    paths: Array<string>;
+    parent?: typeof module;
```
SimenB:
```sh-session
$ node
> module.parent
undefined
```

lib/core.js
```diff
@@ -783,19 +783,21 @@ declare class Proxy<T> {
...
+    children: Array<typeof module>;
 };
 declare var require: {
-  (id: string): any;
+  (id: string): typeof exports;
   resolve: (id: string) => string;
   cache: any;
-  main: typeof module;
+  main?: typeof module;
+  extensions: {[string]: Function}
```
SimenB: this is deprecated, not sure if it should be added?

lib/core.js
```diff
@@ -783,19 +783,21 @@ declare class Proxy<T> {
...
+    parent?: typeof module;
+    children: Array<typeof module>;
 };
 declare var require: {
-  (id: string): any;
+  (id: string): typeof exports;
   resolve: (id: string) => string;
   cache: any;
-  main: typeof module;
+  main?: typeof module;
```
SimenB:
```sh-session
$ node
> require.main
undefined
```

lib/core.js
apsavin: Changes like this have no sense for me.
If code in a file, module.filename is always exists.

Issue #5018, opened 2017-10-03
Intersection type doesn't refine `mixed`


qm3ster:
[Reproducing using https://flow.org/try/](https://flow.org/try/#0C4TwDgpgBAshDO8CGBzaBeKBvAJk4SAXFALYCWAHhDgL5QBk2eBxWAUFFAGZkA2ErHv2IBGGmxriAxgHsAdvGBR5JBMjRRMXAK5ypwMvKgAKAbDWoIASmwcoZLiYgA6ZkmdDrtzuJpA)
```js
type Message = MessageEvent & {data: {
  file: {file: File}
}}
onmessage = function (e: Message) {
  if (e.data.file) {
    // do stuff with File
  }
}
```
## Expected
> :grinning: 
## Actual
> property `file` Property not found in mixed

asolove:
Have you tried using the type object spread operator, lik {...MessageEvent, data: whatever } ? When you want to combine two object types, that will work much better than intersection. Intersection of object types has a complex and unintuitive meaning that is rarely what you mean.

qm3ster:
[This works](https://flow.org/try/#0C4TwDgpgBAshDO8CGBzaBeKBvAdHrAJksEgFxQC2AlgB4QEC+ANFESeVgFBRQBmVAGwgd+Q8gEYGnBlIDGAewB28YFCUUEyNFEy8ArotnAqSqAAphsTaggBKbNyhVe5iDjZIcouw55SGQA), thank you very much @asolove.

## Why it is not as good

1. It is less semantical, in my opinion, to say 
"object that contains everything MessageEvent instances contain, but data is a different type now"
versus
"object is a MessageEvent instance, and also is an object having a more specific `data` property value"(that is `&`-ed with `mixed`, not silently replaces it)
2. It [silently overwrites MessageEvent types](https://flow.org/try/#0C4TwDgpgBAshDO8CGBzaBeKBvAdHrAJksEgFxQB2ArgLYBGEATgL4A0URJ5WAUFFADMAlgBsI3YWPIBGZj2ZyAxgHsK8YFFU0EyNFEwCqFRcCGqoACnGwdqCAEpsfKEIGWIOTkhySHT-nLMQA)!
[...]

PR #5055, opened 2017-10-08
Model MessageEvent as generic (parametric `data`, `source`, and `ports`)

popham:
Over at #5018 a user was attempting to use an intersection type to specify the data parameter of a `MessageEvent`. The `data` property should be parametrized. The `source` and `ports` properties on `MessageEvent` also warrant parametrization. This PR remodels `MessageEvent` and its uses with generics.

I expect that it will break a lot of typed code, but the transition should happen at some point. Unfortunately I had a lot of difficulty in finding per-use specs for the `source` and `ports` properties. I suspect some variation across browsers, but the remodelling is correct for Chromium. I'd appreciate any input to nail it down with more confidence.

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -446,7 +446,9 @@ declare class CloseEvent extends Event {
     wasClean: boolean;
 }
 
-declare class WebSocket extends EventTarget {
+type WSDataT = string | Blob | ArrayBuffer;
+type WSSendT = string | Blob | ArrayBuffer | $ArrayBufferView;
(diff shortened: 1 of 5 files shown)


nmote:
This looks pretty reasonable. Again, I'd like to apologize for the long response time here.

I'd like to see how this affects our main internal codebases. When you get a chance, could you rebase, I'll import, and then we can go from there?

Also, do you think it would be reasonable to have defaults for these type parameters, so that people could use `MessageEvent<>` if they would like to retain the current behavior?

nmote:
Don't worry about re-recording the tests, I can deal with that after I import.

SamChou19815:

PR #5061, opened 2017-10-09
Fix tests/rest/type_generic.js

k15a:
The rest property has to be the last one otherwise this is invalid syntax.


gabelevi:
Wow! Nice find! This is a totally a bug in the Flow parser, that it fails to complain about this! I'm not going to take this PR, but I'm going to fix the parser (and this test). Thanks again for bringing this to our attention!

nmote:
Looks like this was fixed in https://github.com/facebook/flow/commit/d985b109f6de5a75e6ba96c538ab3847c2b038a1 but this PR didn't get closed. Thanks again for finding this issue!

PR #5067, opened 2017-10-10
Update core.js types

apsavin:
Add PropertyDescriptor, update Object, Function, Proxy and setTimeout/setInterval

PR #5070, opened 2017-10-10
Add compression stats to --sharedmemory-log-level 1

gabelevi:


apsavin:
@gabelevi what's with AppVeyor? It seems that this check isn't stable. AppVeyor build failed on my [PR](https://github.com/facebook/flow/pull/5067) as well.

PR #5075, opened 2017-10-12
Add typings for globalAgent of Node.js http and https

shuhei:
Adding missing `globalAgent`s to `http` module and `https` module of Node.js

Corresponding documentations are:

- https://nodejs.org/api/http.html#http_http_globalagent
- https://nodejs.org/api/https.html#https_https_globalagent


gabelevi:
Looks good to me! I'll import and re-record tests.

PR #5081, opened 2017-10-13
Add lib definition for StorageEvent

TLadd:
Emitted when another tab writes to localStorage or sessionStorage

https://www.w3.org/TR/webstorage/#the-storageevent-interface


gabelevi:
Looks good to me! Tests probably need to be re-recorded but I can do that.

Issue #4723, opened 2017-08-26
Document "%checks" (Was: "if" condition type guard does not work when moved to function)


lll000111:
### Description

I moved the boolean "if" condition into a function because it was pretty long. Without any other changes, this if-condition type guard is no longer recognized and Flow complains about using properties that that function checks exist.

After some experiments, I don't see a way to have a type guard be a function?

I understand the "mutability" assumption Flow makes whenever a function is called, but being unable to use a function as type guard, as seems to be the case, seems a little... severe. After all, functions starting with `is...()` (isBoolean, isThis, isThat, etc.) are quite common. I am unable to use a function in a  type guard check?

### Code

```javascript
// @flow
'use strict';

function test1 () {
    // ERROR: Property 'myFn' cannot be assigned on 'data' (mixed)
    function isMyFunction (data: mixed): boolean {
        return typeof data === 'object' && data !== null && data.myFn === 'function';
    }

    function demo (data: mixed): void {
        if (isMyFunction(data)) {
            data.myFn = () => {
                // ...
            };
        }
    }
}

function test2 () {
    // WORKS
    function demo (data: mixed): void {
        if (typeof data === 'object' && data !== null && data.myFn === 'function') {
            data.myFn = () => {
                // ...
            };
        }
    }
}
```

### Error

```
Error: src/main.js:11
 11:             data.myFn = () => {
                      ^^^^ property `myFn`. Property cannot be assigned on
 11:             data.myFn = () => {
                 ^^^^ mixed

Found 1 error
```

### Working example

[...]

PR #5087, opened 2017-10-13
Add some docs about %checks

apsavin:
For #4723

### DIFF
--- a/website/en/docs/types/functions.md
+++ b/website/en/docs/types/functions.md
@@ -213,6 +213,80 @@ var num: number = method.call(42);
 var str: string = method.call(42);
 ```
 
+### Predicate Functions <a class="toc" id="toc-function-checks" href="#toc-predicate-functions"></a>
+
+Sometimes you will want to move condition from `if` statement into a function
+
+```js
+function concat(a: ?string, b: ?string): string {
+  if (a && b) {
+    return a + b;
+  }
+  return '';
+}
+```
+
+However, `flow` will see an error in the code below:
+
+```js
+function truthy(a, b): boolean {
+  return a && b;
+}
+
+function concat(a: ?string, b: ?string): string {
+  if (truthy(a, b)) {
+    // $ExpectError
+    return a + b;
+  }
+  return '';
+}
+```
+
+You can fix it using `%checks` mark like this:
+
+```js
+function truthy(a, b): boolean %checks {
+  return !!a && !!b;
+}
+
+function concat(a: ?string, b: ?string): string {
+  if (truthy(a, b)) {
+    return a + b;
+  }
+  return '';
+}
+```
+
+Such predicate functions must return immediately (i.e. you can not declare variables). 
+But it's possible to call other predicate functions inside a predicate function:
+
+```js
+function isString(y): %checks {
+  return typeof y === "string";
+}
+
+function isNumber(y): %checks {
+  return typeof y === "number";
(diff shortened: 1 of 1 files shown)


jcready:
You should probably also mention that these predicate functions must return immediately (i.e. you cannot declare variables) and that you cannot call other predicate functions inside a predicate function (i.e everything must be inlinined into a single predicate function).

gabelevi:
Our intern who built `%checks`, @panagosg7, just returned to FB full time! I'll run this by him and see if he's cool with some preliminary docs.

website/en/docs/types/functions.md
```diff
@@ -213,6 +213,53 @@ var num: number = method.call(42);
...
+
+function concat(a: ?string, b: ?string): string {
+  if (truthy(a, b)) {
+    return a + b;
+  }
+  return '';
+}
+```
+
+**You can't move such function with `%checks` to another file currently.**
```
panagosg7: Moving `truthy` to a separate file and keeping the intended functionality is possible. You would, however, have  to include type annotations for the arguments, e.g.
```
export function truthy(a: mixed, b: mixed): boolean %checks {
  return !!a && !!b;
}
```

apsavin:
@jcready thanks for the tips!
@panagosg7 fixed.

website/en/docs/types/functions.md
```diff
@@ -213,6 +213,54 @@ var num: number = method.call(42);
 var str: string = method.call(42);
 ```
 
+### Functions with %checks <a class="toc" id="toc-function-checks" href="#toc-function-checks"></a>
```
panagosg7: You can call these functions "Predicate Functions", since you are referring to them later as such. "%checks" can also be in fixed-width font for consistency with the rest of the doc.

website/en/docs/types/functions.md
```diff
@@ -213,6 +213,54 @@ var num: number = method.call(42);
...
+function concat(a: ?string, b: ?string): string {
+  if (truthy(a, b)) {
+    return a + b;
+  }
+  return '';
+}
+```
+
+Such predicate functions must return immediately (i.e. you can not declare variables). 
+Also, it's not possible to call other predicate functions inside a predicate function.
```
panagosg7: This is possible too. See https://github.com/facebook/flow/blob/master/tests/predicates-inferred/simple-predicate-func.js#L38.

apsavin:
@panagosg7 Fixed.

website/en/docs/types/functions.md
```diff
@@ -213,6 +213,79 @@ var num: number = method.call(42);
...
+
+function isNumberOrString(y): %checks {
+  return isString(y) || isNumber(y);
+}
+
+var a = 'a';
+var b = 5;
+var c = [];
+
+function sum(y): string | number {
```
panagosg7: I think this example is a bit trivialized, since the only values that flow to `a`, `b` and `c` respectively are string, number and array, so the `isNumberOrString` check does not filter out any values in the first two cases and filters out everything in the latter. How about the following variation:
```js
function foo(x): string | number {
  if(isNumberOrString(x)) {
    return x + x;
  } else {
  	return x.length; // no error, because flow understands that x has to be an array
  }
}

foo('a');
foo(5);
foo([]);
```

apsavin: Agree

PR #5126, opened 2017-10-17
Add docs for $ElementType

apsavin:
Part of #2464

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -129,7 +129,7 @@ Work in progress
 A $PropertyType is the type at a given key.
 
 As of Flow v0.36.0, `x` must be a literal string. In future versions, `x` may be allowed to be any type, as long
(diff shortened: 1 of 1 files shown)


apsavin:
cc @gabelevi

jcready:
Can anyone explain why one would ever use `$PropertyType` over `$ElementType` since string literals are also valid flow types?

MOZGIII:
I can find a reason to do so in a "perfect" world. So in the "real" world, the only reason may be if `$ElementType` is buggy or not implemented.

mrkev:
If @MOZGIII is right then it makes little sense to document $ElementType, because you'd be informing encouraging a buggy feature. If not, it makes little sense to document because it's essentially like $PropertyType, except more error prone since you might use types that don't resolve to strings.

Either way, I don't think documenting this makes sense.

mrkev:
rip

PR #5160, opened 2017-10-22
Add process.binding

SimenB:
It's not documented, but it's there to allow us to access the internal core node C++ bindings.

One common use case is doing `process.bindings('fs').realpath` to access the native `realpath` or `process.bindings('natives')` to get a list of all built-in modules in node.

(I understand if you don't want to add undocumented functions)

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1917,6 +1917,7 @@ declare class Process extends events$EventEmitter {
   abort() : void;
   arch : string;
   argv : Array<string>;
+  binding(id: string) : typeof exports;
   chdir(directory : string) : void;
   config : Object;
   connected : boolean;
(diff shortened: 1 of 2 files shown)


vkurchatkin:
> but it's there to allow us to access the internal core node C++ bindings

Not really, it's there only for Node.js itself, i.e. it is not a part of public API. Using is causes a lot of problems.

SimenB:
I suppose if things like https://github.com/nodejs/node/pull/15776 are done more often, the use-cases will disappear.

But with https://github.com/nodejs/node-v0.x-archive/issues/624 closed, getting access to a list of built-in modules is a pain (https://github.com/tapjs/stack-utils/blob/459f196b67b01684f6d370ab774944de184a8cfe/index.js#L16-L28 & https://github.com/sindresorhus/builtin-modules/blob/32ab9f2fe30c1cad2b449d7e6058a1b4018d9b88/index.js#L8-L10).

I can agree on them being edge-cases though, and not wanting to make it easier to use/discover `process.binding` than it is

SimenB:
Meh, bad practice

PR #5161, opened 2017-10-22
fix(cssom): add missing webkitTransition property

oliviertassinari:
I have noticed that [TypeScript uses the following property](https://github.com/Microsoft/TypeScript/blob/ceba50750b7bd6c8cfc21a5886e6eb9de7f82a36/lib/lib.dom.d.ts#L2312) that is missing here:
```ts
webkitTransition: string | null;
```


mrkev:
Down to merge this, I'll just need it rebased since it's over a few months old. Dw about the appveyor or re-recording the tests 👍

Issue #4673, opened 2017-08-21
`import * as React from 'react'` causes deprecation warnings


dgcoffman:
With dependencies:
``` bash
"flow-bin": "0.53.1",
"react-test-renderer": "15.6.1",
"react": "15.6.1",
```

Given a test file:
``` javascript
import * as React from 'react';
import renderer from 'react-test-renderer';

test('renders', () => {
  renderer.create(<div />);
});
```

Running the test file with Jest writes the following to the console:
```
  console.warn node_modules/react/lib/lowPriorityWarning.js:40
    Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead. For info on usage, compatibility, migration and more, see https://fb.me/prop-types-docs

  console.warn node_modules/react/lib/lowPriorityWarning.js:40
    Warning: Accessing createClass via the main React package is deprecated, and will be removed in React v16.0. Use a plain JavaScript class instead. If you're not yet ready to migrate, create-react-class v15.* is available on npm as a temporary, drop-in replacement. For more info see https://fb.me/react-create-class
```

Changing `import * as React from 'react';` to `import React from 'react';` causes these warnings to disappear.

The documentation at https://flow.org/en/docs/react/types/ recommends `import * as React from 'react';`

asolove:
Ah, interesting. Looks like the `_interopRequireWildcard` helper added by Babel has to copy properties onto a new object. So naturally it iterates over them and gets them and sets them. The React deprecation warnings are called in getters for `PropTypes` and `createClass`. So doing `import * as React` calls those getters just to set up the required object, even if user code doesn't use them.

I can't think of a good way around this on any level. The require wildcard is needed to know to import the namespace with `React.*` types. The copying is needed to preserve the semantics of imports. And there's really no way for React to know this is just ignorable library stuff rather than user code.

dgcoffman:
Because we want to keep our tests free of warnings, I'd like to find a workaround.

Options:
0. `import React, { type Node } from 'react';`
1. `React$Node`

bodiddlie:
Have the same issue after implementing flow in my project. Explicitly importing the needed types like @dgcoffman did in Option 1 got rid of the warnings.

mkscrg:
Ran into the same issue. At least we should fix the React docs section, which directly suggests

```js
import * as React from 'react';
```

asolove:
[...]

PR #5189, opened 2017-10-25
Docs: Improve the React namespaced import Docs

akiokio:
Just making clear the warnings when babel re-import the react module if you use the wildcard import

@asolove explanation in [here](https://github.com/facebook/flow/issues/4673)


Gpx:
@calebmer in my opinion these false positives make it harder to find deprecated code. I for one encountered this very problem and the only solution I could find was in this PR. I think it would be useful to add a small note in the docs to help future developers 🙇

alexreardon:
This was a pain to find. It would be great if it was called out in the docs

PR #5196, opened 2017-10-26
Fix return types for WebGL is* methods

lbud:
Fixes return types for `WebGLRenderingContext#is*` methods, which should all return booleans (confirmable at the [IDL](https://www.khronos.org/registry/webgl/specs/latest/1.0/webgl.idl) linked from [`dom.js#L1566`](https://github.com/facebook/flow/blob/7f2680cd89fab009b69c03961fc9213a97ff9d05/lib/dom.js#L1566)).

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -2344,13 +2344,13 @@ declare class WebGLRenderingContext {
   getVertexAttribOffset(index: number, pname: number): number;
 
   hint(target: number, mode: number): void;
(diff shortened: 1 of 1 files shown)


calebmer:
Thank you! 😊

lbud:
@calebmer I'm not sure how the Phabricator/import bot works, but it looks like this change is at least still not in [master](https://github.com/facebook/flow/blob/master/lib/dom.js#L2342)/recent releases — is there anything I should do to help this along?

PR #5198, opened 2017-10-27
Update hackfmt.ml

greb22:
spelling correction


mroch:
please reopen after signing the CLA, thanks!

PR #5202, opened 2017-10-27
Update solve_state.ml

greb22:
Spell changes


mroch:
this file was synced from hack, so it'd be better to fix it upstream at https://github.com/facebook/hhvm/blob/master/hphp/hack/src/hackfmt/solve_state.ml

thanks, though!

PR #5235, opened 2017-11-01
spell check in comments

anuolu:


mroch:
looks like this was fixed in #5200 and #5201. thanks

PR #5251, opened 2017-11-02
Add documentation for $Shape - Resolves #5246

### LINKED ISSUE #5246: Best way to copy a type and make every attribute optional
This one is most likely related to #5245 I guess...

What's the best way to copy my `Config` type and make it into an `PartialConfig`, which is just a `Config`, but with every attribute as an optional type.

So
```js
type Config = {
  allowsac: boolean,
  allowssso: boolean,
  email?: string,
  firstname: string,
};
````
will be
```js
type PartialConfig = {
  allowsac?: boolean,
  allowssso?: boolean,
  email?: string,
  firstname?: string,
};
```

aikar:


MichaelDeBoey:
Thanks for this one @aikar! 🙂

MichaelDeBoey:
I think this can be merged after the conflict is solved?

CC: @asolove @samwgoldman @mroch @nmote

mvitousek:
It looks like #6694, which was merged last year, contained similar documentation for $Shape. Thank you for the contribution, and sorry we didn't get a chance to review and merge your diff; feel free to reopen if there's anything in your commit that didn't get included in the docs as they exist now which you'd like us to have a closer look at.

Issue #5314, opened 2017-11-10
Add HTTP methods union instead of a string type to the RequestOptions


NameFILIP:
Can we change the `method?: string` (https://github.com/facebook/flow/blob/master/lib/bom.js#L935) definition to something like `method?: MethodType`, where `MethodType` is:

```
type MethodType = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH';
```

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

PR #5315, opened 2017-11-10
Add type definition for HTTP methods

NameFILIP:
For the issue https://github.com/facebook/flow/issues/5314

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -531,7 +531,7 @@ declare class XDomainRequest {
     ontimeout: (ev: Event) => any;
     responseText: string;
     contentType: string;
-    open(method: string, url: string): void;
+    open(method: MethodType, url: string): void;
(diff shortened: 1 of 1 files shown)


nmote:
If you are still interested in getting this change in, could you link API docs supporting this change?

nmote:

NameFILIP:
Hi @nmote, by API docs do you mean something like https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods ?

nmote:
Something like that, but that looks to be the HTTP spec. I'm looking for something that documents the JS functions you've modified. In this case, I would expect those API docs to say that only a valid HTTP methods could be passed as the `method` argument.

goodmind:
https://fetch.spec.whatwg.org/#concept-method

PR #5375, opened 2017-11-20
Adds support for digest('buffer')

arcanis:
`"buffer"` is a valid parameter for the `digest` function (in fact, it's the default value), but it is currently not recognized by Flow.


mroch:
you snuck this into https://github.com/facebook/flow/commit/d199971cac7500752fec1bd12bac8b8f2a06438b :-P

PR #5390, opened 2017-11-21
Add missing declarations for node copyFile and copyFileSync

chyzwar:
From node 8.5.0 copyFile and copyFileSync were added. This path aims for adding missing functions declarations.

https://nodejs.org/api/fs.html#fs_fs_copyfile_src_dest_flags_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -949,6 +949,26 @@ declare module "fs" {
       flag?: string
     }
   ): void;
+  declare function copyFile(
+    src: string | Buffer | number,
+    dest: string | Buffer | number,
+    flags: number,
+    callback: (err: ?ErrnoError) => void
+  ): void;
+  declare function copyFile(
+    src: string | Buffer | number,
+    dest: string | Buffer | number,
+    callback: (err: ?ErrnoError) => void
+  ): void;
+  declare function copyFileSync(
+    src: string | Buffer | number,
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -949,6 +949,26 @@ declare module "fs" {
...
+  ): void;
+  declare function copyFile(
+    src: string | Buffer | number,
+    dest: string | Buffer | number,
+    callback: (err: ?ErrnoError) => void
+  ): void;
+  declare function copyFileSync(
+    src: string | Buffer | number,
+    dest: string | Buffer | number,
+    flags: number
```
nmote: This can just be optional, and we can avoid the overloading

lib/node.js
```diff
@@ -1028,6 +1048,8 @@ declare module "fs" {
     S_IROTH: number, // 4
     S_IWOTH: number, // 2
     S_IXOTH: number, // 1
+    UV_FS_COPYFILE_EXCL: number, //1
```
nmote: I don't see this constant listed in the docs. Could you provide API docs that include this constant, or remove it?

nmote:
Actually, I just took a look and realized that these have been added since you put up the PR. The typedefs are not exactly the same, so feel free to put up a followup PR if you would like to update them. Feel free to cc me to make sure it gets prompt attention.

PR #5398, opened 2017-11-23
Fix type information for `Buffer.prototype.write`.

arthurschreiber:
`Buffer.prototype.write` can be called in multiple ways which are not supported by the current type definition.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -78,7 +78,9 @@ declare class Buffer extends Uint8Array {
   toJSON(): buffer$ToJSONRet;
   toString(encoding?: buffer$Encoding, start?: number, end?: number): string;
   values(): Iterator<number>;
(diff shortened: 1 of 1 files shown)


arthurschreiber:
Unfortunately, the travis tests are failing because the line number offsets in the `lib/node.js` file changed. 😞

nmote:
@arthurschreiber Sorry for the long delay in responding. If you'd still like to get this in, could you do the following?

* Post a link to some docs (e.g. MDN) which support this change
* Add some tests to `tests/node_tests` which verify that the overloading works correctly? It sometimes doesn't work quite as expected. In this case, it looks like each variant has a different arity so Flow should have no trouble selecting the right one, but it would be nice to be sure.

nmote:
Feel free to reopen if you would still like to get this in. For now I'm closing since I have not yet seen a response.

PR #5484, opened 2017-12-07
Fix markdown typo

eleanombre:
This fixes a typo in the documentation where there is an extra `*` at the end of an ***italicised bold*** sentence which seems to suggest there is more information on the subject! but it's just a typo!!!

Anyway, I fixed it for you :)


calebmer:
Thank you! 😊

PR #5526, opened 2017-12-15
add URLs to flow-upgrade's package.json

jedwards1211:
They should always, always, always be provided

### DIFF
--- a/packages/flow-upgrade/package.json
+++ b/packages/flow-upgrade/package.json
@@ -22,7 +22,7 @@
   "bugs": {
     "url": "https://github.com/facebook/flow/issues"
   },
-  "homepage": "https://github.com/facebook/flow/edit/master/packages/flow-upgrade#readme",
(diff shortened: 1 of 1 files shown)


packages/flow-upgrade/package.json
```diff
@@ -15,6 +15,14 @@
...
     "postpublish": "node ./scripts/postpublish.js"
   },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/facebook/flow.git"
+  },
+  "bugs": {
+    "url": "https://github.com/facebook/flow/issues"
+  },
+  "homepage": "https://github.com/facebook/flow/edit/master/packages/flow-upgrade#readme",
```
mroch: s/edit/tree/: `https://github.com/facebook/flow/tree/master/packages/flow-upgrade#readme`

jedwards1211: good catch!

PR #5527, opened 2017-12-16
Fix the build

popham:
A function's side effects may make it useful as a callback regardless of its return type. `setTimeout` and `setInterval` should admit such functions.


popham:
Fixed by a recent commit (https://github.com/facebook/flow/commit/5fc417c22bb163207eda2a7c475686844c4a2270)

PR #5555, opened 2017-12-22
sketchy-boolean-op lint rule

danwang:
### Status
WIP

### Summary
This adds a lint rule `sketchy-boolean-op` that triggers when the truthiness of the left-hand side of a short-circuiting boolean operator (`&&` or `||`) can be determined statically.

There's four ways this might happen (`{truthy, falsy} × {&&, ||}`) that fall into two categories:

#### Unreachable code
This is either `falsy && ...` or `truthy || ...`. If the static analysis is correct, the expression on the right-hand side should never evaluate. This is usually a bug because the code doesn't do anything. For example:

```jsx
// Precondition: `foo` is falsy
foo && bar(); // The right side should never be evaluated

// Precondition: `foo` is truthy
foo || bar(); // The right side should never be evaluated
```

#### Confusing boolean operators
This is either `falsy || ...` or `truthy && ...`. If the static analysis is correct, the expression on the right-hand side will always evaluate. This is a smell because the value of the LHS expression is never used, but side effects still happen. For example:

```jsx
// Precondition: `foo()` is truthy
foo() && bar(); // The value is always the result of bar()

// Precondition: `foo()` is falsy
foo() || bar(); // The value is always the result of bar()
```

A possible remediation might be to split these into separate statements -- the boolean operators here act like the [comma operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comma_Operator), which can be confusing.

### Motivation
I have a type `Predicate = (Foo) => boolean`, along with some combinators
```jsx
const and: (Predicate, Predicate) => Predicate = ...;
const or: (Predicate, Predicate) => Predicate = ...;
```

I'm afraid users might use `&&` and `||` inappropriately:

```jsx
const isNas: Predicate = ...;
const isCar: Predicate = ...;
// Should be and(isNas, isCar). This only checks if it's a car, but it's not Flow error
const isNascar: Predicate = isNas && isCar;
```

I can't solve this with linter alone because I don't have the information to tell when a dead branch may occur.

### TODO
- [x] truthy && ...
- [ ] truthy || ...
- [x] falsy && ...
- [ ] falsy || ...
- [ ] tests
----

Dear Flow maintainers, this is my first time writing nontrivial OCaml -- any suggestions on style or better ways to do things would be greatly appreciated! 💖


TheSavior:
This seems really useful. What should happen in cases with more than one boolean operator?

For example:
```
const sometimesTrue = Math.random() > 0.5 ? true : false;

var foo = sometimesTrue && false && value;
```

This should still warn about the `false`, right? If so, can you add tests covering these cases?

PR #5560, opened 2017-12-22
WIP: Invert error and warning ordering

popham:
When I use Flow from a terminal I frequently hack `src/common/errors/errors.ml` to reverse the ordering of errors and warnings. This way I don't have to scroll up to the first error message after every check. I would like to expose a `--reverse-errors` flag for `check`, `check-contents`, and `status`. Would such a feature be welcome?

### DIFF
--- a/src/common/errors/errors.ml
+++ b/src/common/errors/errors.ml
@@ -694,18 +694,18 @@ module Cli_output = struct
       let truncate = not (flags.show_all_errors) in
       let one_line = flags.one_line in
       let color = flags.color in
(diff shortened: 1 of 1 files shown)


mrkev:
Oh hmmm, this is an interesting proposal! Sad it didn't get more attention. I am actually a `mosh` user which messes with scroll back, and although I don't bump into this with flow as much I do bump into similar problems with other tools.

I usually just `less`, but reversing the order and displaying the most important stuff at the bottom is clever-- never thought of it.

Do we need to reverse the list though? Is printing the errors after the warnings not enough?

PR #5565, opened 2017-12-24
Model indexer keys as contravariant

### LINKED ISSUE #5566: Covariant indexers are broken on interfaces
When I put a covariant indexer on an interface (analogous to the indexer on `$ReadOnlyArray<+T>`) I get a nonsense error.

```js
declare class Indexed {
  +[key: number]: number
}

declare var indexed: Indexed;
const n: number = indexed[0];
```

```
34: const n: number = indexed[0];
                      ^ computed property. Covariant property `$key` incompatible with contravariant use in
34: const n: number = indexed[0];
                      ^ computed property
```

popham:
Currently Flow applies indexer variance to both indexer key and indexer value. Indexer keys should be modelled contravariant regardless of the indexer variance. Fixes #5566.

This variance stuff is somewhat foreign to me. Here's my reasoning....

Suppose that the variance token only imposes variance on the indexer's value. I intend to show that given an indexer composed of key `k` and value `v`, `k` is contravariant for covariant `v`, contravariant `v`, and invariant `v`.

* **Covariant `v`:** The indexer `+[k]: v` is isomorphic to `get(k): v`. Key `k` only appears in contravariant positions, so covariant `v` implies contravariant `k`.

* **Contravariant `v`:** The indexer `-[k]: v` is isomorphic to `set(k, v): void`. Key `k` only appears in contravariant positions, so contravariant `v` implies contravariant `k`.

* **Invariant `v`:** The indexer `[k]: v` is isomorphic to the combination of `get(k): v` and `set(k, v): void`. Key `k` only appears in contravariant positions, so invariant `v` implies contravariant `k`.

Since key `k` is contravariant regardless of value `v` variance (and equivalently indexer variance), indexer keys should always be contravariant.

*Note: By "isomorphic" I mean that given a data structure containing an indexer, there exists a data structure with identical semantics where the indexer has been replaced with `get(k): v`, `set(k, v): void`, or both `get(k): v` and `set(k, v): void` (mangling may be necessary to avoid name collisions).*

### DIFF
--- a/src/typing/class_sig.ml
+++ b/src/typing/class_sig.ml
@@ -698,10 +698,9 @@ let add_interface_properties cx properties s =
     | Indexer (_, { Indexer.key; value; static; variance; _ }) ->
       let k = Anno.convert cx tparams_map key in
       let v = Anno.convert cx tparams_map value in
-      let polarity = Anno.polarity variance in
       x
-        |> add_field ~static "$key" (None, polarity, Annot k)
-        |> add_field ~static "$value" (None, polarity, Annot v)
+        |> add_field ~static "$key" (None, Type.Negative, Annot k)
+        |> add_field ~static "$value" (None, Anno.polarity variance, Annot v)
     | Property (loc, { Property.key; value; static; _method; optional; variance; }) ->
       if optional && _method
       then Flow.add_output cx Flow_error.(EInternal (loc, OptionalMethod));
--- a/src/typing/flow_error.ml
+++ b/src/typing/flow_error.ml
@@ -1439,7 +1439,7 @@ let rec error_of_msg ~trace_reasons ~source_file =
         | PredicateVoidReturn ->
             "Predicate functions need to return non-void."
         | MultipleIndexers ->
-            "multiple indexers are not supported"
+            "multiple indexer properties are not supported"
         | SpreadArgument ->
             "A spread argument is unsupported here"
       in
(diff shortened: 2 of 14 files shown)


samwgoldman:
This explanation makes sense to me. Will take a closer look when I'm back from the holidays. Thanks!

popham:
Sweet. I've still got a couple of rough edges to mess with this weekend (I think I can remove `report_polarity` from `rec_flow_p` and "$value" mentions shouldn't appear in error messages like they currently do). I assume "holidays" includes New Year, so this should be be done before you're back.

popham:
@samwgoldman, I'm probably going to rebase this PR for a project of mine. Are there any specific criticisms that I should mitigate during that process? In particular, should I narrow its scope?

samwgoldman:
I don't think there's an issue here, except that it has proven difficult to update our codebases to be correct with respect to this change. Sorry for the serious delay here. I'll try to prod my reviewers internally.

popham:
Any progress on this?

popham:
How about some `--indexer-variance` option? I'm mystified how the current implementation can be useful. Otherwise I could maybe contrive a better name for such an option.

My use case is a read-only interface for Uint8Array,

```js
type uint = number;
type u8 = number;
interface BytesR {
  +[position: uint]: u8;
  subarray(begin: uint, end: uint): BytesR;
}
```

I've managed with opaque types, but it's a mess of `any`-casting and stuff like `subarray(bytes, begin, end)`.

nmote:
@popham, I'd like to apologize for how this has gone so far.

If you rebase, I'll take responsibility for getting this merged in a timely fashion. I'll be able to take another look at the effect this has on our codebase, and if we do need to put it behind a flag for now, I'll let you know.

popham:
I'm guilty of far worse, but thanks for the apology. I'll try to get something up in the next couple months.

PR #5574, opened 2017-12-26
fix: return parameter for writable.setDefaultEncoding()

dnalborczyk:
fix return parameter for node Writeable.setDefaultEncoding():

since v6.1.0, method returns 'this'

https://nodejs.org/api/stream.html#stream_writable_setdefaultencoding_encoding

speaking of which, what is the reason that the node.js type library (and others) are not versioned, e.g. with flow-typed?

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1442,7 +1442,7 @@ declare class stream$Writable extends stream$Stream {
     encodingOrCallback?: string | Function,
     callback?: Function
   ): void;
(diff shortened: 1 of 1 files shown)


nmote:
Sorry for the long delay on this. This looks good, I'll get it landed.

We would like to move the core libdefs into flow-typed at some point. Like you said, it would allow us to version them. It would also provide several other benefits. However, we don't know when we will have the time to do this.

PR #5578, opened 2017-12-26
add util.callbackify to node type def

dnalborczyk:
simple type def version for node util.callbackify (similar to promisify), introduced in v8.2.0

https://nodejs.org/api/util.html#util_util_callbackify_original


nmote:
The use of `Function` is unsafe, but it's used throughout the module, so I think it's okay here. We'll have to clean it up at some point, but this additional function shouldn't make it too much harder.

Issue #5584, opened 2017-12-28
$ObjMap output types conflated into union


freddi301:
keywords: $ObjMap

[flow try](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAmBTAxtAhgJzaJAK4B2GALgJawmgawC2ADrAM5oA8AggDSgBCfAMIA+ABSQAXKDEAjafwCUoALwjQQvgHNpY7NK7K1AxQdXqhAbmTkAnkwIAVWClirQAb1CuSaabNhYaDRsEj5yNAAPcmlWclxKEi1QAF9EOwdQAGVybAj3L3IXNmlnVwBtAF0+elII3GkSIgZZNFxU6wyCABk0ElYOLKJZOLy0dRUZRFBQbApqEl1WYdGI6SGR3IijdQ3VtEQd7K2CYxyxy1Bken7yUESMd33zUGeAalAARmtEEFn5misX5gRwAeQAIqDpFo+m1sNBKAAvAhiDChEiwO5dDAAC0wAGtFIgbnF7iQimUSqBev0OJSqhNZi9WC8xF4AHSc1jhYqsaTYMSsdlFVysZQpRTWEl3RJFISwOptaQ0gZNFptRnYZmsjlcmoK8lK2aC9m1Q24cWS4mAu5m8iUrgA-rKvoDemVTVO8zTOiMFjsMSy2CU1h6J2KPhB+WK3CBsiKK2IYhkKg0UDBfpoAbnNaeUDlfFoWyxeKJLSVfkkWypcQ+-brE6HaQAElBsgAVgBZbBMDg+nNoHg+wYrE61mYzZabMbrUdjH1HMRh1OLGRT+vZOfbF57E6LjcDo6HxDqDw+-DkIi4WhiAAMldsVrS1turxOkQbYwKvpj0lvPNFaQqg6a4bXTV0s0idwM3YUN9kiRNpTJOUDXqKDJhgyDTVQtofiQoMQ3Q8DM1YSJhV5PCwKIqMcNwSIxBINQSAAKgAJkUQV3yJIA)

SamChou19815:
The issue seems to be fixed a long time ago

Issue #5366, opened 2017-11-19
ObjMap example broken since 0.57.0


trevorah:
Looks like the changes to `$ObjMap` broke something in the example: https://flow.org/en/docs/types/utilities/#toc-objmap

0.56.0:

<img width="980" alt="screen shot 2017-11-19 at 15 45 06" src="https://user-images.githubusercontent.com/298742/32992314-f80e6812-cd40-11e7-9c81-cf675003e3e2.png">

0.57.0 and higher:

<img width="980" alt="screen shot 2017-11-19 at 15 45 19" src="https://user-images.githubusercontent.com/298742/32992315-038d16fc-cd41-11e7-9476-f127249064a6.png">

I'd love to know if this is a bug or a changing api, as the `$ObjMap` example is almost identical to something I'm trying to type up at the moment.

Thanks!

https://flow.org/try/#0PTAEAEDMBsHsHcBQiSmgUwC4HIDOp4AnAS03VAENRMBPAB3QwDdHRIBXAOwGNNjZO1ABYVM1CgGt0+KgAMAFAEpQAXgB8oAGqzKnACahCWdoU4zQs7aHml8RzCcG0GixM-IBRAB6ZCFXgBKxqYAKvTkKqAAPJpq8kqqGprK6lrIHDx8AoZcUQCCADSgAPIAXKAA3gDaUjTluL7EnADmALrlCal5AL5xsOXFiuUAJMUARgBWALIUdFHFRd6+-phBDqHhGhWIoIbBguMT6LwAdLW48rCKJ0Z67Nzo8f7cRRIpGofHmCcUuLjEzU48meRQqoBq7VAsAhCW6ilBcIA3IhushuAIGlDVJUdpQOu9qIR2OgCrixvjEqBsJBYLBsCjkYh5ESgVcfuUxrSMBROIpEaBUMUJEyWZdruTQA0SC0+btBcKgA

vkurchatkin:
This just seems like a very bad example that is not supposed to work. This works fine:

```js
type ExtractReturnType = <V>(() => V) => V

declare function run<A, O>(o: O): $ObjMap<O, ExtractReturnType>;
    
const o = {
  a: () => true,
  b: () => 'foo'
};

(run(o).a: boolean); // Ok
(run(o).b: string);  // Ok
```

trevorah:
@vkurchatkin that fixes the return type, but the args are no longer enforced:
```js
type ExtractReturnType = <V>(() => V) => V

[...]

PR #5282, opened 2017-11-06
Update $ObjMap example for Flow v0.57

julienw:
Flow v0.57 is more strict around generic typing. The example here fails because in the passed object the properties' functions' return values have different types.

We use `any` here because we don't need to check the functions' return types in the argument, yet ExtractReturnType can still extract it.

See [the current example failing in Try](https://flow.org/try/#0PTAEBsFMBcHIGdQHcBOBLalQENTQJ4AOkUAbiaAGYCuAdgMbRoD2teAFttHtgNaSJcAAwAUASlABeAHygAakJy0AJqBQxqKWoNBCFoERkTromtgWJiAUBawBRAB7QU2RgCUNWgCpEsk0AA8ctIi4lKychIy8lZWNAxMrGp0AQCCADSgAPIAXKAA3gDa-Ph58M5otADmALp5YdGpAL4hzHlZYnkAJFkARgBWALLYhAFZmY7OrtAept6+svlWoGqebH39kIwAdCXwIsxi2+rK1PSQoa70mbxRshtb0NvY8PBoVbQiV5n5oMV1oGY-zCTTEP1BAG4rE1YvRWOVAVICsscPU7ngUNRIOkUb00eFQLBKMxmLBoVCrCJMZ9Ds88r0SVBsLQxBDQCBsrxKdSDkc8aByuhqqyVhyslyOV1HMRGHYUChmChuXRedt+QzmEyWWyOQA5ZjETK9UBoHSCypVKyS6WPOUKpU82n0NkrV1u0VgfWG0D0ZAvUC0ZjcSocLCK96VbDgQEDR5WIA).

[...]

PR #5597, opened 2017-12-31
Fix $ObjMap documentation example.

vicapow:
The `$ObjMap` example in the doc is broken. Looks like this happened between v0.56 and v0.57

Using the existential type was the only solution I could find.

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -354,11 +354,11 @@ getProp({a: 42}, 'b'); // Error: `b` does not exist
 
 `ObjMap<T, F>` takes an [object type](../objects) `T`, and a [function type](../functions) `F`, and returns the object type obtained by mapping the type of each value in the object with the provided function type `F`. In other words, `$ObjMap` will [call](#toc-call) (at the type level) the given function type `F` for every property value type in `T`, and return the resulting object type from those calls.
 
-Let's see an example. Suppose you have a function called `run` that takes an object of thunks (functions in the form `() => A`) as input:
+Let's see an example. Suppose you have a function called `run` that takes an object of thunks as input:
 
 ```js
 // @flow
-function run<A, O: {[key: string]: () => A}>(o: O) {
+function run(o) {
   return Object.keys(o).reduce((acc, k) => Object.assign(acc, { [k]: o[k]() }), {});
 }
 ```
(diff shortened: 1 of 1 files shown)


vicapow:
looks like this is a duplicate of https://github.com/facebook/flow/pull/5282/files but I think this solution is better as it avoid using `any`

vicapow:
is this good to land?

JoelMarcey:
@nmote @mroch @avikchaudhuri @gabelevi

vicapow:
Checking in again. Is there something I forgot to do?

mroch:
I think using `{}` instead of `*` also works, and it's a good idea to avoid using the existential type when possible. I'm gonna import it and ask @samwgoldman and @calebmer to take a look, though. Thanks!

vicapow:
@facebook-github-bot cool but what exactly does "imported" mean?

JoelMarcey:
@vicapow Hi. It means that we have taken this PR and are reviewing it internally since Flow has an internal source of truth that we push out using our `shipit` tool. And I just looked at the review internally and there has been some comments and action on this PR. And I am commenting on it too now.

wchargin:
Friendly ping. It'd be great if this could be merged. The docs are still wrong and causing confusion (source: I was confused). :-)

(+1 to @JoelMarcey 's suggestion to use `{}` instead of `*`, because `$ObjMap` requires that `O` be of object type, and in general the existential is not. It slightly improves the error message when calling `run("not_an_object")`: you additionally get "string is incompatible with object type".)

JoelMarcey:
@wchargin It was @mroch's suggestion, actually 😄 --- but, I have pinged the internal diff on this again.

vicapow:
Hello! Just following up again. Is there anything else you'd like me to change on this?

JoelMarcey:
@vicapow Hi. Heard internally that the consensus is that as written we shouldn't land this. The Flow team wants to remove * from Flow, so adding it to docs is undesirable.

cc @gabelevi

bwestergard:
@JoelMarcey Be that as it may, the question remains: how is `$ObjMap` supposed to behave? Should it be possible for the object properties' functions' return values have different types, as discussed in #5282, #5584, and #5366, among others?

@vkurchatkin [suggested](https://github.com/facebook/flow/issues/5366#issuecomment-345762578) that it _should_ be possible, and that this broken example indicates a regression.

This is an important issue, because this $ObjMap functionality allows one to write validation logic much more concisely than is otherwise possible, as @andreypopp's [validated](https://github.com/andreypopp/validated#object) library demonstrates.

PR #5606, opened 2018-01-03
Support inline annotations of function class fields

samwgoldman:
This helps mitigate an issue that people run into, where the following code is an error:

```
class C<T> {
  f = (x: T): T => x;
  m(x: T): T { return this.f(x); }
}
```

Because `f` is unannotated, we infer it's type. We check the bodies of polymorphic classes (and functions) by instantiating their type parameters with their bounds, usually `empty` and `mixed`.

This is normally fine, but In the code above, `f` is inferred to be `(empty => empty) | (mixed => mixed)`. Before this change, the only reasonable work-around was to add an annotation the code like this:

```
class C<T> {
  f: T => T = (x: T): T => x; // note the T => T annotation here
  m(x: T): T { return this.f(x); }
}
```

The insight here is that the initialization value for `f` is fully annotated. When that is the case, we can "lift" the annotation from the function expression into a function type annotation.

### DIFF
--- a/src/parser/ast_utils.ml
+++ b/src/parser/ast_utils.ml
@@ -54,3 +54,38 @@ let partition_directives statements =
     | rest -> List.rev directives, rest
   in
   helper [] statements
+
+
+let function_type_of_function =
+  let open Option.Monad_infix in
+  let rec param_type_of_pattern ?(optional=false) (loc, patt) =
+    let open Pattern in
+    match patt with
+    | Expression _ -> None
+    | Assignment { Assignment.left; _ } ->
+      param_type_of_pattern ~optional:true left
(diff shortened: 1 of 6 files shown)


samwgoldman:
This got a bit stale, and the landscape has changed a bit since I wrote this. Right now class fields, even initialized, allow widening:

```
class C {
  p = null;
  m() {
    this.p = 0; // allowed
  };
}
```

this is similar to local variables:

```
var p = null;
p = 0; // allowed
```

but distinct from object literals:

```
var o = { p: null };
o.p = 0; // error
```

Current discussions in the team support moving the current behavior to be more like object literals, where the type of the property is fixed, asking for annotations when a fixed type can not be clearly determined.

A change in that direction would subsume this change, so I'm closing this in favor of that larger change down the road.

PR #5629, opened 2018-01-09
Allow passing null or undefined to clear timer functions.

zaygraveyard:
When `clearInterval()` and `clearTimeout()` are passed an invalid ID they silently do nothing; no exception is thrown.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -729,8 +729,8 @@ declare function unescape(str: string): string;
 
 declare opaque type TimeoutID;
 declare opaque type IntervalID;
-declare function clearInterval(intervalId?: IntervalID): void;
-declare function clearTimeout(timeoutId?: TimeoutID): void;
+declare function clearInterval(intervalId?: ?IntervalID): void;
+declare function clearTimeout(timeoutId?: ?TimeoutID): void;
 declare function setTimeout<TArguments: Array<mixed>>(
   callback: (...args: TArguments) => mixed,
   ms?: number,


jedwards1211:
@zaygraveyard yes, but passing an invalid ID may be a developer mistake that many would want Flow to report.  It's an awkward thing but there are many cases of type defs being more strict than what an API actually allows for the sake of catching potential bugs.

zaygraveyard:
@jedwards1211 I see your point, but I'm only allowing `null` and `undefined` to be passed not any value (which is what the spec permits).
I have seen this use case multiple times (like in [preact's Clock example](https://github.com/developit/preact/blob/02f717e2bd2b0461b51aa3a96b927f4de99c23cc/README.md#components)), and it allows omitting the useless `if` guard.

nmote:
I'm inclined to leave this as-is. I understand that having an additional `if` guard in some circumstances is cumbersome, but I think the inconvenience is outweighed by the current definition's potential to catch programmer errors. Thanks anyway, though!

jedwards1211:
@nmote I support the goal of catching programmer errors but what mistakes does this actually catch?  I can't think of any examples.

nmote:
The null check forces the programmer to think about that case. When writing or reading the null check, you are forced to consider what should happen in that case. If you choose to do nothing, that's often a legitimate choice, but in my opinion, that case should be considered.

> Maybe your main motivation is just to reduce changes to the core lib defs?

That's not true at all. I've been merging tons of libdef PRs recently, and I believe this is the first one I've closed simply because I don't agree with the change. I've closed others which were abandoned or already addressed in another commit, but the rest I have either merged or sent back to the author requesting specific changes. There's nothing nefarious going on here -- I just think the current libdef is slightly safer.

Issue #5600, opened 2018-01-02
Buffer#buf.swap64() is not supported


kesor:
https://github.com/facebook/flow/blob/aea3635e9b5cd806a1fe83b664eb1939aceda63d/lib/node.js#L77 is missing `swap64()`

According to https://nodejs.org/api/buffer.html#buffer_buf_swap64 `buf.swap64()` was added in: v6.3.0

nicoffee:
Hello! I made PR for this issue: https://github.com/facebook/flow/pull/5888
Please, take a look at it.

jdecked:
Has this been fixed? Given that #5888 & #5653 are both closed, I assume yes, but otherwise, I'd like to work on this.

PR #5653, opened 2018-01-13
Added Buffer swap64() to node definitions

curtiswilkinson:
A quick PR that addresses this issue: https://github.com/facebook/flow/issues/5600
https://nodejs.org/api/buffer.html#buffer_buf_swap64

Have a great weekend y'all!


nmote:
Ah, I just realized this was already addressed in #5888. Yours came first, so apologies that it turned out this way.

Issue #4560, opened 2017-08-08
On Linux, using `--temp-dir` makes flow fail everytime


julienw:
When I use the option `--temp-dir` on Linux, it seems to make the server unaccessible.

I run the command like this: `./node_modules/.bin/flow --temp-dir ./temp-flow`.

Here is the log in the console:
```$ ./node_modules/.bin/flow --temp-dir ./temp-flow
Launching Flow server for /home/julien/travail/git/perf.html
Spawned flow server (pid=15103)
Logs will go to /home/julien/travail/git/perf.html/temp-flow/zShomezSjulienzStravailzSgitzSperf.html.log
The flow server is not responding (0 retries remaining): [processing] /
Out of retries, exiting!
```
Then any subsequent commands will output the 2 last lines only -- likely it tries to use the still present socket with no server behind.

Here is the full log:
```
[2017-08-08 15:07:28] Initializing Server (This might take some time)
[2017-08-08 15:07:28] executable=/home/julien/travail/git/perf.html/node_modules/flow-bin/flow-linux64-v0.52.0/flow
[2017-08-08 15:07:28] version=0.52.0
[2017-08-08 15:07:28] Parsing
[2017-08-08 15:07:33] Building package heap
[2017-08-08 15:07:33] Loading libraries
[2017-08-08 15:07:34] Resolving dependencies
[2017-08-08 15:07:34] Running local inference
[2017-08-08 15:07:35] Calculating dependencies
[2017-08-08 15:07:35] Merging
[2017-08-08 15:07:39] Done
[2017-08-08 15:07:39] Server is READY
[2017-08-08 15:07:39] Took 11.146234 seconds to initialize.
[2017-08-08 15:07:39] sending (0 errors) and (warnings from 18 files) to 0 subscribed clients (of 0 total)
[2017-08-08 15:07:39] sending (0 errors) and (warnings from 18 files) to 0 subscribed clients (of 0 total)
[2017-08-08 15:07:39] sending (0 errors) and (warnings from 18 files) to 0 subscribed clients (of 0 total)
... More times the same line below
```

Note that running flow on the same project without `--temp-dir` works fine.
Using version 0.52.

mroch:
Ooh! This looks just like a bug that is making our tests unreliable on Travis but we haven't been able to repro on our own.

When you say the last line of the logs is repeated more times below, how many more times? In Travis, it seems to be many many many more times, so many that Travis kills the job... possibly infinite. 

Also what OS are you using?

[...]

Issue #270, opened 2015-02-21
Investigate why Flow doesn't work on 32 bit Ubuntu


gabelevi:
We should look into what exactly doesn't work on 32 bit machines and if it is easy to fix.

[Reported on reddit](https://www.reddit.com/r/javascript/comments/2wn3rf/flow_the_js_typechecker_by_facebook_no_longer/cosdy92).

jwatzman:
My suspicion is that there's some subtle pointer size bug in hh_shared.c where we assume pointers are 8 bytes or something like that. Skimming the code, the hash stuff looks suspicious -- it uses "unsigned long" whose size I think goes from 8 to 4 bytes when you go 64 to 32 bit; that probably should be uint64_t. Possibly a similar error in ocaml land, assuming things about size. This stuff is all fixable.

May also be in the ocaml *runtime* where we integer overflow their integers, due to needing bigger int sizes for hash table stuff or something like that. This will be much harder to deal with.

jwatzman:
Also @pikatchu just explained why -- the dep table is a combo of 2 32-bit hashes, into one 64-bit value so you can CAS it atomically.

TrySound:
Is this still relevant?

Xunnamius:
I use an Odroid (so ARM) Ubuntu 16.04 and it's relevant for me. Fortunately, I can run Flow through ESLint on Sublime on Windows and sidestep having to use the CLI tool at all, but this is still unfortunate.

kiler129:
Also this fails with very weird message which is hard to debug, which I'm going to leave here for anyone googling and wtf-ing:

```
Starting: /usr/bin/yarn run eslint --no-color /opt/jetbrains/TeamCity/buildAgent/work/f613c80626aaf1ff/assets/
in directory: /opt/jetbrains/TeamCity/buildAgent/work/f613c80626aaf1ff
yarn run v1.7.0
$ /opt/jetbrains/TeamCity/buildAgent/work/f613c80626aaf1ff/node_modules/.bin/eslint --no-color /opt/jetbrains/TeamCity/buildAgent/work/f613c80626aaf1ff/assets/	
    "file" argument must be a non-empty string
    TypeError: "file" argument must be a non-empty string
        at normalizeSpawnArguments (child_process.js:383:11)
        at Object.spawnSync (child_process.js:519:38)
[...]

PR #5659, opened 2018-01-14
update README docs about windows compatibility

krishnasaga:
in-order to save time for developers who are working on windows without getting exited and try flow on first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Flow works with:
 
 * Mac OS X
 * Linux (64-bit)
-* Windows (64-bit)
+* Windows (64-bit) ( With some problems )
 
 There are [binary distributions](https://github.com/facebook/flow/releases) for each of these platforms and you can also build it from source on any of them as well.
 


kumarharsh:
It might be useful to list those problems (maybe like to a super-issue or something)

There are a bunch of problems, but I keep coming back every few weeks and trying to get flow running, just to get disappointed.

mrkev:
I'd be down to better document flow's issues on windows. @sagar12861 are you still interested in this? If so, you could sign the CLA and we could keep on developing this PR.

sagar12861:
@mrkev I signed in the CLA.

mrkev:
Sweet. Rather than this entry as parens next to "Windows 64-bit" I think it'd be helpful to have known problems section like I sometimes see in the bottom of other READMEs linking to specific issues or a super-issue that'd make it clear what said problems are. Which ones did you encounter?

kumarharsh:
@mrkev should we create another issue for it? There are about 7 pages worth of stuff broken on Windows: https://github.com/facebook/flow/issues?utf8=%E2%9C%93&q=windows

Some of the issues I found in the non-LSP version is the runaway spawning of multiple flow processes with none of them actually sending any diagnostics or autocomplete data - they just eat up memory.

The Flow LSP works *slightly* better - but after about 5-10 edits the LSP stops responding because some hover or other provider (I don't remember which, exactly) gets stuck and never gives any result back to vscode. Then there is the whole issue of the flow-for-vscode plugin not working as expected on Windows which causes additional problems while diagnosing these issues.

sagar12861:
@kumarharsh this pull request is about updating the README about windows issues. As It helps users to take better desession to use flow in their code base are not.

mrkev:
@kumarharsh your search includes closed issues and PRs. I'm seeing [2 pages](https://github.com/facebook/flow/issues?utf8=✓&q=is%3Aissue+is%3Aopen+windows) (though not all of them seem to be issues with Windows: https://github.com/facebook/flow/issues/4560 is there for example, because someone [mentioned](https://github.com/facebook/flow/issues/4560#issuecomment-320984269) it doesn't work on Linux but does on Windows. Same with https://github.com/facebook/flow/issues/270).

mroch:
> changing it to include no specific list of issues or more information about when said problems surface doesn't seem too useful

I agree. let's put together a concrete list of issues before adding this kind of warning. perhaps the way watchman talks about beta/less maintained platforms works for us: https://facebook.github.io/watchman/docs/install.html#system-requirements

PR #5663, opened 2018-01-15
lib: Improve Object.values and Object.entries

zertosh:
Use `$Keys` and `$Values` to better reflect that actual types of `Object.values` and `Object.entries`. Fixes #2804.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -32,7 +32,7 @@ declare class Object {
     static create(o: any, properties?: any): any; // compiler magic
     static defineProperties(o: any, properties: any): any;
     static defineProperty(o: any, p: any, attributes: any): any;
-    static entries(object: any): Array<[string, mixed]>;
+    static entries<T: Object>(object: T): Array<[$Keys<T>, $Values<T>]>;
     static freeze<T>(o: T): T;
     static getOwnPropertyDescriptor(o: any, p: any): any;
     static getOwnPropertyNames(o: any): Array<string>;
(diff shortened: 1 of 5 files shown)


vkurchatkin:
This is really unsound in general.

zertosh:
@vkurchatkin how so?

vkurchatkin:
Because $Keys and $Values return statically known properties, while an object might have more properties at runtime.

TrySound:
@vkurchatkin Which ones? Prototype is not handled.
> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).

samwgoldman:
Vladimir is right. There are two issues.

Width subtyping:

```
var o: { p: number, q: string } = { p: 0, q: "", foo: "foo" };
var keys: "p" | "q" = Object.keys(o); // lies
var vals: number | string = Object.values(o); // lies
```

Proto vs. own

```
var o: { p: number, q: string } = Object.create({ p: 0, q: "" });
var keys: Array<"p" | "q"> = Object.keys(o); // lies
var vals: Array<number | string)> = Object.values(o); // lies
```

Exact object types are great for the former case, because they by definition exclude values having more properties. For the latter case, we really need to provide better mechanisms to specify object layout as part of types, since so many APIs depend on it. It's on our radar.

Closing this because the implementation is unsound in a way that I'm not comfortable with. I think if we can continue wearing the hair shirt for a bit longer we can come up with some useful+principled solutions to make idiomatic code pass with simple types.

Thanks!

PR #5665, opened 2018-01-15
Fix arity of clearInterval, clearTimeout

STRML:
It's okay for their argument to be null/undefined, which appears to have been the intended type, but `clearTimeout()` does not make sense.

Fixes type errors in 0.63.x where (valid) code that passed an `?IntervalID` or `?TimeoutID` threw errors.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -729,8 +729,8 @@ declare function unescape(str: string): string;
 
 declare opaque type TimeoutID;
 declare opaque type IntervalID;
-declare function clearInterval(intervalId?: IntervalID): void;
-declare function clearTimeout(timeoutId?: TimeoutID): void;
+declare function clearInterval(intervalId: ?IntervalID): void;
+declare function clearTimeout(timeoutId: ?TimeoutID): void;
 declare function setTimeout<TArguments: Array<mixed>>(
   callback: (...args: TArguments) => mixed,
   ms?: number,


mrkev:
Good catch, sad no one looked at this earlier. I'd argue that passing `null` or `undefined` is undesirable too so this shouldn't have `?` at all.

STRML:
Passing `null` or `undefined` is not a bug and sometimes desirable in code where you want to clear an interval if it exists, e.g. on a destructor.

mrkev:
Oh hmm https://github.com/facebook/flow/pull/98

mrkev:
@STRML I personally would +1 this change, though idk what the stance of internal reviewers might be. In any case, @facebook-github-bot is failing to import this PR because it's old. Could you rebase it?

STRML:
Rebased it @mrkev.

kassens:
Allowing `null` means to allow a `?TimeoutID`, that one might have I suppose for people who use `?` over `| void`.

mroch:
my concern was that `setTimeout` returns a number that is an index into the list of timers, and `undefined` coerces to `NaN` so definitely isn't a valid timer ID, but `null` coerces to `0`.

however, https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#list-of-active-timers says that the timeout ID must be greater than zero, so I suppose that means `null` is ok.

PR #5673, opened 2018-01-17
Fix send type

mjesun:
`ChildProcess#send` can accept any object, since it will be serialized to JSON. Passing an `Array` now throws, but by using `any` it will work.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -239,7 +239,7 @@ declare class child_process$ChildProcess extends events$EventEmitter {
   disconnect(): void;
   kill(signal?: string): void;
   send(
-    message: Object,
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -239,7 +239,7 @@ declare class child_process$ChildProcess extends events$EventEmitter {
   disconnect(): void;
   kill(signal?: string): void;
   send(
-    message: Object,
+    message: any,
```
nmote: Could you change this to `mixed`? It won't really make a difference since this is an input position, but I'd prefer to avoid `any` where possible.

nmote:

PR #5680, opened 2018-01-18
Add validationMessage to HTMLSelectElement

kyledetella:
This adds the `validationMessage` property to the `HTMLSelectElement`.

* MDN – https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement
* Apple Webkit – https://developer.apple.com/documentation/webkitjs/htmlselectelement/1632209-validationmessage?changes=latest_major


nmote:
I just realized this change was made in another commit

PR #5706, opened 2018-01-24
Add documentation for $Shape<T>

sanohin:
Part of #2464

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -18,6 +18,7 @@ Table of contents:
 - [`$TupleMap<T, F>`](#toc-tuplemap)
 - [`$Call<F>`](#toc-call)
 - [`Class<T>`](#toc-class)
+- [`$Shape<T>`](#toc-shape)
 - [`$Supertype<T>`](#toc-supertype)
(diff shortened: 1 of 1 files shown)


sanohin:
Is that ok to land?

vicapow:
Is there hesitation in landing this?

mrkev:
Oh, hello

mrkev:
Ok, got it reviewed, and someone from the core team offered some insight. The main points:

- It was implemented for it's utility when typing React in a somewhat ad-hoc way. There might be edge-cases that are not considered, and the intended semantics have not been fully thought out.
- We can't stop people from digging through the code and discovering this stuff. It's used on React libdefs so it's also not going away soon, but there's hesitation on documenting it and making it "official"-- especially with some other work the team is doing around objects at the moment.
- "I don't think we should publicly document it yet."

It'll probably reach a point at which the team will be comfortable documenting it though-- especially considering it's shown it's utility in a number of scenarios.

sibelius:
@mrkev I think flow should be more open on what is the roadmap or what the core team is working on

basically most of what React team does or relay meeting notes 

this could get more people from community engaged with flow

PR #5728, opened 2018-01-27
Type ReactDOM.render to correctly indicate that it may return null

jeffchan:
It's possible for `ReactDOM.render` to return null.

See issue: https://github.com/facebook/react/issues/10309#issuecomment-318392589
And demo: https://jsfiddle.net/x7c7bdh0/

### DIFF
--- a/lib/react-dom.js
+++ b/lib/react-dom.js
@@ -17,7 +17,7 @@ declare module 'react-dom' {
     element: React$Element<ElementType>,
     container: Element,
     callback?: () => void,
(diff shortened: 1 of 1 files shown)


Brianzchen:
react-dom has been moved to flow-typed, if you're still interested in this change you raise a PR to change it there
https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/react-dom_v16.x.x/flow_v0.117.x-/react-dom_v16.x.x.js#L10

PR #5729, opened 2018-01-27
module.system.node.main_field

### LINKED ISSUE #5725: module.system.node.main_field
**TL;DR:** A new `module.system.node.main_field` option for trying to resolve different `package.json` fields (other than `package.json#main`)

```ini
[options]
module.system.node.main_field='source'
module.system.node.main_field='main'
```

---

## `package.json#main` or `package.json#module`

It is not uncommon to see `package.json` files that look like this:

```json
{
  "name": "my-package",
  "main": "dist/cjs/index.js",
  "module": "dist/esm/index.js"
}
```

Tools like Webpack, Rollup, and others will look at fields like `package.json#module` in addition to `package.json#main` when resolving dependencies.

## Workspaces

The specific reason that I want Flow to have something similar is to support [Workspaces](https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/).

Right now in order to support workspaces you need to have some very brittle `module.name_mapper` configuration:

```ini
# .flowconfig
[options]
module.name_mapper='^@scope\/\([a-zA-Z0-9_\-]+\)$' -> '<PROJECT_ROOT>/packages/[^/]+/\1/src/index'
```

> Effectively: `@scope/*` &rarr; `/packages/*/src/index.js`

This can break in a number of different ways, but I won't dive into all those.

## `module.system.node.main_field`

If we could configure Flow like:

```ini
[options]
module.system.node.main_field='source'
module.system.node.main_field='main'
```

We could write our workspace `package.json` files like this:

```json
{
  "name": "my-package",
  "main": "dist/index.js",
  "source": "src/index.js"
}
```

And Flow could check for `package.json#source` before `package.json#main`.

### Implementation

The basic resolve process in Flow would look like this:

For each `main_field`:
1. Check if the `package.json` has a field with that name
2. If the `package.json` does not have the field, continue to next `main_field`
3. If it field exists, check for a file at that location.
4. If no file at that location exists, continue to the next `main_field`
5. If the file exists, use that file as the resolved module and stop looking

The relevant bits of code seem to be:

https://github.com/facebook/flow/blob/9c75e15cc8809170933236c2add89a4646b04f44/src/parser_utils/package_json.ml#L40-L63
https://github.com/facebook/flow/blob/2ff56f4e7620b2d3d28cb862047449d3cbfd615b/src/services/inference/module_js.ml#L328-L371

gabelevi:
(Copy pasting docs for commit message 😛 )

Flow reads `package.json` files for the `"name"` and `"main"` fields to figure
out the name of the module and which file should be used to provide that
module.

So if Flow sees this in the `.flowconfig`:

```
[options]
module.system.node.main_field=foo
module.system.node.main_field=bar
module.system.node.main_field=baz
```

and then it comes across a `package.json` with

```
{
  "name": "kittens",
  "main": "main.js",
  "bar": "bar.js",
  "baz": "baz.js"
}
```

Flow will use `bar.js` to provide the `"kittens"` module.

If this option is unspecified, Flow will always use the `"main"` field.

See [this GitHub issue for the original motivation](https://github.com/facebook/flow/issues/5725)

Resolves #5725

### DIFF
--- a/newtests/package_json_changes/irrelevantChangeMainField.json
+++ b/newtests/package_json_changes/irrelevantChangeMainField.json
@@ -1,4 +1,4 @@
 {
   "name":"foo",
-  "other_main":"foo",
+  "other_main":"foo"
 }
--- a/newtests/package_json_changes/irrelevantChangeMainField.json
+++ b/newtests/package_json_changes/irrelevantChangeMainField.json
@@ -1,0 +1,4 @@
+{
+  "name":"foo",
+  "other_main":"foo",
+}
--- a/newtests/package_json_changes/node_flowconfig_with_main_field
+++ b/newtests/package_json_changes/node_flowconfig_with_main_field
@@ -1,0 +1,11 @@
+[ignore]
+
+[include]
+
+[libs]
+
+[options]
+all=true
+module.system=node
+module.system.node.main_field=main
+module.system.node.main_field=other_main
--- a/newtests/package_json_changes/test.js
+++ b/newtests/package_json_changes/test.js
@@ -69,4 +69,12 @@ export default suite(({addFile, removeFile, exitCode, flowCmd}) => [
       .waitForServerToDie(2000)
       .serverRunning(true)
   ]).flowConfig('haste_flowconfig'),
+
+  test('node - When using main_fields, a change which resolves to '+
+    'the same main file should NOT kill the server', [
+    addFile('start.json', 'package.json'),
+    addFile('irrelevantChangeMainField.json', 'package.json')
+      .waitForServerToDie(2000)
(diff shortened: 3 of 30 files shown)


kentcdodds:
Hi folks! This feature suddenly became kinda important [for me](https://twitter.com/kentcdodds/status/969644458160676872). What can I do to help get it merged?

goodmind:
@gabelevi are you still plan to merge this?

mischnic:
I would love to see this feature get added and have resolved the merge conflicts at https://github.com/mischnic/flow/tree/main_fields. @gabelevi  Do you want to update this PR or should I open a new one?

PR #5763, opened 2018-02-03
Add "destroy()" to stream$Stream

piglovesyou:
I'm getting this error in my project.
`Error:(124, 5) property `destroy`. Property not found in stream$Writable` 
Let me fix it.

https://nodejs.org/api/stream.html#stream_writable_destroy_error

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1213,7 +1213,6 @@ declare class http$Agent {
 }
 
 declare class http$IncomingMessage extends stream$Readable {
-  destroy(error?: Error): void;
   headers: Object;
   rawHeaders: Array<string>;
   httpVersion: string;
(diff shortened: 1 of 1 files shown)


piglovesyou:
> mixed

Sure, I force pushed to mine. Sorry, I didn't understand the next one.

> It also looks like the object itself (stream$Stream).

piglovesyou:
fixing

piglovesyou:
@mrkev Done. Glad if you check this again.

nmote:
Looks good, I will get this merged

nmote:
Actually, when I took a closer look I realized that the base `stream` class doesn't actually have the `destroy` method. In both the implementation and the docs, `destroy` appears on the individual subclasses of `stream`. I think the Flow libdefs should remain faithful to the implementation, even if it doesn't seem to make a difference right now.

Furthermore, `destroy` returns `this`, not `void`.

There are a few other open PRs that address this issue and seem a bit closer to the ideal solution. Even though they were opened later, I think it will be fastest to close this and work with the other PR authors.

PR #5779, opened 2018-02-06
Fix for setState definition

kikoseijo:
Does not match the specs removed to pass latest flow check.

  setState(
    partialState: $Shape<State> | ((State, Props) => $Shape<State> | void),
    callback?: () => mixed,
  ): void;


mrkev:
Wait, this did match the spec, and the change makes it invalid javascript?

Closing 😅

Issue #5594, opened 2017-12-30
Unsupported option specified! (unsafe.enable_getters_and_setters)


dsherida:
I'm getting the message `.flowconfig:72 Unsupported option specified! (unsafe.enable_getters_and_setters)` which is an option defined in the default .flowconfig while using create-react-native-app. This option is also specified in the [Flow documentation](https://flow.org/en/docs/config/options/#toc-unsafe-enable-getters-and-setters-boolean) so why am I getting this error?

idiostruct:
Its been removed in [0.62.0](https://github.com/facebook/flow/blob/master/Changelog.md#0620). Getters and setters are enabled by default.

aikar:
That was a horrible idea to make everyones process fail hard instead of simply ignoring it or printing a deprecation warning.

TrySound:
@aikar It's major release. Just read changelog.

Aravindhan12:
I'm using flow bin version 0.63.1. Same error as @dsherida mentioned. Tried Downgrading to 0.62.0 but still the error remains the same.

While the yarn run flow command runs, following error appears,
.flowconfig:72 Unsupported option specified! (unsafe.enable_getters_and_setters)
error Command failed with exit code 8.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

TrySound:
@Aravindhan12 Just remove that option from your config.

Aravindhan12:
Finally I have fixed this by downgrading to Version 0.56.0.

@TrySound  As you said I have removed the option. Now when the yarn run flow command runs, I received a error as  Launching Flow server for Wrong version of Flow. The config specifies version ^0.56.0 but this is version 0.62.0 Could not start Flow server!  error Command failed with exit code 78. 

So I downgraded to 0.56.0. After that when I run the  yarn run flow command, lot of errors was thrown in all the files with the message  get/set properties not yet supported. 

To fix it I have added the unsafe.enable_getters_and_setters in the .flowconfig (which I have removed earlier) and it works.

Hope this helps someone.

TrySound:
@Aravindhan12 Downgrading doesn't help anybody. Kill all flow processes, remove version mark from config, remove unsafe.enable_getters_and_setters and run `flow`.

TrySound:
@Aravindhan12 Also make sure you have latest versions installed globally and locally.

Aravindhan12:
[...]

PR #5782, opened 2018-02-07
Docs: Remove `unsafe.enable_getters_and_setters`

iray-tno:
The `unsafe.enable_getters_and_setters` option was already removed in 0.62.0.

### DIFF
--- a/website/en/docs/config/options.md
+++ b/website/en/docs/config/options.md
@@ -46,7 +46,6 @@ can be overridden with command line flags.
 * [`suppress_type`](#toc-suppress-type-string)
 * [`temp_dir`](#toc-temp-dir-string)
 * [`traces`](#toc-traces-integer)
-* [`unsafe.enable_getters_and_setters`](#toc-unsafe-enable-getters-and-setters-boolean)
 
 #### `all` _`(boolean)`_ <a class="toc" id="toc-all-boolean" href="#toc-all-boolean"></a>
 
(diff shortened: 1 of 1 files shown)


iray-tno:
Refs #5594

pronebird:
Please merge since `unsafe.enable_getters_and_setters` is not available anymore.

iray-tno:
Should we add caution instead of just deleting description maybe?

pronebird:
@iray-tno good point. can be a deprecation notice.

mroch:
I could go either way on adding a deprecation warning. It's nice to document it if people are still using this version, but at some point it's so old that it's just clutter... 0.61 might already be that old. I don't have a good sense of whether these old versions are still used, so what do you guys think?

To add a notice, check out the `{% until 0.61 %}` tag. I'd say also reword the description to talk about the option in past-tense.

iray-tno:
@pronebird @mroch Thanks for your replies.

I was thinking the same thing.
A deprecation warning can be useful. But also it might be confusing for newcomers.

I couldn't find another example of a deprecation warning. And um... Im not so confident in English so you guys can overtake my pull request. 😅

I guess we better to add a deprecation warning to an error message instead of documents.

mroch:
sorry, something went wrong and this failed to land. trying again!

mvitousek:
Looks like this change was already made with #6336. Thanks for the PR, sorry we didn't land this sooner!

PR #5820, opened 2018-02-15
correct type for readdir

rgbkrk:
Includes potential for readdir returning an `Array<Buffer>`, which happens when encoding is set to `'buffer'`. https://nodejs.org/api/fs.html#fs_fs_readdir_path_options_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -779,16 +779,16 @@ declare module "fs" {
   declare function readdir(
     path: string,
     options: string | { encoding: string },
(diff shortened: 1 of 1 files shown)


nmote:
I'm worried that this will make it a lot harder to use this function in the common case where it actually does return a `string`. Could you overload this function declaration to capture the type more precisely, and add tests in `tests/node_tests` to verify that it behaves as desired?

rgbkrk:
I've switched over to TypeScript so I can't justify putting more work into this PR. Feel free to use the code in the commits, with or without credit.

PR #5837, opened 2018-02-18
Add WebAssembly libdefs

pnevyk:
Although WebAssembly is still in a working draft stage, its API is *stable* enough. Typescript types can be found [here](https://github.com/01alchemist/webassembly-types/blob/master/webassembly.d.ts).

Also, I am willing to add some tests.


mrkev:
Oh sweet. This should've gotten more attention! Thanks for linking the typescript too! I'll try to make sure I don't forget to import this later in the week 😅

PR #5842, opened 2018-02-19
Fix header formatting in intersections guide

alexshpilkin:
This marks up what should be a header in `intersections.md` and also brings all the other section headers there to the same level. Even if I didn’t get the latter right (I’m sure “Impossible intersection types” shouldn’t a subsection of “Intersection type syntax”, but the analogous markup of `unions.md` seems to think otherwise), the former should be fixed.


mroch:
seems good to me, thanks!

Issue #5647, opened 2018-01-11
Flow just stops reporting errors


alexeyraspopov:
I like using Flow since it provides amazing extension to existent JavaScript code. The type system is efficient, mostly straightforward, and error reports are user-friendly.

I want to thank anyone who's involved in development, it's a great tool.

There is only one thing that bugs be for long time already. You can find it partially reported: https://github.com/facebook/flow/issues/5619, https://github.com/facebook/flow/issues/5586, https://github.com/facebook/flow/issues/5316, https://github.com/facebook/flow/issues/5180, https://github.com/facebook/flow/issues/5495. Sometimes, for some unexpected reasons (mostly after version update), Flow just stops seeing errors (even explicitly made). Me and my colleagues tried tons of ways to work around this: manually killing server instances, looking for temp files that should be cleaned up, etc. Still, Flow can just not report errors.

This is really frustrating since you can't really rely on this tool output. I hope maintainers can prioritize this issue and not just close this ticket as duplication. I'm not familiar with OCaml so there is no even a way to debug this thing locally from my side.

Thank you again for your work and wish we can continue improving Flow.

Alexey

apsavin:
#5586 is closed, #5316 is obvious duplicate of #5180 - but #5180 has a solution and can be closed too. #5495 does not provide enough info.

Can you show your .flowconfig?

alexeyraspopov:
My `.flowconfig` is:

```
[libs]
typings/

[options]
module.system.node.resolve_dirname=node_modules
module.system.node.resolve_dirname=src
esproposal.class_instance_fields=enable
esproposal.class_static_fields=enable
```

I use Node 8.6, macOS Sierra, Flow 0.63.1

apsavin:
You need to remove this part from your config:

```
[options]
module.system.node.resolve_dirname=node_modules
module.system.node.resolve_dirname=src
esproposal.class_instance_fields=enable
esproposal.class_static_fields=enable
```

apsavin:
And instead of `module.system.node.resolve_dirname=src` use something like

`module.name_mapper='^components/\(.*\)$' -> '<PROJECT_ROOT>/src/components/\1'`

for each directory in `src`

alexeyraspopov:
Could you please explain why do I need to remove that part of my config? It is what I found in Flow docs and used based on what I’m able to do in JS with CRA setup.

apsavin:
You don't have to declare stuff like these because it works by default:

```
[...]

PR #5850, opened 2018-02-20
Update config docs to warn about antipattern

### LINKED ISSUE #5180: flow v0.57.x don't shows no errors if use `module.system.node.resolve_dirname=src`
I need help!

flow don't show any errors, if write `module.system.node.resolve_dirname=src` in *.flowconfig*

https://github.com/gin0606/flow-v0.57.1-trouble

Please see this repository, that can see my trouble.

Thanks!

Maybe related to https://github.com/flowtype/flow-bin/issues/91 and https://github.com/facebook/flow/issues/5105

### LINKED ISSUE #5316: Strange flow-cli behavior with errors appearing and disappearing
I'm running into a very frustrating issue where Flow sometimes reports errors and sometimes does not. I'm utterly baffled and I would love help identifying the cause of the issue.

Here's the process I'm currently using to find this behavior:

1. Run `flow`, to start the flow server and get a status check
  _Result:_ no errors
2. Run `flow check` to sanity check the server results
  _Result:_ no errors
3. Make an edit which introduces an obvious Flow error, run `flow`
  _Result:_ 24 errors
4. Run `flow check` with obvious error
  _Result:_ 0 errors 🤔 
4. Make an edit which undoes the obvious Flow error, run `flow`
  _Result:_ 23 errors 🤔 
5. Run `flow check`
  _Result:_ 0 errors
6. Run `flow stop`, then `flow`
  _Result:_ 0 errors 🤔 

Here's what it looks like to me...
1. Spawning a flow server reports errors correctly until an error is introduced
   1.1 If the error is introduced, a bunch of "ghost errors" appear out of nowhere
   1.2 Fixing the error still leaves the ghost errors
   1.3 Restarting the server makes the ghost errors go away
2. Throughout this whole process, `flow check` is failing to type-check code at all

callumlocke:
## Problem

This is a common pattern in the Flow community:

```
[options]
module.system.node.resolve_dirname=node_modules
module.system.node.resolve_dirname=src
```

It enables you to write `import foo from 'common/foo'` to get hold of `src/common/foo` from anywhere within your codebase, even from a deeply nested module, so you can avoid having to do `../../../` paths.

The problem is, Flow v0.57.0 added a [new feature](https://github.com/facebook/flow/blob/master/Changelog.md#0570):

> * Flow will now only check the files in `node_modules/` which are direct or transitive dependencies of the non-`node_modules` code.

This is a good feature. But for anyone using the above technique, it breaks Flow in a very confusing way: it stops checking all your `src` files reliably. It reports errors in *some* files, but misses many others. Then while you're developing, the watch-driven Flow server will occasionally discover more errors, which mysteriously disappear next time you do a full `flow check`. It feels totally indeterminate and unpredictable, and it led to days of debugging in my case.

From the number of issues relating to this, it seems it's easy to miss the problem when you first upgrade past 0.57.0 – the weird behaviour only surfaces much later, because everything seems to be working fine at first (`No errors!`).

## Solution

Update the docs to explain that it is now an antipattern to use `module.system.node.resolve_dirname` for internal code, and that you should use `module.name_mapper` instead.

Closes #5180 
Closes #5316 
Relates to #5647
Also closes https://github.com/flowtype/flow-bin/issues/91

(There have been many more issues stemming from this lack of documentation, but most have been closed as duplicates.)

### DIFF
--- a/website/en/docs/config/options.md
+++ b/website/en/docs/config/options.md
@@ -219,6 +219,8 @@ module.system.node.resolve_dirname=custom_node_modules
 Then Flow will look in directories named `node_modules` or
 `custom_node_modules`.
 
+It is important to use `module.system.node.resolve_dirname` only for external dependency sources such as `node_modules`. If your aim is to enable your internal modules to import each other via absolute paths (to avoid long `../../..` relative paths), use [`module.name_mapper`](https://flow.org/en/docs/config/options/#toc-module-name-mapper-regex-string) instead. This is because using `module.system.node.resolve_dirname` opts into a new feature added in v0.57.0 intended for `node_modules`, whereby Flow skips checking any files that aren't direct or transitive dependencies of other files outside this directory—probably not what you want.
+
 > **Note:** you can specify `module.system.node.resolve_dirname` multiple times
 
 #### `module.use_strict` _`(boolean)`_ <a class="toc" id="toc-module-use-strict-boolean" href="#toc-module-use-strict-boolean"></a>


STRML:
This one drove us nuts for a long time. Unsure if breaking `resolve_dirname` was intended or not, but we had to move everything to `module.name_mapper` to compensate for it (which is far more verbose).

callumlocke:
> we had to move everything to `module.name_mapper` to compensate for it (which is far more verbose).

Yeah it's a shame it's more verbose. There's also a trick where you can use a dot-slash in `resolve_dirname` (i.e. `./src/` instead of `src`), and this currently opts out of the new 0.57.0 behaviour. But it seems to be an undocumented hack, and it might stop working in future, so using multiple `name_mapper` lines seems safest. If I'm wrong on this, and the `./` is in fact a supported opt-out, please can a Flow team member let me know and I will update this PR.

TrySound:
@STRML This was done because everybody asked to not check whole node_modules, but only direct and transitive dependencies.

I think both alias and root resolve are bad practices because every tool can support this features differently or not support at all. This features do not have a spec and they are not even convention. I usually force to not create deep folders. Also there's solution via monorepos.

callumlocke:
> I think both alias and root resolve are bad practices because every tool can support this features differently or not support at all. This features do not have a spec and they are not even a convention. I usually force to not create deep folders. Also there's solution via monorepos.

I agree they're not great practices. I prefer to use `../../../` (and accept the fact they are ugly and need to be edited sometimes when files get moved around). But in reality, many projects use either `resolve_dirname` or `name_mapper` for nice clean paths. And one of the two methods (`resolve_dirname`) is catastrophically bad since Flow 0.57.0, so the docs need to warn against it. My added paragraph does not _recommend_ doing any path magic, it just says you should use `name_mapper` if you're going to do it.

mrkev:
I think this'd be good, but the import is failing. Could you rebase so I can give it a shot?

mrkev:
bump @callumlocke

PR #5892, opened 2018-02-28
Process .mjs files by default

shauns:
Modules in node > 9.6.1 can use import syntax when ran with a flag on and using`.mjs` extension


fishythefish:
Note that this [feature](https://nodejs.org/api/esm.html) is currently marked [`Stability 1: Experimental`](https://nodejs.org/api/documentation.html#documentation_stability_index).

shauns:
Maybe include just the docs change? Make the example '.mjs' and '.some-other-standardish-extension' rather than '.foo'/'.bar'? I found this relatively obscure.

PR #5896, opened 2018-03-01
Website: Wrap long text

chrisself:
On narrow viewports, longer headings and inline code snippets result in horizontal overflow. This can cause some weirdness when scrolling vertically or when zooming in and out. As an example, you can check out the [type reference for React](https://flow.org/en/docs/react/types/) article, or watch [this video](https://cl.ly/0O2u0N2q1Q0u). From what I'm seeing, the simplest fix is to wrap words that overflow the article container.

### DIFF
--- a/website/_assets/css/_article.scss
+++ b/website/_assets/css/_article.scss
@@ -1,5 +1,6 @@
 .article {
   line-height: 1.7;
+  word-wrap: break-word;
 
   p {
     margin-bottom: 1.2rem;


fishythefish:
Thanks for the PR, @chrisself! Can you include a demo of what you expect this to look like with your fix? The naive `word-wrap: break-word` behavior also looks bad to me; breaking on logical word boundaries might be more appropriate.

chrisself:
Hey! Thanks for checking this out. I took some screenshots demonstrating how this addition would affect the documentation.

Original:
https://cl.ly/1G3R3t0X1X0Y

With `word-wrap: break-word;`:
https://cl.ly/3P0D433Y3V0X

From what I'm seeing, `word-break: break-word;` results in similar behavior, but it appears this property is [mainly for CJK characters](https://www.w3.org/TR/css-text-3/#word-break-property). You mentioned that we could try breaking on "logical word boundaries." Did you have any suggestions that could point me in the right direction if `word-wrap: break-word;` isn't ideal? 

Thanks!

fishythefish:
`word-wrap: break-word` (or equivalently, `overflow-wrap: break-word`) is fine, but it doesn't care where in the word it inserts the break. Using `<wbr>` elements would allow us to choose where the line breaks occur.

chrisself:
My apologies for the delay. I'll start adding `<wbr>` elements where necessary in the coming weeks.

Issue #5889, opened 2018-02-27
Referencing Fragment from the default React export generates an error


nathantalewis:
Using `React.Fragment` as seen in the [official Fragments documentation](https://reactjs.org/docs/fragments.html) creates a flow error.

https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDGMA0cDecBiUKA5iEgHbwC+cAZlBCHAORSobMDcAUNwPR84AdWgBrYOWJwA7gAsKcUJFhIAJnACuAZwlS8hEmUpwq3WhvIZgEcnABCKKAAoAlLm5w4bGBqi3k6DAAdGhsKDBIAKIANkhGME4GpBQwLjym-IKRUAxQMvK26GhIWlpqmjqSiOzBSfFmFlY2BBAQru6e3r7+NSFhETFxKU4BGEF1KWncpkA

idiostruct:
`React.Fragment` doesn't exist on the [default export](https://github.com/facebook/flow/blob/master/lib/react.js#L241-L255). Feel free to open a pull request.

nathantalewis:
PR opened.

TrySound:
This probably can be closed.
/cc @vkurchatkin

PR #5913, opened 2018-03-05
Add Fragment to the default export

nathantalewis:
Solves #5889 

I'm not interested in signing the Contributor License Agreement, so anyone who wants to can go ahead and fork what I have to get this merged in.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -252,6 +252,7 @@ declare module react {
     +Component: typeof Component,
     +PureComponent: typeof PureComponent,
     +Children: typeof Children,
+    +Fragment: typeof Fragment,
   |};
 }
 


TrySound:
Maybe just stop using default exports? And stop mixing both, it's a hack. This way is much cleaner.
```js
```

nathantalewis:
@TrySound, thank you for sharing that workaround. It might be just what some people are looking for. Perhaps you could also share it with the maintainers over at prettier.io and other opinionated code formatting projects?

However, since default exports are part of the ES6 spec and it seems that Flow is trying to provide type information for the default export for React, is there harm in adding `Fragment`?

TrySound:
It's not related to formatting. In the future react will not be available as default export. Forcing yourself and others to use named exports with react will simplify transition.

nathantalewis:
> In the future react will not be available as default export.

Interesting. I hadn't heard about that and my googlefu is failing me at the moment. Would you mind sharing where you found it?

Until the default export is removed, however, it is still just one opinion between valid options. If people want to continue using the default export up to the last minute (or up until there is a codemod script to do the heavy lifting for them), shouldn't they be allowed to?

TrySound:
Well flow force you to start using named exports. It doesn't make sense to add even more properties which still will be deprecated.

SleepWalker:
@TrySound this is about React's actual export. It is good, when flow forces me to `import type { Node } from 'react';`. But `Fragment` is **not a magical type**. It is an export of the library and it is **really there**. 

Babel relies on that **valid export** and in my case I can not force it to use my local `Fragment` instead of `React.Fragment`. As result I get a valid code, which flow reports as invalid. The goal of `flow` is to provide types for the actual implementation (and the docs, of coarse https://reactjs.org/docs/fragments.html. There is no info about imports). Flow should not hide implementation details and confuse the users. Am I right?

One more to note. There are some folks (like me), that believe, that one should explicitly declare dependencies instead of saying 'I want to use here all the React can provide for me' (moreover `React.Node` is not that nice comparing to simply `Node`). I can't follow that code style, because it will break existing tooling, like babel, which will use `React.Fragment` behind the scenes and produce the code, that flow thinks to be invalid.

SleepWalker:
The most recent docs continue to show the pattern, that is considered to be 'deprecated':

https://reactjs.org/docs/strict-mode.html

```jsx
import React from 'react';

function ExampleApplication() {
  return (
    <div>
      <Header />
      <React.StrictMode>
        <div>
          <ComponentOne />
          <ComponentTwo />
        </div>
      </React.StrictMode>
      <Footer />
    </div>
  );
}
```

TrySound:
This probably can be closed.
/cc @vkurchatkin

PR #5920, opened 2018-03-06
Update React libdefs for createContext, createRef

sophiebits:
Tested by checking this script:

```js
// @flow

var React = require('react');

var X = React.createContext('div');

class Foo extends React.Component<{}> {
  // Removing this type annotation should give an error on line 25 (div isn't
  // an HTMLImageElement) but it doesn't.
  divRef: {value: null | HTMLDivElement} = React.createRef();

  render() {
    return (
      // Changing this to 'spam' correctly triggers an error
      <X.Provider value='span'>
        <div ref={this.divRef}>
          <X.Consumer>
            {(Tag: 'div' | 'span' | 'img') => <Tag />}
          </X.Consumer>
        </div>
      </X.Provider>
    );
  }

  componentDidMount() {
    var div: null | HTMLImageElement = this.divRef.value;
  }
}
```

This correctly gives a type error in componentDidMount. But removing the divRef annotation also should have an error but it doesn't. What did I do wrong?

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -171,6 +171,7 @@ declare type React$Key = string | number;
  * The type of the ref prop available on all React components.
  */
 declare type React$Ref<ElementType: React$ElementType> =
+  | {value: React$ElementRef<ElementType> | null}
   | ((React$ElementRef<ElementType> | null) => mixed)
   | string;
 
@@ -185,8 +186,6 @@ declare module react {
   declare export var PropTypes: ReactPropTypes;
   declare export var version: string;
 
-  declare export function initializeTouchEvents(shouldUseTouch: boolean): void;
-
   declare export function checkPropTypes<V>(
     propTypes: $Subtype<{[_: $Keys<V>]: ReactPropsCheckType}>,
     values: V,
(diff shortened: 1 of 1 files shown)


TrySound:
Does `divRef: React.Ref<HTMLDivElement> = React.createRef()` also work?

TrySound:
I guess you don't get errors because dom elements are not fully covered with types
https://github.com/facebook/flow/blob/master/lib/react-dom.js#L277

Did you tried this with class components?

lib/react.js
```diff
@@ -196,11 +195,19 @@ declare module react {
...
+    Provider: React$ComponentType<{value: T}>,
+    Consumer: React$ComponentType<{children: (value: T) => React$Node}>,
+  };
   declare export var createElement: React$CreateElement;
   declare export var cloneElement: React$CloneElement;
   declare export function createFactory<ElementType: React$ElementType>(
     type: ElementType,
   ): React$ElementFactory<ElementType>;
+  declare export function createRef<ElementType: React$ElementType>(
+  ): {value: null | React$ElementRef<ElementType>};
```
gaearon: Need to change this to `current`.

goodmind:
Export separate `Context` type for return in `createContext`?

zaynv:
Is there a reason new stuff like this and `Fragment` isn't getting added to the default export as well? Does this mean people should always be doing

```js
```

instead of

```js
import React from 'react'
```

TrySound:
@zaynv Yep, default export won't work when react will migrate to esm. There should be only one way to use things.

billyjanitsch:
@samwgoldman this should not have been merged without addressing https://github.com/facebook/flow/pull/5920#discussion_r178429127. The existing commit (`value`) doesn't match the actual API (`current`).

samwgoldman:
@billyjanitsch Thanks. Fix is landing now.

billyjanitsch:
Thanks 😊

zaynv:
It seems like `Fragment` is still not on the default export. This means that you can't use the fragment syntax unless you do

```js
```

samwgoldman:
@zaynv Thanks! I'll put up a change to add `Fragment` as well.

lib/react.js
briandastous: Is there any difference between `null | React$ElementRef<ElementType>` and `?React$ElementRef<ElementType>` ?

sophiebits: The latter includes `undefined`. For basic code questions please ask on Stack Overflow or another site so we can keep this uncluttered and avoid spamming people.

PR #5928, opened 2018-03-08
Update flow --help output example from 0.66 output

iamdustan:
Maybe don’t need to show _all_ of the output, but it’d be nice if the documentation page for the help wasn’t quite so out of date.


gkz:
Closing as out of date

PR #5961, opened 2018-03-15
bash -e -> bash -xe

gregory-nisbet:
more verbose logging for setup scripts


mroch:
probably fine but we should make sure this doesn't expose any of the secrets in the logs first

panagosg7:
Closing since we've moved from travis to circleci

PR #5984, opened 2018-03-19
[RFC][Documentation only] Document that Class is a type

tomasreimers:
I couldn't find this anywhere in the documentation, so drafted something quick documenting how Class is a type; however, I could totally believe this is documented elsewhere and I just missed it.


goodmind:
https://flow.org/en/docs/types/utilities/#toc-class

PR #5988, opened 2018-03-19
Add destroy method to streams

### LINKED ISSUE #3204: child_process streams missing destroy property
```
> c = chp.spawn('python', ['-v']);
> c.stdout.destroy()
undefined
```

Right now `child_process$ChildProcess`'s `stdout` and `stderr` are declared as `stream$Readable`. Each of these, including `stdin` have some additional properties beyond `stream$Readable/Writable` including `destroy`.

hiikezoe:
Fixes #3204.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1180,7 +1180,7 @@ declare class net$Socket extends stream$Duplex {
   bytesRead: number;
   bytesWritten: number;
   connect(options: Object, connectListener?: Function): void;
-  destroy(exception?: Error): void;
+  destroy(exception?: Error): net$Socket;
   end(
     chunk?: string | Buffer,
     encodingOrCallback?: string | (data: any) => void,
@@ -1437,6 +1437,7 @@ declare class stream$Readable extends stream$Stream {
   unshift(chunk: Buffer | string): void;
   push(chunk: ?(Buffer | string), encoding? : string): boolean;
   wrap(oldReadable: any): stream$Readable;
+  destroy(error?: Error): void;
 }
 
 type writableStreamOptions = { highWaterMark? : number, decodeString? : boolean, objectMode? : boolean };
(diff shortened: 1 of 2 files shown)


mrkev:
Could you link to the spec backing up this change?

hiikezoe:
@mrkev, thanks for the reply.  I am afraid I don't quite follow what the spec means and what the backing up mean.

hiikezoe:
Oh thanks.  This change is for node APIs,  so it might be sufficient to put the links to the API references?

https://nodejs.org/api/stream.html#stream_writable_destroy_error
https://nodejs.org/api/stream.html#stream_readable_destroy_error
https://nodejs.org/api/stream.html#stream_transform_destroy_error
https://nodejs.org/api/net.html#net_socket_destroy_exception

mrkev:
Hey @hiikezoe 

I don't see a return value for destroy on https://nodejs.org/api/stream.html#stream_transform_destroy_error, but there's one on the diff. Is this an omission in the docs or does it have to be changed in the diff?

I do see one on https://nodejs.org/api/stream.html#stream_readable_destroy_error but the definition you added returns void.

nmote:
@hiikezoe, if you are still interested in getting this in, I'll work with you to make sure it gets landed promptly.

One useful thing here is the `this` type. You can use that instead of spelling out the type explicitly, e.g. `stream$Readable`.

Could you fix up the return types? I can update the tests when I import, so feel free to just remove the changes to the `.exp` file.

hiikezoe:
hey @nmote, would you mind fixing up them on behalf of me?  I am afraid I don't have the environment for that at this moment.
Thank you!

nmote:
Sure, this is pretty close so I'll just import and then make a few tweaks.

nmote:
I ended up just adding the right `destroy` method declarations to `stream$Writable` and `stream$Readable` and removing them from children of those classes, including the other classes changed in this PR. Once I get a second pair of eyes on this it should land, assuming we don't discover a problem.

PR #6019, opened 2018-03-22
Declare g:ale_linters to complete config per Ale docs

kevinold:
Update Vim docs, per correct config as outlined in https://github.com/w0rp/ale/issues/607#issuecomment-305570159


mvitousek:
From the link you posted, it looks we already recommend (one possible) correct .vimrc configuration, don't we? It looks like the fix you're suggesting would help if we had been suggesting
```
let g:ale_fixers.javascript = [ ... ]
```
but we were already suggesting 
```
let g:ale_linters = {
\  'javascript': [ ... ]
\}
```
which is one of the valid configurations in your link. 

I'm closing, but if I got this wrong please feel free to re-open and correct me: I'm not a vim expert :)

Issue #6056, opened 2018-03-28
Example from docs for $ObjMap broken


jwickens:
$ObjMap mixes up return values when mapping functions. By "mixes up" i mean that I get errors showing that flow is considering the union of all return types collectively for each function individually.

* [docs im referring to](https://flow.org/en/docs/types/utilities/#toc-objmap)
* [try it in the playground](https://flow.org/try/#0PTAEAEDMBsHsHcBQiSmgUwC4HIDOp4AnAS03VAENRIBXAOwGNNjY7RMBPAB3MwAsKmdhQDW6fFQAGACgCUoALwA+UADVJlOgBNQhLDUJ0JoSetDTS+PZgNtOPWYnvkAogA9MhCkwBK+wwAq3OQKoAA8qkrScooqqvLKasi0jMysuvRhAIIANKAA8gBcoADeANpiHMW4nsR0AOYAusUxiVkAvlGwxfmyxQAk+QBGAFYAshRcYfl57p7emH42gcEqJYiguv5swyPoTAB0lbjSsLIHelo0DOjR3gx5Igkqu-uYBxS4uMT1dNL3eRKoAqzVAsBBMXaskBUIA3Ih2sgGKwamDFKUNpQWs92IQaOgcpihtjYqBsJBYLBsAj4YhpHi-mcPsUhpSMBQ6LJYaBUPkRHSGadzsTQDUSA0uZtefzUP13DwmC5CIRYIQBfQhQcRazYOzOdzUAA5WA8PJDUDEYxiur1FBgOVuBWYJUqtWCpkMbmbb0+qVgY2m0AMAifUB0WBCOrsPjkVU-OoUaBg0ZvIA)

Expected: 

only errors on line 18 and 20 showing usage. This is the case with v56

Actual:

 Using version 68 the following errors are displayed in addition to 18 and 20. 
```
11:   a: () => true,
               ^ Cannot return `true` because string [1] is incompatible with boolean [2].
References:
12:   b: () => 'foo'               ^ [1]
11:   a: () => true,
               ^ [2]
12:   b: () => 'foo'               ^ Cannot return `'foo'` because boolean [1] is incompatible with string [2].
References:
11:   a: () => true,
               ^ [1]
12:   b: () => 'foo'               ^ [2]

```

Impact:

Anyone wanting to map an object type containing functions to a transformed set of functions will be affected. I discovered this while making types to transform from types for our API to backend types for our controllers, which are basically the same as the API types but with additional parameters.

[...]

PR #6059, opened 2018-03-29
test for issue #6056

jwickens:
Test to show how object map conflates types.

I also wanted to add tests for property and element type in type-desctructoring but it does not seem to pick these tests up. Any idea why?
```js
/tests/type-destructors/element_type_for_functions.js
/**
 * @format
 * @flow
 */

type Funcs = {
  funcA: (a: number) => string,
  funcB: (a: string) => number,
};

const funcaAImplmentation = (a: number) => 'a' + 'b';

(funcaAImplmentation: $ElementType<Funcs, 'funcA'>); // ok
```

### DIFF
--- a/tests/objmap/objmap.diff
+++ b/tests/objmap/objmap.diff
@@ -1,0 +1,37 @@
+--- objmap.exp	2018-03-29 09:37:46.000000000 +0200
++++ objmap.out	2018-03-29 09:37:50.000000000 +0200
+@@ -314,6 +314,28 @@
+                  ^^^^^^ [2]
+ 
+ 
++Error ----------------------------------------------------------------------------------------- objmap_functions.js:17:2
++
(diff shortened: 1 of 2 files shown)


mvitousek:
Sorry we didn't take a look at this sooner! It looks like you're adding a ".diff" file to record the result of the test. The correct way our tests are formatted, at least today, is with a ".exp" file that records the expected output of the test. The ".diff" file records the difference between the actual and expected outputs, and isn't meant to be included in any tests as far as I know. If you want to regenerate the .exp file with "./runtests.sh -r" and resubmit I'll take a look.

mvitousek:
I'm going to close this PR for now but please feel free to message me or reopen if you plan on taking a look at it again, or if you need any help in doing so. Sorry again that we took so long to initially get to this.

PR #6076, opened 2018-04-01
fix(lib/core.js): remove required @@iterator method from Iterator

jedwards1211:
`Iterators` are not required to be `Iterables`:
http://devdocs.io/javascript/iteration_protocols

> An object is an iterator when it implements a next() method with the following semantics

> *Some* iterators are in turn iterables


nmote:
Good catch, thanks

jedwards1211:
Closing per discussion in #6075.  Thanks for your consideration @nmote!

PR #6078, opened 2018-04-02
Add React.StrictMode

aweary:
https://reactjs.org/docs/strict-mode.html

`React.StrictMode` has the same API as `React.Fragment`, a component that takes no props and renders its children.


davidmccabe:
It looks like there's some kind of snapshot test that needs to be updated for this to land. @aweary  Would you be so kind?

jbrown215:
Sorry for getting to this so late. This is in the libdefs already. I have a diff internally (expected to land in 0.92) that moves it to AbstractComponent

PR #6100, opened 2018-04-05
Provide correct type of the `onmessage` event handler (DedicatedWorkerGlobalScope)

mobily:
This PR fixes type of the `onmessage` event handler in the DedicatedWorkerGlobalScope object.


nmote:
Could you provide a link to API docs which support this change?

nmote:
This one was so obviously a bug that I just went ahead and found the docs myself: https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/onmessage

nmote:
Looks like this ended up getting addressed in #6853 though, so I'm closing

PR #6118, opened 2018-04-09
Expose `implements` and `mixins` of `DeclareClass` AST

### LINKED ISSUE #5463: flow-parser: `mixins` silently dropped
See [sample](https://astexplorer.net/#/gist/7e956da8da57d5eafdde0e04de9b95a1/5b6fb6f97bbd1e61134d323f2e8b320a51206e9e).

I'm guessing the problem is that `estree_translator.ml` is [ignoring it](https://github.com/facebook/flow/blob/94f53077b74980de1b723a4462903c120ec0bbbe/src/parser/estree_translator.ml#L643-L647) even though it's [there in the original ast](https://github.com/facebook/flow/blob/94f53077b74980de1b723a4462903c120ec0bbbe/src/parser/ast.ml#L378-L385); it doesn't seem to have been appropriately updated when mixins [were added](https://github.com/facebook/flow/commit/b72d1c61487f9e95ddbe68c4f994d9699b7b71c9).

Edit to add: this is a problem for [prettier](https://github.com/prettier/prettier/pull/3391).

### LINKED ISSUE #6096: flow-parser: `implements` dropped in `DeclareClass`
Minimal example: [flow playground](https://flow.org/try/#0JYOwLgpgTgZghgYwgAgELIN4F8BQOAmECANnFCiXAM5XICCywAtgA7ERMTi3oY65A)

The `implements` is not included in `DeclareClass` like in `ClassDeclaration`.

It's similar to #5463, it's also an issue for [prettier](https://github.com/prettier/prettier/pull/4244)

duailibe:
Closes #5463
Closes #6096

Updates `estree_translator.ml` to expose those two fields of `DeclareClass` as we need those for Prettier. I fixed esprima compatibility tests that broke, but I'm not sure if/how I should add regression tests when the declaration includes an interface or mixin.

I'm not familiar with this codebase, let me know if I should do anything differently.

Thanks!

cc @mroch

### DIFF
--- a/packages/flow-parser/test/esprima_test_runner.js
+++ b/packages/flow-parser/test/esprima_test_runner.js
@@ -267,6 +267,15 @@ function handleSpecialObjectCompare(esprima, flow, env) {
     case 'CallExpression':
       delete flow.optional;
       break;
+
+    case 'DeclareClass':
+      if (esprima.implements === undefined) {
+        esprima.implements = [];
+      }
+      if (esprima.mixins === undefined) {
+        esprima.mixins = [];
+      }
+      break;
   }
 
   switch (esprima.type) {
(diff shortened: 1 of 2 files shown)


vjeux:
cc @mroch

mroch:
thanks!

PR #6119, opened 2018-04-09
Add BeforeUnload event support

Connormiha:
Add `beforeunload` event for `addEventListener`

https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload


mrkev:
Could you link to the spec that backs up this chagne?

Connormiha:
Added link to mdn

lib/dom.js
```diff
@@ -303,6 +310,10 @@ declare class FocusEvent extends UIEvent {
   relatedTarget: ?EventTarget;
 }
 
+declare class BeforeUnloadEvent extends Event {
+  returnValue: mixed;
```
nmote: According to the linked spec, it looks like this actually has type `string`. `mixed` is perfectly safe, but we might as well go with the more precise type here. Could you make that change, then I'll make sure it gets prompt attention?

nmote:
Feel free to reopen if you would still like to get this in. For now I'm closing since I have not yet seen a response.

Issue #2626, opened 2016-10-13
Intersection doesn't work for exact object types


philipp-spiess:
Given the following example:

``` js
declare type Foo = {| foo: string |} & {| bar: string |}
const example: Foo = {foo: 'foo', bar: 'bar'}
```

I expect this to work since the intersection type should be allowed to further extend the exact object. But in the latest version, it errors.

Is there an easy way to make this work or is this intentional?

vkurchatkin:
If value `v` has type `A & B` that means that it has type `A` and has type `B` at the same type. If  `A` is `{| foo: string |}` it means that `v` has exactly one property `foo` and if `B` is `{| bar: string |}` it means that `v` has exactly one property `bar`. Obviously, you can't satisfy these two constraints at the same time, so `A & B` is an empty type.

philipp-spiess:
Thank you for the answer. Your explanation totally makes sense but I was hoping to build an exact object type using intersection somehow. The above example obviously works with a regular object type, so I expect the exact object to behave in a similar matter (so that the outcome equals `declare type Foo = {| foo: string, bar: string |}`.

Do you know of another why I could model this behaviour? I have two types, `A` and `B`, who define keys and values and I want to build an intersection `A & B` as an exact object, so that _only_ the keys defined in either `A` or `B` are valid.

vkurchatkin:
I'm pretty sure there is no way do it at the moment.

dsimmons:
Chiming in -- just ran into this myself.

Terminology aside, do you think "combining exact types" is a possibility at some point?

Going back to the explanation above, let's say I have an exact object type `A` and I want to create a new type `C` that includes a superset of `A`, plus some new fields (we'll call them object type `B` -- whether or not `B` is also exact doesn't matter).

Let's also say that `A` is used elsewhere. If I want to create this new type without duplicating all of the same fields included as part of the exact object type `A`, I'm forced to relax the strict/exact check.

To provide a more concrete example, I'm attempting to do the following:

```Javascript

// Used elsewhere as a standalone type.
export type MyStrictType = {|
  uri: string,
  name: string,
  type: string
|};

export type Action =
  // ... others
  | ({ type: 'SOME_CONST', anotherAttr: string } & MyStrictType)
;
```

Admittedly, in addition to relaxing the strict/exact object requirement, one way I could fix the above is by instead introducing a nested attribute of `{ type: 'SOME_CONST', anotherAttr: string }` rather than trying to combine the two.

[...]

PR #5973, opened 2018-03-16
Documentation for object type spread

caoilte:
This PR adds documentation for object type spreading to the website.

It is largely derived from some excellent comments by @vkurchatkin in #2626 and #3534. 

I am also greatly indebted to @zhenyong for [his version of the documentation](https://zhenyong.github.io/flowtype/), particularly his explanation of [Constructor Functions and Prototype Objects](https://zhenyong.github.io/flowtype/docs/objects.html#constructor-functions-and-prototype-objects). This helped me understand the effect of object property inheritence in Flow enough to document its effect on object type spreading. 

I struggled to get the website building locally so there may be some problems with my markdown.

PR #6124, opened 2018-04-11
Usage of object spread for exact object types

mattgstevens:
Came across this issue: https://github.com/facebook/flow/issues/2626 and decided it could be in the documentation :)

### DIFF
--- a/website/en/docs/types/objects.md
+++ b/website/en/docs/types/objects.md
@@ -229,6 +229,19 @@ properties to an exact object type.
 var foo: {| foo: string |} = { foo: "Hello", bar: "World!" }; // Error!
 ```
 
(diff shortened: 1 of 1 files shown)


saadq:
Instead of

```ts
type FooBarT = {| ...FooT, BarT |}
```

did you mean

```ts
type FooBarT = {| ...FooT, ...BarT |}
```

mattgstevens:
Good catch @saadq updated and expanded the example since the failure wont be at the type definition, but when using it for a variable binding.

mattgstevens:
@fishythefish is there anything else to do for this? think it is ready for a merge. 

mentioning you as it seems you have merged other PRs recently

apsavin:
Related to #5973

PR #6132, opened 2018-04-12
Implements custom resolvers

arcanis:
This diff adds a new configuration option to Flow: `module_resolver`. It points to an executable that consumes a stream of JSON objects under the form `[request, issuer]`, and outputs a stream of paths in response.

Since it's been my first work with OCaml since quite a long time there's probably errors and odity in my code, feel free to point them 🙂 Some of them I have in mind (maybe they can be fixed in a followup pass, you tell me):

- The JSON payload is crafted with sprintf, I wasn't sure how to properly generate a JSON string
- For the same reason, the resolver returns a stream of paths instead of an actual JSON object because I wasn't sure how to parse them in Flow
- Related to the previous point, the resolver doesn't have a way to notify Flow that a request cannot be satisfied (it would be solved if the return object was a `[err, result]` entity).
- If the `module_resolver` is invalid, Flow will currently hang (I thought `create_process` would "throw", but maybe it doesn't make sense in OCaml?)


arcanis:
cc @mroch

arcanis:
The Appveyor error seems unrelated (sqlite3), and the Travis one is one that I cannot reproduce locally (I had it some time ago, but then I updated the mli file to include module_resolver and it worked). Could it be possible that the mli file is cached somewhere?

fatfatson:
could this solve #6305 ?

jakesyl:
@arcanis, if you have a second it'll be super helpful if you could look at #7014

PR #6176, opened 2018-04-20
Update JSX typing to allow exact object splats

danwang:
## Summary
Adjusts rules for JSX so that splats can work with components that expect exact object types.

I updated the tests:
- There were flow errors that are no longer flow errors because the "incompatible with exact type ..." error is not surfaced anymore. I updated the `Error: ...` comment and expectation file.
- There were some existing lines in the **.exp** file that had `// OK` comment, but resulted in a Flow error. They still have a Flow error, but I changed the comment to reflect the behavior. 
- Added some test cases to check for the new behavior

Thanks for your time!

## Motivation
Some examples ([flow try link](https://flow.org/try/#0JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQduEAdqvACQCuzoYANlhBZmMLABMAXCmYBPOAF44ANnowZYLHACiADzwwACsTDpFAbwA+yKWyjBmAcwA0cAEZTmnEK6xQLAX3omVnhdfQBhEkhmYRgpbH0AOkjwFliAFXUsAB4w-CMIEwA+BTguHnABIRFxejoAenq4dIALYHQxCCxUZjJ4AHdoAGsXV054YHhUFohOPjE6XL18FOjYuDNErbNrcmQyUakARn9-OHqihkbmto6unr64QagRt3G4Sbhp2fnGFjY4GBjKgpHlDMDSjspGR9oc4Cd6EsIlE0iINltEkDCqgzhcrk1Wu04J1ur0BsNRu9Pt85gtggD7ARfJQxJDdjCDm5joFFmDVqj4JstozmeJcZcGgTbsT7mSnhS3hMpjNaX8QoDkLBgMg+GzofseUiViiYmihZjNTBtXwzq55GYTucikA)):
```jsx
type ExactProps = {|a: string, b: number|};
const ExactComponent: React.ComponentType<ExactProps> = ...;

// This doesn't work, but it should
<ExactComponent {...{a: 'a', b: 1}} />

// This doesn't work, but it should
const props: ExactProps = {a: 'a', b: 1};
<ExactComponent {...props} />

// This doesn't work, but it should
const inferred = {a: 'a', b: 1};
<ExactComponent {...inferred} />

// This doesn't work, but it should
const partial = {a: 'a'};
<ExactComponent {...partial} b={1} />
```

# TODO
Order of statements matter now (even though they didn't use to). I think this may be because I'm using **Context.find_resolved**, which returns different results at different points in time of the type checking?

```jsx
// If ExactProps is defined before, it works, but after fails
<ExactComponent {...(props: ExactProps)} />
```


fishythefish:
Closing since these issues are already resolved in 0.71.

danwang:
Yes, but 0.71 displays "No errors!" more often at the expense of failing to identify bugs. For example:

https://flow.org/try/#0JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQuEAdqvACQCuToYANliCxMYWACYAuFEwCecALxwALPToxpYLHAAKxMOgUBvAD504KSayjAmAcwA0dIwF8VjFvADCJSEyExJ2HgwAHRe4Mx+ACrqWAA8OhB6AHzycJzc4PyCwmKuzKxwYLqokgbIFjBWtk6pZZJkyGR2cABG9TAAFsDoANZYsqgdEBy8onBMEPBYAB7dMGQudLFhPkKmZhsEXPjAzHAAYhAQABRYWZJMHCAtWFDNRYkl2sUAlHAGTusbcG4FAELIKCpACMACYAMwqb5mX7wB56UrlOCWaw2GqGJENJqteotBb0aFwWIAoEGYLk+GoGoAeiSNCAA

fishythefish:
@danwang This isn't about JSX; see [this example](https://flow.org/try/#0C4TwDgpgBACgTgezAZygXigbwD4CgpQCGAXFMsHAJYB2A5gDS7YC+A3LrgMYLXlRiIUpTCTIUatZuiyiA5IVn0oAI1KzgAC0qoA1hBBkNCAK4AbACZRqCYFAgAPbcFltcACkwA6bwKTJmpPB+AJSsQA).

If you remove the `{a: string}` annotation, you get an error, as expected.

(Side note: we would rather falsely claim there are no type errors and then improve our gradual typing to prevent this than falsely claim there is a type error and prevent valid code.)

fishythefish:
@danwang Yeah, we're working on unifying spreads in JSX and object literals, so although different code handles them at the moment, that's why they have similar implementations and errors. The fact that the example I linked doesn't error is a known problem that we're working on, so fixing it should resolve both cases.

samwgoldman:
@danwang You are very observant to notice this. I am working on a bunch of changes around object types which will make spreads sound in this respect (and in other respects where it is currently unsound). Will share our plan when things start coming together.

PR #5153, opened 2017-10-20
Added Buffer property to buffer module

CMCDragonkai:
https://github.com/facebook/flow/issues/5133

PR #6179, opened 2018-04-21
Add Buffer property to buffer module

motiz88:
Previously discussed or implemented in #5153, #3495, #2938, #3723, #3724 and maybe elsewhere.

This PR refines previous efforts by adding:

1. A correct, unambiguous type for `require("buffer").Buffer`:
```jsx
declare var Buffer: typeof global.Buffer;
```

2. A test.

To be clear: Per the Node [docs](https://nodejs.org/api/buffer.html), explicitly importing `Buffer` is rarely needed, given its availability in the global scope, but correct. It happens to be a viable option in environments where "magic" support for Node globals is either undesirable or broken (the latter is [currently](https://github.com/webpack/webpack/issues/7032) the case for ES modules in Webpack).


motiz88:
PR tests might fail due to the current state of `master` - this branch does pass tests on its own.

PR #6193, opened 2018-04-24
Make event handlers on Worker nullable

tiffon:
Seems like the `onmessage`, `onerror`, and `onmessageerror` event handlers on `Worker` should be able to be set to null.

Apologies for the hassle If this is not something flow is inclined to allow.


mrkev:
Ah, for future reference it'd be helpful if you linked a spec backing up libdef changes. Could you find said link for this change?

tiffon:
Hey @mrkev, no problem.

The `Worker` interface [event handlers spec](https://html.spec.whatwg.org/multipage/workers.html#dedicated-workers-and-the-worker-interface:event-handlers).

And, from [8.1.5.1 Event handlers](https://html.spec.whatwg.org/multipage/webappapis.html#event-handler-attributes):

>An event handler **has a value, which is either null, or is a callback object**, or is an internal raw uncompiled handler. The EventHandler callback function type describes how this is exposed to scripts. Initially, an event handler's value must be set to null.

Sounds like this would mean all `on*` event handlers should be nullable. Looking through [source-graph](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/facebook/flow%24+%5Cson%5Ba-z%5D%2B:+file:bom) seems like it's a bit of a mix.

Let me know if more info is warranted.

Issue #5897, opened 2018-03-01
Can't use 100+ bytes literal types


aatatuzov:
It seems that Flow doesn't support variable-length string literal types. See this [flow/try](https://flow.org/try/#0C4TwDgpgBAysBOBGADMqBeKByZiBMAzACwCsAbAOwAcAnLoaZbfcedXfq0xw28543YtB-Xt2F8eXdlgBQoSLAQ0aGbAMkTxG7WKE79e0dOMipZrTNmyANhGBQAzghTIAXEqSo1OA6c1+5gFGQbomoYbhlrRydg7O8CoecImqmFiB0RH+YRaZ+SFZWXJAA).

```
type Str100 = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'
type Str99 = '123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'

let str100: Str100 = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'
let str99: Str99 = '123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'

```

```
4: let str100: Str100 = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'                        ^ Cannot assign `'0123456789...'` to `str100` because string [1] is incompatible with string literal `0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789` [2].
References:
4: let str100: Str100 = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'                        ^ [1]
4: let str100: Str100 = '0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789'
               ^ [2]
```
At first glance you may think that such a long literal types do not make sense. However one (me) can use a JSON with phrase-based unicode fields for multilingual support. In this case Flow is supposed to check if there are i18n literals in code not defined in the JSON.

mrkev:
Welp, nvm not a bug; it's right [here](https://github.com/facebook/flow/blob/master/src/typing/statement.ml#L3330). It's expensive to keep track of long strings so literals strings are capped at 100 characters.

Perhaps a way to set this constant this with a flag or through `.flowconfig` would come in handy for edge cases like yours. Unfortunately that is not on the roadmap, but if you'd like to PR I'd be happy to have a look at the branch.

aatatuzov:
[...]

PR #6215, opened 2018-04-26
Add max_literal_length option

aatatuzov:
Replaces magic 100 in `typing/statement.ml` with new `.flowconfig` option `max_literal_length`. Default value is set to 100. Zero value disables the check. #5897

There are a lot of files involved. I don't know if I really need to duplicate default values in `flow_dot_js.ml` and `testgen/flowtestgen_utils.ml`. I did it just by following the case of `enable_const_params`.

Changes were tested in [v70 based branch](https://github.com/aatatuzov/flow/tree/max_literal_length_v70).

### DIFF
--- a/src/commands/commandUtils.ml
+++ b/src/commands/commandUtils.ml
@@ -794,6 +794,7 @@ let make_options ~flowconfig ~lazy_mode ~root (options_flags: Options_flags.t) =
       |> min Sys_utils.nbr_procs;
     opt_suppress_comments = FlowConfig.suppress_comments flowconfig;
     opt_suppress_types = FlowConfig.suppress_types flowconfig;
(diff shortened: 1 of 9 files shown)


gabelevi:
Out of curiosity, why do you want this change? Do you use really long string literals and need Flow to track their literal types? How high do you want to set this option?

aatatuzov:
@gabelevi I use dictionary object with phrase based multilingual field names (they 1. are quite long and 2. consume several bytes per character) for i18n support. It is handy to verify its consistency during type checking stage.

I started the issue #5897 on this problem and now I'm just going with the flow.

mrkev:
@aatatuzov will merge this soon. Although not necessary, it would be nice if you submitted another PR documenting this option in https://flow.org/en/docs/config/options/

PR #6222, opened 2018-04-28
add ECDH class

eleith:
adding the [ECDH class](https://nodejs.org/docs/latest-v9.x/api/crypto.html#crypto_class_ecdh)


eleith:
unclear if the failing appveyor build is related to this PR. fyi.

mrkev:
It's not, dw about it 👍

Issue #4330, opened 2017-07-07
Please never remove the `--version` argument


w0rp:
When you run `flow --version`, it tells you that it is deprecated in favour of `flow version` in newer Flow versions. This implies that `--version` will eventually be removed. I would like to suggest that `--version` is _never_ removed from `flow`.

I integrate `flow` with my Vim plugin `ALE`, and I need to check the version of Flow to know which arguments are supported, as someone might be running version `0.49.1`, or `0.27.0`. If `--version` is ever removed, then at best I'll have to write `flow --version; flow version` and try and parse the version number from the combined output.

andreypopp:
In addition to that, `cmd --version` is a widely adopted convention and I usually automatically reach for `cmd --version` when trying to check the command's version.

eedrah:
I completely agree - it seems like it's a pretty strong standard among almost all CLIs that `--version` will give you the version number. Many tools and people use this without thought. A user should not have to consult the documentation or help to look for the command when there seems to already be a community standard.

(Though I do appreciate why there is also a command `version` to allow for further options.)

I've created a pull request to un-deprecate this.

nmote:
Okay

PR #6249, opened 2018-05-04
[PR] Remove deprecation of flow --version

eedrah:
#4330 discusses keeping the command `--version`. This PR removes the deprecation notice from the command line and the documentation.


eedrah:
I just rebased the commit back to the last commit where the build was succeeding, to ensure the checks all pass.

nmote:
Let me see what the rest of the team thinks

gabelevi:
I think I originally added the deprecation warning to `--version` since I didn't like the duplication between `flow version` and `flow --version`. Furthermore, I felt that `flow version` was superior since it could have its own flags, like `flow version --json` which I use quite often.

But many other common utilities (like `git`, `hg`, `watchman`) also support both, so I suppose this is fine.

PR #6252, opened 2018-05-05
Add index and input to String.prototype.match result

bjorn3:
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match


bjorn3:
Should have read the CODE_OF_CONDUCT not willing to sign CLA.

PR #6266, opened 2018-05-08
Improve core js Object.values typing.

Mtnt:


mrkev:
Hmmm, this is actually not correct. For more info see

https://github.com/facebook/flow/pull/5663#issuecomment-357593210

PR #6279, opened 2018-05-09
Remove files in hack that are no longer used

fangyi-zhou:
Some files are no longer used in hack directory, either because they
have been removed from Hack repo and not updated here, or because they
are imported to this repo incorrectly. This commit removes these files.


fangyi-zhou:
Are LSP tests flaky?

gabelevi:
Ok, I resolved this in https://github.com/facebook/flow/commit/687a5bd12bc1ad8f02164b68a25bd5be3a15db17

PR #6280, opened 2018-05-09
Update react-dom/server

marudor:
Add streaming API.
Also added the version string.


mrkev:
Could you link to some sort of docs that verify this is correct?

marudor:
https://reactjs.org/docs/react-dom-server.html#rendertonodestream

mrkev:
I don't see any reference to the version string, can we confirm in exists or remove it from the PR?

marudor:
image

PR #6303, opened 2018-05-14
Add 'as' property to HTMLLinkElement

tmcw:
As documented in https://developer.mozilla.org/en-US/docs/Web/API/HTMLLinkElement/as 

> The as property of the HTMLLinkElement interface returns a DOMString representing the type of content being loaded by the HTML link, one of "script", "style", "image", "video", "audio", "track", "font", "fetch".

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -3098,6 +3098,7 @@ declare class HTMLLinkElement extends HTMLElement {
   rel: string;
   sizes: DOMTokenList;
   type: string;
+  as: string;
 }
 
 declare class HTMLScriptElement extends HTMLElement {


mrkev:
Sweet! Is this supposed to be read-only? If so, it wouldn't hurt to mark it so. Check [here](https://flow.org/en/docs/types/interfaces/#toc-interface-property-variance-read-only-and-write-only) and [here](https://medium.com/@gcanti/immutability-with-flow-faa050a1aef4) to learn more.

tmcw:
Nope, it's a readable & writable property both in the spec and in the browser implementations.

PR #6309, opened 2018-05-15
Add react-preset note to compiler-babel.md

davidgilbertson:
I've seen a few instances of people using the `flow` as well as `react` presets unnecessarily. Since many users will already be using the React preset is it worth pointing this out here?

Or has a conscious effort been made to not mention React to counter the preconception that Flow can only be used with React?

### DIFF
--- a/website/en/docs/_install/compiler-babel.md
+++ b/website/en/docs/_install/compiler-babel.md
@@ -9,6 +9,9 @@ First install `babel-cli` and `babel-preset-flow` with either
 {{include.install_command}} babel-cli babel-preset-flow
 ```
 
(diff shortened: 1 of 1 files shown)


nsfmc:
i'm not a team member, but it's worth looking at the upcoming @babel/preset-react for babel 7, here https://www.npmjs.com/package/@babel/preset-react and specifically this comment: 

> Note: Flow syntax support is no longer enabled in v7. For that, you will need to add the Flow preset.

so i think it's probably a good idea for people to keep using it (as this will almost certainly be a problem for people upgrading from babel 6->7)

davidgilbertson:
What an excellent catch! I'm gonna close this.

PR #6311, opened 2018-05-16
Update objects.md

davidgilbertson:
Changed 'fewer' to 'more'.

This was the wrong way around, right? The above is about Flow cleverly inferring type, and this will improve even more over time.


mrkev:
Hmm, that line in general doesn't make much sense to me tbh

PR #6320, opened 2018-05-17
Ensure value is added to estree ast for ObjectTypeInternalSlot

existentialism:
While implementing support for new additions to Flow 0.72 in Prettier, I noticed `value` was missing from `ObjectTypeInternalSlot`.

/cc: @samwgoldman


samwgoldman:
Thanks @existentialism!

PR #6335, opened 2018-05-21
Add TextEncoder and TextDecoder to the module 'util' (Fixes #6326)

### LINKED ISSUE #6326: TextDecoder and TextEncoder missing from node's "util" module
Flow has a module definition for `util`:
https://github.com/facebook/flow/blob/4d8376ab1541c7ed70fa57b243bfe377609cd801/lib/node.js#L1711-L1723

But it lacks the definition for TextDecoder and TextEncoder that node has since v8.3.

Code to reproduce:
```javascript
import { TextEncoder, TextDecoder } from 'util';
```
Error:
```
Cannot import TextDecoder because there is no TextDecoder export in util.
Cannot import TextEncoder because there is no TextEncoder export in util.
```

julienw:


lib/node.js
```diff
@@ -1720,6 +1720,8 @@ declare module "util" {
   declare function inherits(constructor: Function, superConstructor: Function): void;
   declare function deprecate(f: Function, string: string): Function;
   declare function promisify(f: Function): Function;
+  declare export var TextEncoder: Class<TextEncoder>;
+  declare export var TextDecoder: Class<TextDecoder>;
```
julienw: If I don't use the `export` (like for others module), flow doesn't find `TextEncoder`... but finds `TextDecoder`, for my test file. Super strange.

tests/node_tests/node_tests.exp
```diff
@@ -1334,8 +1334,8 @@ References:
    process/nextTick.js:27:3
```
julienw: only line number changes

tests/node_tests/util/text-utils.js
```diff
@@ -0,0 +1,12 @@
+import { TextDecoder, TextEncoder } from 'util';
```
julienw: I manually checked that when I use `require` this works as well.

julienw:
Please do not review for now, I noticed I broke other exports for 'util'.

julienw:
OK, should be fixed

lib/node.js
```diff
@@ -1709,17 +1709,19 @@ type util$InspectOptions = {
...
-  declare function format(format: string, ...placeholders: any): string;
-  declare function log(string: string): void;
-  declare function inspect(object: any, options?: util$InspectOptions): string;
-  declare function isArray(object: any): boolean;
-  declare function isRegExp(object: any): boolean;
-  declare function isDate(object: any): boolean;
-  declare function isError(object: any): boolean;
-  declare function inherits(constructor: Function, superConstructor: Function): void;
-  declare function deprecate(f: Function, string: string): Function;
-  declare function promisify(f: Function): Function;
```
julienw: TBH I don't know hiw this could work before :/

nmote:
Looks like the CLA was never signed, feel free to reopen if this changes.

PR #6380, opened 2018-05-30
Make `Array.of` typesafe

### LINKED ISSUE #6359: Array.of does not recognise clearly what type of argument was passed in them
No errors are shown in next code:
```
const myStringArr: Array<number> = Array.of('foo');
const myNumberArr: Array<string> = Array.of(5);
```

An example in the playground: https://flow.org/try/#0MYewdgzgLgBAtgTwMpQE4EswHMCCrUBcMeqAhggDxgCucARgKaoB8MAvMfuQHQgBmACgDkfECCEBKANwAoUJFiIAcrUaoSREuQrQM2Vhy0JeggKzSgA

wchargin:
Summary:
The static method `Array.of` can have a well-defined type, without
resorting to `any`. Fixes #6359.

Test Plan:
Added a test case to the `arraylib` suite. Running `make test` passes,
and the tests properly fail when the patch to `lib/core.js` is reverted.

wchargin-branch: polymorphic-array-of

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -282,7 +282,7 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     static from<A>(iter: Iterator<A>, mapFn: void): Array<A>;
     static from<A>(arrayLike: {length: number}, mapFn: (elem: void, index: number) => A, thisArg?: any): Array<A>;
     static from(arrayLike: {length: number}, mapFn: void): Array<void>;
-    static of(...values: any[]): any[];
+    static of<T>(...values: Array<T>): Array<T>;
 }
 
 declare class String {
--- a/tests/arraylib/array_lib.js
+++ b/tests/arraylib/array_lib.js
@@ -55,3 +55,10 @@ function from_test() {
     return String(val);
   });
 }
+
+function of_test() {
+  var emptyArrayOkay: Array<empty> = Array.of();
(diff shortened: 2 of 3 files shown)


mrkev:
This looks good. Small nit: could you change this to use generic (`Array<...>`) syntax for consistency?

wchargin:
Done. (I had used `T[]` for consistency with the previous code, but
changing to `Array<T>` makes it more consistent with the surrounding
code, which makes sense to me.)

deecewan:
how does this work with things like `document.querySelectorAll`, where the item you're passing in isn't actually an Array?

wchargin:
@deecewan: You’re probably thinking of `Array.from`, which takes
`Iterable<T>` to `T[]` and already has proper Flow type definitions. By
contrast, `Array.of(x)` is always the same as just `[x]`.

You can check this by noting that each of the following evaluates to an
array of length `1` whose sole element is a `NodeList`…
```js
Array.of(document.querySelectorAll("p"));
Array.of(document.querySelectorAll("definitely-not-a-real-tag"));
```
…or by reading [the spec for this method][1], which is pretty readable.

[1]: https://www.ecma-international.org/ecma-262/6.0/#sec-array.of

deecewan:
I am indeed thinking of `Array.from`. Apologies!

PR #6393, opened 2018-05-31
Fix some definitions in node Stream

BastienAr:
Fix chunk type (can be anything in ObjectMode) according to the [documentation](https://nodejs.org/api/stream.html#stream_transform_transform_chunk_encoding_callback)

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1464,7 +1464,7 @@ declare class stream$Readable extends stream$Stream {
   resume(): stream$Readable;
   unpipe(dest?: (stream$Writable | stream$Duplex)): void;
   unshift(chunk: Buffer | string): void;
-  push(chunk: ?(Buffer | string), encoding? : string): boolean;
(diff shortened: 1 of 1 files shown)


mrkev:
Hmmm, I wonder if there's another way to implement these "Object Mode" semantics. 

First of all, `any` should be avoided in favor of `mixed`. `mixed` can't be used here because it includes `null`, which is not a valid parameter in object mode.

And in any case, it'd be good to be able to make the distinction between a stream in object mode and one that is not, so as to avoid type errors by passing random objects to non-object mode streams. My guess here is perhaps these could be two different classes, and there could be two declarations for functions creating streams: one outputting a normal stream when object mode is false, and an object mode stream when it's true.

nmote:
I just took a look at this. It looks like object mode streams are not the common case. I don't think it's a good idea to risk encountering runtime type errors in the common case in order to support the less common case.

@mrkev's suggestion of using a separate class to model a stream in object mode is reasonable. If you'd like to pursue that route, feel free to put up an issue to discuss (CC me) or join our Discord and we can discuss there.

PR #6420, opened 2018-06-05
css properties are nullable

antonk52:
I came around a scenario where I might have discovered a bug, by the spec we can reset the styles on an HTML element by assigning an empty string or `null`. Here is a quote from an [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style)

```
A style declaration is reset by setting it to null or an empty string, e.g. elt.style.color = null.
```

At the moment if we imagine a scenario where we have an element and we perform some style modifications on it and later on decide to reset the styles flow will error out. Here is an example

```javascript
const myDiv: HTMLElement = document.querySelector('div');
myDiv.style.color = 'red';
// Cannot assign `null` to `myDiv.style.color` because null is incompatible with string
myDiv.style.color = null;
```

For now I used a group type with `string` and `null`. And I am not sure if this is the best way to fix the issue.

There are multiple ways to approach similar result. We could use maybe types but that will also allow use of `undefined` which does work in browsers yet I am not sure if it should be used as a good practice.  

Another point is that even after assigning a `null` value when reading we will always get an empty `string` value. Therefore perhaps having getters and setters here could be another solution.

I want to leave this PR and see if someone else might have a more elegant solution for this issue.

Thanks


fishythefish:
@antonk52 This does look like the libdefs don't match the specification at first glance, but you should still be able to reset a style by assigning `""` instead of `null`.

Note that the [CSSOM specification](https://drafts.csswg.org/cssom/#the-cssstyledeclaration-interface) specifies that these properties are all (non-nullable) strings. `null` can be supplied as an input, but is treated as `""`. Flow doesn't currently support different input and output types for a property, so we generally stick with the output type, which is `string` here. Making this change would require every *read* of these properties to be checked for `null`.

PR #6473, opened 2018-06-12
Add scrollIntoView option enums

mvanlonden:
Adds enums according to https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView


mvanlonden:
Closing in favor of https://github.com/facebook/flow/pull/5520

PR #6476, opened 2018-06-13
Change wording on the landing page - ‘Coworkers’

robinpokorny:
Currently, on the landing page it reads:
‘Flow can help prevent bad rebases. Flow can help **protect** your carefully designed library **from your coworkers**.’ (Emphasis mine)

I do not want to protect my library from coworkers. I want them to be able to quickly and safely contribute.

In this PR I suggest ‘protect […] from errors’. However, I'd be happy with other changes.

### DIFF
--- a/website/_data/i18n/en.yml
+++ b/website/_data/i18n/en.yml
@@ -37,7 +37,7 @@ homepage_code_bigger_img_alt: ""
 homepage_code_bigger_description: >
   Working in a codebase with lots of developers can make it difficult to keep
   your master branch working. Flow can help prevent bad rebases. Flow can help
-  protect your carefully designed library from your coworkers. And Flow can help
+  protect your carefully designed library from errors. And Flow can help
   you understand the code you wrote six months ago.
 
 homepage_featurettes:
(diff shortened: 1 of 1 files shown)


mrkev:
wow just noticed it hahah. Yeah I can see this being change-worthy, thanks!

mrkev:
I think I'd prefer this be "Flow can help protect your carefully designed library from misuse and misinterpretation.", which is probably more on in spirit of the original text.

The "misuse" bit: JavaScript is very dynamic-- you can add and remove properties from pretty much any object at any point, browse up and down the prototype chain as you'd like, pass in objects that act like other objects or pretend to be functions/arrays... the way I read it is: if I meant to write a library that is moving an object around, and that shouldn't be mutated, I don't need to worry that someone is going to pass objects that mutate, or fake objects, or is going to inject their own methods into my library to handle their use-cases.

The "misinterpretation" bit: you can name types. If you're coding a big project it's not rare that things get bolted on the wrong places, because people understand things differently. For example, if I have a configuration object, it's well typed, defined in a single place everyone can see, and it's named `type Configuration = ...`, that helps me ensure people will understand if they want to pass in data this is not the object that should house it, but rather they should look somewhere else.

I think errors is a bit too much like the "bugs" sentence on "Code faster".

Thoughts?

robinpokorny:
@keithamus Sure! `misuse and misinterpretation` sounds much better. Thanks. Should I change that?

mrkev:
Yeah, can u? Ping me when done and I'll re-import 👍

robinpokorny:
@mrkev, I updated it 👍 

(Yes, I typed ‘Kevin’ and blindly accepted the first suggestion 😄)

PR #6485, opened 2018-06-17
Website EOT Document Tweak: Explicitly Mention Vertical Bar in EOT

cyan33:
[The "Exact Object Type" doc page](https://flow.org/en/docs/types/objects/#toc-exact-object-types) doesn't even have a SINGLE word that mentioned "pipe operator", which makes a new comer like me difficult to search what I'm looking for. (Actually it's not mentioned across the whole website:))

Also, accessibility-wise, my suggestion is we should at least mention "what the syntax looks like", rather than only about "what it does". This is the thing people are based on when they do searching and find related materials.

Open to any re-wording suggestions.

### DIFF
--- a/website/en/docs/types/objects.md
+++ b/website/en/docs/types/objects.md
@@ -215,7 +215,7 @@ method({
 > **Note:** This is because of ["width subtyping"](../../lang/width-subtyping/).
 
 Sometimes it is useful to disable this behavior and only allow a specific set
-of properties. For this, Flow supports "exact" object types.
+of properties. For this, Flow supports "exact" object types through enabling to add a pair of vertical bars.
 
 ```js
 {| foo: string, bar: number |}


Huxpro:
I am not quite sure would u wanna call it pipe operator, which usually refer to `|>` in JavaScript/Hack/OCaml etc and only refer to `|` under Unix context. What about just "vertical bar" according to <https://en.wikipedia.org/wiki/Vertical_bar>? 😂

cyan33:
That's a good point. Since "vertical bar" is used in [Union Type](https://flow.org/en/docs/types/unions/#union-type-syntax-), I'll change it to keep it consistent.

website/en/docs/types/objects.md
```diff
@@ -215,7 +215,7 @@ method({
 > **Note:** This is because of ["width subtyping"](../../lang/width-subtyping/).
 
 Sometimes it is useful to disable this behavior and only allow a specific set
-of properties. For this, Flow supports "exact" object types.
+of properties. For this, Flow supports "exact" object types through enabling to add a pair of vertical bars.
```
mrkev: Oh, sad this didn't get more eyes, I like this change! I would change this to 

> Flow supports "exact" object types. Their syntax is just like that for objects, but with vertical bars after the opening curly brace and before the closing one:

or something along those lines

PR #6510, opened 2018-06-23
Add definition for React.forwardRef()

### LINKED ISSUE #6103: Missing React.forwardRef definition
React 16.3 introduced a new [`forwardRef()`](https://reactjs.org/blog/2018/03/29/react-v-16-3.html#forwardref-api) API, for which Flow does not currently have a definition.

motiz88:
Closes #6103. There is some context re: soundness in https://github.com/facebook/flow/issues/6103#issuecomment-399657238.

~The type of the `render` argument here diverges very slightly from that in React `master` at the moment, but see https://github.com/facebook/react/pull/13100 for the reason why.~

UPDATE: With https://github.com/facebook/react/pull/13100 now merged, this is aligned with the React source code's own definition of `forwardRef`.


motiz88:
Unfortunately I'm not able to build this branch merged with Flow `master`. Have tracked the issue down to d268d0d4726f54702c5fd7fd9f76d4982f992952, so I've merged its _parent_ commit here, but the vast majority of conflicts are from later in `master`'s history. Help?

motiz88:
Hey @AlmeroSteyn,

1. I think https://github.com/facebook/react/issues/12553 gives a complete picture of React 16.3 features in Flow and that's probably the best place to discuss them as a whole. I wish I could tell you anything about timelines, but your guess is as good as mine I'm afraid.

2. https://flow.org/en/docs/react/refs/ is probably due for an update documenting `createRef` (and `forwardRef` for that matter, when this is merged) since it only mentions ref functions at the moment. I'm not aware of any other source of documentation for this, but I'm sure a docs PR will be welcome 😉

motiz88:
@mrkev Any chance you could give this a look? I'm not overjoyed about the https://github.com/facebook/flow/issues/6103#issuecomment-399657238 situation but I think it's the best we can get out of the current mechanics of Flow, and the drawbacks are outweighed by the benefits of providing _some_ form of support for this React feature, 5 months now since its release.

therealparmesh:
I think there may need to be more work done here.

I needed the `React.forwardRef` flow type in a project so I made this wrapper and copy pasted from this PR.

```js
// @flow

// $FlowFixMe
import { forwardRef as ReactForwardRef } from 'react';

export default function forwardRef<Props>(
  render: (props: Props, ref: React.Ref<React.ElementType>) => React$Node
): React$ComponentType<Props> {
  return ReactForwardRef(render);
}
```

The problem is that `React.forwardRef` doesn't actually return `React$ComponentType`. [It returns an object whose `render` key is of `React$ComponentType`.](https://github.com/facebook/react/blob/v16.6.3/packages/react/src/forwardRef.js#L49)

So if I want to assign something static like `defaultProps`, I have to do so on `React.forwardRef(/* ... */).render`. Using the current type definition, this code will throw a flow error.

I think this implies that other flow types in the `react` module need to be adjusted to handle this object.

samwgoldman:
This was reviewed internally a while back, but we never updated this PR. Sorry about that. The proposed types here are not quite correct and lead to missed errors when the `ref` pseudo-prop of the forwardRef component is actually used.

Supporting forwardRef properly requires some deeper changes to Flow's React support, which @jbrown215 is working on.

PR #6514, opened 2018-06-24
Add dgram$Socket$rinfo to lib/node.js

hczhcz:
According to [dgram docs](https://nodejs.org/api/dgram.html#dgram_event_message)

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -597,7 +597,7 @@ declare module "crypto" {
 }
 
 type net$Socket$address = {address: string; family: string; port: number};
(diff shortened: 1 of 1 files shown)


mrkev:
Sweet thanks, two things:

1. This new type is defined, but isn't used anywhere?
2. family could be made `'IPv4' | 'IPv6'`

hczhcz:
@mrkev
1. It can be used in the event `.on('message', (msg: Buffer, rinfo: dgram$Socket$rinfo) => void)`.
2. Looks good; Does it also apply to `net$Socket$address.family`? Node's documentation does not have enough information.

PR #6561, opened 2018-07-03
Add Symbol.observable

mmiszy:
I was wondering if this is a the right place to add this.
Symbol.observable is just an ECMAScript Draft proposal.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -88,6 +88,7 @@ declare class Symbol {
   static hasInstance: $SymbolHasInstance;
   static isConcatSpreadable: $SymboIsConcatSpreadable;
   static iterator: string; // polyfill '@@iterator'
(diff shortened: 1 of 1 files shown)


mrkev:
Mmm little split on adding proposals to the core definitions right now. 

To be honest, ideally there should be different versions of core, dom, etc. Right now these core libs are kind of all over the place-- I don't doubt there's stuff there that ideally shouldn't, and stuff missing that should, but we should really do a better job about keeping it in line with stable.

I've been thinking of copying the current definition, adding a few draft proposals and PR-ing it to  [flow-typed](https://github.com/flow-typed/flow-typed) as core-es7.js or something but just haven't had time or bandwidth to really flesh out the workflow for users, implications, etc.

Issue #4647, opened 2017-08-18
flow-upgrade not working under Windows


tirli:
I've tried to run flow-upgrade under Windows 10 and it fails
![capture](https://user-images.githubusercontent.com/7783308/29453378-13de40b2-8412-11e7-8ae7-7dc1d4e61e5f.PNG)

Then I tried to do it with Windows Ubuntu Bash and everything works. So it seems like it's only Windows specific problem.

mjomble:
As a quick workaround for anyone who doesn't want to install Windows Ubuntu Bash, you can open the runCodemods.js file shown on the first line of the call stack and edit `fs.mkdtempSync('/tmp/flow-upgrade-')` to something like `fs.mkdtempSync('C:/tmp/flow-upgrade-')`.
Then make sure the `C:\tmp` dir exists and run flow-upgrade again. Worked for me on Windows 7.

lll000111:
> for anyone who doesn't want to install Windows Ubuntu Bash

If you install git for Windows you get a `bash` plus the usual Unix command line tools for free.

tirli:
@lll000111 my screenshot is from git bash ;) It's not working. As well as PowerShell

lll000111:
Okay... so this problem can be reproduced with this simple command (taken from the actual code) in a node.js console:

    fs.mkdtempSync('/tmp/flow-upgrade-')

tirli:
![image](https://user-images.githubusercontent.com/7783308/29513776-023f068e-866f-11e7-9a01-7e0ebd747a75.png)

yep.

lll000111:
See the node.js issue I created referenced above - the absolute path on Windows gets translated to `C:\tmp\...`

JMontagu:
Can confirm, same issue occurring - @lll000111's workaround above resolved it.

mjomble:
I had all my line endings converted as well.

mrkev:
@lll000111 if anyone wants to make a PR of this and ping me, the fix seems simple. It seems to be just changing the line here: https://github.com/facebook/flow/blob/5aaf9b9177514c7ed0c2b65af7efff4fea5c5775/packages/flow-upgrade/src/codemods/runCodemods.js#L27

From `fs.mkdtempSync('/tmp/flow-upgrade-'),` to `fs.mkdtempSync(path.join(os.tmpdir(), 'flow-upgrade-'),`.

Busy with lots of other stuff so it's in the backlog for some time. Easy GitHub square though (:

[...]

PR #6575, opened 2018-07-09
Fix for windows enoent on flow-upgrade

jeffersoneagley:
Changed the lines specified in  #4647 by @mrkev and added associated require statement

This is to fix flow-upgrade throwing enoent errors due to Windows' security settings against allowing modification of /tmp (which node translates to C:\tmp)


jeffersoneagley:
Side note: In case it's not showing up yet, I did sign the contribution agreement

PR #6579, opened 2018-07-09
Docs: fix subtype example indention

gustavpursche:
Fixes a misaligned code example on the subtypes page: https://flow.org/en/docs/lang/subtypes/


gabelevi:
My guess from the context is that this might have been intentional, but I agree that it's not very helpful and that removing the indentation is a good thing to do. Thanks!

PR #6584, opened 2018-07-10
Create an explicit type for Contexts 

EdwardDrapkin:
Sometimes you want to interact with a React Context in an abstract way, such as an implementation for a higher order component that spreads a context consumer into a component.  The way that `createContext` is currently typed makes this a little more difficult than it probably needs to be, so I created a type called `React$Context<T>` and exported the type from the react package as well which seems to be the standard approach for types like this.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -192,6 +192,15 @@ declare type React$Ref<ElementType: React$ElementType> =
   | string;
 
 /**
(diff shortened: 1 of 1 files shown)


gabelevi:
This seems reasonable. I'll get it merged...

Issue #2957, opened 2016-12-06
FileReader bug


caperavensoftware:
```
            let reader = new FileReader();
            reader.readAsBinaryString(file.slice(0, 10));
```

flow states that readAsBinaryString is not a property of FileReader.

caperavensoftware:
if the intent is to say that the function is depreciated, the message should just be a bit more clear.

PR #6585, opened 2018-07-10
added readAsBinaryString to FileReader #2957

parviz-93:


gabelevi:
Hmm, [the docs](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsBinaryString) seem to say that `readAsBinaryString` can take a `Blob` or `File`. But they say that for [readAsArrayBuffer too](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsArrayBuffer), so I think this is fine for now.

For future reference, [the FileAPI draft says](https://w3c.github.io/FileAPI/#blobAndFileParams)

> This is a Blob argument and must be a reference to a single File in a FileList or a Blob argument not obtained from the underlying OS file system.

Issue #3502, opened 2017-03-11
Array<Array<string | number>> does not allow Array<Array<number>> as an input


gajus:
```js
type DatabaseQueryParametersType =
  Array<
    Array<string | number>
    >;

type DatabaseConnectionType = {
  query: (query: string, paramters: DatabaseQueryParametersType) => void
};
  
const getGenresByMovieIds = (connection: DatabaseConnectionType, movieIds: Array<number>): void => {
  connection
    .query('', [
    	movieIds
    ]);
};

```

gives error:

```
3:     Array<string | number>
             ^ string. This type is incompatible with
10: const getGenresByMovieIds = (connection: DatabaseConnectionType, movieIds: Array<number>): void => {
                                                                                     ^ number
```

https://flowtype.org/try/#0C4TwDgpgBAIghsOAjOBnCBFArhATiABTlzgFsJg9UAVcaAXgCgooBBXEkAHmZbY7jdUwXAEsAdgHMoAHyjispJHgB8vFioDcjRqEiwEyNBADCAe3HiIAY2CiLtffSgBvXgEcc+AFxQAFJ54IL7CYlIANFBgxGSUuKi+8Igo6NhBRCTkcTR0AJRQ9CpQAG5mogAmjAC+2iyM1hbCUJIUAOIQ4rgQqABCIACyZsWiEACS5agF-g2WNnYWiYYpphZWtvbijhCRpEMj4wn8nFwKSqq5vqUVBUVuLDNr8+LqUAB0gfh+AORfkQDaLwAkLthmMJi8ALq5bQ1RhAA

gajus:
Interestingly, a literal `Array<number>` does work:

[...]

PR #6595, opened 2018-07-13
Document `$ReadOnlyArray<T>`

### LINKED ISSUE #5033: $ReadOnlyArray needs to be listed in the documentation (Depth Subtyping?)
EDIT:
$ReadOnlyArray needs to be listed in the depth subtyping documentation section to explain why Arrays that contain generic types cannot be set to arrays that contain specific types, as it risks the specific-typed array being changed to contain a generic type value.

e.g. variable C will give an error while variable D will not:
```
const a: Array<number> = [1, 2, 3];
const b: Array<number | string> = [1, 'two', 3];
const c: Array<number | string> = a;
const d: $ReadOnlyArray<number | string> = a;

// Why C is invalid:
c.push("string"); // a: Array<number> would now contain [1, 2, 3, "string"]
```

This is not documented anywhere, and really needs to be.

Original question left below, but not as relevant as the above solution:

[example](https://flow.org/try/#0PTAEAUFcBdoSwHYHNQGpSNAMwE4HsFpQ8tQBDchAE1ACNQcBTAWzwDdGBnUaAdz1CMc+HNyx4coTnmaMGjMtIQAaAFAg6MUAElQVAgHIiCRgGMunMjjgAbAJ6heZQjwAWciXCSIyN4rQArM2huaAFaOThmZhgyWhtGADpVdTAAZQAHMzgsOFMeOyzVaEK5TOzc-IBeUABvcgAuUARIZgicZTomzmhrZFAAXwBuFI0AUThod0kSotm5ADk8aHLTHLzQGvrUAG0Aa0Y7bt7EJABdJpa2oVAAHykT-uHRsCxIBFN4AjcyIjJTcwZEL+IKfbgkQSTaYFOalUAAYRkGQIjEI4HwGW4W1UoD0vzITQAgsIyHYADxLFZZNaVAB8qmepgIPVATOYyJMLhqAAoMhjOE1EeyUWj+QBKTa0uqgYAAKj0AgQy1cp1AsrAz1SIOCrII4hwzFVvChUmp63y804qiZCBZjAAHmR2QkACL4pqrc2baUE0AARk6tCaRn4BkGIy1-Bwe24xqmeC0VhwpKtNrtjudjDd0DIACYmpTPZVvQ6nRlXfiIxoozGtUnSaymL9GDREDZEIxrUiRdBubUcXic00dqXM9myGcGWKq2Aa1aNPWHBDEFghEwaJbQNzXHheBhQiruFQcqumC4ps43HITPaiKPy4wAPxiru2u8Zh-E5NHUBf0lk2UpRqEcPwrHMzhGNkOVRXt+1xKh3UEUDGD-OwpxnBUuAQIxHAkPY6xJJdSE4M1i0tV90zLBJUPzX9CLJIs8iA0AQKorN8QgrthU5WCBwQockLYmj0NUIA)

```
// Putting + in front of a and b removes two errors for some reason,
// but I don't necessarily want the original objects to be immutable.

// Specific type
type Specific = { a: number, b: string };

// Either type
type NotSpecific = { +[key: string]: number | string };

// function that accepts objects of either type
type ComponentProps = {
  data: Array<NotSpecific>
};
const component = (props: ComponentProps) => { /* do nothing */ };

// object conforming with specific types
const exampleData: Specific = { a: 1, b: 'two' };

// works without arrays
const exampleData2: NotSpecific = exampleData;

// works
// array created inline
component({
  data: [exampleData]
});

// works
// array of inferred types (how is this different than the next example?)
const exampleArray: Array<*> = [exampleData];
component({
  data: exampleArray
});

// doesn't work
// array of specific types
const exampleArray2: Array<Specific> = [exampleData];
component({
  data: exampleArray2
});

```

I am trying to use the last example here. I have an array of objects of a specific type (e.g. { a: number, b: string }. I want to pass them as a prop to a component that accepts an array of any objects, as long as the values are number or string; it doesn't matter what the key is.

This works if the array is inline or inferred (I must be confused as to what the * means, because I thought it meant it would grab the type from the code, in this case exampleData). But it does not work if the type is declared, even though it works without arrays.

That is to say:
```
// works
const exampleData: Specific = { a: 1, b: 'two' };
const exampleData2: NotSpecific = exampleData;

// doesn't work
const exampleArray: Array<Specific> = [exampleData];
const exampleArray2: Array<NotSpecific> = exampleArray;
```

I'm not having any luck using +Array<type> or Array<+Type> to state that the array is covariant, but really I'm not sure what I'm supposed to be doing here to make this work.

reyronald:
Been using this myself for a while despite it not being documented. Can we formally include it in the Utility Types documentation?

https://github.com/facebook/flow/blob/d390d54460ef86ce18dfa7d270be8f8de119da04/src/typing/type_annotation.ml#L249-L256

Related to #3502
Closes #5033


motiz88:
Drive-by comment, sorry 😅 

1. Would it perhaps be better to document this as part of [Array Types](https://flow.org/en/docs/types/arrays/) rather than the Utility Types page? Thinking of discoverability mostly.
2. Should we also clarify that `$ReadOnlyArray` _doesn't_ propagate read-only-ness into the element type? It might not be obvious to everyone, depending on which type system they're coming from.

reyronald:
No need to apologize, all comments and suggestions are welcome!

1. Sure. If it was up to me  I would leave the extended documentation of $ReadOnlyArray in Utility Types and include a reference to it from the Array Types page. Something like "There's also a $ReadOnlyArray, click here to read more about it" and the link would take you there, thoughts?
2. I have nothing against that, how would you phrase it? Could you give me something that I can just copy/paste in? :P

motiz88:
My personal preference would be to _move_ this to Array Types. I see `$ReadOnlyArray` as an array type, with the only thing it shares with the other Utility Types being a leading `$`. In other words, to me it has more similarities with `Array<T>` than with `$ReadOnly<T>`.

As for the note on mutable elements, I would go with something like this:

`````markdown
Note that an array of type `$ReadOnlyArray<T>` can still have mutable _elements_:

```js
// @flow
const readonlyArray: $ReadOnlyArray<{x: number}> = [{x: 1}];
readonlyArray[0] = {x: 42}; // error
readonlyArray[0].x = 42; // ok
```
`````

**Comment by reyronald** [comment]
Thanks for the snippet! Included it in the docs. Also moved to Array Types as you suggested!

**Comment by callumlocke** [comment]
I see the reasoning for putting this on the Array Types page, as it's not really a utility, it's a type in itself. But it seems wrong not to mention it on the Utility Types page. Can you add a note at the end of the `$ReadOnly` section, something like:

> **Note:** `$ReadOnly` is only for making read-only **object** types. See the Array docs for [how to type read-only arrays](../arrays/#toc-readonlyarray).

**Comment by reyronald** [comment]
@callumlocke Thank you for the suggestion! I've pushed the requested change.

Does everybody feel more comfortable with the current way it is documented now?

File: website/en/docs/types/utilities.md [hunk 1]
```diff
@@ -145,6 +145,8 @@ type Obj = {
 type MappedObj = $ReadOnly<$ObjMap<Obj, TypeFn>> // Still read-only
 ```
 
+> Note: $ReadOnly is only for making read-only object types. See the Array docs for [how to type read-only arrays](../arrays/#toc-readonlyarray).
```
roryokane: You should mention the name `$ReadOnlyArray` here in case someone searches the Utility Types page for that name. I suggest wording the second sentence like this:

~~~markdown
See the Array docs for how to [type read-only arrays](../arrays/#toc-readonlyarray) with `$ReadOnlyArray`.
~~~

website/en/docs/types/arrays.md
```diff
@@ -116,3 +116,27 @@ if (value !== undefined) {
...
+
+Similar to `$ReadOnly<T>`, it is the supertype of all arrays and all tuples and
+represents an immutable array. It does not contain any methods that will allow an
+object of this type to be mutated (no `push()`, `pop()`, etc.)
+
+```js
+// @flow
+const readonlyArray: $ReadOnlyArray<number> = [1, 2, 3]
+
+const first = readonlyArray[0] // Ok to read
```
roryokane: `Ok` should be capitalized: `OK`

website/en/docs/types/arrays.md
```diff
@@ -116,3 +116,27 @@ if (value !== undefined) {
...
+readonlyArray.push(4)          // Error
+```
+
+Note that an array of type `$ReadOnlyArray<T>` can still have mutable _elements_:
+
+```js
+// @flow
+const readonlyArray: $ReadOnlyArray<{x: number}> = [{x: 1}];
+readonlyArray[0] = {x: 42}; // error
+readonlyArray[0].x = 42; // ok
```
roryokane: capitalize `ok` as `OK` here too

website/en/docs/types/arrays.md
```diff
@@ -116,3 +116,27 @@ if (value !== undefined) {
 
 As Flow is made to be smarter it may be possible in the future to fix this
 problem, but for now you should be aware of it.
+
+## `$ReadOnlyArray<T>` <a class="toc" id="toc-readonlyarray" href="#toc-readonlyarray"></a>
+
+Similar to `$ReadOnly<T>`, it is the supertype of all arrays and all tuples and
```
roryokane: Make `$ReadOnly<T>` a link to https://flow.org/en/docs/types/utilities/#toc-readonly.

website/en/docs/types/arrays.md
```diff
@@ -116,3 +116,27 @@ if (value !== undefined) {
...
+represents an immutable array. It does not contain any methods that will allow an
+object of this type to be mutated (no `push()`, `pop()`, etc.)
```
roryokane: Add a period `.` after the closing parenthesis.

roryokane:
Also, it would be nice to mention in this documentation what seems to me the primary use-case for `$ReadOnlyArray`: fixing confusing Flow errors where Flow thinks you might mutate an `Array`. But that could be added later if you don’t feel like writing it up for this PR.

The problem is described in [issue #3502](https://github.com/facebook/flow/issues/3502) (which was linked in the PR description), and the way `$ReadOnlyArray` solves it is described in [this comment](https://github.com/facebook/flow/issues/3502#issuecomment-286259843). Discussing the possibility of errors related to Array mutability in this documentation will hopefully lead people to the fix more quickly.

reyronald:
@roryokane Thanks for your suggestions! Just pushed your required changes.

I also explained the use-case you mentioned with the simplest example I could think of. Obviously I'm using my own words and I'm biased so if someone else can think of a better way to phrase it please come forward!

website/en/docs/types/utilities.md
```diff
@@ -145,6 +145,8 @@ type Obj = {
 type MappedObj = $ReadOnly<$ObjMap<Obj, TypeFn>> // Still read-only
 ```
 
+> Note: $ReadOnly is only for making read-only object types. See the Array docs for [how to type read-only arrays](../arrays/#toc-readonlyarray) with `$ReadOnlyArray`.
```
roryokane: Suggestions I didn’t think of before:

- `Note: $ReadOnly` should be ``Note: `$ReadOnly` ``, with backticks
- Emphasize the word “object”: `read-only _object_ types`

Comments on your added text:

- I think the link would sound better if its bounds were changed from “how to type read-only arrays” to “type read-only arrays with `$ReadOnlyArray`”:

> See the Array docs for how to [type read-only arrays with `$ReadOnlyArray`.](../arrays/#toc-readonlyarray)

This is more natural because `$ReadOnlyArray` is the only way to do it, not just an example of how to type read-only arrays.

mrkev:
Wow this is great, has explanations, examples, multiple entry points, amazing! 😄

reyronald:
@gabelevi Done! Let me know if that's what you had in mind :)

website/en/docs/types/arrays.md
```diff
@@ -142,10 +142,11 @@ readonlyArray[0] = {x: 42}; // Error!
...
 
-If you use `$ReadOnlyArray` instead of `Array` for arrays that are never mutated,
-you can prevent Flow from raising spurious errors. For mutable arrays of generic
-types that are set to arrays of specific types there's the possibility of adding
-values to it that don't match the type of the specifically-typed array.
+The main advantage to using `$ReadOnlyArray` instead of `Array` is that `$ReadOnlyArray`'s
+type parameter is *covariant* while `Array`'s type parameter is *invariant*. That means that
+`$ReadOnlyArray<number>` is a subtype of `$ReadOnlyArray<number | string>` while
+`Array<number>` is NOT a subtype of `Array<number | string>`. So it's often useful to use
+`$ReadOnlyArray` as a type annotation to allow various different types of arrays.
```
roryokane: > So it's often useful to use `$ReadOnlyArray` as a type annotation to allow various different types of arrays.

I find this sentence confusing, because “various different types of arrays” is ambiguous. I would word the sentence like this:

> So it's often useful to use `$ReadOnlyArray` in type annotations for arrays of various types of elements.

mrkev:
(@gabelevi liked it, merging soon!)

website/en/docs/types/arrays.md
mrkev: Hmmm yeah, I could see that. I personally don't think it's that ambiguous given the context but the change would be strictly more clear.

This PR has been out for a while and I don't want to keep @reyronald going back and forth. Merging it as is but if you're so inclined @roryokane feel free to PR a change in the wording.

roryokane: @mrkev I made the PR with that change: https://github.com/facebook/flow/pull/6816

PR #6614, opened 2018-07-19
Add TransitionEvent.

koddsson:
Add `TransitionEvent` as a type and accept it as a parameter to `addEventListener` and other such functions.

I think this is correct? I might have missed something. I tried following the code for the other events such as the `ClipboardEvent`.

I also followed the interface on https://www.w3.org/TR/2017/WD-css-transitions-1-20171130/#interface-transitionevent to do this.


koddsson:
There's a lot of output but I'm unsure why the tests are failing 👀

PR #6632, opened 2018-07-24
Improve react (React$Context.Provider) typings

Jabher:
add optional  `children` prop to React$Context.Provider

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -196,7 +196,7 @@ declare type React$Ref<ElementType: React$ElementType> =
  * createContext() with a default value.
  */
 declare type React$Context<T> = {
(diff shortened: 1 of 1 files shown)


TrySound:
Does optional children make any sense?

Jabher:
Do not think so in terms of common development, but it works in that way (`<context.Provider val="bar" />` is handled by react properly), and I suggest that flow, as type system should check not "saneness", but "realism" of what's written.

Also, I can imagine some complex system behavior with HOC composition pattern that will cause such cases from time to time, e.g. wrapper for some component with logic similar to `<this.props.ProviderComponent {...this.props.children ? {children: this.props.children} : {}}/>`.

PR #6660, opened 2018-07-30
callback function updated

smhmayboudi:
normally, we never call the callback function with the null for the result. In-case of error, we just call the callback with error.


mrkev:
Oh, I think this is this way on purpose. We assume someone completely new to the typesystem is reading this, and at this point we haven't talked about optional parameters yet (if you keep reading, optional parameters are introduced a couple of sections down).

Rather than fitting real-world patterns this are just examples and using a feature we haven't talked about yet could confuse very new readers. Thanks for the contribution though!

PR #6674, opened 2018-08-01
Add 'path' to Event

Ricardo-Marques:


mrkev:
Hello 👋

Could you provide documentation in the description backing this change? How do we know this is the correct definition? 

Also, style-wise, we prefer `Array<T>` over `T[]` in the libdefs 👍

Ricardo-Marques:
(Seems like it's a non-standard actually)[https://stackoverflow.com/questions/39245488/event-path-undefined-with-firefox-and-vue-js].

Apologies, closing.

mrkev:
No worries, thanks for checking!

PR #6675, opened 2018-08-01
Update lib type definition

hczhcz:
See also:

[Crypto](https://nodejs.org/api/crypto.html#crypto_crypto_scrypt_password_salt_keylen_options_callback)


nmote:
Could you split this into two PRs so that the change to `String` can be reviewed separately? Feel free to cc me to make sure it gets prompt attention.

PR #6683, opened 2018-08-02
Fix function signature for String.prototype.codePointAt method

dozzman:
The ECMAScript spec defines that the String.prototype.codePointAt
function should return undefined if there is no element at the provided
index. This means that the function signature should be:

(index: number) => (number | void)

instead of currently:

(index: number) => number

The following code should produce an 'undefined is
incompatible with number' type error but currently does not:

```
function f(s: string) : number {
  return s.codePointAt(0);
}
```

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -290,7 +290,7 @@ declare class String {
     anchor(name: string): string;
     charAt(pos: number): string;
     charCodeAt(index: number): number;
-    codePointAt(index: number): ?number;
+    codePointAt(index: number): number | void;
     concat(...strings: Array<string>): string;
     constructor(value?: mixed): void;
     endsWith(searchString: string, position?: number): boolean;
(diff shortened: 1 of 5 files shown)


dozzman:
References:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt

https://tc39.github.io/ecma262/#sec-string.prototype.codepointat

lib/core.js
```diff
@@ -290,7 +290,7 @@ declare class String {
     anchor(name: string): string;
     charAt(pos: number): string;
     charCodeAt(index: number): number;
-    codePointAt(index: number): number;
+    codePointAt(index: number): ?number;
```
motiz88: `?number` is `void | null | number`. I think `number | void`  would be more accurate.

mrkev: +1, ping me after having fixed this

dozzman: @mrkev have updated this

dozzman:
Hi @mrkev do you have additional feedback for this PR? I have made all the changes requested thus far.

mrkev:
Not that I can remember-- just didn't get to it because I was merging a lot of PRs that day 😅 I'll look at this next probably.

PR #6694, opened 2018-08-04
Add doc details for $Shape.

vicapow:


website/en/docs/types/utilities.md
```diff
@@ -578,6 +578,23 @@ function makeParamStore<T>(storeClass: Class<ParamStore<T>>, data: T): ParamStor
 (makeParamStore(ParamStore, 1): ParamStore<boolean>); // failed because of the second parameter
 ```
 
+## `Shape<T>` <a class="toc" id="toc-shape" href="#toc-shape"></a>
+
+Copies the shape of the type supplied, but marks every field optional.
```
rickyp-uber: Maybe not use "Shape" in the definition of "Shape"? How does "Copies an existing type and makes each of its properties optional" sound?

website/en/docs/types/utilities.md
```diff
@@ -578,6 +578,23 @@ function makeParamStore<T>(storeClass: Class<ParamStore<T>>, data: T): ParamStor
...
+  age: number,
+  name: string,
+}
+type PersonDetails = $Shape<Person>;
+
+const person1: Person = {age: 28}; // Error due to incorrect type of Person.
+const person2: Person = {name: 'a'}; // Error due to incorrect type of Person.
+const person3: PersonDetails = {age: 28}; // Ok
+const person4: PersonDetails = {name: 'a'}; // Ok
+const person5: PersonDetails = {age: 28, name: 'a'}; // Ok
```
rickyp-uber: Nit: Possibly add an example of it's major use case merging:
```
const originalPerson: Person = {
   age: 25,
   name: 'Original Person'
};
const partialPersonUpdates: PersonDetails = {
   name: 'Updated Person'
}
const updatedPerson: Person = {
   ...originalPerson,
   ...partialPersonUpdates
}
```

rickyp-uber: Maybe a bit confusing, not sure, I'll let you decide.

PR #6707, opened 2018-08-08
docs(utilities.md): actualize table of content, fix broken markup

nodkz:


mrkev:
Thanks! Looks like some of the errors got resolved before, but $NonMaybeType still isn't in the index. Merging

PR #6708, opened 2018-08-08
Fix docs for `$Shape`

wchargin:
Summary:
Docs for `$Shape` were introduced in #6694, but that PR had four major
errors:

  - the section was not linked from the table of contents;
  - the type `$Shape<T>` was misspelled as `Shape<T>` in the header;
  - the code block did not have `// @flow`, so lacked error annotations;
  - the code block was not terminated, so the entirety of the rest of
    the page was marked as code, [like this][1]—clearly no one previewed
    the page before deploying!

[1]: https://archive.fo/CqH7D#87%

This patch fixes those errors. While in the area, we clean up the
comments reading "Error" and "OK" to (a) eliminate the horizontal
scrollbar and (b) match the style elsewhere on the page.

Closes #6692.

Test Plan:
After much effort, managed to get the website to build locally. Visited
<http://localhost:8080/en/docs/types/utilities/#toc-shape> and checked
that the four issues mentioned above are all fixed.

(For posterity, steps taken to get the build working are listed here:
<https://github.com/facebook/flow/pull/6708#issuecomment-411308129>.)

wchargin-branch: fix-shape-docs

PR #6714, opened 2018-08-09
Fix bad markup in the Utility Types documentation

roryokane:
Also:

- fix the name of the `$Shape` utility type
- fix grammar within the `$Shape` code example

You can see that currently the page is messed up after the `Shape` section because of the unclosed code block: https://flow.org/en/docs/types/utilities/#toc-shape


roryokane:
You’re right, #6708 is a better PR for these fixes. Closing.

PR #6724, opened 2018-08-10
Add NonMabyeType to utilities.md table-of-contents

Spacerat:
I did a quick check, this seems to be the only thing missing.


mrkev:
This is being fixed in #6707 😅 thanks for the contribution anyway though!

PR #6750, opened 2018-08-16
Remove shadowed generics in Proxy$traps

talbenari1:
In v0.67.0, `Proxy$traps.getOwnPropertyDescriptor` and
`Proxy$traps.defineProperty` were erroneously modified with an
additional generic, which shadowed the surrounding generic. By removing
it, these two methods will now correctly refer to the proxy target's
type, instead of an independently inferred type.

MDN docs for `getOwnPropertyDescriptor`:
<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/getOwnPropertyDescriptor>

MDN docs for `defineProperty`:
<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/defineProperty>


nmote:
This change definitely looks right, but could you include some docs (e.g. MDN) to support this change?

talbenari1:
Sure thing, I originally put this together hastily since I was working on something else at the time. Before I get around to updating this, here's an example of the behavior this PR fixes (which I'll add a test for when updating):

https://flow.org/try/#0C4TwDgpgBAYg9nKBeKBvAUFKAPAXFAOwFcBbAIwgCd0BfAbnXQGM4CBnYKAMwX3kRSoc+AIxR6jFu0484ABUpxsIZIQgB3KAqUgAFLIA0aTFAAmELgEsCEbZEqhdwAIaUA5hGB8EASmNYsSk8iSgJuZwAbNggGLBpaHwYgA

nmote:
Yeah, that makes sense, I'd just like to be able to verify the change against some official docs. Let me know when you get a chance to provide them.

talbenari1:
Just updated with some MDN docs for the fixed methods. 

As an aside, do you guys have a place for regression tests (or whatever you'd think is a good place for the example I provided above)? I wasn't able to find any `Proxy`-related tests in the repo, so there weren't any existing tests for me to fix. If that's not something you guys generally test, that's fine too.

nmote:
Looks good, thanks! Right now, we don't consistently test the builtin libdefs, but tests are definitely welcome. If you'd like to add tests for definitions (like this one) defined in `core.js`, look at the `tests/core_tests` folder.

I'd like to take this as-is to avoid further delays, but feel free to CC me if you put up tests for this in another PR, and I'll make sure it gets prompt attention.

Issue #3567, opened 2017-03-25
Parsing anonymous class expressions with type params


zertosh:
Anonymous function expressions can have type params, but anonymous class expressions can't (named are fine):

```
(function F<T>() {});
(function <T>() {});
(class C<X> {});
(class <X> {});
```

```
4: (class <X> {});
          ^ Unexpected token <
4: (class <X> {});
            ^ Unexpected token >
4: (class <X> {});                 ^ Unexpected token ;
5:    ^ Unexpected end of input
```

Babylon handles this just fine.

PR #6765, opened 2018-08-18
Add support for type parameters to anonymous class expressions

nicolo-ribaudo:
This is done by unifying the parsing logic for class expressions and
class declarations, since anonymous class declarations already suppoted
type arguments.

Fixes https://github.com/facebook/flow/issues/3567

---

Hi, this is my first PR :slightly_smiling_face:
I was hit by this bug while working on a personal project and I tought that it was a great opportunity to learn something new.

I have some problems running the tests locally: if I run `make && make js` and then `make test` in the `src/parser` folder, for some reason it doesn't consider my changes to the code. I double-checked by deleting whole sections of the parser files both nothing changed :confused:.
I tested my changes using `echo "(class <T> {});" | bin/flow ast | prettier`, so I'm not sure yet that this PR didn't break anything. I'll wait for the CI result.


mrkev:
Oh, this is cool @dsainati1

PR #6791, opened 2018-08-23
Add additional fs constants

Gudahtt:
Update list of fs constants to include all defined in Node.js v10.9.0


Gudahtt:
It looks like I need to update the expected output for the `node_tests`. I'll update this.

PR #6801, opened 2018-08-27
Fix nullable methods on HTMLCollection

muan:
[`namedItem()`](https://dom.spec.whatwg.org/#dom-htmlcollection-nameditem-key) and [`item()`](https://dom.spec.whatwg.org/#dom-htmlcollection-item) can be `null`, and `[]` can be `undefined`.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -536,9 +536,9 @@ declare class Attr extends Node {
 declare class HTMLCollection<Elem: HTMLElement> {
   @@iterator(): Iterator<Elem>;
   length: number;
-  item(nameOrIndex?: any, optionalIndex?: any): Elem;
-  namedItem(name: string): Elem;
-  [index: number]: Elem;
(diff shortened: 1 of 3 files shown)


mrkev:
[x] clear description
[x] pointers to spec
[x] tests (?!) 

10/10

mrkev:
We're discussing this one for a similar reason to #6823. `HTMLCollection` tends to be used with similar semantics to `Array` and in that case we consciously allow `[]` to not be `undefined`, though it also has Object-like semantics since you can use a string id to access it (I'm trying to having merged as-is but I'll let you know if there's any change necessary).

mrkev:
Hmmm, looks like the conclusion is to not change the return type of the indexer for now, so I'll be changing it to `[index: number | string]: Elem;`. This is to keep it in line with the use of `Arrays` as described above and on #6823.

PR #6806, opened 2018-08-28
Allow use of string and array parameters in toLocaleLowerCase, toLocaleUpperCase

tufantunc:
String.toLocaleLowerCase and String.toLocaleUpperCase methods allow parameters to specify locale to be used to convert. 

Here is the details for toLocaleLowerCase: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleLowerCase
And toLocaleUpperCase: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -311,8 +311,8 @@ declare class String {
     startsWith(searchString: string, position?: number): boolean;
     substr(from: number, length?: number): string;
     substring(start: number, end?: number): string;
-    toLocaleLowerCase(): string;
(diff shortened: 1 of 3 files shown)


tufantunc:
I didn't figure out why its giving error on build. Can anyone help?

mrkev:
It's probably fine. Importing and checking it out 👍

PR #6823, opened 2018-08-30
Fix possibly undefined return values on Array

muan:
- ~~[`[]`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Accessing_array_elements)~~
- [`.pop()`](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.pop)
- [`.shift()`](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.shift)

### DIFF
--- a/tests/arraylib/array_lib.js
+++ b/tests/arraylib/array_lib.js
@@ -62,3 +62,13 @@ function of_test() {
   var upcastOkay: Array<string | number> = Array.of("hello", "world");
   var incompatibleTypeNotOkay: Array<string> = Array.of(1, 2);
 }
+
+function undefined_test() {
+  var emptyArray: Array<number> = []
+
+  var pop: number = emptyArray.pop() // error, undefined ~> number
+  if (pop) pop.toFixed() // ok
+
+  var shift: number = emptyArray.shift() // error, undefined ~> number
+  if (shift) shift.toFixed() // ok
+}
--- a/lib/core.js
+++ b/lib/core.js
@@ -235,7 +235,7 @@ declare class $ReadOnlyArray<+T> {
     slice(start?: number, end?: number): Array<T>;
     some(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => any, thisArg?: any): boolean;
     values(): Iterator<T>;
-    +[key: number]: T | void;
+    +[key: number]: T;
     +length: number;
 }
 
@@ -275,7 +275,7 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     unshift(...items: Array<T>): number;
 
 
-    [key: number]: T | void;
+    [key: number]: T;
     length: number;
     static (...values:Array<any>): Array<any>;
     static isArray(obj: any): bool;
--- a/lib/core.js
+++ b/lib/core.js
@@ -235,7 +235,7 @@ declare class $ReadOnlyArray<+T> {
     slice(start?: number, end?: number): Array<T>;
     some(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => any, thisArg?: any): boolean;
     values(): Iterator<T>;
-    +[key: number]: T;
+    +[key: number]: T | void;
     +length: number;
 }
 
@@ -249,7 +249,7 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     findIndex(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): number;
     forEach(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): void;
     map<U>(callbackfn: (value: T, index: number, array: Array<T>) => U, thisArg?: any): Array<U>;
-    pop(): T;
+    pop(): T | void;
     push(...items: Array<T>): number;
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: Array<T>) => T,
@@ -268,14 +268,14 @@ declare class Array<T> extends $ReadOnlyArray<T> {
       initialValue: U
     ): U;
     reverse(): Array<T>;
-    shift(): T;
+    shift(): T | void;
     some(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): boolean;
     sort(compareFn?: (a: T, b: T) => number): Array<T>;
     splice(start: number, deleteCount?: number, ...items: Array<T>): Array<T>;
     unshift(...items: Array<T>): number;
 
 
-    [key: number]: T;
+    [key: number]: T | void;
     length: number;
     static (...values:Array<any>): Array<any>;
     static isArray(obj: any): bool;
--- a/tests/predicates-inferred/sanity-ordering.js
+++ b/tests/predicates-inferred/sanity-ordering.js
@@ -12,6 +12,7 @@ function dotAccess(head, create) {
   const stack = path.split('.');
   do {
     const key = stack.shift();
+    if (!key) continue
     head = head[key] || create && (head[key] = {});
   } while (stack.length && head);
   return head;


tests/arraylib/array_lib.js
```diff
@@ -62,3 +62,16 @@ function of_test() {
   var upcastOkay: Array<string | number> = Array.of("hello", "world");
   var incompatibleTypeNotOkay: Array<string> = Array.of(1, 2);
 }
+
+function undefined_test() {
+  var emptyArray: Array<number> = []
+
+  var indexed: number = emptyArray[0] // error, undefined ~> number
```
muan: This line is actually passing, and I don't get why... 🤔 Any idea?

mrkev: Hmmm not sure either

mrkev:
Oh boy, this opens a can of worms. Technically any array access can return undefined, but the decision to skip the void check is a conscious one afaik (albeit, still contested). Updating the array indexer's type to reflect undefined would trigger a ton of errors in pretty much every codebase-- it's probably more trouble than it's worth since a lot of array accesses are done in loops though the length of the array which prevents this kind of error.

Is it correct? Technically no, but it makes adding flow to a codebase easier.

This would be a good candidate for a `flow strict` lint, and I personally hold the opinion that while it might be a hassle to jump into it on your pre-existing codebase, it would be useful to be able to enable this on a brand new project where you don't have a ton of old code to migrate.

I'll import this to try to spark some conversation about a possible `flow strict` lint and to see how many errors it surfaces in facebook's codebase for fun and curiosity (that is, if it finishes type-checking at all), but I don't expect this one to be merged. Thanks for the PR though, this one also looks great (:

muan:
Thanks @mrkev! I only discovered this when I noticed an exception in production where we're just assuming `arr[0]` exists, and was bummed that flow didn't catch it for me. 😬

> it's probably more trouble than it's worth since a lot of array accesses are done in loops though the length of the array which prevents this kind of error.

Ah true. This did not occurred to me.

> This would be a good candidate for a `flow strict` lint, and I personally hold the opinion that while it might be a hassle to jump into it on your pre-existing codebase, it would be useful to be able to enable this on a brand new project where you don't have a ton of old code to migrate.

`flow-strict` makes sense 👌.

tests/arraylib/array_lib.js
wchargin: This is passing because indexer property access is also unsafe. As
@mrkev noted, this is by design, for the same reasons as with arrays.
See docs:

  - <https://flow.org/en/docs/types/objects/#toc-objects-as-maps>
  - <https://flow.org/en/docs/types/arrays/#array-access-is-unsafe->

muan:
@mrkev Can this PR land without the index accessor change? or is the possible `flow-strict` lint more preferable?

vicapow:
Would love to see this myself, but also wonder if it will be contentious / noisy for existing, larger codebases. It might be less straightforward, but I like the idea of adding this and other additional checks as part of configurable lint options.

nmote:
We just had a discussion on Discord about this. I'd like to try to gauge the rollout burden of this change, without the indexer part. If you are still interested in getting this in, could you update to remove that? Don't worry about re-recording tests -- feel free to just drop the changes to `.exp` files when you rebase, if you like.

muan:
> could you update to remove that? 

Done,

> feel free to just drop the changes to `.exp` files when you rebase

and Done!

Thanks!

mrkev:
hello, I'm alive 👋

Tested this and with just `pop` and `shift` it still causes a whole lot of of errors internally. This makes rollout very complicated— it'd be cool if we had a better way of suppressing errors. Let me explain:

We could rollout this change and deal with the errors by suppressing them with`$FlowFixMe`s like so:

```
const arr = [0,1,2];
//$FlowFixMe to fix arr.pop()
const result = x + arr.pop();
```

The way `$FlowFixMe`s work though, they silence ALL errors coming from the next line. This is dangerous, since there could be a future change that triggers other bugs but flow wouldn't catch them. For example: 

```
const arr = null;
//$FlowFixMe, to fix arr.pop()
const result = x + arr.pop(); // No error, but throws!
```

This change would be much easier to roll-out if we had a way to limit `$FlowFixMe`s to specific errors. Say, if each error had a "code" and we could disable just that specific code (kind of how eslint does it):

```
const arr = null;
//$FlowFixMe(error: expected-number-got-void)
const result = x + arr.pop(); // Still throws a "calling-method-on-null" error
```

This way we could roll it out without spending months and months just fixing pre-existing errors or running the risk of not catching thousands of new ones 😬.

mrkev:
Another alternative would be having a way of describing `normal` libdefs and `strict` libdefs. Or enable this in new projects but not old ones, like discussed before. I'd say this is still blocked until one of those options becomes available.

jedwards1211:
@mrkev being able to have `normal` and `script` libdefs is a great idea, there's very similar contention about the `flow-typed` defs for certain `lodash` functions 👍

AndrewSouthpaw:
One other point that's kind of interesting to me: even statically typed languages e.g. Java don't return maybe types for `.pop` etc.

Putting maybe return types on something as fundamental as array/object access into static analysis of types always seemed overkill to me.

wchargin:
@AndrewSouthpaw: In Java, `empty.pop()` and `empty.get(0)` throw
exceptions. In JavaScript, they just return `undefined`, which is a
significant difference. Java’s type system also includes `null` as a
value of every reference type (which Flow’s does not), so the fact that
`emptyMap.get(1)` returns `null` of type `V` is consistent.

C# fits into the same category as Java. Rust does return `Option` types
from `Vec.get` and `BTreeMap.get`. Haskell and OCaml are of course
likewise sound.

> Putting maybe return types on something as fundamental as array/object
> access into static analysis of types always seemed overkill to me.

If you can’t trust the fundamentals, how can you trust anything built on
top of them?

AndrewSouthpaw:
Thanks @wchargin, I stand corrected, you're quite right and I hadn't considered that fact. :-)

> If you can’t trust the fundamentals, how can you trust anything built on top of them?

For me? With tests.

Looking at the original example, if I had to do something like

```js
const arr = [0,1,2];
const val = arr.pop();
if (typeof val === 'undefined') throw new Error('not there')
const result = x + val;
```

every single time I'd rather stop using Flow (or, more likely, just stop upgrading past when this was released). It would, IMO, make my code way too verbose and having to check things that are self-evident and not be worth the many advantages Flow offers.

I don't see a point in trying to make a 100% Correct type system for a dynamic language that wasn't written to be that. I like JS for what it is, try to embrace its oddities like returning `undefined` for `[].pop()`, and appreciate that Flow has fit in with the Not 100% Right spirit of JS while adding a ton of value. I have always seen Flow as fitting in with the language to help people make sense of the myriad types being passed around without trying to "correct" JS for language design choices/mistakes. I think that's impressive and beautiful.

But that's just me. I recognize it's kind of hard to defend a position that essentially says, "I don't want to make changes that make a system More Correct." 😆 I get that I'm only one user of the many, many people who use Flow and perhaps this is what people want.

jedwards1211:
@wchargin keep in mind that in Java those are `RuntimeExceptions` that you aren't forced to catch.  That's very analagous to Flow not forcing you to do null checks in this case.

wchargin:
@jedwards1211: The analogy is a bit tenuous, I’m afraida not least
because surely the JavaScript analogue of Java’s runtime exceptions
would be… JavaScript exceptions, for which Java and Flow have the same
typing semantics.

The fact that a method is declared to return `T` means that _if_ a call
to the method evaluates to a value, then that value has type `T`. An
invocation that throws an exception does not evaluate to any value, and
so the contract stipulates nothing about this case. The Java methods
`pop`, `get`, etc. follow this contract: sometimes they evaluate to
values of the correct type, and sometimes they throw. The Flow-typed
methods `pop`, `shift`, etc. do not follow this contract: sometimes they
evaluate to `T`, but sometimes to `undefined`, which (again, in Flow’s
semantics) is _not_ of type `T`. This is unsound—by design, but still
unsound.

In short, exceptions are neither here nor there.

FireyFly:
@AndrewSouthpaw I think that concern is why @mrkev suggested `normal` and `strict` libdefs might be a way forward, so that you can opt in to proper soundness, but default to weaker, more "conventionally JavaScript" libdefs. So, I don't think you need to worry too much about this suddenly changing in Flow from one version to the next...

(That said, I'm probably in "camp soundness" here, in that I wouldn't mind having to do extra checks as appropriate when using methods that may return `undefined`... although, it would be nice if asserting `arr.length > 0` was enough for pop/unshift to return non-`undefined`, but that's probably not viable to expect here)

AndrewSouthpaw:
@FireyFly Ah, yeah! I didn't mention that I liked the idea of `normal` vs `strict` libdefs for Flow built-ins as well as flow-typed. That would allow people to opt into the level of type safety that they want.

I also have no concept of how difficult that would be for the flow implementation side of things. :-) On flow-typed, it'll be similarly tricky I could imagine, but also doable. I'm imaging something like reading off a flow config for strictness, and then where possible pulling down the `[lib]_strict` version instead of `[lib]` from flow-typed repo.

jedwards1211:
@wchargin okay, never mind about the analogy - 
what I mean is neither Java nor Flow have *compile-time* errors about potential index out of bounds errors.  Having compile-time Flow errors would be a difference from Java in that regard. 
Imagine writing a sort algorithm in Java, and getting compiler errors because you didn't check the index bounds on each array access (checks you wouldn't want in a properly-written, high-performance algorithm).

nmote:
Unfortunately we never reached consensus either here or in discussions internally. I don't object to this change, but I also don't feel strongly enough about it to push past the objections that have been raised. For now, let's leave `pop` and `shift` consistent with ordinary array accesses.

PR #6829, opened 2018-08-31
Remove void from React$Node type declaration to prevent React.render(undefined)

### LINKED ISSUE #6818: React render() cannot return undefined: React$Node shouldn't be undefined
Hello! 

Small problem we stumbled upon while developing: In React, the `render()` method of a `React.Component` does not accept undefined. 

However, in Flow, the `render()` method returns a `React$Node`, which can be `void` (=`undefined` ? I'm not sure of that).

#### Minimal Working Example:

```jsx
const UndefinedComponent = () => undefined;
```
When you render it:

image

But as you can see on [this Try Flow](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBcwAlAUwEMBjYqfOLMAcn0pqYG50qYKBnPmADCDPADsyYwuTEATMvgCqcslAwTZYMgA9Ck2YPLVCAEhG44EqQB4QAPjABvMKxX4wAXjAAKAJSeHAFcVNQ0wAF8gA), Flow doesn't pick it up.

The simplest solution is to remove `  | void ` from the `React$Node` type declaration. (`lib/react.js`, l.13)
May I open a PR to fix that, or is there another way to do it to prevent side effects?

Thanks :smiley: 

---

Environment:

| lib | version | 
| ------------------------------ | ------------------- | 
| Flow                    | 0.79.1         | 
| React                    | 16.3.2         |

Ojisama:
Fix #6818 

> Small problem we stumbled upon while developing: In React, the render() method of a React.Component does not accept undefined.
> However, in Flow, the render() method returns a React$Node, which can be void

```jsx
// Example 1

const Foo = (props: {}) => {} 
<Foo /> // Error from React: Nothing was returned from render.

// Example 2

class Bar extends React.Component<{ foo: string }> {
  render = () => this.props.foo;
}
<Bar /> // Error from React: Nothing was returned from render.
```
---

I split `React$Node`, into:
```jsx
declare type React$RenderableNode =
  | null
  | boolean
  | number
  | string
  | React$Element<any>
  | React$Portal
  | Iterable<React$Node>;


declare type React$Node =
  | void
  | React$RenderableNode
```

And I replace `React$Node` by `React$RendableNode` in the return type of `render()` and `React$StatelessFunctionalComponent`

**Why not simply removing `void` from `React$Node` type declaration?** Because
```jsx
// Example 1

const Parent = (props) => <div>{props.children}</div>

<Parent/> // No error from React
<Parent>{undefined}</Parent> // No error from React
<Parent>{[undefined, undefined]}</Parent> // No error from React

// Example 2

const IterableComponent = (props: {}) => ['foo', undefined, 'bar'];
<IterableComponent /> // No error from React

```

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -9,8 +9,8 @@
  * A UI node that can be rendered by React. React can render most primitives in
  * addition to elements and arrays of nodes.
  */
-declare type React$Node =
-  | void
+
+declare type React$RenderableNode =
   | null
   | boolean
   | number
@@ -18,6 +18,16 @@ declare type React$Node =
   | React$Element<any>
   | React$Portal
   | Iterable<React$Node>;
+
+/**
+ * If a node is not rendered, it can be undefined
+ * e.g: an element from the children array may be undefined
+ * React just won't render it.
+ */
+
+declare type React$Node =
+  | void
+  | React$RenderableNode
 
 /**
  * Base class of ES6 React classes, modeled as a polymorphic class whose main
@@ -41,7 +51,7 @@ declare class React$Component<Props, State = void> {
   // lifecycle methods
 
   constructor(props?: Props, context?: any): void;
-  render(): React$Node;
+  render(): React$RenderableNode;
   componentWillMount(): mixed;
   UNSAFE_componentWillMount(): mixed;
   componentDidMount(): mixed;
@@ -135,7 +145,7 @@ declare class LegacyReactComponent<Props, State>
  * can type check.
  */
 declare type React$StatelessFunctionalComponent<Props> = {
-  (props: Props, context: any): React$Node,
+  (props: Props, context: any): React$RenderableNode,
   displayName?: ?string,
   propTypes?: $Subtype<{[_: $Keys<Props>]: any}>,
   contextTypes?: any
--- a/lib/react.js
+++ b/lib/react.js
@@ -10,24 +10,14 @@
  * addition to elements and arrays of nodes.
  */
 
-declare type React$RenderableNode =
+declare type React$Node =
   | null
   | boolean
   | number
   | string
   | React$Element<any>
   | React$Portal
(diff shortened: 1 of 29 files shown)


Ojisama:
I don't know why CI fails upon spawning flow :confused:

mrkev:
Not sure about the CI, but I'll have a look. Thanks for the PR!

For reference to anyone seeing this PR, React complains on a `void` return from `render()`.

> Invariant Violation
Photo(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.

pakoito:
Hi @Ojisama. Thanks for the PR! I'd like to keep the `render` methods in files like `tests/react_modules/createclass-callsite.js` still annotated with a type. Would it be possible to address it in the PR before we proceed to merging?

Ojisama:
@pakoito Sure I'll do that! But in this case, shouldn't we also expose the `RenderableType` introduced in this PR (or something equivalent) so people can type their own `render` methods?

mrkev:
@Ojisama what if we change every instance of `React$Node` for `?React$Node` except for `render()` (which we keep as `render(): React$Node;`, and then remove `void` from the definition of `React$Node`.

`React$Node` already includes null so everything should be equivalent (`?T` === `T | null | void`) except for `render()`. Thoughts?

pakoito:
If @mrkev proposal works and we keep the methods annotated, I'm happy :)

Ojisama:
@mrkev @pakoito sounds good, I'll try that and I'll update my PR today :)

Ojisama:
@pakoito `render()` types are back in
@mrkev your solution works like a charm, and is neat :+1:

pakoito:
Thank you both!!

mrkev:
@pakoito hope to see you around in a couple more (:

PR #6853, opened 2018-09-06
Fix type declaration for DedicatedWorkerGlobalScope.onmessage

omninonsense:
It's just a two-character fix. :slightly_smiling_face:

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -496,7 +496,7 @@ declare class SharedWorker extends EventTarget {
 declare function importScripts(...urls: Array<string>): void;
 
 declare class WorkerGlobalScope extends EventTarget {
-    self: WorkerGlobalScope;
+    self: this;
     location: WorkerLocation;
     navigator: WorkerNavigator;
     close(): void;
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -510,7 +510,7 @@ declare class WorkerGlobalScope extends EventTarget {
 }
 
 declare class DedicatedWorkerGlobalScope extends WorkerGlobalScope {
-    onmessage(): (ev: MessageEvent) => any;
+    onmessage: (ev: MessageEvent) => any;
     postMessage(message: any, transfer?: Iterable<Object>): void;
 }
 


omninonsense:
To be fairm the definitions for a lot of the worker-related stuff is wrong. If I have permission, I don't mind correcting them myself.

mrkev:
Hey @omninonsense; this seems to be a field, not a function. Your change suggests it's a function you run to get back another function, but what you want to do is acutally assign a function to it (: 

See https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope#Event_handlers and https://developer.mozilla.org/en-US/docs/Web/API/DedicatedWorkerGlobalScope/onmessage

omninonsense:
Hi @mrkev!

> Your change suggests it's a function you run to get back another function

No, my change doesn't do that. I think you misunderstood the diff. What you are describing is the current type declaration, which is what I am trying to fix.

Can you please double-check to make sure? :slightly_frowning_face: 

> this seems to be a field, not a function... what you want to do is acutally assign a function to it (:

Yes exactly, and this is what my commit does. :slightly_smiling_face:

mrkev:
Oh shoot, how did that happen. I looked at the before and was the after hahah

I'll try to make a mental note to import this on Monday. Yeah, this is a very welcome fix, sorry!

omninonsense:
No problem! It happens sometimes.

I might have ago at fixing some mistyped stuff. I have a question regarding that, though: Should it be a series of smaller PRs, or a larger one?

jdetle:
Hey, I think you should also have to add `self: DedicatedWorkerGlobalScope` here.

omninonsense:
Hi @jdetle 

> Hey, I think you should also have to add `self: DedicatedWorkerGlobalScope` here.

Could you point out where to put it, and why it's required? I don't know what "here" means in this context.

lib/bom.js
```diff
@@ -510,7 +510,7 @@ declare class WorkerGlobalScope extends EventTarget {
 }
 
 declare class DedicatedWorkerGlobalScope extends WorkerGlobalScope {
```
jdetle: you need to override `self` here to ensure that `self` in the `DedicatedWorkerGlobalScope` refers to `DedicatedWorkerGlobalScope` with `self: DedicatedWorkerGlobalScope`

jdetle: Otherwise if someone writes self.postMessage() in a WebWorker, it is still going to resolve to the definition for the WorkerGlobalScope.

mrkev:
@omninonsense more, smaller PRs are preferred. Makes reviewing them easier,

mrkev:
Sorry it took a while @omninonsense, I've been out of the loop a couple of months!

omninonsense:
No worries @mrkev! :grin: Glad you're back!

PR #6867, opened 2018-09-07
Add controlsList property on HTMLMediaElement

### LINKED ISSUE #6859: Add property 'controlsList' on `HTMLMediaElement`
According to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList), `controlsList` is a property on `HTMLMediaElement`. I assume this was not included in flow because it is experimental, but maybe it should be included as `controlsList?: DOMTokenList`

gregeinfrank:
resolves #6859 

According to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/controlsList) `controlsList` is an experimental feature, but does have full support on Chrome and Opera. I'm not sure what flow's official position is with experimental features, but it seems like having `controlsList?` as an optional property makes sense.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -3049,6 +3049,7 @@ declare class HTMLMediaElement extends HTMLElement {
   volume: number;
   muted: boolean;
   defaultMuted: boolean;
+  controlsList?: DOMTokenList;
 
   // tracks
   audioTracks: AudioTrackList;


gregeinfrank:
@facebook-github-bot I signed!

mrkev:
I'll import and check what people thing 👍

gregeinfrank:
@mrkev Any update here? Thanks!

Issue #5475, opened 2017-12-05
getBoundingClient should return a DOMRect not a ClientRect


gregtatum:
The spec defines `getBoundingClient` as returning a `DOMRect`.
https://drafts.csswg.org/cssom-view/#dom-element-getboundingclientrect

MDN also uses the `DOMRect` in its examples:
https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect

Flow defines it as a `ClientRect`:
https://github.com/facebook/flow/blob/166b0c202b778539832f61fc7f98a4a7e94dfa9d/lib/dom.js#L995

And has the Class declaration as `ClientRect`:
https://github.com/facebook/flow/blob/166b0c202b778539832f61fc7f98a4a7e94dfa9d/lib/dom.js#L3045-L3052

You can create a `DOMRect` directly using:
```
new DOMRect()
DOMRect { x: 0, y: 0, width: 0, height: 0, top: 0, right: 0, bottom: 0, left: 0 }
```

But it is not compatible with the ClientRect.

gregtatum:
flow.org/try does not appear to have the DOM definitions in it, but I believe this would error:

```js
function acceptClientRect(rect: ClientRect) {}

// Fine:
acceptClientRect(element.getBoundingClientRect());

// Bad:
acceptClientRect(new DOMRect());
```

julienw:
`[1] IE and Edge support the non-standard MSDN: ClientRect which does not define x and y properties`

Maybe that's why there's this difference, with Flow taking the safe approach ?

julienw:
> But it is not compatible with the ClientRect.

It actually is :) a DOMRect _is_ a ClientRect too (it has all the right properties)

TrySound:
This probably can be closed.
/cc @vkurchatkin

julienw:
So, nothing changed here as far as I know. GetBoundingClientRect still returns ClientRect. Is the rationale from my previous comment right? Maybe it should return `ClientRect | DOMRect` ?

TrySound:
I would just define it as interface to prevent instantiating. It still does not give any functionality.

airqb:
use "top" instead of "y" and "left" instead of "x" I think the "top" has more Availablity in most of the browser, chrome support "y" and "top" but others maybe are not. this why the "top" is most supported cross-browser  than "y"

julienw:
[...]

PR #6885, opened 2018-09-13
getBoundingClient should return a DOMRect not a ClientRect

ryota-murakami:
## Introduction
@gregtatum is already noticed this issue at #5475, and his explain is still correct today.

but above issue was closed without fix PR.
So I created PR along to #5475 reports.

Thanks any review😀

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -3568,38 +3568,6 @@ declare class ClientRectList { // extension
   [index: number]: ClientRect;
 }
 
-declare class DOMRectReadOnly {
-  static fromRect(rectangle?: {x: number, y: number, width: number, height: number}): DOMRectReadOnly;
-  constructor(x: number, y: number, width: number, height: number): DOMRectReadOnly;
-  +bottom: number;
-  +height: number;
-  +left: number;
-  +right: number;
-  +top: number;
-  +width: number;
-  +x: number;
-  +y: number;
-}
-
-declare class DOMRect extends DOMRectReadOnly {
-  static fromRect(rectangle?: {x: number, y: number, width: number, height: number}): DOMRect;
-  bottom: number;
-  height: number;
-  left: number;
-  right: number;
-  top: number;
-  width: number;
-  x: number;
-  y: number;
-}
-
-declare class DOMRectList {
-  @@iterator(): Iterator<DOMRect>;
-  length: number;
-  item(index: number): DOMRect;
-  [index: number]: DOMRect;
-}
-
 // TODO: HTML*Element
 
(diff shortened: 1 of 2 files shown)


lib/dom.js
```diff
@@ -3568,7 +3568,35 @@ declare class ClientRectList { // extension
   [index: number]: ClientRect;
 }
 
-// TODO: HTML*Element
```
motiz88: I'm not a maintainer here, but I think this `// TODO` comment should probably be kept - since this PR doesn't actually address it.

ryota-murakami: @motiz88 Thanks your review.
Yeah, I agree with you and I've  pushed revert commit.

lib/dom.js
```diff
@@ -3568,6 +3568,36 @@ declare class ClientRectList { // extension
...
+declare class DOMRectReadOnly {
+  +bottom: number;
+  +height: number;
+  +left: number;
+  +right: number;
+  +top: number;
+  +width: number;
+  +x: number;
+  +y: number;
+  fromRect(rectangle?: {x: number, y: number, width: number, height: number}): DOMRect;
```
mroch: couple of things:

1) should be `static`
2) if i'm reading [the draft](https://drafts.fxtf.org/geometry/#domrectreadonly) correctly, `DOMRectReadOnly.fromRect()` should return a `DOMRectReadOnly`, and `DOMRect.fromRect()` should return a `DOMRect`. so add one to DOMRect as well

lib/dom.js
```diff
@@ -3568,6 +3568,36 @@ declare class ClientRectList { // extension
   [index: number]: ClientRect;
 }
 
+declare class DOMRectReadOnly {
```
mroch: missing the constructor: https://drafts.fxtf.org/geometry/#dom-domrectreadonly-domrectreadonly

lib/dom.js
```diff
@@ -3568,6 +3568,38 @@ declare class ClientRectList { // extension
...
```
fishythefish: Seems like these classes are already declared in [`bom.js`](https://github.com/facebook/flow/blob/master/lib/bom.js#L395)?

ryota-murakami: Oh I didn't realize it.
I see, surely that's not DOM things BOM is property place.

next my plan is 
1. fix Exists DOMRect definition based on @mroch review.
2. add DOMRect(already declared in bom.js) to getBoundingClient() return type.

ryota-murakami: @fishythefish I moved these classes to bom.js!

PR #6887, opened 2018-09-13
Document $Subtype and $Supertype

### LINKED ISSUE #6886: Unsoundness when using $Subtype
The documentation only says "Work in progress", so I'm not sure exactly how `$Subtype` is supposed to work.

A colleague found it and thought it was a nice way to get around a type error. (not realizing it was unsound to do so)

Here's an [example](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AFVALygDeokssAXKAHYCuAtgEZYBOoAvgNyqY4CqhEmQrV6zNgBpQTAIatqAZxSsAljQDmnHogDGsGstAohpclVAAGbXoNHZ1AIKtWstAB4AJAGU6TdNjuuAB8wUIA2igAujz6hiZ01IJEsuGWUchxRjSwKAD6snnKappKKupaRHQAdHKsOlkmsjR5WAAeulgYKKoGYows7EQ5+YXFFdXQWJooABZcQA)

```
// @flow

type T = { foo: number };
type U = { foo: number, bar: string };

const t = { foo: 0 };
const a: Array<$Subtype<T>> = [t];
const u: U = a[0]

const not_a_string: string = u.bar;

const an_exception: number = not_a_string.length;
```

aij:
Short description taken from `type_annotation.ml` and formatted.

Unsoundness warnings per those in `objects.md` and `functions.md`

Close #6886

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -604,11 +604,20 @@ const person5: PersonDetails = {age: 28, name: 'a'};  // OK
 
 ## `$Supertype<T>` <a class="toc" id="toc-supertype" href="#toc-supertype"></a>
 
-Work in progress
+`$Supertype<T>` acts as `any` over supertypes of `T`.
+
+If you need to opt-out of the type checker, and don't want to go all
+the way to `any`, you can instead use `$Supertype`. **`$Supertype` is
+unsafe and should be avoided.**
 
 ## `$Subtype<T>` <a class="toc" id="toc-subtype" href="#toc-subtype"></a>
 
-Work in progress
+`$Subtype<T>` acts as `any` over subtypes of `T`.
+
+If you need to opt-out of the type checker, and don't want to go all
+the way to `any`, you can instead use `$Subtype`. **`$Subtype` is
+unsafe and should be avoided.**
+
 
 ## Existential Type (`*`) <a class="toc" id="toc-existential-type" href="#toc-existential-type"></a>
 


wchargin:
Your helpful example at #6886 demonstrates that it is absolutely the
case that `$Subtype`/`$Supertype` can be used in unsound ways. But I’m
trying to understand whether these type constructors are _always_
unsound, or whether it’s just easy to use them in cases that induce
unsoundness—especially in light of the rather aggressive warning
introduced by this PR (not that I object!). Perhaps you can help me
understand?

For instance, consider the following function:

```js
/**
 * Merge maps without mutating the arguments.
 *
 * Merges multiple maps, returning a new map which has every key from
 * the source maps, with their corresponding values. None of the inputs
 * are mutated. In the event that multiple maps have the same key, an
 * error will be thrown.
 */
export function merge<K, V>(
  maps: $ReadOnlyArray<Map<$Subtype<K>, $Subtype<V>>>
): Map<K, V> {
  // Implementation elided (available upon request). It has no casts.
}
```

The use of `$Subtype`, instead of fixed type parameters `InK` and `InV`,
is useful because it enables heterogenously typed maps with a common
upper bound:


```js
    it("allows upcasting the type parameters", () => {
      const numberMap: Map<number, number> = new Map().set(1, 2);
      const stringMap: Map<string, string> = new Map().set("one", "two");
      type NS = number | string;
      const _unused_polyMap: Map<NS, NS> = MapUtil.merge([
        numberMap,
        stringMap,
      ]);
    });
```

When I wrote the type for the above function, I certainly didn’t think
of it as “opting out of the type system”; rather, it’s just expressing
the natural constraint.

It seems to me that this function is sound, despite the use of
`$Subtype`, similar to how a function `(x: Array<any>) => x.length` is
sound. Do you agree? Is there a nice way to characterize under which
circumstances these types can be used safely?

zeorin:
Does this mean that `$Subtype<() => mixed>` is the same as `Function` and `$Subtype<{}>` is the same as `Object`?

dsainati1:
This pull request is redundant now, since these utilities have been deprecated.

Issue #4529, opened 2017-08-03
Wrong annotation on the WeakMap.


Mtnt:
The WeakMap`s key can be only an object typed, no way for primitives. But if you specify a string as a key type, it will not throw an error: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZhRxxgC8YA6ngIYDWAstTgDwDOGATgJYB2A5gBow1HlgB8AbnQw8GQgC5CxMmB54kVOoxwAKAJRSoAOlaydAcnNCAjACYDQA

PR #6908, opened 2018-09-19
Enforce that WeakMap keys are Object types

rtsao:
WeakMap keys must be of Object type (primitive types such as string and number are not allowed).

Resolves https://github.com/facebook/flow/issues/4529

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -545,7 +545,7 @@ declare class Map<K, V> {
     [key: $SymbolToStringTag | $SymbolSpecies]: Function;
 }
 
-declare class WeakMap<K, V> {
+declare class WeakMap<K: Object, V> {
     constructor(iterable: ?Iterable<[K, V]>): void;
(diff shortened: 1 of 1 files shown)


vicapow:
Hello,

Following up. Think we can get this merged?

Best,
Victor

vicapow:
Seems like this was imported into phabricator internally but not landed. Is there some internal discuss or issues there that could be communicated here?

mwiencek:
This seems to break using objects with null prototypes as keys https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzCjjjAF4wB5AIwCsBTDfAOgwCd6BDfegCmwCuMGAEoA3Oix5C1Tm3Jhs9JAHUuAawCynAA69x6WW2a56+XsTgAaMAHJbBoA (compare v0.83.0 to master)

PR #6935, opened 2018-09-25
Add destroy and _destroy methods to stream and stream in lib/node.js

jkomyno:
This is a copy of #5001 which fixes issue #4999.
cc @DullReferenceException @mrkev


nmote:
This actually has a syntax error, could you fix it, build Flow, and do a quick sanity check to make sure the libdef behaves as you expect?

PR #6948, opened 2018-09-28
feat: add flat/flatMap

keithamus:
Just taking a punt on types for `flat` and `flatMap`. These are part of a [TC39 Stage 4 proposal](https://github.com/tc39/proposal-flatMap) that has been [implemented in all browsers](https://github.com/tc39/proposal-flatMap/issues/71).

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -211,6 +211,9 @@ declare class $ReadOnlyArray<+T> {
     keys(): Iterator<number>;
     lastIndexOf(searchElement: mixed, fromIndex?: number): number;
     map<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => U, thisArg?: any): Array<U>;
+    flatMap<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => U|U[], thisArg? any): Array<U>;
+    flat(depth?: number): Array<T>;
+
 
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => T,
@@ -245,6 +248,8 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     findIndex(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): number;
     forEach(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): void;
     map<U>(callbackfn: (value: T, index: number, array: Array<T>) => U, thisArg?: any): Array<U>;
+    flatMap<U>(callbackfn: (value: T, index: number, array: Array<T>) => U|U[], thisArg? any): Array<U>;
+    flat(depth?: number): Array<T>;
     pop(): T;
     push(...items: Array<T>): number;
     reduce(


chicoxyzzy:
It's not standardized yet and possible `flat` will be unshipped because of web compatibility issues
https://bugs.chromium.org/p/chromium/issues/detail?id=888128#c15

wchargin:
drive-by thought: can you do anything with

```js
type Join<T> = Array<$ElementType<$ElementType<T, number>, number>>;
```

?

lydell:
Would it be possible to add only `flatMap` for now, and `flat` later when it is been figured out how to do so? I regularly miss `flatMap` in my code, but not `flat`.

chicoxyzzy:
`flat` is landed to ES2019 so it's safe now to add it

nmote:
I'll merge this once my concerns described above are addressed. I suspect that the best we can do is `Array<mixed>` for `flat`. I have similar concerns around `flatMap`.

goodmind:
@nmote React.ChildrenArray does flattening

FezVrasta:
Could someone from the Flow team take a look at this PR? Not having support for standard JS features is pretty bad...

goodmind:
`flatMap` can be accepted, only `flat` is difficult one, but this PR needs tests

goodmind:
https://github.com/facebook/flow/pull/7854

goodmind:
https://flow.org/try/#0CYUwxgNghgTiAEA3W8Bm0AuAueAoe8AFADwCqAfIVDgCQBKIUwA8gHYQCeAgjDFB8XqMW7br36CGTNpx58BFcuQCU8ALzl4pANoBdVQDJ8RMpWrwh00XImWRs8QqWqNWvYeMkKVWlPtj5UxdNHX0AblxcMAB7VgBnDHgoAAZ1eG0MgEZdXQAadO0AJhz8jIBmHN0I3AgQRIAjNPQoDCpk5TCgA

It doesn't really work even as static function

Bannerets:
[Here](https://github.com/microsoft/TypeScript/pull/32131/files) is a accurate `Array#flat` implementation in TypeScript via recursion.

```typescript
type Dec = [-1, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20]
type Flat<A, D extends number> = {
    '1': A,
    '0': A extends ReadonlyArray<infer T>
        ? Flat<T, Dec[D]>
        : A
}[D extends -1 ? '1' : '0']
```

If someone knows how to rewrite it in Flow, feel free to propose ideas

nmote:
I've merged #7854 which added a declaration for `flatMap`, but my concerns around `flat` still stand. I'd consider a PR which added `flat` with a return type of `Array<mixed>`. That is not likely to be especially useful, but I don't think we can currently type it more precisely.

nnmrts:
Another great example of "uhm yeah this is a standard js feature but you just can't use it and you basically have to decide between using flow or not"

At least let me freaking disable certain parts of flowlib, or let me extend it locally, or whatever.

The way you guys handle these kind of "don't think we can type it currently more precisely" issues is mindbogglingly frustrating. Then don't type it and just ignore it. Or use `any` or `mixed` OR ANYTHING. Or idk, add a unsafe mode. Or a flowconfig option.

It's not like flow is unusable, no, flow actually makes 20% of javascript unusable.

Correct me if I'm wrong, but my only options in these cases would be:

1. downloading an extra library like lodash and using their untyped functions (suddenly flow has no problem with that (?) (except that coverage will go down))
2. refactoring all instances of a unusable standard feature like `flat` in with some other, less efficient and more verbose method like `flatMap` (or spread with `Object.assign`, or `Array.reduce` with `Object.entries`, etc.)
3. enabling `no_flowlib` in my flow config, copying the flowlib locally and changing or adding definitions after learning how that even works because writing library definitions isn't really easy and when I'm finished with that, flow probably got a new minor patch update that's useful that I can only use if I start this whole process all over again
4. sprinkle dozens of `$FlowFixMe` comments over my codebase 🆗🆒
5. enabling `suppress_type` and somehow adding it to my code 😮💯
6. just ignoring all these funny wavy red underlines in my code 🧠

nmote:
You've correctly identified your options: either work around this missing libdef or submit a PR that addresses the issues I've raised here. If you do so, feel free to cc me and I'll make sure to take a look at it quickly.

PR #6952, opened 2018-09-30
Update babel packages to new @babel/package 

Rohitkrops:
I have updated the old package names to the new @babel/__ ones. 

The reason: 

If you use the old packages with @babel/core v7+ then it gives the error "plugin should return a function not an object".  

This change is important as most of the packages are now available in this format. e.g. @babel/preset-react, @babel/preset-env, etc.

### DIFF
--- a/website/en/docs/_install/compiler-babel.md
+++ b/website/en/docs/_install/compiler-babel.md
@@ -2,11 +2,11 @@
 support for Flow. Babel will take your Flow code and strip out any type
 annotations.
 
-First install `babel-cli` and `babel-preset-flow` with either
+First install `@babel/cli` and `@babel/preset-flow` with either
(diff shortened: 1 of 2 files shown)


mroch:
whoops, it failed to import.

the babel.md parts of this landed in #7221, which also updated the `"presets": ["flow"]` line.
one of the two lines in the compiler-babel.md now conflicts, due to #7856 (sorry I saw that one before this one).
#7049 makes a similar fix to use `@babel/cli`, but also adds `@babel/core`. i don't know whether it's necessary.
#7933 makes a partial fix to `@babel/preset-flow` but not `@babel/cli`.

I'm going to close this one since it needs to be rebased, at least. I'm going to try to resolve all of these this week, though.

Issue #6329, opened 2018-05-20
Flow cannot resolve name Audio


EreHuc:
Hi,

It's very simple and I didn't find anything about it, I just want to use a sound for an JS Vanilla project using HTMLAudioElement via `new Audio(src: string)`, but flow can't resolve name Audio.

Here's a link using [Try Flow](https://flow.org/try/#0MYewdgzgLgBAtgTwMogK5gCYwLwzAUwHcYBBVDASxAAoByABwEMoALAfShDcfKrYDMKAG3y0AlAG4gA) with the current behavior

Here's MDN doc about [HTMLAudioElement](https://developer.mozilla.org/fr/docs/Web/API/HTMLAudioElement)

Thx

mrkev:
The library definitions are probably missing. Feel free to submit a PR for this, if that’s the case :+1:

hecontreraso:
Any workaround here? I'm experiencing the same problem

jamesisaac:
Added in v0.93

PR #6963, opened 2018-10-03
Add Audio declaration

vldvel:
The Audio declaration is missing Flow.
The issue for this commit: https://github.com/facebook/flow/issues/6329

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -667,10 +667,6 @@ type PositionOptions = {
     maximumAge: number;
 }
 
-declare class Audio extends HTMLAudioElement {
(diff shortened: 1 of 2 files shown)


nmote:
I think this belongs in `dom.js`, not `bom.js`. Could you move it or make a case that it belongs where it currently is?

vldvel:
> I think this belongs in `dom.js`, not `bom.js`. Could you move it or make a case that it belongs where it currently is?

I agree Audio belongs to dom. Fix it.

nmote:
Looks good, thanks

PR #6998, opened 2018-10-14
Add React.StrictMode

mxstbr:
Rebase of #6078


jbrown215:
Thanks for the PR! We've since added this as part of the change that added the react hooks API libdefs.

PR #7011, opened 2018-10-17
Use more specific type for navigator.serviceWorker

dhui:


dhui:
I'm not familiar with the flow build system or the OCaml ecosystem so any pointers to fixing the broken build would be appreciated

nmote:
Could you provide a link to docs which support this change?

dhui:
@nmote 
> Could you provide a link to docs which support this change?

Not sure what you mean by link to docs. Without this change, I was getting flow type errors.
The `ServiceWorkerContainer` type definition is here: https://github.com/facebook/flow/blob/v0.91.0/lib/serviceworkers.js#L109

nmote:
I'm looking for something like MDN docs which show that this change is in line with the specification.

dhui:
Ahh, here's the [MDN doc](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/serviceWorker)

nmote:
Looks good, thanks

PR #7083, opened 2018-10-23
add <PROJECT_ROOT> token support for node.resolve_dirname

tanhauhau:
add <PROJECT_ROOT> token support for `node.resolve_dirname` in `.flowconfig`

### DIFF
--- a/src/commands/commandUtils.ml
+++ b/src/commands/commandUtils.ml
@@ -468,6 +468,13 @@ let collect_flowconfig_flags main ignores_str untyped_str declarations_str inclu
   let raw_lint_severities = list_of_string_arg lints_str in
   main { ignores; includes; libs; raw_lint_severities; untyped; declarations; }
 
+let expand_project_root_token root str =
+  let root = Path.to_string root
+    |> Sys_utils.normalize_filename_dir_sep in
+  str
+    |> Str.split_delim Files.project_root_token
+    |> String.concat root
+
 let file_options =
   let default_lib_dir ~no_flowlib tmp_dir =
     let root = Path.make (Tmp.temp_dir tmp_dir "flowlib") in
@@ -481,11 +488,8 @@ let file_options =
   let ignores_of_arg root patterns extras =
     let patterns = Core_list.rev_append extras patterns in
     Core_list.map ~f:(fun s ->
-     let root = Path.to_string root
-       |> Sys_utils.normalize_filename_dir_sep in
      let reg = s
-       |> Str.split_delim Files.project_root_token
-       |> String.concat root
+       |> expand_project_root_token root
        |> Str.regexp in
       (s, reg)
     ) patterns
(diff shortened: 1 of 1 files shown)


nmote:
Overall this looks pretty reasonable. I'm not familiar with this option so I'll have to spend a bit of time reviewing to make sure I'm comfortable with this change.

If you are still interested in contributing, could you rebase and I'll take a closer look?

Also, it looks like the change to trim_lines is extraneous, could you either remove that change or split it out into a separate pull request?

Sorry for the delay here. If you decide to move forward, I'll personally make sure this gets reviewed.

tanhauhau:
hi @nmote i've rebased my changes. 😄

tanhauhau:
My main motivation is for my lerna based monorepo project setup:

packages:
  - pkgA
    - dependencies:
        - pkgB
        - ext_A
  - pkgB
    - peerDependecies:
        - ext_A
    - dependencies:
        - ext_B

dependencies within the project are symlinked, so you get:

- pkgA
  - node_modules/
      - pkgB _(symlink)_
         - node_modules/
            - ext_B
         - src/
            - code_B
      - ext_A
   - src/
      - code_A
- pkgB _(actual code)_
   - node_modules/
      - ext_B
   - src/
      - code_B

in code_B, I imported `ext_A`, (since it's a peer dependency). when I run flow in `pkgA`, it throws error because it couldn't resolve `ext_A`, because `pkgB` is resolved to `/pkgB` (the actual location) instead of `/pkgA/node_modules/pkgB`, and then `resolve_dirname` will recursively look upwards, and couldn't find `ext_A` in the `/node_modules/`.

So, I wish to add `<PROJECT_ROOT>` token, which is available to other options, to resolve from the absolute path of `ext_A/node_modules`.

nmote:
My point though is that Flow, today, appears to resolve these paths relative to the project root (that is, the location of the `.flowconfig`) It appears that setting the option to `ext_A/node_modules` today would be equivalent to setting the option to `<PROJECT_ROOT>/ext_A/node_modules` after your PR. Am I missing something?

tanhauhau:
for my case, my `.flowconfig` resides in `pkgA/`

nmote:
So are you thinking you would use `<PROJECT_ROOT>/../pkgB/node_modules`? Does simply writing `../pkgB/node_modules` not have that effect today? `<PROJECT_ROOT>` will resolve to the location of the `.flowconfig`, which in your case would be `pkgA/`.

tanhauhau:
It does not work previously when I made the PR, I need to check whether it
works with the latest flow version as of today.

nmote:
Okay, I'll try it out myself as well. Let me know what you find.

nmote:
I checked and Flow today does not support that usecase. Could you test your PR and confirm that it it allows you to use `<PROJECT_ROOT>` to traverse outside of the root directory?

nmote:
@tanhauhau, I'm closing for now since I haven't heard whether this PR addresses your problem. Feel free to reopen it if you would like to move forward

Issue #1413, opened 2016-02-16
arrow function. This type is incompatible with EventHandler...


Gozala:
Flow incorrectly fails on following code:

``` js
/* @flow */

const receive = (worker/*:Worker*/, handler/*:Function*/)/*:void*/ => {
  const onMessage = event => {
    event.target.removeEventListener("message", onMessage);
    handler(event.data);
  };

  worker.addEventListener("message",  onMessage);
}
```

```
src/listener.js:9
  9:   worker.addEventListener("message",  onMessage);
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `addEventListener`
  9:   worker.addEventListener("message",  onMessage);
                                           ^^^^^^^^^ arrow function. This type is incompatible with
 66:     addEventListener(type: string, listener: EventListener, useCapture?: boolean): void;
                                                  ^^^^^^^^^^^^^ union: object type | EventHandler. See: /private/tmp/flow/flowlib_3b5ce42a/dom.js:66
```

Gozala:
It turns out the reason flow actually fails is access to `event.data` that is expected on message event:
https://www.w3.org/TR/2011/WD-webmessaging-20110317/#dom-messageevent-data

But the error produces by flow is not only invalid but also completely misleading.

Gozala:
Presumably problem is the interface definitions for EventTarget / EventListener:
https://github.com/facebook/flow/blob/master/lib/dom.js#L61-L70

I guess it would make sense to change EventTarget to:

```js
declare class EventTarget <event:Event> {
    removeEventListener(type: string, listener: EventListener<event>, useCapture?: boolean): void;
    addEventListener(type: string, listener: EventListener<event>, useCapture?: boolean): void;
    detachEvent?: (type: string, listener: EventListener<event>) => void;
    attachEvent?: (type: string, listener: EventListener<event>) => void;
    dispatchEvent(evt: event): boolean;
}
```

And then also make each `EventTarget` also polymorphic, so `Worker`:
https://github.com/facebook/flow/blob/master/lib/bom.js#L236-L242
will also change to something like:

```js
type WorkerEvent <data>
  = MessageEvent<data>
  | ErrorEvent

declare class Worker <data> extends EventTarget <WorkerEvent>  {
    constructor(stringUrl: string): void;
    onerror: (ev: Event) => any;
    onmessage: (ev: Event) => any;
    postMessage(message: any, ports?: any): void;
[...]

PR #7097, opened 2018-10-25
Add type definitions for message events

wachino:
Add type definitions for message events


claramunt:
This should fix #1413

PR #7111, opened 2018-10-28
Add React.memo

jamesgpearce:
See https://reactjs.org/docs/react-api.html#reactmemo

This function wraps a stateless functional component to memoize the result based on the props passed in.

Adds a test (and of course lots of line number noise in the others)

Addresses #7093


jamesgpearce:
cc @gaearon ;)

lib/react.js
```diff
@@ -230,8 +230,11 @@ declare module react {
   ): React$ElementFactory<ElementType>;
   declare export function createRef<ElementType: React$ElementType>(
   ): {current: null | React$ElementRef<ElementType>};
-
   declare export function isValidElement(element: any): boolean;
+  declare export function memo<Props>(
+    type: React$StatelessFunctionalComponent<Props>,
```
sebmarkbage: Technically any react element type is allowed. Not sure functional components.

jbrown215:
There are a few more fundamental issues blocking memo that I'm working on. See https://github.com/facebook/flow/issues/7093 for details. A comprehensive solution is in the works! It will work for `memo`, `lazy`, and `forwardRef`

jamesgpearce:
You got it :)

PR #7118, opened 2018-10-30
Make Geolocation PositionOptions properties optional

atabel:
This should be allowed:
```js
navigator.geolocation.getCurrentPosition(onSuccess, onError, {timeout: 2000});
```


ivanloveit:
??

PR #7122, opened 2018-10-30
lib/node.js: stream: Allow Uint8Array in a few places.

cakoose:
Support for Uint8Array was added in Node 8.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1417,7 +1417,7 @@ declare class net$Socket extends stream$Duplex {
   destroy(exception?: Error): void;
   destroyed: boolean;
   end(
-    chunk?: string | Buffer,
+    chunkOrEncodingOrCallback?: Buffer | Uint8Array | string | (data: any) => void,
     encodingOrCallback?: string | (data: any) => void,
     callback?: (data: any) => void
   ): void;
@@ -1435,7 +1435,7 @@ declare class net$Socket extends stream$Duplex {
   setTimeout(timeout: number, callback?: Function): net$Socket;
   unref(): net$Socket;
   write(
-    chunk?: string | Buffer,
+    chunk: Buffer | Uint8Array | string,
     encodingOrCallback?: string | (data: any) => void,
     callback?: (data: any) => void
   ): boolean;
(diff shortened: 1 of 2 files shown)


fishythefish:
This change causes too many type errors in the libdefs. You can see these by running the tests, for example. You'll most likely have to update the libdefs for `Uint8Array` support everywhere, not just in a few places.

cakoose:
@fishythefish: I updated the method signatures for the `stream.Duplex` and `net.Socket` subclasses.

BTW, it seems strange that the subclasses re-declare all the base class's methods.  Maybe it would be better to omit them and just inherit?

cakoose:
@fishythefish: Can you take a look at the updated diff?

cakoose:
**Docs**

- `net$Socket`: [`.end`](https://nodejs.org/api/net.html#net_socket_end_data_encoding_callback), [`.write`](https://nodejs.org/api/net.html#net_socket_write_data_encoding_callback)
- `stream$Readable`: [`.unshift`](https://nodejs.org/api/stream.html#stream_readable_unshift_chunk), [`.push`](https://nodejs.org/api/stream.html#stream_readable_push_chunk_encoding)
- `stream$Writeable`: [`.end`](https://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback), [`.write`](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback)
- `stream$Duplex`: has `.end` and `.write` because it [extends `stream$Writable`](https://nodejs.org/api/stream.html#stream_class_stream_duplex)

**Callback function type**

The base class (`stream$Writeable`) uses `Function` and the derived classes (`stream$Duplex` and `net$Socket`) use `(data: any) => void`.  I changed them all to `Function` to be consistent.  Options:
1. Leave them as they were.  Maybe it was intentional that the derived classes had slightly different signatures?
2. Remove the signatures from the derived classes entirely.  And then perhaps change the base class to use `(data: any) => void`.

nmote:
Thanks, I'll take a look at the docs when I get a chance, and assuming they match, we're good there.

For now, could you do #1 for the callbacks? I think we should switch the base classes to use the safer callback (and even better would be to get rid of `any` here altogether), but I think those changes belong in another PR. This one is close to merging so I wouldn't want to delay it while code elsewhere is perfected.

cakoose:
Changed the callback type and rebased.

Issue #6749, opened 2018-08-16
[untyped] not in docs


chrisdrackett:
As far as I can tell you can still use `[untyped]` in config but it is not documented in the flow docs.

mrkev:
Oh, true. Anyone welcome to PR on this one.

PR #7129, opened 2018-10-31
Add `[untyped]` documentation

dominicfraser:
[This commit](https://github.com/facebook/flow/commit/1fdf33de5e46d9cf15fd5cdae9e7d53e27594d44) introduced the `[untyped]` config section. 

As in https://github.com/facebook/flow/issues/6749 this is present and is suggested to be used in several issues, but is not documented. 

This is a first pass at adding documentation for it. I only started looking at Flow yesterday, so hopefully it is OK! 🤞

### DIFF
--- a/website/en/docs/config/untyped.md
+++ b/website/en/docs/config/untyped.md
@@ -1,0 +1,58 @@
+---
+layout: guide
+---
+
+Flow needs to know which files to parse and of those how to handle any Flow types within them. For third party libraries you may wish to not parse, parse but not preserve types, or parse but not raise errors, depending on the quality and compatibility of their Flow implementation.
+
+Different sections are available to specify different behaviours, and by combining them most use cases are expected to be met.
+
+### `[untyped]` <a class="toc" id="toc-untyped" href="#toc-untyped"></a>
+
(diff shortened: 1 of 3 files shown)


website/en/docs/config/untyped.md
```diff
@@ -0,0 +1,58 @@
...
+.*\.untype\.js
+```
+
+This `[untyped]` section will parse:
+
+1.  Any file or directory under a directory named `third_party`
+2.  Any file or directory under `.*/src/foo` or under `.*/src/bar`
+3.  Any file that ends with the extension `.untype.js`
+
+Starting with Flow v0.23.0, you may use the `<PROJECT_ROOT>` placeholder in
```
jketcham: IMO this section from here to the end could probably be dropped, and might be better under a "using paths" section on the index page of the flow config doc. This section already is duplicated between the `[declarations]` and `[ignore]` doc pages.

jketcham:
Just ran into an issue where I needed to use `[untyped]` and noticed it wasn't documented, so this would be nice to have

nmote:
@dominicfraser I made a couple minor fixes and put this up for review internally. As soon as someone else on the team takes a look I should be able to merge. Thanks for your contribution, and sorry it has taken so long to get to!

Issue #5934, opened 2018-03-09
React.cloneElement fails because of props in union


danrot:
We have a React component which can take a few different components as children. Based on the parent component we want to set a property on all child components, therefore we use `React.cloneElement`. However, if we try use clone on this element it does not really seem to work. I've built something to show that:

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgWwA5wE4AuYASgKYCGAxkVHnJmAOR4XWMDcGO+RBAntlJgA3gGEAFuhgATFgDsAgnjzk+AGjABRGKUyk5BAL5ha9JiyoEOqVP0FgJU6QAU62AM4BGMAF4RqMDACUncCTwAuMDkAV0wAI1I8NVRDTlsBIUcZVzgPACZff0Dg0LzI0Lx0OQBzZNSbShhyd3cHSRlvUgAPYLlpVrJLADpRelw5fQIAHiyXNy8APn9DBqaWtqcC7t7+klYCEbG4CYMZ9rnc9zyl4RSbOyEAFRCCHI9C28DKc-lI7V1JlMHnAoBsOksAD5aHR6U7A0Gza4pNKNZqtZ6hMDbfS7QbUQ7cE7TDGveY3AJgeTSRIACgAlEVApTSARong5GAptJ0AA3BYUpkiPEHRrHUj-WEEGkESTuIbYeZDb5OeQaGXoOUKy5Kn76eXzOkrQVTYDcvmcQIrQxAA

Shouldn't the following line always succeed:

```
{React.cloneElement(this.props.children, this.props.children.props)}
```
Because the set of props the child has itself should always be correct, shouldn't it?

SamChou19815:
Closing since we now banned React.Element inspection

Issue #5839, opened 2018-02-19
React.cloneElement function and false positive about read-only property


pepakriz:
When I try clone component with $ReadOnly props, flow throws error: `Cannot call React.cloneElement because property foo is not writable in property position.`

See: https://flow.org/try/#0JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQduANmugMImQB2WXMAgnCwAPGLwAm6bHhgA6DuAg8+AHgDecAhAgAuOKhhRgXAOZwAvgBo4ANwjBxAPjhrz9ZqzgLuvGACEhUQkpKjlvJV91OABqLV19Q2MzK1t7Jxc3BnEsD0pbZCg4XE4Ivn49aXxZAFEmLBBIthZUVBVw5QFHR3ps3Kx8wuLFDr8K0Jq6htUm1jaSka76Okq5Zgja+t8ACiGfMusNOL0yZAAjXDILAEp6Fdk1ng2pmB353z8DzW1js4vrmjgAHpAZpkExUP1IKhgDBgDZ+lgoMQoHQgA

Example:

```
import * as React from 'react';

class ComponentA extends React.Component<{ foo: string }, void> {};
class ComponentB extends React.Component<{ +foo: string }, void> {};

declare var componentA: React.Element<Class<ComponentA>>;
declare var componentB: React.Element<Class<ComponentB>>;

React.cloneElement(componentA, { foo: 'abc' }); // OK
React.cloneElement(componentB, { foo: 'abc' }); // Cannot call `React.cloneElement` because property `foo` is not writable in property `foo`.
```

Issue #4824, opened 2017-09-07
React.cloneElement props errors when applied to unions of class components


avaly:
It seems that `React.cloneElement` is throwing some unexpected errors when applied on children defined as `React.ChildrenArray<React.Element<typeof Foo | typeof Bar>>`.

1. The first error is that the props from `Foo` are not found in `Bar` and vice-versa.
1. The second error is that the new prop added via `cloneElemenet` (`ham`) is not found in the props of `Foo` or `Bar`.

This only happens when `Foo` and `Bar` are declared as class components.

```js
class Foo extends React.Component<{ foo: string }> {
  render() {
  	return <div>{this.props.foo}</div>;
  }
}

class Bar extends React.Component<{ bar: string }> {
  render() {
  	return <div>{this.props.bar}</div>;
  }
}

type ChildElement = React.Element<typeof Foo | typeof Bar>;

class Parent extends React.Component<{
  children: React.ChildrenArray<ChildElement>
}> {
  render() {
    return <div>{
      React.Children.map(this.props.children, child => {
        return React.cloneElement(child, { ham: 123 });
      })  
    }</div>;
  }
}
```

[...]

PR #7139, opened 2018-11-06
Custom type definition for React.cloneElement

EdwardDrapkin:
The built in type definition for React.cloneElement occasionally errors out erroneously.  This should fix that.

Should fix these issues:
 https://github.com/facebook/flow/issues/5839 
 https://github.com/facebook/flow/issues/5934
 https://github.com/facebook/flow/issues/4824


SamChou19815:
We do not plan to make changes to the typing. In most cases, you shouldn't be using cloneElement anyways

PR #7146, opened 2018-11-07
Fix type definition of Node.js dns.lookup()

shuhei:
- The `options` argument is optional. When it's not specified, the `callback` is the second argument.
- The `callback` is not optional. When it's not specified at all, the function throws a `TypeError`.
- Add specific fields to the `options`.
- Add `dns.ADDRCONFIG` and `dns.V4MAPPED`, which are useful for `hints` in the `options`.

https://nodejs.org/dist/latest-v11.x/docs/api/dns.html#dns_dns_lookup_hostname_options_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -706,8 +706,12 @@ declare module "dns" {
 
   declare function lookup(
     domain: string,
-    options?: ?number | ?Object,
-    callback?: (err: ?Error, address: string, family: number) => void
+    callback: (err: ?Error, address: string, family: number) => void
+  ): void;
+  declare function lookup(
+    domain: string,
+    options: number | Object,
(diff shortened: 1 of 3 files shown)


nmote:
This looks reasonable, but unfortunately overloading doesn't always work as expected in Flow. Could you add a few tests to `tests/node_tests` to verify that this has the expected behavior?

shuhei:
Sure, I'll add tests for the change and update the PR.

shuhei:
I added tests and rebased the branch against `master`. I am not sure about a diff file from `make test`, but I added it as I saw in other pull requests like https://github.com/facebook/flow/pull/7387. Is it what is supposed to do?

shuhei:
Thank you for taking care of it!

PR #7147, opened 2018-11-07
Fix unsound equivalence between Dates and numbers

### LINKED ISSUE #5539: Major: Function type with Date argument accepts a function with number argument
### Flow verisons
Seems to happen on any version I've tried so far:
* 0.37
* 0.40
* 0.42
* 0.50
* 0.57.1
* 0.61

### Code [(Try flow link)](https://flow.org/try/#0C4TwDgpgBAIghsCBhOAbVAjOBjA1lAXigAoA3NAVwgC5YEIBKQgPigAUAnAewFsBLAM4QAPHAB2IZgCgoUKQDMKY7MD5cxUeVy5lKNKGIo8MEDg1qdegkeMlQA3jKgcIwChw2X+QgHQuBXKikEMQMUgC+UlLEWly08Igo6Fh4DADcQA)
```js
type DateCallback = (value: Date) => Promise<any>
  
function foo(value: number): Promise<any> {
  return Promise.resolve()
}

(foo: DateCallback);
```

### Expected
```
7: (foo: DateCallback);
     ^ function. This type is incompatible with
7: (foo: DateCallback);
          ^ function type
This parameter is incompatible:
1: type DateCallback = (value: Date) => Promise<any>
                               ^ Date. This type is incompatible with
3: function foo(value: number): Promise<any> {
                       ^ number
```

### Actual
```
No errors!
```

NilSet:
fixes #5539

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -5836,7 +5836,7 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
     (* arithmetic/bitwise/update operations besides addition *)
     (*********************************************************)
 
-    | _, AssertArithmeticOperandT _ when numeric l -> ()
+    | _, AssertArithmeticOperandT _ when numberesque l -> ()
     | _, AssertArithmeticOperandT _ ->
       add_output cx ~trace (FlowError.EArithmeticOperand (reason_of_t l))
 
@@ -6621,7 +6621,7 @@ and flow_comparator cx trace reason flip l r =
   let (l, r) = if flip then (r, l) else (l, r) in
   match l, r with
   | DefT (_, StrT _), DefT (_, StrT _) -> ()
-  | (_, _) when numeric l && numeric r -> ()
+  | (_, _) when numberesque l && numberesque r -> ()
   | (_, _) ->
     let reasons = FlowError.ordered_reasons (reason_of_t l, reason_of_t r) in
     add_output cx ~trace (FlowError.EComparison reasons)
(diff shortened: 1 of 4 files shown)


tests/date/date.js
```diff
@@ -24,3 +24,9 @@ new Date(2015, 6, 18, 11, 55, '42');
 new Date(2015, 6, 18, 11, 55, 42, '999');
 new Date(2015, 6, 18, 11, 55, 42, 999, 'hahaha');
 new Date('2015', 6);
+
+var b:boolean = d > 0;
+var n1:number = d - 12;
+var n2:number = d & 255;
+var n3:number = d + 12; // expect error
+var s:string = d + 12; // fixme? in js this coerces both to string and concats them
```
ilias-t: Assuming the maybe FIXME is that this should throw an error, but does not. Should we file a separate issue for this & reference in the comment?

NilSet: It throws confusing errors, and it possibly should be allowed.

src/typing/flow_js.ml
```diff
@@ -7308,12 +7308,15 @@ and any_propagated_use cx trace use_op any = function
 and numeric = function
   | DefT (_, NumT _) -> true
   | DefT (_, SingletonNumT _) -> true
+  | _ -> false
 
+and dateiform = function
```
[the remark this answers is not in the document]
ilias-t: Ok so -iform is like -esque 👍

mroch:
I'm trying to remember why we went out of our way to support this. I'm going to import it to test on our internal repositories to see if that surfaces the reason.

NilSet:
Based on the discussion in #5539 it may have been an unintended consequence of allowing numeric operations and comparisons on Dates. This PR preserves that quality of life feature, and still maintains a type level distinction between Date and number, which is important since it very much is not a number.

An example of the unsoundness is
```
const a: number = new Date() + 12;
```
leaving a string tagged in the type system as a number.

NilSet:
I think at least some of this problem was introduced in a7cb6da61d26f8590272571c1f8ca2ad762b08ba

mroch:
thanks! one common pattern we used was `new Date(date)` to clone a date, so I added a constructor for that. otherwise, whatever motivation we had wasn't worth it because this PR uncovered a bunch of legit bugs.

goodmind:
But this still allows multiplying `Date`?
https://flow.org/try/#0MYewdgzgLgBAhgLhmArgWwEYFMBOMC8yWA7jACJxRYAUAlDAFQwCMATANxA

PR #7149, opened 2018-11-07
Add React hooks (and tests)

bvaughn:
* Added Flow defs for new [React hooks](https://reactjs.org/docs/hooks-intro.html) API along with tests for each new hook.
* Fixed an error in the `React.createRef` definition. (Refs are for any type of value, not just React elements.)
* Added missing `calculateChangedBits` param to `React.createContext` definition.
* Added `React.ConcurrentMode` (16.7+) and tests.
* Added tests for `React.createRef`, `React.createContext`, and `React.StrictMode` (since they didn't have any coverage previously).

Commits are atomic.


bvaughn:
I've spoken with Jordan who offered to be my POC on this. I'll be pushing tests for the remaining hooks this evening, hopefully.

bvaughn:
While adding tests for the `useImperativeMethods` hook, I noticed that the definition for `React.createRef` is incorrect. Refs can hold _any_ type, not just `React$ElementType`. I'm going to make this change too, but let's talk if it breaks anything subtle that I'm unaware of.

bvaughn:
Hooks and tests have been added!

I also added some basic sanity tests for older `createContext` and `createRef` features.

lib/react.js
```diff
@@ -228,8 +228,8 @@ declare module react {
   declare export function createFactory<ElementType: React$ElementType>(
     type: ElementType,
   ): React$ElementFactory<ElementType>;
-  declare export function createRef<ElementType: React$ElementType>(
-  ): {current: null | React$ElementRef<ElementType>};
+  declare export function createRef<T>(
+  ): {current: null | T};
```
bvaughn: This definition was outdated. Refs can be used for any type of value.

lib/react.js
```diff
@@ -222,14 +222,19 @@ declare module react {
   declare export var createClass: React$CreateClass;
   declare export function createContext<T>(
     defaultValue: T,
+    calculateChangedBits: ?(a: T, b: T) => number,
```
bvaughn: The `createContext` API supports `calculateChangedBits` as an advanced perf optimization technique for libs like Relay and Redux. I'm going to add this and also add a small test case for it too.

bvaughn:
Rebased and tweaked the `useRef` return type of `current` to `T | null`

lib/react.js
```diff
@@ -269,6 +271,58 @@ declare module react {
...
+    inputs: ?$ReadOnlyArray<mixed>,
+  ): void;
+
+  declare export function useLayoutEffect(
+    create: () => MaybeCleanUpFn,
+    inputs: ?$ReadOnlyArray<mixed>,
+  ): void;
+
+  declare export function useCallback<T>(
+    callback: () => T | void,
```
AntonPuko: @bvaughn Hi.  Is this supposed that  callback function is always without args?
`const onSubmit = useCallback((values) => {}); return <Form onSubmit={onSubmit} />` - how should we type this case, where values passed as args to the callback from inside child component?

TrySound: Values are passed from parent scope. Nothing is passed to callback.
```
const onSubmit = useCallback(() => {
  fetch('/update', id)
  // this is only invalidation param
}, [id])
```

AntonPuko: As I said, i need to pass callback which accept args from child component. It doens't make sense that its impossible.

something like that.

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H2BuTXAWIBvMAFdyNAKJQoNRmAC+YZqw5cGhPunpwAduWIAFUlz3EAvGAAUASjAWAfGGGowYXQeL6AymIBGWBiW1gBupDBiNOT2Ti5u7mDAwHKE9AAWCYoJCVyEYvh6YAA8AGIEbL4BQYQWwlWBwcrAjqiK-KiehmDl+GxW1vV6fo2EirHOru7FUBVgDTV1C02tiSW9WKUYNDAAJmB6pFg0FgBEs3CnSaslKRWtikA

it can be form onSubmit with values..or whatever other callback with args we can imagine.

its absolutely valid from react side, but not from these typings.

AntonPuko: and there is one more issue with those typings, we can't pass `async` keyword to callbacks in the hooks, like here:
https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H2BuTXAWIBvMAFdyNAKJQoNRmAC+YZqw5cGhPunpwAduWIAxOHDABeMAAoAlBYB8YYajDjJMuYysUAnnvrWduaOwooANGAA2gC6NvyuXIRi+HpgADwAJhgAbmDA9qiKqEA

bvaughn: Ah, good catch. The callback function passed to `useCallback`**can** accept parameters. (We have some tests for this in React, e.g. [1](https://github.com/facebook/react/blob/88ada98198b22d2b8cd056cfba5ffb113c6a50d3/packages/react-dom/src/__tests__/ReactDOMServerIntegrationHooks-test.internal.js#L520-L522), [2](https://github.com/facebook/react/blob/88ada98198b22d2b8cd056cfba5ffb113c6a50d3/packages/react-reconciler/src/__tests__/ReactHooksWithNoopRenderer-test.internal.js#L1494-L1500)).

bvaughn: I _believe_ I've fixed the broken `useCallback` definition via e33b4b57e

AntonPuko: cool, thanks! 👍

thebuilder:
It would be nice with an example of how to fix the issues caused by this breaking change to `createRef`.🙂

Is the correct way of defining the type for `React.createRef()` calls to assign an object with the current value?

```js
divRef: {current: null | HTMLDivElement} = React.createRef();
```

There isn't a fixed type you can use instead? Something like:

```js
divRef: React.Ref<HTMLDivElement> = React.createRef();
```

TrySound:
@thebuilder 
```js
divRef = React.createRef<HTMLDivElement>()
```

TrySound:
@thebuilder Check that you use the latest version of babel and eslint

thebuilder:
Just tried it with the latest version of `eslint-config-react-app`, and also running Babel 7.

```
  62:50  error  Parsing error: Unexpected token

  60 |   }
  61 | 
> 62 |   helpPanelRef = React.createRef<HTMLDivElement>()
     |                                                  ^

```

thebuilder:
Quick example from Codesandbox:
https://codesandbox.io/s/kwyrxykz7r

thebuilder:
@jbrown215 `flow-parser` wasn't part of the project, and `yarn why flow-parser` didn't find any references.

Adding it as a dev dependency did solve the issue. What controls what version of Flow `babel-eslint` uses? Ideally this should be updated in the relevant package, so it'll work without requiring the user to force an install of `flow-parser`

jbrown215:
@thebuilder: I agree with you. I have no idea what controls the version of the flow parser that's used. There might even be something that checks to see if flow-parser is available, and if not defaults to @babel/parser.

We're basically taking advantage of the fact that whoever _is_ setting the flow-parser version is probably using `^0.xx.0` in their package.json, so you can add a higher version without that package caring.

I'm pretty shocked that this hack worked given that you have no deps that depend on the flow parser.

bvaughn:
> We're basically taking advantage of the fact that whoever is setting the flow-parser version is probably using `^0.xx.0` in their package.json, so you can add a higher version without that package caring.

Unless I'm misunderstanding what you'er saying– semver ranges don't work like that for 0.x version numbers. For instance, `^0.3.0` will match any `0.3.x` version– but not `0.4.0` or later.

jbrown215:
Hmm.. then in that case I have no idea why adding this specific dep would fix things.

thebuilder:
A search for `flow-parser` inside my node_modules didn't turn up anything related to eslint. The closest result was `eslint-jsx-a11y`.

MaksimShakavin:
Do you plan to add examples of using flow with hooks into the Flow documentation?

jbrown215:
Was there something specific you wanted to see? The hooks have no special behavior in Flow and the types do not use any complicated built-in types.

PR #7150, opened 2018-11-08
fix definitions of net$Server and tls$Server

bradennapier:
Fixes the definitions of net$Server and tls$Server so that they have the appropriate `listen` overloads. Currently, for example, `server.listen(3106, err => {})` is an error as it is thinking the first parameter should be a string (path). 

Also adds the `options` object as an option to pass to `listen`.


bradennapier:
cla signed

Issue #7180, opened 2018-11-15
Not supported Node.isConnected nor Node.getRootNode()


jorgecasar:
Attending to the Node API are missing isConnected readonly property and getRootNode method

jbrown215:
Feel free to put up a PR

jbrown215:
Sorry I've taken a while to merge this in-- there's been a whole bunch of updates to the libdefs recently which cause conflicts. I will make sure this is in the 0.89 release.

PR #7181, opened 2018-11-15
Add `isConnected` and `getRootNode()` to Node

jorgecasar:
Solve #7180

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -485,6 +485,7 @@ declare class Node extends EventTarget {
   baseURI: ?string;
   childNodes: NodeList<Node>;
   firstChild: ?Node;
+  isConnected: boolean;
   lastChild: ?Node;
   nextSibling: ?Node;
   nodeName: string;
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -500,6 +501,7 @@ declare class Node extends EventTarget {
   cloneNode(deep?: boolean): this;
   compareDocumentPosition(other: Node): number;
   contains(other: ?Node): boolean;
+  getRootNode(): Node;
```
jbrown215: https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode

Can you add the optional object argument?

jbrown215: `void | {composed: boolean}` is what you want, since it's either omitted or or an object with a `composed` field.

jorgecasar: Thanks, I'm not familiarized with flow, sorry.

Just to be sure, this: `getRootNode(options: void | {composed: boolean}): Node;` is the same than `getRootNode(options?: {composed: boolean}): Node;`

jbrown215: No problem! Best way to learn is to write libdefs like this.

Those two types are equivalent to callers, yes. I'm not sure how our function subtyping handles optional arguments vs. void union arguments, but `getRootNode(options?: {composed: boolean}): Node;` is totally fine for what we need here.

jbrown215:
The tests need to be re-recorded, since error messages have been updated.
I can handle that. Thanks for this again!

jorgecasar:
Thanks!

Issue #6966, opened 2018-10-05
flow experience needs a lot of work around flowconfig and flow-typed


indigoviolet:
This is not going to be a very _specific_ bug report, and I'm sure none of this will be a surprise to the flow team, but despite that I believe it is worth re-emphasizing. No disrespect is intended!

I'm a big fan of flow's core feature set - which is awesome - but the developer experience around it can really use a lot of improvement. Some of the notably painful things I've encountered in a couple years of using flow over several projects (with pretty smart developers who are motivated to use flow):

- Setting up a flowconfig on a project can be extremely frustrating. Specifically, the `module.name_mapper` `module.system.node.resolve_dirname` options are quite poorly documented (eg https://github.com/facebook/flow/issues/5180 and related issues), have surprising and opaque behavior. It is surprisingly difficult to teach flow the structure of a project (woe betide you if you have a slightly weird shared folder at a different level from your flowconfig or something), and the `unable to resolve module` errors are not very helpful to fix them (e.g I just struggled with one problem where a bad `module.file_ext` call was hiding some important modules). Flow's configuration for these unfortunately has very little in common with the other tools that do similar things (eg. webpack). 

- It's really easy to get into a broken state where it looks like flow is working and there are no errors, but it's actually your flowconfig that's broken. This is pretty damaging to developer trust, and a huge pain to recover from. We've resorted to constructing a broken file in CI and verifying that flow breaks with that expected error, but I'm not sure that is sufficient. 

- flow-typed is just not there IMO. I recognize how incredibly difficult this problem is, of getting all the major libraries typed and keeping them up to date - but at the same time, flow is only achieving some small fraction of its potential benefit in the current situation. Adding a flow-typed library will probably add hundreds of errors into your codebase, making it quite an unattractive proposition, and then it will randomly break and be wrong in many situations. 

- A corollary of the previous point: I bet most developers have no idea what fraction of their codebase is well typed. The gradual typing premise makes a lot of sense, but it then falls to flow to guide developers to completion. flow needs to do a lot more to (a) demonstrate to users how much type information is missing (b) how to fill in the gaps in the most efficient way - for example, which modules are the biggest blind spots; how to quickly add visibility without having to fix everything, but still fix things when they are touched.

Flow is such an awesome tool, I really hope that it grows out of these limitations so that people can take full advantage of its potential. Much love and thanks for the work that has gone into this!

jbrown215:
This is really good feedback. Thank you!

[...]

Issue #5860, opened 2018-02-22
Files skipped intermittently


ajs139:
Sometimes the files are being checked and have their errors reported, other times they are ignored by Flow (even though the file is still present in the list from `flow ls`). If a file that is prone to not being reported ever stops being ignored (which seems to happen if a file it depends on generates an error), any change to the file (such as adding an empty line) will causes it to disappear from the the output.

This behavior is seen even for isolated errors:

```
18: const foo:number = '';
                        ^^ string. This type is incompatible with
 18: const foo:number = '';
               ^^^^^^ number
```

This was introduced in v0.57.x, and was not a problem in v0.56.0.

kogai:
You might seek #5180 or #5850

ajs139:
Thanks @kogai -- #5180 looks like it's exactly the problem I have.

Issue #5819, opened 2018-02-15
In a large codebase with several errors, Flow reports "No errors!" until triggered by a watch event to recheck


callumlocke:
We've been using Flow in a fairly large codebase (flow says it parses 50344 files, including node modules), for a year or so. Unfortunately it's closed source so I can't share it.

In the last few weeks, Flow has started behaving strangely when I follow these steps:

- Run `flow stop && flow` – it performs a check lasting several seconds, and eventually reports `No errors!`
- Edit a *comment* in a random file somewhere in our codebase (just to trigger a filesystem watch event)
- Run `flow` – now it reports a bunch of errors that weren't reported before (in various files, not necessarily connected to the one you edited).
- Run `flow stop && flow` again, and the errors go away.

The errors appear to be real type errors in our code. So why didn't Flow notice them before? Some of these errors turn out to have existed in our codebase for months. They just never got reported until now, and still often don't get reported until seemingly random events. Then what's extra weird is, if we fix those errors and perform the above steps again, `flow` eventually finds more errors in other files, which it didn't report before. It's as if Flow is reaching some kind of internal limit on how many files it can traverse, and gives up after a while and just says `No errors!`

---

EDIT: just to clarify, running `flow check` also results in `No errors!`. The only way to see the errors is to run `flow` to get a server going, then edit a file to trigger a recheck, then run `flow` again to see the status. Also, if you do further no-change edits to other files, it sometimes finds even more errors in still other files.

aikar:
I think im seeing similar issues. I had a code bug that should of been caught by flow, missing argument to a function that wasnt marked optional, and i get no errors.

kogai:
Could you share your `.flowconfig` or making sure #5180?
I guess perhaps such behavior of the Flow seems similar to the issue above reported.

aikar:
```
[ignore]
<PROJECT_ROOT>/dist/

[include]
<PROJECT_ROOT>/src/

[libs]
<PROJECT_ROOT>/flow-typed/

[options]
module.ignore_non_literal_requires=true
```

Mine's pretty trivial.

kogai:
Thanks for sharing 🙏
But sorry in your case, the `.flowconfig` seems not a cause of the problem.
(At least to the caused I guessed)

Problems between @callumlocke 's one and @aikar 's possibly might be from another cause. 🤔 

@aikar Did you receive an error which indicates missing argument when you edit something in the code?

callumlocke:
Mine is:

```
[...]

Issue #5778, opened 2018-02-06
Flow starts listing dozens of errors after 5-10 minutes - without any code changes


callumlocke:
In a large codebase I'm working on, I can run `flow` and get `No errors`. But then if I wait a few minutes and run `flow` again (**without any changes to my code**), it lists a bunch of errors (about 7). If I wait a bit longer and run `flow` again, I sometimes get a lot more errors (often 28 or more).

If I run `flow stop && flow`, it shows "No errors" again.

Unfortunately the project is closed source, but I can show 28 dubious errors here:

<details>
 <summary>Flow errors</summary>

```
Error: app/common/ModalRoutes.jsx:0
inconsistent use of library definitions
133:   declare type ComponentWithDefaultProps<DefaultProps: {}, Props: {}> =
                                                                       ^^ undefined property `intl`. This type is incompatible with. See lib: flow-typed/npm/react-intl_v2.x.x.js:133
 56: type $npm$ReactIntl$IntlShape = $npm$ReactIntl$IntlConfig &
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^ $npm$ReactIntl$IntlConfig. See lib: flow-typed/npm/react-intl_v2.x.x.js:56

Error: app/common/ModalRoutes.jsx:0
inconsistent use of library definitions
133:   declare type ComponentWithDefaultProps<DefaultProps: {}, Props: {}> =
                                                                       ^^ undefined property `intl`. This type is incompatible with. See lib: flow-typed/npm/react-intl_v2.x.x.js:133
 57:   $npm$ReactIntl$IntlFormat & { now: () => number };
       ^^^^^^^^^^^^^^^^^^^^^^^^^ $npm$ReactIntl$IntlFormat. See lib: flow-typed/npm/react-intl_v2.x.x.js:57

Error: app/common/ModalRoutes.jsx:0
inconsistent use of library definitions
133:   declare type ComponentWithDefaultProps<DefaultProps: {}, Props: {}> =
                                                                       ^^ undefined property `intl`. This type is incompatible with. See lib: flow-typed/npm/react-intl_v2.x.x.js:133
 57:   $npm$ReactIntl$IntlFormat & { now: () => number };
                                   ^^^^^^^^^^^^^^^^^^^^^ object type. See lib: flow-typed/npm/react-intl_v2.x.x.js:57

Error: app/common/components/AddressForm/index.jsx:44
               v-------------
 44:           <ManualAddress
 45:             country={country}
[...]

PR #7185, opened 2018-11-16
[FEATURE] Add `module_alias` resolution capabilities

bradennapier:
Hey Y'all! 

So first of all, never used OCAML in my life so excuse any flaws in style.  Mostly just tried to mimic styling I saw done elsewhere as I implemented this feature.  I did add tests and it is definitely fixing the issues I have been having thus far!

So there are a ton of people that use things like `Webpack`, `Rollup`, or something like `babel-plugin-module-resolver` (including Flow team recently I noticed ;-)).  

## The Feature

This adds an option that is very similar to `module.system.node.resolve_dirname` but with slightly different handling.  It is meant to be used in situations where `Webpack`'s `resolve.modules` or `babel-plugin-module-resolver` (as well as multiple other bundlers which offer this type of functionality).  

It basically allows a separated place to define paths that should be aliased but type-checked normally since its actually part of the main codebase in this case. 

> **Note:** aliases are resolved before any `resolve_dirname` configurations.

```
[options]
module.system.node.resolve_alias=shared
module.system.node.resolve_alias=src
```

In the case of something like the image below:

image

We can then do the following in `ComponentOne` or `ComponentTwo`:

```
/* @flow */
import A from 'A';
import B from 'B';
import C from 'C';
```

this would then mimic the `Webpack` configuration of:

```javascript
module.exports = {
  //...
  resolve: {
    modules: ['shared', path.resolve(__dirname, 'src'), 'node_modules']
  }
};
```

or the `babel-plugin-module-resolver` configuration of:

```
{
  "plugins": [
    ["module-resolver", {
      "root": ["shared", "./src"]
    }]
  ]
}
```

> Order of precedence would be `C` (shared), `B` (shared), then `A` (./src)

## The Problem This Fixes

The general idea is to remove the need to use relative paths everywhere and anywhere.  At this point I can't even program without it!  So essentially people will do a few possible things:

### Have a Resolution to their Root 

```
module.system.node.resolve_dirname=./src
module.system.node.resolve_dirname=node_modules
```

This way they can import from their root as modules (@see [Webpack Documentation](https://webpack.js.org/configuration/resolve/#resolve-modules)) (@see [babel-plugin-module-resolver Documentation](https://github.com/tleunen/babel-plugin-module-resolver#getting-started))

### Use a customized styling for `node_modules` 

In my case, I generally use a `shared` folder and set it up to resolve the same as `node_modules` would.  This is especially useful in `React` applications as it provides the ability for components to easily share pieces of code at any level.  

While the `resolve_dirname` kind of works better here (with the other example it breaks the code completely), the problem is that when you're building a file it won't work until its been imported from `outside` a shared folder.

https://github.com/bradennapier/flow-modules-bug-repro

This shows the issue when using it this way.

### Why `resolve_dirname` doesn't work for this

The issue with `resolve_dirname` is that doing this will generally break the entire application due to a change in `0.57.0` 

```
New Features:
* Flow will now only check the files in `node_modules/` which are direct or transitive dependencies of the non-`node_modules` code.
```

This is extended to any defined `resolve_dirname` folders even though it's largely used for this purpose.  We have since discovered that using a relative path can fix it but there are still lingering issues in these cases.

### Examples of People having issues that this solves

Just did some quick searches for examples

#5180 
https://github.com/flowtype/flow-for-vscode/issues/289
https://github.com/facebook/flow/issues/4103
https://github.com/facebook/flow/issues/5494
https://github.com/facebook/flow/issues/5819
https://github.com/facebook/flow/issues/5647
https://github.com/facebook/flow/issues/5586
https://github.com/facebook/flow/issues/5385
https://github.com/tleunen/babel-plugin-module-resolver/issues/233
https://github.com/facebook/flow/issues/5778
https://github.com/facebook/flow/issues/5860
https://github.com/facebook/create-react-app/pull/4105
https://github.com/facebook/flow/issues/6966
https://github.com/tleunen/babel-plugin-module-resolver/issues/235

As well as many more if you search through the various flow repos.

## Testing on OSX

If you don't want to build the source and run OSX you can use the binary I built.  There is a compiled binary for OSX located in the repro i linked within a branch: https://github.com/bradennapier/flow-modules-bug-repro/tree/feature/fixed-by-pr-7185/bin


ugeng:
Hi @bradennapier! Looks like you've done very useful and highly-demanded task here!
I didn't see your code and have a question regarding of its capabilities in lerna-managed monorepos.
Will it work if I have only one `.flowconfig` in monorepo's root, where I describe all the aliases (say, all my packages have same structure)?
Thanks!

bradennapier:
Hmm well theoretically yes it would work fine for a monorepo I do not see why not.  You’d have to use a non relative “src” so it would bubble up in each repo to find the right root path for each but no reason it wouldn’t work great to solve that issue. 

I’d imagine if not then this at least provides the ground work for what would be needed to support it.  I do have a monorepo I’ll test it on at some point when I have time.

ugeng:
@bradennapier, sounds wonderful! No doubt in that!
I've mentioned that issue because when I use `name_mapper` like below:
```
module.name_mapper='^containers' ->'src/containers'
```
Flow throws famous 'can not resolve module' error in every package. 

I definitely sure that you predict and avoid such issues in your PR, but it should be noticed though.

Thank you! Can't wait for PR's being approved.

bradennapier:
Have you tried the following:

```
module.name_mapper='^containers\/\(.*\)$' -> '<PROJECT_ROOT>/src/containers/\1'
```

Not sure if the substitution is at all necessary but its how all mine are done :-P

ugeng:
hi, @bradennapier! I've just tried a new binary in my lerna-managed monorepo with yarn workspaces enabled.
I've tuned my `.flowconfig` like following
image
and checked that i'm on new version (currenly i'm on 0.84 because of react-redux typings issue)
image

All seems fine.

And it definitely works in my "host" package
image

Like you can see, it provides path to module which, as I can assume, tells us that the module resolved.
But in my "library" packages it works a bit differently.
image

It does not provides neither module path, nor any module info.. But vscode plugin doesn't complain about it, so it probably works! 😄

PS. In order to simplify webpack config issues I use [`nwb`](https://github.com/insin/nwb) starter tool both in main (hosting) package and in library packages. I don't know how it could impact on situation but probably it could.

PPS. You set version=0.86 but such version already released and have issues with react-redux typings

PPPS. Thank you for the great job you've done. Seems like all almost perfect )

bradennapier:
Yeah I leave it out of the List which would define the alias as a “module” to Flow so it probably is the reason it won’t show up there as a module.  

It’s an alias so it’s simply resolving the given alias in a node_module way while preserving standard flow type checking that is removed for modules. 

Thanks for testing!  Glad it works for your needs :)

bradennapier:
@jbrown215 any chance i can get a review on this PR?  I think its a really solid feature and i miss it already since i updated to latest and havent yet built my version of it! 

I updated and fixed conflicts with newest version

jbrown215:
Hey @bradennapier:

I really appreciate that you took the time to make this PR. Unfortunately, I'm not a good person to review it. I'll import it and see if anyone else can take a look, but I can't guarantee you someone will get to it.

bradennapier:
Thanks @jbrown215 !  Looks like build is failing with my conflict resolutions.  If someone could review the concept and let me know they'd merge it I will absolutely fix it right away.  I probably will anyway but would be nice to have someone with the proper authority or w/e approve the spec.

jbrown215:
I recommend waiting for someone to respond before you put more work into this. I don't want you to keep putting time into something that we may decide is something we don't want or cannot put in the time to support.

bradennapier:
Will do, thanks again!

vjpr:
> You could also try `module.system.node.resolve_dirname=./src` which should resolve all directories in src as modules if that is a desired effect.

FYI: This means `modules/foo.js` won't work, but `modules/foo/index.js` will, for `import foo from 'modules/foo`.

vjpr:
@jbrown215 Can you please review this as discussed [on Twitter](https://twitter.com/flowtype/status/1089056221703950336).

bradennapier:
I would note that this part of the codebase has changed a little bit so would need to adjust accordingly, but happy to look into that should the general concept of the feature be approved for further acceptance (which I would personally like to see). 

We do run into it on every project and it's fairly common practice to use `module alias` for webpack setups.

src/services/inference/module/module_js.ml
```diff
@@ -407,7 +407,14 @@ module Node = struct
...
       lazy (
         if SSet.mem dir node_modules_containers then
-          lazy_seq (Files.node_resolver_dirnames file_options |> Core_list.map ~f:(fun dirname ->
+          lazy_seq (Files.node_resolver_aliases file_options |> Core_list.map (fun dirname ->
+            lazy (resolve_relative
+              ~options
+              loc ?resolution_acc dir (spf "%s%s%s" dirname Filename.dir_sep r)
+            )
+          ))
+        else if SSet.mem dir node_modules_containers then
```
nmote: This doesn't look right to me. This is checking the same exact thing as the `if` above, so this code path will never be traversed. Am I missing something?

nmote:
I chatted briefly with @samwgoldman who had some concerns that this might cause problems with rechecks. I'm still looking into this potential issue and I'll have more to say on Monday. I'm optimistic that we can get this merged in some form in the near future.

nmote:
I spent some time investigating this today, and I think this should work out. If I'm missing something, hopefully @samwgoldman corrects me.

Basically, the concern was that complicating module resolution in this way could lead us to miss rechecks. Specifically, if `foo.js` depends on `bar.js`, but another `bar.js` elsewhere is added, and due to module resolution rules, `foo.js` now depends on the *new* `bar.js`, we have to make sure to recheck `foo.js` against the newly created `bar.js`.

Fortunately, it looks to me like our logic for tracking this should generalize quite easily. In fact, I don't think this PR will need to do anything special for everything to fall into place.

I could be misunderstanding something (this is my first time looking at this logic), but I'm optimistic that this can be merged shortly. @bradennapier, when you get a chance could you rebase this so that I can play around with it and make sure that everything behaves the way I expect?

src/services/inference/module/module_js.ml
bradennapier: Modified the handling here but absolutely needs a review :-D

bradennapier:
Alright so rebased and cleaned up the commits a bit here, confirmed build worked and the feature still operates as expected.

There is one deviation that could come up from the other options.  I largely did not incorporate things this way because it would have made for a much more significant shift in the way things are done (and may not even be possible since our goal is to isolate the `node_modules` resolution from our custom alias resolution).  

In webpack alias you can do something like

```javascript
resolve: [
   'shared',
   'node_modules',
   'backup_modules'
]
```

which would then resolve in the order provided if possible.  in our scenario (which is the only way i have ever even considered using this feature, and the way i believe most people do use it), our custom aliases would also resolve **before** `node_modules` so having a folder locally which shadows a `node_module` would instead resolve our alias.

bradennapier:
## Failure Case (Non Critical but Important)

@see [this commit](https://github.com/facebook/flow/pull/7185/commits/961d65533d91766dc299f2fb54cd9df5319ddb69)

So the above example actually led me to finding a failure case in how this should work (and it also fails in the case of using `module_dirname` which was my suspicion all along.  The reason using `./src` in `module_dirname` is a workaround is because of this bug.

Essentially it is not honoring relative paths properly.  I think I found the reason for this when I was putting this PR together but also think it should properly handle `<PROJECT_ROOT>` - a bit beyond me as-to how to do this properly.  I assume its just about calling `expand_project_root_token_to_string` at the right time, just not familiar enough with `OCAML` to fix this case (and the actual relative path case `./`).

nmote:
Hey, just to make sure that everyone is on the same page here: I'm planning to review the rest of the PR once @samwgoldman is satisfied with the part that he has started reviewing.

bradennapier:
I spoke with Sam and the only thing needed is to do the extra tests which I haven’t had time for due to moving across the country.  Hope to get to it soon.

nmote:
Okay, no rush on our end, I just wanted to make sure you weren't waiting for me.

goodmind:
@bradennapier are you available for this PR?

davidroeca:
Any updates on this PR? The current work-around is to have a messy flow config with a bunch of module name mappers, which becomes cumbersome.

nmote:
I haven't heard anything from @bradennapier in some time, so I'm going to close this. Anyone, feel free to ping me on a replacement PR if you are interested in driving this to completion.

PR #7203, opened 2018-11-22
Substitute 'Type' with 'typeof Component'

superburrito:
Not 100% sure if a typo was made here -- but I think this reads somewhat clearer since it'll be consistent with the rest of page


pakoito:
Thank you @superburrito! It took a while, and that's my bad :D

PR #7213, opened 2018-11-27
Add React.Suspense component, remove React.useMutationEffect hook

### LINKED ISSUE #7219: React.lazy has an incorrect type definition
I am getting an error when using React.lazy:
```
Cannot call `React.lazy` with function bound to `component` because in the return value:
 - Either a callable signature is missing in `Promise` [1] but exists in `React.StatelessFunctionalComponent` [2].
 - Or `Promise` [1] is incompatible with statics of `React.Component` [3].
```
as an example:
```
const asyncGraphiQL = React.lazy(() =>
  import(/* webpackChunkName: "graphiql" */ './containers/graphiql')
);
```

Should [line](https://github.com/facebook/flow/blob/9d5d7ed0b9ca3677186d4ba444973277add29404/lib/react.js#L288) be returning a promise instead?

```
declare export function lazy<P>(
    component: () => Promise<{ default: React$ComponentType<P> }>,
  ): React$ComponentType<P>;
```

bvaughn:
Resolves #7219

- [x] 2fdcda5: Removed `React.useMutationEffect` hook (facebook/react/pull/14336) and updated tests. This hook was only ever exposed in an alpha release of 16.7 and so it should be safe to remove in a non-backwards-compatible way.
- [x] dabf49e: Add `React.Suspense` component and tests. (No final public documentation exists for `Suspense` yet– although we're working on it. For the time being, [tests like this](https://github.com/facebook/react/blob/master/packages/react-reconciler/src/__tests__/ReactSuspense-test.internal.js) can serve as a reference.)
- [x] 8df05bd: Fixed invalid `useCallback` definition and tests.
- [x] 063b022: Fixed invalid `React.lazy` signature to reflect that a `Promise` with a "default" property should be returned (rather than a React component).


bvaughn:
The type-at-pos test kept crashing for me locally but I don't _think_ that's related to this PR.

lib/react.js
```diff
@@ -250,6 +250,12 @@ declare module react {
   declare export var ConcurrentMode: ({children: ?React$Node}) => React$Node; // 16.7+
   declare export var StrictMode: ({children: ?React$Node}) => React$Node;
```
sophiebits: (Should these also be using React$ComponentType? You can't call them as a function.)

lib/react.js
```diff
@@ -250,6 +250,12 @@ declare module react {
...
 
+  declare export var Suspense: React$ComponentType<{
+    children?: ?React$Node,
+    fallback: React$Node,
```
sophiebits: fallback is optional (for nested Suspense components); missing, undefined, and null should all typecheck I think

bvaughn: Ah thanks! I thought it was a required prop.

lib/react.js
```diff
@@ -250,6 +250,12 @@ declare module react {
...
+    maxDuration?: number
+  }>; // 16.6+
```
sebmarkbage: Does this imply that this is a class? I guess that isn't technically correct since it's one of those internal special symbols like Fragment but maybe good enough for now.

Should it use the same hack as Fragments and pretend to be a functional component? That way maybe refs are invalid maybe?

bvaughn: > Should it use the same hack as Fragments and pretend to be a functional component? That way maybe refs are invalid maybe?

Then people might think they can call it as a function, which would error?

lib/react.js
jbrown215: Good catch. I'm ok with adding that fix to this PR or doing it in a separate one.

lib/react.js
jbrown215: ComponentType is the type alias that (currently) represents function components and class components.

bvaughn: Is there an alias for something that more accurately represents `StrictMode`, `Suspense`, etc?

lib/react.js
jbrown215: @bvaughn actually brought up a good point below, it's not quite a ComponentType. It _will_ be an AbstractComponent, though!

bvaughn: Let's leave it for now?

lib/react.js
jbrown215: Not yet, but AbstractComponent will fulfill that role!

TrySound: Is React$AbstractComponent made for this?

lib/react.js
jbrown215: Yeah, sounds good

lib/react.js
jbrown215: AbstractComponent was made to represent components more abstractly than we do today. When it's done, it'll be able to represent this, the component returned from forwardRef, and hopefully many more components that React releases in the future.

It's still a WIP, but I will release a blog post when it's fully landed in master.

jbrown215: > Then people might think they can call it as a function, which would error? (Doesn't seem like that would obviously be better than this.)

If the instance type is void, it's more precise to use the function hack. Since we're already using that hack today, and since it will be fixed as part of my work on AbstractComponent, I think we should use the function representation for now.

bvaughn: Ok~

lib/react.js
```diff
@@ -250,6 +250,12 @@ declare module react {
   declare export var ConcurrentMode: ({children: ?React$Node}) => React$Node; // 16.7+
   declare export var StrictMode: ({children: ?React$Node}) => React$Node;
 
+  declare export var Suspense: React$ComponentType<{
+    children?: ?React$Node,
```
jbrown215: Is this needed?
IIUC, this says that it may or may not have children, and they have to be react nodes. [I think that's the default](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAxnAdgZwC5gBKApgIboEC8YATsQI4CuAlnQBQDkd5eHAlAG5UURlgrNsYAMJwAtgAdsxLHjYAPAFxgA3gF8+WkjwB0AOTgATYjtrE8jGljBZGMGALC6M2fGAD6pGDUADwyCkoqYMAAfEKYuAR+AEZBYKFyiljKeNHBFswAbtHaAEQAFsRucCW6wcD5RXVhmdmxQA).

sophiebits: whoa, that seems like a bug – composite components should only take children if it's explicitly specified (it's only host components like `'div'` that always take React$Node children)

bvaughn: Based on the context provider and consumer types above, I assumed an explicit `children` prop was needed.

lib/react.js
bvaughn: I must be missing something, but changing it to a function declaration seems break the validation.

For example, [this test](https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocgEyTQBsUi4kAPSWOAN0bgGUBXAM5gkAO0FIAXImIwAJAGFckUWJgAVAJ4iAPAG9ycOGgAWwOlSKiA-NOvJ08gHIQaAGkNxMKOnQBG6ADWtjKOci7uniAoLAAi-FAoMMAQNtKi-CB+SFDkAL4AfKRwAPQlcACMAGwAdFUA1JQGRjoCwmIScNFxCUkpogC8AEQofmhDpQWl5QCiUDhQ0oIwUMCiAOZwwIJbomjKfX50SHAA7sAwJnAZWTn55EA) fails, as I would expect.

But [this one](https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocgEyTQBsUi4kAPSWOAN0bgGUBXAM5gkAO0FIAXHAAUAb3Jw4aABbA6VIqID807cnQwAJADkINADSK4mFHToAjdAGtdiYsbOXrIFCwAi-FAoMMAQOtKi-CAOSFDkAL4AlHAAvAB87oam5kikcAD0BXAAjABsAHRlANSK5ApKADwCwmIScL4BQSFhoqkARCgOaP2FmUVwAKJQOFDSgjBQwKIA5nDAguuiaLhgPQ50SHAA7sAwKnBRMXGJ5EA) and [this one](https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocgEyTQBsUi4kAPSWOTAVwDsNgIPOAGUuAZzBIeYpHAAUAb3Jw4aABbA6VIjwD8ALji7k6GABIAchBoAaZZxR06AI3QBrA4mLmrt+yBQWABEuKBQYAT1DHi4QZyQocgBfAEpDEwxLayRSOAB6PLgARgA2ADoSgGpKJRUAHlEJKRk4AODQ8MiAXgAiFGc0HvyAPnzCgFEoHChDMRgoYB4AczhgMVW+XDBO5zpZAHdgGDU4GLiE5PIgA) do not.

bvaughn: Calling `Suspense` directly (as a plain function) e.g. `Suspense({ maxDuration: 'abc' })` properly catches the error, but wrapping it in a `createElement` call does not.

bvaughn:
I pushed a change to make `Suspense` "fallback" prop optional but I didn't change the component type yet because it looks broken.

lib/react.js
jbrown215: FWIW, children is the part of our react model that I'm still not totally familiar with. I'm glad we're having this conversation because it's forcing me to look into it more.

@sophiebits: I'll have to do some more investigation. Requiring an explicit children prop would be an extremely disruptive change at this point.

@bvaughn: I just took a look at our react docs to see what we say about children types. It looks like we have conflicting advice [here](https://flow.org/en/docs/react/types/#toc-react-node) and [here](https://flow.org/en/docs/react/children/). I think the `?React.Node` bit is necessary, but I'm unclear about whether or not the `?children` part is also necessary. I'll have to look into it more, but either way I'm ok with what you have here.

jbrown215: @sophiebits: Width subtyping allows extra properties to be passed into an object, so if the Props type is _not_ exact then we allow children even if they are not specified explicitly. On the other hand, if the Props type _is_ exact then [we don't allow the extra properties](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAxnAdgZwC5gBKApgIboEC8YATsQI4CuAlnQBQDkd5eHAlAG5UURlgrNsYAMJwAtgAdsxLHjYAPAFxgA3gB9dAXz5aSPAHQA5OABNiO2sTyMaWMFkYwYAsAYzZ8YAD6pGDUADwyCkoqYMAAfEKYuASBAEahYBFyiljKeHFh1swAbnHaAEQAFsSecOUGYcBFpY2ROXkJQA).

Mystery solved!

sophiebits: for posterity: React's treatment of `children`:

1. React.createElement takes varargs that become `props.children` (JSX children syntax turns into this)
2. host component (`'div'` etc) `children` prop needs to be react nodes and React recurses into them
3. composite component `children` are *not* special in any way; by convention people often use it to contain React nodes then render `<div>{props.children}</div>` or similar, but they can also be other data types (eg: the "render props" pattern often has `children` be a function)

jbrown215:
Yea, that looks like a bug to me.

The only downside I see with ComponentType is that we won't correctly check the `ref` pseudo-prop. Other than that, it's just a strange representation. I'd rather not block people from using it just because it's strange and has unsafe ref typing.

AbstractComponent is far enough along that I don't think this problem will be around for long.

bvaughn:
Sweet! Thanks for the udpate!

jbrown215:
@bvaughn: I think the tests need to be re-recorded again.

bvaughn:
Yeah, I noticed. I guess something landed in master that caused conflicts.

I'm in the process of rebasing and rerecording now.

bvaughn:
I noticed something wrong with the `useCallback` def that I'm going to go ahead and toss in here too.

bvaughn:
Okay. I think I fixed the bad `useCallback` definition (and tests) via e33b4b57e. Another round of review, pls!

bvaughn:
Okay. `useCallback` updated (again) as well as `lazy`

TrySound:
Can useRef current be not nullable since it has default value?

jbrown215:
The default value isn't required :(. This is something Brian and I did discuss too

bvaughn:
FB internal lint failure because of the lack of a newline at the end of the `lazy` test :( so I rebased once more and fixed that along the way.

PR #7214, opened 2018-11-27
Update http(s).request|get to support url and options

leebyron:
See: https://nodejs.org/api/http.html#http_http_request_options_callback


nmote:
Could you add a few tests in `tests/node_tests` to verify that this libdef behaves as expected? Sometimes Flow exhibits surprising behavior with overloading.

nmote:

PR #7215, opened 2018-11-27
lib/node.js: 'https' interfaces should use tls$TLSSocket instead of net$Socket

cakoose:
The issue:

```js
// @flow
const http = require('http');
const https = require('https');

http.request({}, res => {
    console.log(res.socket.getPeerCertificate());  // <-- true positive
});
https.request({}, res => {
    console.log(res.socket.getPeerCertificate());  // <-- false positive
});
```

Before, Flow would flag both with the message: Cannot call `res.socket.getPeerCertificate` because property `getPeerCertificate` is missing in `net$Socket` [1].

After this change, Flow only flags the first.

My fix adds a `SocketT` type parameter to a few base classes.  Not sure if that's the "proper" fix.


nmote:
At a glance, this looks reasonable. Could you link API docs that support this change?

cakoose:
It's not spelled out explicitly, but:
- The "http" module doc says that response.socket references the "underlying socket" ([link](https://nodejs.org/api/http.html#http_response_socket))
- The "https" module doc ([link](https://nodejs.org/api/https.html))
    - Has several instances of of "https.blah() is like http.blah() but for HTTPS"
    - The "https.request()" doc says that options from "tls.connect()" are accepted
- There's a Node.js bug report relating to `getPeerCertificate()` and the responses from the Node.js team members seem ok with relying on it from `https.request()` ([link](https://github.com/nodejs/node/issues/7672)).

cakoose:
Added the `SocketT` default and rebased on to latest origin/master.  Feel free to tweak/squash/replace my commits; authorship is not important to me.

(Too bad we can't write `Agent` instead of `Agent<>` when all type parameters have defaults.  Oh well...)

cakoose:
Oh, that makes sense.  Java did something similar when to aid the transition to generics in 1.5.

nmote:
I think we're good here. My only concern was that anybody who references these types would have to add the `<>` on rollout, but I don't think that will be a huge issue. I only found a few references to these types in our main internal codebases, and I asked in the Discord to see if anybody had concerns and nobody seemed worried about it.

I'll make sure it gets mentioned specifically in the release notes when it goes out.

I just need to get someone to approve this internally and then I'll land it. Thanks for the contribution!

nmote:
FYI I made some tweaks since this PR ended up causing us some rollout pain internally. The specific issue and the fix are in the commit message: https://github.com/facebook/flow/commit/6a4c5a04c65825499f2754b70b515f544400b2d9

This change will go out with v0.94.0. In the meantime we are just going to suppress the errors.

PR #7225, opened 2018-12-03
createContextualFragment returns a DocumentFragment

wincent:
[Both MDN](https://developer.mozilla.org/en-US/docs/Web/API/Range/createContextualFragment) and the [DOM Parsing and Serialization Working Draft](https://w3c.github.io/DOM-Parsing/#extensions-to-the-range-interface) indicate that this method returns a `DocumentFragment`, not a `Node`.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1195,7 +1195,7 @@ declare class Range { // extension
   setStartAfter(refNode: Node): void;
   extractContents(): DocumentFragment;
   setEndAfter(refNode: Node): void;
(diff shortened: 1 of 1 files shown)


skeggse:
This build should be passing fine, as this change is identical to #7259 and that build passed. Might just need a rebuild?

wincent:
> This build should be passing fine, as this change is identical to #7259 and that build passed. Might just need a rebuild?

Just rebased, which should be enough to trigger a re-build.

PR #7245, opened 2018-12-08
precise return types for Object.entries and Object.keys on typed objects

### LINKED ISSUE #5838: Incorrect mixed type from value of Object.entries
Flow assumes that `value` as returned by `[key, value]` for [Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) is mixed. Not really sure where to start looking but willing to check it out / try to help!

```js
const obj: {
  	[string]: number
} = {
	a: 1,
  	b: 2,
  	c: 3
};
for (const [char, num] of Object.entries(obj)) {
	const copy: number = num;
}
```

Error:
```
11: 	const copy: number = num;
                          ^ Cannot assign `num` to `copy` because mixed [1] is incompatible with number [2].
References:
[LIB] static/v0.66.0/flowlib/core.js:35:     static entries(object: any): Array<[string, mixed]>;
                                                                                         ^ [1]
11: 	const copy: number = num;
                 ^ [2]
```

[Try Flow](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzDgCMArALjAG9UwwBIAbQICcBLbAcwF1LsBXALbEApi1QBfMAF5qqegENKARgA0tBsUoAmdXXoZKAZkkBuVFDgswACix5CjDAAsFLVWAGDuRKGADyZCIY+AB0Itj47CK4NiSkAJQJcgY4BGBYAA4AnnxCotayXuYSqEA)

mxmul:
Fixes #5838.

Our codebase makes extensive use of [objects as maps](https://flow.org/en/docs/types/objects/#toc-objects-as-maps), and we currently lose that type information when passing into `Object.entries` or `Object.values`, since they return `Array<string, mixed>` and `Array<mixed>` respectively. I've added some additional annotations for these functions when the object passed in has an exact type – this lets us do things like this:

```js
const obj: {
  	[string]: number
} = {
	a: 1,
  	b: 2,
  	c: 3
};
for (const [char, num] of Object.entries(obj)) {
	const copy: number = num;
}
```

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -45,6 +45,7 @@ declare class Object {
     static create(o: any, properties?: PropertyDescriptorMap): any; // compiler magic
     static defineProperties(o: any, properties: PropertyDescriptorMap): any;
     static defineProperty<T>(o: any, p: any, attributes: PropertyDescriptor<T>): any;
+    static entries<K, V>(object: { [K]: V }): Array<[string, V]>;
     static entries(object: any): Array<[string, mixed]>;
     static freeze<T>(o: T): T;
     static getOwnPropertyDescriptor<T>(o: any, p: any): PropertyDescriptor<T> | void;
@@ -59,6 +60,7 @@ declare class Object {
     static preventExtensions<T>(o: T): T;
     static seal<T>(o: T): T;
     static setPrototypeOf(o: any, proto: ?Object): any;
+    static values<K, V>(object: { [K]: V }): Array<V>;
     static values(object: any): Array<mixed>;
     hasOwnProperty(prop: any): boolean;
     isPrototypeOf(o: any): boolean;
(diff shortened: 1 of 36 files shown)


nmote:
This comes up quite a bit. Unfortunately we can't make these libdefs more precise without introducing unsoundness. I don't remember the specifics but maybe @samwgoldman would be willing to elaborate.

samwgoldman:
This change works correctly for objects as maps, but not for other kinds of objects, due to width subtyping.

```
declare function values<K, V>(o: { [any]: V }): Array<V>;
var o: { p: string } = { p: "foo", q: number };
var vs: Array<string> = values(o); // lies!
```

TrySound:
@samwgoldman Can this work for exact types as a special case? It will be more relevant since you are gonna make all types exact by default.

samwgoldman:
It's possible, yes, but it's not clear to me whether we should allow it. The resulting union of all value types in a heterogeneous object could lead to expensive type-checker behavior. (We have observed this, as people use a version of this definition internally as a work around.)

PR #7248, opened 2018-12-10
Fix Node.js assert.fail libdefs

rtsao:
The [non-deprecated unary/nullary form of `assert.fail`](https://nodejs.org/api/assert.html#assert_assert_fail_message) should be allowed. 

Additionally, when using the [deprecated `assert.fail` form](https://nodejs.org/api/assert.html#assert_assert_fail_actual_expected_message_operator_stackstartfn), `message` and `operator` parameters should be optional and there should be an optional fifth `stackStartFn` parameter.

Furthermore, the message parameter can be either `string` or `Error`.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2081,7 +2081,8 @@ declare module "assert" {
   declare module.exports: {
     (value: any, message?: string): void;
     ok(value: any, message?: string): void;
-    fail(actual: any, expected: any, message: string, operator: string): void;
+    fail(message?: string | Error): void;
+    fail(actual: any, expected: any, message?: string | Error, operator?: string, stackStartFn?: Function): void;
     equal(actual: any, expected: any, message?: string): void;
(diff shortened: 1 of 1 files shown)


mroch:
thanks for the fix.

can you add some tests to https://github.com/facebook/flow/blob/master/tests/node_tests/assert/assert.js please? some that pass and some that cause errors, to make sure it's not just falling back to allowing anything. overloading is tricky for Flow to choose between, so it's not so much to make sure it works properly now as it is a regression test to make sure we don't break it when we improve overloading.

mroch:
run the test with `make; ./runtests.sh bin/flow node_tests`. to update the expected output, add `-r` (has to be the first arg, before `bin/flow`)

nmote:
@rtsao, are you still planning to work on this?

nmote:
Closing for now, feel free to reopen if you would like to add the necessary tests.

PR #7259, opened 2018-12-12
Fix createContextualFragment declaration

skeggse:
Per the [DOM parsing spec][1], `Range#createContextualFragment` returns a `DocumentFragment`, and can't return other `Node`s.

[1]: https://w3c.github.io/DOM-Parsing/#extensions-to-the-range-interface


skeggse:
Duplicate of #7225.

PR #7262, opened 2018-12-12
Update types for WeakMap

nmn:
WeakMap doesn't currently accept Arrays as keys because Arrays are not subtypes of Objects in Flow's type system. This explicitly add support for Array keys to WeakMap.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -556,12 +556,12 @@ declare class Map<K, V> extends $ReadOnlyMap<K, V> {
     [key: $SymbolToStringTag | $SymbolSpecies]: Function;
 }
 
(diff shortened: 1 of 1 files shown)


benmosher:
ah sweet! I was about to file this but saw it was fixed in master via https://flow.org/try/.

thanks @nmn!

benmosher:
FWIW for any others until this ships: I ran into this using a `Tuple` type as my `WeakMap` key.

To work around, I found that doing a type union with `{}` 
```jsx
type MyTuple = {} & [ /* whatever */ ];
```
defuses the `WeakMap` error without damaging the type too much.

doasync:
How about a function as a key?

Issue #7128, opened 2018-10-31
enum indexer property not inferred properly


villesau:
[try flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1ASQHYAmWAHlgE6gC8oA5AEa0Dcy62oAgtaAN4A+AQwBcofAFcAtvQp8AvizY4AQt34A6DRwA0oANoBrLGhENaAXREBnFOQCW+AOag5LEKEWgVNdZp30rNvZOLqAg8LDk+pbIAMaw+NbuWNYiXrzCoADMAEwAjH4mAvQxtPKgQA)

This prevents writing certain libdefs, e.g this: [try flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAmBTAxtAhgJzaFtgM7GgBi2GALrLgJ4A8AKgDSgCCGGArgLY8ctXKAC8oAN4AfKQF8AfJMShQ2FCg7VquRgGkAXKAAkutPWIt57AGqGjAUWho+aAHbVm9AA5oW7XfLyABTKKqAA1maGuqyhKgBu2NA8aIbWoQCUhpQ0dExskgB0xVy8AkJ07ADaugC6aaAKANyhAOZo1ABKaMSC1EFZnNz8gtjCiLLIGLCuxNSgkFTC9GKgrmjwFEt5jNLYhq78AEZouOxHhnO4AJaurXLBGS2I07Pzi7kMAEyrH8uFag0WlwQQA5EdQexQcQ+NQABZ3ACEoKeLxmcwW2wYAGZflj6AD1JptGDsJDQNivgBGVGIECgIJ-PJfQrtLo9PoDQzSC6gK63VrsfZrY6nORPUD0niuaZ8FzuBZ0UCnXB0YhAA)

This has never worked. It's especially important when moving exact by default types since `A & B` would not work in that kind of cases anymore. Now the example is possible to make working with limitation that exact properties are not supported, and using `A & B` approach.

Would be very interesting to try to fix this by my self, but Flow codebase is a bit frightening. Any pointers what lines should I look at and would this be big task to fix?

wchargin:
This seems like a special case of indexers being totally broken on exact
objects. Changing `key: "b"` to `key: string` in your example raises the
same error.

The simplest repro is

```js
({a: 1}: {|[string]: number|});  // fails! :-(
```

@jbrown215: When objects become exact-by-default, this will be a
crippling pain point. Indexer properties will become basically unusable.
[...]

Issue #7240, opened 2018-12-07
exact objects with indexer properties fails to type check


villesau:
[try flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AeQEagC8oA3gD6ICQA1ANoDWWaAXKAM4oBOAlgHYBzALps+AVwC2+LF0TkAvsgDGsPp1Cx8bAsTIBDNgEYATAGZ5QA)

following code:

```
// @flow

type Ob = {|
	+[key: string]: number
|}

const ob: Ob = {a: 123}
```

fails. There are valid reasons to use exact indexer: You want to define that all the possible properties have certain type what ever the key is. If the `Ob` type would be inexact, that would mean that it's possible that the object has what ever in it. 

Actually having inexact type 

```
type Ob = {
	+[key: string]: number
}

```

type does not make much sense: What ever key you have should be number, but the object also might have some other properties? There is clear conflict in the logic: the type _needs_ to be something, but it also _might_ be something else.

SamChou19815:
Now it succeeds

Issue #4181, opened 2017-06-15
Debugging Flow


agentcooper:
I am using Visual Studio Code with [vscode-ocaml](https://github.com/hackwaly/vscode-ocaml) extension. `./bin/flow` is compiled using `make build-flow-debug` (see also: https://github.com/facebook/flow/issues/4177).

`.vscode/launch.json`:

```js
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "OCaml",
      "type": "ocamldebug",
      "request": "launch",
      "program": "${workspaceRoot}/bin/flow",
      "console": "internalConsole",
      "stopOnEntry": false,
      "arguments": [
        "${workspaceRoot}/local-test"
      ]
    }
  ]
}
```

`local-test/test.js`:

```js
// @flow

const a: number = 'z' + 'd';
```

Debugging `main` seems to work:

<img width="794" alt="screen shot 2017-06-15 at 15 28 09" src="https://user-images.githubusercontent.com/794591/27183582-4f28422c-51df-11e7-9a48-af9d5b69d253.png">

However If I set a breakpoint somewhere in `src/typings/`, for example `statement.ml`, inside `variable_decl` function, then the breakpoint is skipped.

I suspect this is because of the server-client architecture - actual type checking happens inside a different process.

Is there any good way around this?

kogai:
I faced the same problem and I didn't solve it.

For the reference, I tried `flow check` command.
It seems to be checking type at foreground, but when using `ocamlbuild` with settings like below,
debugger responds with `cannot connect to debugger at 127.0.0.1 error: No such file or directory`.

```javascript
{
    "program": "${workspaceRoot}/bin/flow",
    "arguments": ["check", "path/to/mycode.js"],
}
```

The error occurred at `spawn` function in `hack/utils/daemon.ml`(yes, the debugger worked correctory), and I guess that it related to spawning server process.

kogai:
Still examining, but it seems to get some progress.

As above comment I guess, the problem caused by some process that created behind `flow`.

Then, I try out arguments "--max-workers" that described with `Maximum number of workers to create (capped by number of cores)`

I don't understand it in detail, but it seems related to some process behind `flow`.
So, I modified my settings like follows and executed `ocamldebug`.
 
```
{
    "program": "${workspaceRoot}/bin/flow",
    "arguments": ["check", "--max-workers", "0", "path/to/mycode.js"],
}
```

And the result looks as I expected.
[...]

Issue #7210, opened 2018-11-27
 Cannot spread same exact immutable types: null or undefined [1] is incompatible with string [2].


villesau:
[Try Flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMVrKALygDeAPqANSQEBcoA-AM4oBOAlgHYDmo5AXwDciAMawurUCixSSACgBGAQzYN8sAJTqCxAHyh5pULVgMA5KfMAaUADoHKtgM1CgA)

```
// @flow

type Foo = {| +foo: ?string |};
const test = (bar: Foo): Foo => ({ foo: 'foo', ...bar});
```

causes:

```
3: type Foo = {| +foo: ?string |};
                       ^ null or undefined [1] is incompatible with string [2].
References:
3: type Foo = {| +foo: ?string |};
                       ^ [1]
4: const test = (bar: Foo): Foo => ({ foo: 'foo', ...bar});
                                           ^ [2]
```

This is very strange and frustrating issue, I think there is no any scenario that would make the error valid.

The error location is also very odd: It's in the _type_, not in the usage. this makes suppressing the error pretty non-optimal.

If I don't spread, the problem disappears.

This happens with all Flow versions.

[Happens also without exactness or read-onlyness](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMVrKALygDeokBAXKAPwDOKATgJYB2A5qAL4DciAY1htGoFFlEkAFACMAhkxr5YASiUFiAPlBTylWDQDk+wwBpQAOivym3FbyA)
[possibly undefined version works](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMVrKALygDeAPqANSQED8AXKAM4oBOAlgHYDmo5AXwDciAMawurUCixSSACgBGAQzZN8sAJTqCxAHyh5pULVhMA5KfMAaUADoHKtgM1CgA) Unfortunately this is not what our type really is :/

Gisleburt:
I'm suffering the same problem. This appears to have been a long standing bug in flow. See #5243

nmote:
This no longer errors on master

PR #7271, opened 2018-12-16
Fix exact object indexer issue 7128 and 7240

villesau:
This fixes https://github.com/facebook/flow/issues/7128 and https://github.com/facebook/flow/issues/7240

Root cause for both issues were same.

In overall I think that singleton string indexer properties should be treated as a regular keys from the get go but that's probably bigger change. The change introduced in this PR is probably fine for now as nothing broke, and test harness around exact indexer properties is now slightly better.

Having debugger setup in place helps _a lot_: https://github.com/facebook/flow/issues/4181#issuecomment-447596374

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -6741,7 +6741,7 @@ and flow_obj_to_obj cx trace ~use_op (lreason, l_obj) (ureason, u_obj) =
         (Field (None, lv, lpolarity), Field (None, uv, upolarity))
     | _ -> ());
 
-  if rflags.exact && rflags.sealed = Sealed && not (is_literal_object_reason ureason)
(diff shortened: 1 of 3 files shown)


villesau:
This actually also fixes this error: https://github.com/facebook/flow/issues/7210

jedwards1211:
Is there any reason this didn't get merged?  Would be nice 😄

villesau:
@jedwards1211 I had totally forgotten this PR :D  I think here is the most comprehensive discussion about the topic: https://github.com/facebook/flow/issues/7128

Maybe @jbrown215 could comment on this?

SamChou19815:
Closing this PR since all mentioned issues have been fixed. Most recently, https://github.com/facebook/flow/issues/7128 is addressed with newly introduced mapped type.

PR #7285, opened 2018-12-19
Upgrade to bootstrap 4.0.0

mroch:
Pretty significant changes from 4.0.0.alpha.5 to the 4.0.0.beta's to 4.0.0 final, and then even more to get up to latest 4.1.

I had a hard time figuring out what changed since it was using undocumented alpha features, so had to dig through bootstrap commits. This PR goes from alpha5 to 4.0.0, taking it in small steps so if something breaks we can bisect more easily.

### DIFF
--- a/website/_assets/css/_article.scss
+++ b/website/_assets/css/_article.scss
@@ -25,8 +25,8 @@
     }
 
     em code {
-      background: $code-bg;
-      color: $code-color;
+      background: $gray-200;
+      color: #bd4147;
     }
   }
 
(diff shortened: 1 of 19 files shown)


MarwaAbdallah:
Hi @mroch. Would  you know if there is a plan to upgrade bootstrap to a non alpha version of 4.x major ? Thank you.

SamChou19815:
I'm closing this since we will soon migrate to Docusaurus.

Issue #7242, opened 2018-12-07
readonly object spread fails: property `bar` is read-only in object type [1] but writable in object literal [2].


villesau:
[Try flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQiCmAPADrATgF1LgJ4YqgDyARqALygDeAPoqKANSSywBc9DbFAQ2w8AzrmwBLAHYBzUAwC+iRQG5k6LHlABjWFLGgArhgAmA3CgBiElNBMBhPZAlzaACmNmLPSgEoaAHyIAJBudKAc3PSggsKgAOQAbgLQEibxoAoANKAAdPme5qQKvkA)

Following code: 

```
// @flow

export type Ob = {|
  +foo: {| +bar: string |}
|};

export const updateFieldConfig = (update: Ob) =>
	({ foo: { bar: 'valid' }, ...update })
```

results following error:

```
4:   +foo: {| +bar: string |}           ^ property `bar` is read-only in object type [1] but writable in object literal [2].
References:
4:   +foo: {| +bar: string |}           ^ [1]
8: 	({ foo: { bar: 'valid' }, ...update })
            ^ [2]
```

As seen from the snippet, there is no writing happening. The nested object is what's failing. The error location is also very awkward: It's in the type, not in the usage which makes suppressing very inconvenient.

There has been similar bug that's been already fixed in 0.83.0 version, but most likely very related and would lead to proper fix in this case too: https://github.com/facebook/flow/issues/6089

nmote:
This no longer errors in master

Issue #6751, opened 2018-08-17
Problem with object spread syntax where the object type has optional values


jonleighton:
[Example at Try Flow](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCYAvGAN5hRxwBcYA-AM4YBOAlgHYDmYAvgG50UAK6cAxhnZxOVGgAoAhvUIBKCqj7pxMlmGVFSFOXTAByanDP8hOznoBGRypfoWa1wcIWUAdP8UAGhM3S09VATBgYDBOODA8VlY4VlRLeRcaUI9g-19FfgiomMTk1LSfMDyHYNdzMMLI6Nj40pSKuAyQ+pyq-yc+Iua4hKT2oA)

The following code:

``` js
/* @flow */

type A = { foo: ?string };

function foo(a: A) {
}

const a: A = { foo: 'foo' };
const b = { foo: 'foo' };

foo({ ...a, foo: 'foo' }); // no error
foo({ foo: 'foo', ...a }); // error

foo({ ...b, foo: 'foo' }); // no error
foo({ foo: 'foo', ...b }); // no error
```

Produces this error:

``` js
    3: type A = { foo: ?string };
                       ^ null or undefined [1] is incompatible with string [2].
        References:
        3: type A = { foo: ?string };
                           ^ [1]
        12: foo({ foo: 'foo', ...a }); // error
                       ^ [2]
```

It also produces an error if I define the type as:

``` js
type A = { foo: string | null | void };
```

In that case I get:

``` js
    3: type A = { foo: string | null | void };
                                ^ null [1] is incompatible with string [2].
        References:
        3: type A = { foo: string | null | void };
                                    ^ [1]
        12: foo({ foo: 'foo', ...a }); // error
                       ^ [2]
```

jonleighton:
Here's a workaround:

``` js
foo({ ...null, foo: 'foo', ...a });
```

richardbarrell-calvium:
> Here's a workaround:
> 
> ```js
> foo({ ...null, foo: 'foo', ...a });
> ```

Thank you so much for this, you've saved me from having to put a heinous `: any` in my codebase.

levibuzolic:
Seems like it might be a bug with the type refinement process, it's like the type is being refined too early and then being compared against.

When doing:
```js
type A = { foo: ?string };
const a: A = { foo: 'foo' };
const bar = { foo: 'foo', ...a };
```

[...]

Issue #6108, opened 2018-04-06
Cannot override values when spreading


villesau:
[Try flow link](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCYAvGAN6phgCGAXGAHYCuAtgEZ4BOqAvgNzps+MACFSFKrQYBnDFwCWjAOZ9BqAMZxGcsBjxyAYowkAKJTmYYGhAJQNxJAHxhT5AHSeLVgDTSwAOTMABZgcKF0wAG8tvxAA)

```
type A = {
  a: number
};

type B = {
  a: string
};

const testFn = (input: A): B => ({...input, a: 'uh oh :/'});
```

As you can see, the property `a` is overridden with string in the function. Both, input and output are type checked, but flow complains that output and input are not compatible. This is very basic and idiomatic ES6 behaviour and should be fully supported.

vicapow:
You can override values when spreading it's just that when the two types differ, flow infers the type to be an union of the two. This seems like reasonable behavior to me.

````js
const a: { a: (number | string) } = {...{a: 10}, a: 'no erro/'};
```

villesau:
I think it's clear from the code that the result type of `a` can't be number anymore and flow should be able to infer it as string. It just can't be number after spreading anymore.

theres:
Imho this should work. If you need to replace a few of fields out multiple, then you can use rest operator as a _good engough_  [workaround](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCYAvGAN6phgCGAXGAHYCuAtgEZ4BOANFWOwYBnDFwCWjAOZ9qAYwbs4cGHhqNUAXwDc6bPjAAhUhX70wI8VJkDhoidP7yBSlWs07UsuIxFgMeEQAxRmMACnIaHjAAOliuAIwNBkIASgYjEgA+MHDY6PiRKLMAcgQ4LgBrIQBCYo0UrSA):

```js
type A = {
  a: number,
  b: string,
  c: boolean
};

type B = {
  a: string,
  b: string,
  c: boolean
};

const testFn = ({a, ...rest}: A): B => ({...rest, a: 'works!'});
```

wchargin:
[...]

Issue #5243, opened 2017-11-02
Spread Operator isn't working correctly


MichaelDeBoey:
[...]

PR #7298, opened 2018-12-26
Fix object spreading, issues 6108, 7242, 5243, 7210 and 6751

villesau:
This fixes bunch of issues with object spreading. Now the value is replaced, when before flow made union type out of it, which is not how it should be. In most of the cases Context.set_prop is enough. The old logic is needed for 

This fixes at least following issues, likely even more that I didn't find:
https://github.com/facebook/flow/issues/6108
https://github.com/facebook/flow/issues/6751
https://github.com/facebook/flow/issues/7242
https://github.com/facebook/flow/issues/7210

and potentially https://github.com/facebook/flow/issues/5243

The tests follows this naming convention: https://github.com/facebook/flow/blob/master/tests/spread/issue_4152.js

The react_jsx tests had actually wrong expectations, which I changed as well.


src/typing/flow_js.ml
```diff
@@ -4870,14 +4870,16 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
-          let propref = Named (reason_prop, x) in
-          let t = filter_optional cx ~trace reason_prop t in
-          rec_flow cx trace (to_obj, SetPropT (
-            unknown_use, reason_prop, propref, Normal, t, None
-          ))
-        | None ->
+        match Property.read_t p, to_obj with
+        | Some _, (DefT (_, ObjT ob)) when not ob.flags.frozen ->
+          Context.set_prop cx ob.props_tmap x p
+        | Some t, to_obj ->
```
villesau: this branch was still needed for these to work: https://github.com/facebook/flow/blob/master/tests/destructuring/destructuring.js#L61
https://github.com/facebook/flow/blob/master/tests/object_assign/non_objects.js

tests/friendly_errors/issue_7242.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+type Ob = {|+foo: {| +bar: string |}|};
+declare var update: Ob;
+({ foo: { bar: 'valid' }, ...update }: Ob); // Ok
+
+type NestedOb = {|+foo: {| +bar: {| +foo: number |} |}|};
+declare var update2: NestedOb;
+({ foo: { bar: { foo: 123 } }, ...update2 }: NestedOb); // Ok
```
villesau: not sure if firendly_errors folder is correct folder for these, but https://github.com/facebook/flow/blob/master/tests/friendly_errors/prop-variance.js is perhaps the closest to above tests that I found.

src/typing/flow_js.ml
villesau: https://github.com/facebook/flow/blob/master/tests/destructuring/destructuring.js#L61 does not work without `| Some t, to_obj ->` branch because [this](https://github.com/facebook/flow/blob/99e058a7d3f2fc6f816ee9d8616e0814a91932cb/src/typing/flow_js.ml#L4987) generates OpenT instead of ObjT. Otherwise `| Some t, to_obj ->` branch could be unnecessary altogether.

gajus:
Ping @gabelevi @avikchaudhuri

Is there anything else that needs to be done before merging this?

vicapow:
Wow! Amazing work, @villesau. Thank you for your contribution. Hopefully this gets merged soon.

villesau:
Perhaps @jbrown215 or @samwgoldman could comment on this?

avikchaudhuri:
Thanks a bunch! :) We have to pull this and investigate its effect on FB codebases before merging. We'll let you know soon.

villesau:
@avikchaudhuri any progress with this?

samwgoldman:
Hey @villesau -- so sorry for the delay here. Please understand that this time of year is a bit busy for the team, as we have our twice-yearly reviews and planning meetings which consume a lot of time. Thank you for your patience.

And thank you for your extensive research into this problem! I will take a close look at this.

The use of Context.set_prop does jump out at me as a potential concern here. This function performs a destructive operation -- it mutates the property map associated with an object -- which is an unsafe operation in the context of the type checker in general.

Why is it unsafe? Well, the type checker works by adding constraints into a big graph, then solving those constraints. A constraint is either trivially true or false (`string <: number` is false, `empty <: mixed` is true) or decomposes into simpler constraints (`A => B <: C => D` is true if `C <: A` and `B <: D`).

These constraints are "evaluated" eagerly as they are seen, but the order they are seen is roughly the source order -- which is not the same as program order. As a design point, any logic that deals with constraints needs to be "monotonic." In other words, the result of type inference / checking should not be sensitive to the order that constraints are evaluated.

A destructive operation like Context.set_prop, can thwart monotonicity, so it needs to be applied carefully. Unsealed objects use it, but are still monotonic, for example (see the ShadowWrite case of LookupT).

It is possible that the change here does preserve monotonicity, but it will require some careful consideration to prove. Which, unfortunately, means that I need to ask for your continued patience here.

As a first step, I will import this change and see what effect it has on our internal build infrastructure. Will keep you updated. Thanks again!

villesau:
@samwgoldman thanks for the thorough answer! 

> The use of Context.set_prop does jump out at me as a potential concern here. This function performs a destructive operation -- it mutates the property map associated with an object -- which is an unsafe operation in the context of the type checker in general.

What would be the alternative? replacing the whole object map instead? I wanted to bypass `SetPropT` since I think it has way too much too fragile and unrelated logic in case of object spreading. There might be some good middle ground though? 

What I found out, is that spreading creates completely new object map reference, which I _think_ makes Context.set_prop safer. But my understanding of OCaml and Flow is too limited to be sure about that.

Other approach could be to find the place in `SetPropT` where extending the type is done, and change that to replacement so that `{...{a: 123}, a: string}` does not become `{ a: number | string}` but `{ a: string}` instead.

I can also write thorough tests for object spreading, but I didn't want to invest more time on this before you guys consider the change. I went trough all the issues linked in the PR and validated the fix I made against the provided repro steps.

Please note that I didn't solve _everything_ related to spreading in this PR: everything that does not match `| Some _, (DefT (_, ObjT ob)) when not ob.flags.frozen` still falls back to old logic. I found two such cases: frozen objects and certain ways of spreading classes. Frozen object works well with old logic, but spreading classes might still be buggy, although I find it so minor thing that I wouldn't concentrate fixing that in this PR.

samwgoldman:
> What would be the alternative? replacing the whole object map instead? I wanted to bypass SetPropT since I think it has way too much too fragile and unrelated logic in case of object spreading. There might be some good middle ground though?

Any destructive operation, including modifying the context itself, needs to be monotonic. And to be super clear, this change may very well be correct! There are lots of destructive operations in the type checker, but we have also (informally, mostly) proven them to be monotonic. That proof needs to happen here as well.

samwgoldman:
(And since proving these properties does require deep knowledge of the type checker, I am going to prove it to myself, then probably @jbrown215 and @avikchaudhuri.)

villesau:
@samwgoldman any progress or findings yet? :)

Skurkify:
@villesau Thanks for putting this change set together. Hopefully it's receives some attention from contributors soon. We have around 2600 types across 2200 flowed files, and I'm already making a list of things I can clean up once this is merged. 👏

villesau:
I actually found an issue with this approach. The original behaviour is ~correct if the overriding value is optional, see:

```
// @flow

type A = { a: number };
type B = { a: string };
type C = { a?: string };
declare var inputA: A;
declare var inputB: B;
declare var inputC: C;
({...inputA, a: 123}: A); // correct
({...inputA, a: 'text'}: B); // correct
({...inputA, ...inputB}: B); // correct
({...inputA, ...inputC}: C); // wrong but succeeds. result should be { a: number | string }
```

Anyhow, the last assertion succeeds when it should fail. I _think_ that it's fairly easy to fix though, but would be very nice to get feedback before continuing with this. This is still already closer to correct behaviour than current Flow behaviour.

FezVrasta:
@villesau may you take a quick look at [this repro](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgAKATnDgM5gC8YA3qmGAgJYAmGAFgPwBcYAdgFcAtgCM8xMAB8w5DMWb8A5gBpUAXwDc6KIP4BjDMzj8wu-gAp6jFuw7UwAcgDMABjzDHa9XxJlyAJR0DGDEeBiCxKa2nNrqQA) and tell me if your PR is going to fix this behavior also?

Thanks for the great PR!

villesau:
@FezVrasta That's unfortunately different issue.

peter-leonov:
Looking forward to fix hundreds of new errors once this gets merged! 😌

gsklee:
@facebook Could the community get some update on this critical bug? We were promised a really promising, communicative 2019 and got told this issue will be code reviewed soon, but then it's another month of total silence.

It actually makes me feel pretty bad that we have to do something a bit pushy like this to get some progress update, kinda creates a tension between the community and the maintainers. Instead of trying to recruit more devs, I'd say it might actually be more valuable in finding a competent project manager/owner that can help kicking start a healthy communication loop.

villesau:
This PR is in a bit limbo state at the moment I believe. What I've understood, Flow team is working on totally refactored object model which should also fix spreading issues as well as many other problems with current way of handling objects in Flow. For this reason I don't want to put more effort in this at the moment. Maybe @jbrown215 can shed some light on where the refactoring is going?

FezVrasta:
@villesau thanks for the update!

May I ask how did you get these info? I would have expected the Flow team to communicate here, after the Medium post they published about the 2019 promises.

Thanks!

jbrown215:
@gsklee @FezVrasta:

> May I ask how did you get these info?

We've been in touch with @villesau on discord. We actually interact with him almost daily there to talk about a variety of issues, and in general we've found the discord to be a source of very good discussions. If you want the more frequent and less formal updates from the Flow team, I recommend joining the discord: https://discord.gg/8ezwRUK 

> Could the community get some update on this critical bug?

Yes. We are in the middle of a large refactor of our object model. @samwgoldman or I will publish a larger blog post that covers the necessity of the changes we're making. The reason we haven't moved forward on this PR is because of a significant performance regression and because we are still unsure about the correctness.

Fixing object spreads is also my current project (and the vast majority of my day-to-day work). There are a _lot_ more issues than just this one, and I'm aiming to fix all of them. I've been making good progress but I don't want to give an estimate for when it will be released because there are still a few roadblocks that I'm not sure how to get past.

frontendphil:
@jbrown215 I'm sure there is more to it but I'd still like to point out that releasing some fixes usually is better than releasing no fixes ;) I'm saying this without knowing anything about the implications but sometimes I also need to be reminded of the simple fact that working in small iterations generates more value for end-user than releasing things in large batches.

jbrown215:
> releasing some fixes usually is better than releasing no fixes

I agree! We just aren't positive this is a fix that doesn't introduce other issues, but we also know it introduces a performance issue.

In particular, this fix relies on the constraints flow generates being evaluated in a certain order. There are some cases where that invariant may hold, but it's something we usually assume is false unless a convincing argument is given.

That could lead to something like `const x: {foo: number} {...{foo: 'string'}, ...{foo: 3}}` to be a type error, which it shouldn't be.

pkyeck:
So is the buggy object-spreading the reason why this isn't working? Or am I doing something else wrong?

```javascript
type A = { a: string };
type B = A & { b: string };
type C = A & { b: number };

const testB: B = { a: 'test', b: 'test b' };

function test(obj: B): C {
 return {
   ...obj,
   b: 23,
 }; 
}

test(testB);
```

```
    8:  return {
                ^ Cannot return object literal because string [1] is incompatible with number [2] in property `b`.
        References:
        4: type B = A & { b: string };
                             ^ [1]
        5: type C = A & { b: number };
                             ^ [2]
```

kangax:
Pretty sure this is it. It also doesn't work with type spreads (which are generally more "predictable" than `&`'s):

https://flow.org/try/#0C4TwDgpgBAglC8UDeAfKBDAXFAzsATgJYB2A5lCgL4DcAUKJFAEILJoB0nMANFAEbY8RMhRr1w0AMKtUUTux79sxAK4BbPhHyi6tAMYB7YnijAIeJthaIkGbAHIzee7wFRH54P3tQxtAGYqxHrAhEamngAUBnwAVlYAlNjSSLRQ+BDAKvjEyGlQcpwxsdz5SlAATADMpb7UULSUtPRRTsBMCdRAA

amccloud:
@FezVrasta 👋 heya, did you happen to find the right issue for 

https://flow.org/try/#0C4TwDgpgBAglC8UDeVgQM7AFxUwJwEsA7AcygB8oiBXAWwCMI8oBfAbgCgOBjAeyMxQAhjjiIkaTDgBEARmnse-QUIDqBYAAsA8gDcmhACYRRCZJOxRpAJmkAaKADpnQxUA

or 

https://github.com/facebook/flow/pull/7298#issuecomment-458562888

Seems like the type if conflicting with the inferred type of the object? I'm also just trying to provide default options for an argument

Maggie199:
Any updates here?

NamPNQ:
Relate issue
```js
const obj = { ...{ hi: () => 1 }, ...{ hi: (x: number) => 'hi' } };
obj.hi(5);
```
https://flow.org/try/#0MYewdgzgLgBCBGArGBeGBvGA6HmAWAlgFwwAUAlKgHwwCMMAvgDTa4yEmkAeJYArgFt4AUwBOlFDQDkhKY0YBuAFAJEWQqQCs5BUA

nmote:
This week, @jbrown215 landed some major fixes to spread which will come out in Flow v0.111.0, scheduled for release next week. I haven't checked every associated issue to make sure that they were all fixed, but these changes address most (or all?) of the known issues with spread.

Jordan is working on a blog post with more details.

I think that Jordan's changes accomplish the goals of this PR, so I'm going to close it. @villesau, please let us know if that's not the case.

nmote:
Update: I went through the list of associated issues and @jbrown215's recent changes address all but one of them.

PR #7339, opened 2019-01-10
Fix docs for checking all files

aryzing:
I believe the `flow check --all` no longer works, so I've updated the docs to suggest configuring the `all` option Instead.


nmote:
I just tried it and `flow check --all` works just fine for me. Let me know if I have missed something.

aryzing:
My bad, works as expected 👍 . What's the best place to document the --all flag? It's not mentioned in the [CLI docs](https://flow.org/en/docs/cli/) page.

PR #7358, opened 2019-01-15
Add oncontextmenu to HTMLElement

### LINKED ISSUE #7357: oncontextmenu is misisng from HTMLButtonElement
I'm seeing this error: 

> Cannot assign onContextMenu to bp.oncontextmenu because property oncontextmenu is missing in HTMLButtonElement [1].

which is surprising to me because `HTMLButtonElement` has other properties like `onblur` and `onclick`

jasonLaster:
Fixes #7357

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1570,6 +1570,7 @@ declare class HTMLElement extends Element {
   oncanplaythrough: ?Function;
   onchange: ?Function;
   onclick: ?Function;
+  oncontextmenu: ?Function;
   oncuechange: ?Function;
   ondblclick: ?Function;
   ondurationchange: ?Function;
--- a/tests/bom/bom.exp
+++ b/tests/bom/bom.exp
@@ -22,8 +22,8 @@ with `HTMLFormElement` [2].
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 References:
-   <BUILTINS>/dom.js:669:36
-   669|   createElement(tagName: 'input'): HTMLInputElement;
+   <BUILTINS>/dom.js:667:36
+   667|   createElement(tagName: 'input'): HTMLInputElement;
                                            ^^^^^^^^^^^^^^^^ [1]
    <BUILTINS>/bom.js:330:24
    330|     constructor(form?: HTMLFormElement): void;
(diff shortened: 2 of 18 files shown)


jasonLaster:
Out of curiosity, is it documented that the flow [repl] doesn't work with `document`?

[repl]:https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAodUCuA7AxgFwEs5swo44AKASjAG9UwxcSBnfMAUxjAF4wAJnFyYAtp2z4AdLgBOnAIb5OAURidxkygCIBhAG7bqjLjCklcMQrgDWfMDT4A+egF8T3c3hLKAHvk1Me0deFzp3VyA

jbrown215:
The flow repl doesn't load all of the libdefs, so it's possible that the one where document is defined is not loaded.

You may need to re-record tests here. I use `./runtests.sh` at the root of the flow repo to do that.

nmote:
Could you link supporting docs?

jasonLaster:
I'm getting tripped up on the getting started guide

I was seeing this

```
➜  flow opam upgrade
[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
```

which was fixed by doing


```
opam switch create flow 4.05.0
```

now `make` works, but `make js` does not and `./runtests.sh` fails...

jbrown215:
What's the error message you get with `make js`?

jasonLaster:
https://gist.github.com/be1245924d4302f28118e902e3836589


runtests.sh fails with this

```
      10
      11 Stderr:
      12 ./runtests.sh: line 493: /Users/jlaster/src/_os/flow/: is a directory
      13 flow check return code: 126
      14 [✗] ERRORED: annot2
      15
```

nmote:
@jasonLaster You have to pass the location of the built Flow binary as the argument to `./runtests.sh`... it should fail with a better error message though if you don't.

nmote:
Anyway, I'm willing to take care of re-recording the tests, although it's certainly simpler if you are up for it. The main thing I need here is a link to supporting docs so I can verify that this change is correct.

jasonLaster:
here are some docs
https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oncontextmenu

jasonLaster:
@nmote thanks i was able to run `./runtests.sh -r -b bin/flow`. It looks like there are still some failing tests. I'd love a bit of help to see what else i should do...

nmote:
Thanks, I'll take a look

nmote:
I ran the tests and got a much smaller and more reasonable test diff. I'm not sure what went wrong, but I'll get it committed with the updated test output.

jasonLaster:
weird. Anyways, thanks for the help.

As a sidenote, it would be nice to have some improvements to the docs on using `opam switch` and running tests

nmote:
Understood. If you have specific suggestions, feel free to tag me in a PR and I will make sure to take a look.

PR #7369, opened 2019-01-17
Fix MouseEvent type definitions

### LINKED ISSUE #7354: relatedTarget property in the MouseEvent constructor's parameter is badly typed
See https://github.com/facebook/flow/blob/f7e2adc834f02f29041b6a6c8e895f3a5a1dcd2c/lib/dom.js#L299 compared to https://github.com/facebook/flow/blob/f7e2adc834f02f29041b6a6c8e895f3a5a1dcd2c/lib/dom.js#L324.

jamesisaac:
Fixes a few inaccuracies in the `MouseEvent` typedefs:

- Correct `relatedTarget` in constructor ([docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/MouseEvent#Values))
- Replace unnecessary Maybe type with `| null` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/region))
- Add missing `x` and `y` properties ([docs](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/x))

Fixes #7354

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -296,7 +296,7 @@ type MouseEvent$MouseEventInit = {
   button?: number,
   buttons?: number,
   region?: string | null,
-  relatedTarget?: string | null,
+  relatedTarget?: EventTarget | null,
 };
 
 declare class MouseEvent extends UIEvent {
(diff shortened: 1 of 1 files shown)


jbrown215:
If I don't get to this today, I'll be able to get to it next week.

jbrown215:
looks good. just need to rerecord the tests and it'll land. you can expect to see this in 0.93 at the latest, 0.92 if we get lucky with internal reviews :)

PR #7381, opened 2019-01-20
Add support to Server-Sent Events (EventSource)

SlIdE42:
Add support to Server-Sent Events (EventSource) : 
- https://www.w3.org/TR/eventsource/
- https://developer.mozilla.org/en-US/docs/Web/API/EventSource
- https://html.spec.whatwg.org/multipage/server-sent-events.html

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -146,6 +146,8 @@ type AnimationEventHandler = (event: AnimationEvent) => mixed
 type AnimationEventListener = {handleEvent: AnimationEventHandler} | AnimationEventHandler
 type ClipboardEventHandler = (event: ClipboardEvent) => mixed
 type ClipboardEventListener = {handleEvent: ClipboardEventHandler} | ClipboardEventHandler
+type MessageEventHandler = (event: MessageEvent) => mixed
+type MessageEventListener = {handleEvent: MessageEventHandler} | MessageEventHandler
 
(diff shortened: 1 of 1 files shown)


lib/dom.js
```diff
@@ -158,6 +160,7 @@ type DragEventTypes = 'drag' | 'dragend' | 'dragenter' | 'dragexit' | 'dragleave
 type PointerEventTypes = 'pointerover' | 'pointerenter' | 'pointerdown' | 'pointermove' | 'pointerup' | 'pointercancel' | 'pointerout' | 'pointerleave' | 'gotpointercapture' | 'lostpointercapture';
 type AnimationEventTypes = 'animationstart' | 'animationend' | 'animationiteration';
 type ClipboardEventTypes = 'clipboardchange' | 'cut' | 'copy' | 'paste';
+type MessageEventTypes = string;
```
jbrown215: Is there a reason this is just `string` and not a union of string literals like the rest? Are the MessageEvents not defined here: https://developer.mozilla.org/en-US/docs/Web/Events ?

SlIdE42: Browser is allowed to listen to specific events choosed by developer :

```JavaScript
evtSource.addEventListener("ping", function(e) {
  var newElement = document.createElement("li");
  
  var obj = JSON.parse(e.data);
  newElement.innerHTML = "ping at " + obj.time;
  eventList.appendChild(newElement);
}, false);
```

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

jbrown215: Thanks!

PR #7386, opened 2019-01-23
fix(BOM): XDR

Mouvedia:
- the callbacks don't have a parameter
- `addEventListener` wasn't supported
- 2 properties are readonly
- only two verbs were supported


jbrown215:
Can you link to relevant documentation to make it easier to review?

Mouvedia:
@jbrown215 Iv tested myself on Sauce Labs or Windows directly while making [cb-fetch](https://github.com/Mouvedia/cb-fetch).

I have also corrected documentation before mdn/browser-compat-data#2169

PR #7403, opened 2019-01-28
docs: add `declarations` page to config

pascalduez:
Hi,

`[declarations]` is missing in the page config. Hence no TOC, no title. 
https://flow.org/en/docs/config/declarations

### DIFF
--- a/website/_data/guides.yml
+++ b/website/_data/guides.yml
@@ -86,6 +86,7 @@
 - id: config
   pages:
     - path: "/docs/config/"
+    - path: "/docs/config/declarations/"
     - path: "/docs/config/include/"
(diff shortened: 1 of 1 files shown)


jbrown215:
Did you manage to build the site? I've not been able to build it by following our OSS instructions and I want that to be fixed. If you figured out how to do that, I'd appreciate it if you could give me pointers :)

pascalduez:
> Did you manage to build the site? 

I tried, but that was a mess, I had to install a bunch of ruby things, then error after error, install after install, I gave up :)

mroch:
I think this is right. fingers crossed.

@jbrown215, i'd say we should open a separate issue to fix/improve the website install.

PR #7420, opened 2019-01-31
[react] New useReducer signature

threepointone:
React is updating the signature for the `useReducer` hook https://github.com/facebook/react/pull/14723 This PR reflects those changes. We'll also update the official react docs soon.

I'm doing a build and tests right now locally (installing everything from scratch atm), but figured I'd start the PR as well.


threepointone:
(updating tests, will update PR soon)

tests/react/useReducer_hook.js
```diff
@@ -32,12 +32,13 @@ const initialState = { count: 0 };
...
-    payload: 123
-  };
+  function init(arg){
+    return {
+      count: arg.payload
+    }
+  }  
 
-  const [state, dispatch] = React.useReducer(reducer, initialState, initialAction);
+  const [state, dispatch] = React.useReducer(reducer, { type: "reset", payload: 123 }, init);
```
acdlite: Not super important, but why does the init arg in this example have a type and payload? Implies that it's the same type as the action, but that's not necessarily true.

threepointone: No reason, just translated what was already there. I’ll change it. I specifically wanted it to *not* have the signature of the state, but I should have gone with something totally different. Thx!

lib/react.js
```diff
@@ -316,11 +316,24 @@ declare module react {
...
-  ): [S, A => void];
+  ): [S, Dispatch<A>];
+
+  declare export function useReducer<S, A>(
+    reducer: (S, A) => S,
+    initialState: S,
+    init: void,
+  ): [S, Dispatch<A>];
+
+  declare export function useReducer<S, I, A>(
```
sebmarkbage: Is this the best ordering of the generics for when you specify them explicitly or would `<S, A, I>` be better so that the action type line up in the same place? Is there any practical difference?

threepointone: I don't think there's a practical difference, happy to align it. If I squint, I could make the argument that an initialArg would always 'occur' in source before any dispatch calls, so it's chronologically sound?

sebmarkbage: It seems like the TypeScript defs need S and A to line up in a separate type for some obscure reason and therefore adds `I` at the end. So switching it would line up with TS.

threepointone: done and done

lib/react.js
```diff
@@ -316,11 +316,24 @@ declare module react {
...
     reducer: (S, A) => S,
     initialState: S,
-    initialAction: ?A,
-  ): [S, A => void];
+  ): [S, Dispatch<A>];
+
+  declare export function useReducer<S, A>(
+    reducer: (S, A) => S,
+    initialState: S,
+    init: void,
```
nmn: How is this type overload different from the previous one?

jbrown215:
@bvaughn: That or stacking them will work! In either case, they'll need to be re-recorded.

Will review these tomorrow

Issue #7421, opened 2019-01-31
Object.fromEntries type defs missing


dnalborczyk:
## Missing APIs
Object.fromEntries

## Relevant documentation
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

https://github.com/tc39/proposal-object-from-entries

related: https://github.com/facebook/flow/issues/7183

`stage 4` now, and therefore added to the 2019 spec.

vicapow:
Thanks for reporting! I think it makes sense to support it now if it's stage 4 although adoption still seems missing in Chrome and Node 10.

I believe https://github.com/facebook/flow/blob/master/lib/core.js#L38 is the relevant code if you wanted to try and tackle this.

PinkaminaDianePie:
I doubt that it's possible to implement it with current flow limitations. Here is the example:
https://flow.org/try/#0C4TwDgpgBMD2AKAnWYA8BpANFAagPigF4oBvAbXQF0AuXAXwG4AoAEwgGMAbAQ0WgDMArgDt2wAJaxhUfsgC2AZXHCA5pwgBRYcEQgM2fAAoKBygEpacJCn248TKFCbspAZ2AzYsIjPlLV6lo6IMYA5Nyh2ACM5kz8XgB0AEY+AEzMjs5uHkm8PrKwispqmtq6hmER0TRQZOGRUMKCckkQiOZmTLmIyWkMQA

check the difference between `foo` and `bar`:
`foo` is `{[string]: number}`
`bar`:  `{['a']: number}`
there is no way to ask flow treat the first element in the tuple as a literal, not as a `string`
and if you will somehow avoid this limitation, you will fall down into another problem - since property will be computed, instead of having result type `{a: A, b: B}` you will have `{['a' | 'b'] : A | B}`. 
such libded require some improvements of flow itself, otherwise it will be just something like 
`([string, any][]) => any`

vicapow:
I'm guessing given the dynamic nature of `fromEntries`, it will likely mostly be used in the context of objects as maps, not records. With that in mind, I think something like the following would be useful:

````js

declare function fromEntries<K:string, T>(entries: Array<[K, T]>): {[K]: ?T};

// fromEntries(['foo', 'bar']); // error.
fromEntries([['foo', 'bar']]); // no error.
// fromEntries([[10, 'bar']]); // errors
// fromEntries([['bar']]); // errors
[...]

PR #7425, opened 2019-02-01
Add `Object.fromEntries`

vicapow:
Note that this doesn't yet seem to have wide adoption outside of FireFox.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries


vicapow:
fixes #7421

TrySound:
@vicapow Open PR from fork

dnalborczyk:
> Note that this doesn't yet seem to have wide adoption outside of FireFox.

it's actually implemented in Chrome 73 (behind a flag), Chrome 74 (without a flag), and Safari 12.1.

PR #7435, opened 2019-02-01
Add Object.fromEntries

vicapow:


pakoito:
Hello hello, I'll be your liaison today. Let me get the process started, and I'll come back and update you once we have something ;)

arcanis:
Quick notes:

- `Object.entries` should accept an iterable, not only an array (it matters for cases like `Object.fromEntries(map.entries())`
- The keys can also be numbers (even if they end up as strings) and symbols

lib/core.js
```diff
@@ -47,6 +47,7 @@ declare class Object {
     static defineProperty<T>(o: any, p: any, attributes: PropertyDescriptor<T>): any;
     static entries(object: any): Array<[string, mixed]>;
     static freeze<T>(o: T): T;
+    static fromEntries<K:string, T>(entries: Array<[K, T]>): {[K]: ?T};
```
rubennorte: `fromEntries` accepts any iterable, not just arrays, and the entries can be any object having 0 and 1 keys: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries.

I think this would be closer to the spec:

```suggestion
    static fromEntries<K, V>(entries: Iterable<{0: K, 1: V}>): {[K]: V};
```

rubennorte: Actually... there's a limitation in Flow that doesn't allow us to use arrays when the type is defined as an object with numeric keys. I'll check this and go back to you.

pakoito:
As Ruben said, this new version doesn't seem to be compatible. Maybe the `[K, V]` version is enough?

lib/core.js
rubennorte: A combination of them is enough:

```javascript
static fromEntried<K, V>(entries: Iterable<[K, V] | {'0': K, '1': V}>): {[K]: V};
```

rubennorte:
Please see https://github.com/facebook/flow/pull/7435#discussion_r254126403

goodmind:
@vicapow do you still doing this PR?

vicapow:
@goodmind I've updated the diff based on @rubennorte comment but I haven't tested this in any projects.

pakoito:
Cool, I'll aim to merge it this week.

pakoito:
Turns out the outage was not only external. Importing again today hoping it works now :D

pakoito:
Aaaaand donzo. Thanks @vicapow !!

sebholl:
@vicapow - a question about the typing for `Object.fromEntries`. Why is its return type annotation inexact?

```
    static fromEntries<K, V>(entries: Iterable<[K, V] | {
        '0': K,
        '1': V,
        ...
    }>): { [K]: V, ... };
```

https://github.com/facebook/flow/blob/d848405d03938f7a67c637d14c4445178c217602/lib/core.js#L61

I can't think of a reason it wouldn't be just:

```
    static fromEntries<K, V>(entries: Iterable<[K, V] | {
        '0': K,
        '1': V,
        ...
    }>): {| [K]: V |}; // or { [K]: V } in strict by default mode
```

PR #7443, opened 2019-02-04
fix: add missing ocaml includes for flow-parser-bin

lxe:
flow-parser-bin requires opam and ocaml static libs to be installed, from what I gather. Maybe it requires a dynamic lib instead, but I was unable to find it for osx. Suggestions welcome!

### DIFF
--- a/packages/flow-parser-bin/binding.gyp
+++ b/packages/flow-parser-bin/binding.gyp
@@ -13,12 +13,16 @@
         "src/flow_parser_node.cc",
       ],
       "include_dirs": [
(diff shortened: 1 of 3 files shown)


nmote:
I think @mroch is the most familiar with this

mroch:
nice, thanks. how did you run into this? wondering why it built for me and on circle without these changes.

vicapow:
@mroch we just built it on os x. Not sure if I'm talking nonsense but maybe it's setup for you on your workstations automatically as an available dynamically linked lib?

PR #7447, opened 2019-02-05
Update the JSON.stringify

nmn:
To disallow undefined and functions from being passed in.
JSON.stringify will return undefined if undefined or a function is passed in.

if we allow undefined and functions to be passed in, we would need to change the return value to be `void | string` to be correct. Since that would cause way more trouble for users of JSON.stringify, disallowing the problematic inputs seems like the correct approach here.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -475,7 +475,7 @@ declare class URIError extends Error {
 declare class JSON {
     static parse(text: string, reviver?: (key: any, value: any) => any): any;
     static stringify(
-      value: null | string | number | boolean | Symbol | {} | $ReadOnlyArray<mixed>,
+      value: null | string | number | boolean | {} | $ReadOnlyArray<mixed>,
       replacer?: ?((key: string, value: any) => any) | Array<any>,
       space?: string | number
     ): string;
--- a/lib/core.js
+++ b/lib/core.js
@@ -475,7 +475,7 @@ declare class URIError extends Error {
 declare class JSON {
     static parse(text: string, reviver?: (key: any, value: any) => any): any;
     static stringify(
-      value: any,
+      value: null | string | number | boolean | Symbol | {} | $ReadOnlyArray<mixed>,
       replacer?: ?((key: string, value: any) => any) | Array<any>,
       space?: string | number
     ): string;


lib/core.js
```diff
@@ -475,7 +475,7 @@ declare class URIError extends Error {
 declare class JSON {
     static parse(text: string, reviver?: (key: any, value: any) => any): any;
     static stringify(
-      value: any,
+      value: null | string | number | boolean | Symbol | {} | $ReadOnlyArray<mixed>,
```
bcherny: A few thoughts:

- Interestingly, the spec doesn't mention `undefined`: https://www.ecma-international.org/ecma-262/5.1/#sec-15.12.3
- TypeScript allows `any`: https://github.com/Microsoft/TypeScript/blob/2f9218f346bbc606cccc4aebfca1c2c4f3167fee/lib/lib.es5.d.ts#L1042
- What do you do with custom types, like `class` instances?

```ts
JSON.stringify(undefined) // undefined

class A { constructor(){ this.b = 3 }}
JSON.stringify(new A) // '{"b":3}'
```

- A good compromise might be an overload:

```ts
declare class JSON {
  static stringify(value: typeof undefined): typeof undefined;
  static stringify(value: mixed): string;
}
```

nmn: Replies to each point
1. This is why I didn't want to add the overload.
2. Typescript used any till now, however, https://github.com/Microsoft/TypeScript/pull/29744 made a PR there too. Also, https://twitter.com/fdsmars/status/1092634853378383872.
3. custom classes should be accounted for with `{}`.
4. An overload is unnecessary based on my points above. It's also redundant.

kassens:
Does this type allow passing in an exact object (such as a value of type {| foo: 1 |} )?

goodmind:
@kassens yes

lib/core.js
stryju: FWIW, classes can provide custom behavior by implementing `toJSON()` method:

```js
class Foo {}
JSON.stringify(new Foo()) // "{}"

class Bar {
  toJSON() {
    return undefined;
  }
}
JSON.stringify(new Bar()) // undefined
```

vicapow: Good point to consider. I think this should still be okay since it accepts `{}`

nmote:
@nmn Could you import it and see what effect this would have on our internal codebases? It's a good way to gauge what the deploy burden would be.

This looks pretty reasonable, but we should make this change with care.

lib/core.js
goodmind: @stryju something like this? 
```js
declare function stringify<T: string | void>(
  object: { toJSON(): T }
): T
declare function stringify(
  object: null | string | number | boolean | Symbol | {} | $ReadOnlyArray<mixed>
): string
```

goodmind:
Crazy typelevel stuff with JSON.stringify and toJSON
https://flow.org/try/#0AQhQBcE8AcFNgFIGUDyA5AJAZ3AJwJYB2A5vgGaTAC8wAPACoB8AFAPoBcw9AlNY8BgDCAQwA2o2qBAhmU6SABkwZgxYcuwAD7AA3uAD2ydJ2a8q-egF8ANMAAenHZbP8cBEtzkhu1r11-yQmISsvKKymycAG76+AAmWroGRmgmLsAx8c58GbFxnmHASipMkVzpTtw+fiD0foygDXGwAMaiwrjwbcJYWIioaACsulI4wuD4LcBuRKQUqsz6AEYAVpz0ReWcQeKS0inYeLPklEoHMyQnAbVSjADcoJagoC36hDjAhPrgAGp51Mp0ilBgA6C5zSDMSoPV7vcDAfBYP7xAGmHLAsFHS4UZgAV0IzTIRFg+Rhbw+iORcQATKigQNQeCTlDkgM0jl8YTiXFnA8gA

nmn:
Possibly better types:

https://fburl.com/l8kkhrvs

stryju:
@nmn If we're trying to explicitly block `JSON.stringify` from returning `undefined` - which is not what JS does...

nmn:
@stryju - JS returns undefined when undefined is passed in or when a function is passed in. This type bans those arg types and so you will only get string as output.

If you accidentally pass in undefined or a function into stringify it’s probably a bug and rather than trying to add every edge case to the types it makes more sense to disallow incorrect use cases in the types.

Second, it does not block the stringification of classes with the `toJSON` function. They still work.

Currently, functions are not banned because functions are also objects, and also classes that have a `toJSON` method returns undefined is not caught.

But the link I posted in my last comment does fix more of those problems. It’s a higher risk change though, so it’s not quite ready to be merged yet.

vicapow:
any new info on this one?

nmote:
I've accepted the internal diff, @nmn can now choose when/if to land it

PR #4091, opened 2017-06-02
Tweaks to built-in http module.

STRML:
In particular, it was not possible for libdefs to import `http$Server` (making it difficult to type e.g. Express), because imports in libdefs appear to coerce the module into a broken type.

This makes `http$Server` and `https$Server` globals so they can be easily used in libdefs without importing `'http'`. 

Additionally, added a third `listen()` form without the `backlog` arg.

PR #7448, opened 2019-02-06
Tweaks to built-in http module.

STRML:
In particular, it was not possible for libdefs
to import http$Server (making it difficult to type e.g. Express),
because imports in libdefs appear to coerce the module into a
broken type.

This makes http$Server and https$Server globals so they can be
easily used in libdefs without importing 'http'.

Additionally, added a third listen() form without the `backlog` arg.

Continued from #4091, ping @nmote.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1324,23 +1324,38 @@ declare class http$ServerResponse extends stream$Writable {
   writeHead(status: number, headers?: {[key: string] : string}): void;
 }
 
-declare module "http" {
-  declare class Server extends net$Server {
-    listen(port?: number, hostname?: string, backlog?: number, callback?: Function): Server;
-    // The following signatures are added to allow omitting intermediate arguments
-    listen(port?: number, backlog?: number, callback?: Function): Server;
-    listen(port?: number, hostname?: string, callback?: Function): Server;
-    listen(port?: number, callback?: Function): Server;
-    listen(path: string, callback?: Function): Server;
-    listen(handle: Object, callback?: Function): Server;
-    listening: boolean;
-    close(callback?: (error: ?Error) => mixed): Server;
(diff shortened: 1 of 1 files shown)


nmote:
This looks reasonable. I'm about done for the night but I will take a closer look tomorrow.

lib/node.js
```diff
@@ -1324,23 +1324,38 @@ declare class http$ServerResponse extends stream$Writable {
...
-    listen(handle: Object, callback?: Function): Server;
-    listening: boolean;
-    close(callback?: (error: ?Error) => mixed): Server;
-    maxHeadersCount: number;
-    keepAliveTimeout: number;
-    setTimeout(msecs: number, callback: Function): Server;
-    timeout: number;
-  }
+declare class http$Server extends net$Server {
+  listen(port?: number, hostname?: string, backlog?: number, callback?: Function): Server;
```
nmote: Looks like all the mentions of `Server` here need to be updated. You could change them to `http$Server`, but that would cause errors if somebody did something equivalent to `(server.listen(...): Server)`. Probably best to just update all of these to `this`, which should avoid causing new errors.

Once you have this internally-consistent (i.e. running `flow` on an empty project reports no errors), let me know and I'll take a closer look.

STRML:
@nmote Updated.

nmote:
One quick clarification before I get someone else to review this internally:

> Additionally, added a third listen() form without the `backlog` arg.

It looks like this doesn't apply to the current PR. I'm guessing that in the original one, you added a `listen` variant, but since then, the variant that you wanted was added elsewhere. Is that correct, or did I overlook something?

STRML:
Yeah that's correct and the note is now incorrect. It looks like all the
forms were added some time in late 2018.

PR #7460, opened 2019-02-08
website: chore: fix typo

kveerapen:
In my opinion, I think it reads better by putting "first" before "try", given the way the sentence has been constructed.


callumlocke:
The original wording sounds better to me.

jbrown215:
I agree with @callumlocke

Issue #6639, opened 2018-07-25
add support for new type: bigint


dnalborczyk:
```javascript
(1n: bigint)
```

https://flow.org/try/#0BQRgdgXABARglgczmALgSiA

goodmind:
Parsing side of this feature was imported in #7471

dsainati1:
This should be resolved, so I'm going to close this.

goodmind:
@dsainati1 there is still typechecking part

lo-corp:
When can we expect this?

seanlangbrown:
BigInt support would really help with a project I'm working on.
I see there are 2 PRs for BigInts that have been open for 1 year.
Are these just waiting for review or is there anything someone (like me :) ) can contribute to help resolve this issue?

nnmrts:
Can we first have support for `Array` please?

Brianzchen:
I'm still getting the error `BigInt bigint [1] is not yet supported. [bigint-unsupported]` in latest version.

cc @mroch any insight? 🙏

gkz:
BigInt is now supported as of 0.194

PR #7471, opened 2019-02-13
Add support for BigInt parsing

goodmind:
Proposal: https://github.com/tc39/proposal-bigint
Discussion: https://github.com/facebook/flow/issues/6639

This probably can be discussion?

Also `typeof` should support `bigint` but I don't know how to add it

Parsing errors:
image

Type errors:
image


nmote:
Could you link the TC39 proposal so it's easily available from this PR?

goodmind:
@nmote done

nmote:
Thanks! @gabelevi and @mroch are probably the best people to review this -- hopefully one of them finds the time soon.

goodmind:
@nmote do you know what's wrong with CI?

nmote:
I believe @gabelevi agreed to review this in the Discord, so assigning to me. Gabe, feel free to undo this if my recollection is incorrect.

src/typing/statement.ml
```diff
@@ -4261,6 +4261,12 @@ and literal cx loc lit = Ast.Literal.(match lit.Ast.Literal.value with
   | Number f ->
       DefT (annot_reason (mk_reason RNumber loc), bogus_trust (), NumT (Literal (None, (f, lit.raw))))
 
+  | BigInt _ ->
+      let reason = annot_reason (mk_reason (RBigIntLit lit.raw) loc) in
+      Flow.add_output cx (Error_message.EBigIntNotYetSupported reason);
+      AnyT.why AnyError reason
+      (* DefT (reason, bogus_trust (), NumT (Literal (None, (f, lit.raw)))) *)
```
gabelevi: You can delete this comment

src/parser/flow_lexer.ml
```diff
@@ -1641,28 +1804,39 @@ let type_token env lexbuf =
...
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    Token (env, mk_bignum_singleton NORMAL num)
+  | Opt neg, (scinumber | scibigint), word ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | Opt neg, scinumber ->
       let num = lexeme lexbuf in
       Token (env, mk_num_singleton NORMAL num)
     | _ -> failwith "unreachable"
```
gabelevi: This is no longer unreachable. I suspect you need to add a `scibigint` case

src/parser/flow_lexer.ml
```diff
@@ -724,65 +860,77 @@ let token (env: Lex_env.t) lexbuf : result =
...
 
-  | binnumber, (letter | '2'..'9'), Star alphanumeric ->
+  | binbigint ->
+    Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
+  | (binnumber | binbigint), (letter | '2'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | binnumber ->
       Token (env, T_NUMBER { kind = BINARY; raw = lexeme lexbuf })
     | _ -> failwith "unreachable"
```
gabelevi: Also no longer unreachable

src/parser/flow_lexer.ml
```diff
@@ -724,65 +860,77 @@ let token (env: Lex_env.t) lexbuf : result =
...
     Token (env, T_NUMBER { kind = BINARY; raw = lexeme lexbuf })
 
-  | octnumber, (letter | '8'..'9'), Star alphanumeric ->
+  | octbigint ->
+    Token (env, T_BIGINT { kind = OCTAL; raw = lexeme lexbuf })
+  | (octnumber | octbigint), (letter | '8'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | octnumber -> Token (env, T_NUMBER { kind = OCTAL; raw = lexeme lexbuf })
     | _ -> failwith "unreachable"
```
gabelevi: Also no longer unreachable

src/parser/flow_lexer.ml
```diff
@@ -58,6 +58,14 @@ let scinumber = [%sedlex.regexp?
 let wholenumber = [%sedlex.regexp? underscored_digit, Opt '.']
 let floatnumber = [%sedlex.regexp? Opt underscored_digit, '.', underscored_decimal]
 
+let binbigint = [%sedlex.regexp? '0', ('B' | 'b'), underscored_bin, 'n']
```
gabelevi: I'm not too well versed in sedlex, but can you write these regexes in terms of the above regexes? Like

```
let binbigint = [%sedlex.regexp? binnumber, 'n']
let octbigint = [%sedlex.regexp? octnumber, 'n']
...
```

src/parser/flow_lexer.ml
```diff
@@ -284,6 +292,107 @@ end = struct
       with No_good -> raise e
 end
 
+module IntOfString : sig
```
gabelevi: It's not 100% clear to me why this module exists

1. `int_of_string` returns a `float`. So how does this differ from `FloatOfString`?
2. There's a lot of copied code from `FloatOfString`. Can this be refactored into `FloatOfString` if we really need it?

src/parser/flow_lexer.ml
```diff
@@ -284,6 +292,107 @@ end = struct
...
+          |> parse_hex_symbol
+          |> parse_body
+          |> float_of_t
+      with No_good -> raise e
+end
+
+let bigint_strip_n raw =
+  let size = String.length raw in
+  let str =
+    if (raw.[size - 1]) == 'n'
```
gabelevi: If `size` is `0` then this is an exception. Not sure when `raw` is the empty string, but FYI

src/parser/flow_lexer.ml
```diff
@@ -1579,46 +1727,62 @@ let type_token env lexbuf =
...
   | Opt neg, legacyoctnumber ->
     let num = lexeme lexbuf in
     Token (env, mk_num_singleton LEGACY_OCTAL num)
 
-  | Opt neg, hexnumber, non_hex_letter, Star alphanumeric ->
+  | Opt neg, hexbigint ->
+    let num = lexeme lexbuf in
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    begin try Token (env, mk_bignum_singleton NORMAL num)
```
gabelevi: You shouldn't need this `try-catch`. `mk_bignum_singleton` should handle win32, no?

src/parser/flow_lexer.ml
```diff
@@ -1579,46 +1727,62 @@ let type_token env lexbuf =
...
+  | Opt neg, binbigint ->
+    let num = lexeme lexbuf in
+    Token (env, mk_bignum_singleton BINARY num)
+  | Opt neg, (binnumber | binbigint), (letter | '2'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | Opt neg, binnumber ->
       let num = lexeme lexbuf in
       Token (env, mk_num_singleton BINARY num)
     | _ -> failwith "unreachable"
```
gabelevi: Also now reachable

src/parser/flow_lexer.ml
```diff
@@ -1579,46 +1727,62 @@ let type_token env lexbuf =
...
+  | Opt neg, octbigint ->
+    let num = lexeme lexbuf in
+    Token (env, mk_bignum_singleton OCTAL num)
+  | Opt neg, (octnumber | octbigint), (letter | '8'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | Opt neg, octnumber ->
       let num = lexeme lexbuf in
       Token (env, mk_num_singleton OCTAL num)
     | _ -> failwith "unreachable"
```
gabelevi: Also reachable

src/parser/flow_lexer.ml
```diff
@@ -1579,46 +1727,62 @@ let type_token env lexbuf =
...
-
   | Opt neg, legacyoctnumber ->
     let num = lexeme lexbuf in
     Token (env, mk_num_singleton LEGACY_OCTAL num)
 
-  | Opt neg, hexnumber, non_hex_letter, Star alphanumeric ->
+  | Opt neg, hexbigint ->
+    let num = lexeme lexbuf in
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
```
gabelevi: I can't figure out why this is invalid. Can you clarify? `-0x1n` seems fine to me AFAICT from the spec

gabelevi:
Also apologies for the delayed review. I have no excuse, I totally spaced on it. Sorry!

src/parser/flow_lexer.ml
goodmind: Add case for `*bigint` everywhere? I never understanded how this `Numbers cannot be immediately followed by words` should work with bigint if it can be followed by `n`

src/parser/flow_lexer.ml
```diff
@@ -724,65 +860,77 @@ let token (env: Lex_env.t) lexbuf : result =
...
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | floatbigint ->
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | (wholenumber | floatnumber | wholebigint), word ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | wholenumber | floatnumber ->
       Token (env, T_NUMBER { kind = NORMAL; raw = lexeme lexbuf })
```
goodmind: Should I add `floatbigint` here too?

src/parser/flow_lexer.ml
goodmind: Yeah, removed it

src/parser/flow_lexer.ml
goodmind: This is error, probably should be `legacyoctbigint`, but not sure if this should error early

src/parser/flow_lexer.ml
goodmind: Yes, it works

src/parser/flow_lexer.ml
goodmind: Should `hexnumber` case also remove `try-catch`?

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
-  | scinumber, word ->
+  | scibigint ->
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | (scinumber | scibigint), word ->
     (* Numbers cannot be immediately followed by words *)
     recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
     | scinumber -> Token (env, T_NUMBER { kind = NORMAL; raw = lexeme lexbuf })
+    | scibigint -> Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
```
gabelevi: Should this emit an error? Like a

    let env = lex_error env loc Parse_error.InvalidBigInt in

?

src/parser/flow_lexer.ml
gabelevi: Yeah, seems like a good idea!

src/parser/flow_lexer.ml
gabelevi: Hmm...let's see...

I see

```
begin try Token (env, mk_num_singleton NORMAL num)
    with _ when Sys.win32 ->
```

And then `mk_num_singleton` has

```
    begin try FloatOfString.float_of_string num
    with Failure _ -> failwith ("Invalid number "^num)
    end
```

and `float_of_string` has 

```
  let float_of_string str =
    try Pervasives.float_of_string str
    with e when Sys.win32 ->
```

So it looks like Windows is already explicitly handled there too. Though maybe the Windows float parsing fails sometimes and we needed to handle that?

Should `mk_num_singleton` throw `Failure` exceptions? Probably not...it should probably emit an error and return a number. I'm guessing that we never hit those `failwith`'s so, it hasn't been a problem yet.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
       is_tail
     ))
 
-  | binnumber, (letter | '2'..'9'), Star alphanumeric ->
+  | binbigint ->
```
gabelevi: I mentioned it in the top level review, but this case needs to come after the `(* Numbers cannot be immediately followed by words *)` case. You need to do that here and everywhere else.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
+    Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
+  | (binnumber | binbigint), (letter | '2'..'9'), Star alphanumeric ->
```
gabelevi: And actually this is incorrect. This needs to be split into two cases:

1. `| binnumber, (letter | '2'..'9'), Star alphanumeric ->`
2. `| binbigint, Plus alphanumeric ->`

Why? It's because adding a `0` or `1` to the end of a `binnumber` is ok (e.g. `0x111`) but adding `0` or `1` to the end of a `binbigint` is NOT ok (e.g. `0x11n1`)

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
     | _ -> failwith "unreachable"
     )
-
   | binnumber ->
     Token (env, T_NUMBER { kind = BINARY; raw = lexeme lexbuf })
 
-  | octnumber, (letter | '8'..'9'), Star alphanumeric ->
+  | octbigint ->
+    Token (env, T_BIGINT { kind = OCTAL; raw = lexeme lexbuf })
+  | (octnumber | octbigint), (letter | '8'..'9'), Star alphanumeric ->
```
gabelevi: Ditto, this needs to be split up.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
     | _ -> failwith "unreachable"
     )
-
   | octnumber ->
     Token (env, T_NUMBER { kind = OCTAL; raw = lexeme lexbuf })
 
-  | legacyoctnumber, (letter | '8'..'9'), Star alphanumeric ->
+  | legacyoctbigint ->
+    Token (env, T_BIGINT { kind = LEGACY_OCTAL; raw = lexeme lexbuf })
+  | (legacyoctnumber | legacyoctbigint), (letter | '8'..'9'), Star alphanumeric ->
```
gabelevi: Ditto

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
     | _ -> failwith "unreachable"
     )
-
   | legacyoctnumber ->
     Token (env, T_NUMBER { kind = LEGACY_OCTAL; raw = lexeme lexbuf })
 
-  | hexnumber, non_hex_letter, Star alphanumeric ->
+  | hexbigint ->
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | (hexnumber | hexbigint), non_hex_letter, Star alphanumeric ->
```
gabelevi: Ditto

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
-
   | hexnumber ->
     Token (env, T_NUMBER { kind = NORMAL; raw = lexeme lexbuf })
 
-  | scinumber, word ->
+  | scibigint ->
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | (scinumber | scibigint), word ->
```
gabelevi: It's actually ok for this to be the same case. However, it DOES need to come before both the `| scibigint` and `| scinumber` cases.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,86 @@ let token (env: Lex_env.t) lexbuf : result =
...
     Token (env, T_NUMBER { kind = NORMAL; raw = lexeme lexbuf })
 
-  | (wholenumber | floatnumber), word ->
+  | wholebigint ->
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | floatbigint ->
+    let loc = loc_of_lexbuf env lexbuf in
+    let env = lex_error env loc Parse_error.InvalidBigInt in
+    Token (env, T_BIGINT { kind = NORMAL; raw = lexeme lexbuf })
+  | (wholenumber | floatnumber | wholebigint), word ->
```
gabelevi: This is also ok to be one case. But you should add `floatbigint` and you should move this case before the others.

src/parser/flow_lexer.ml
```diff
@@ -1579,46 +1643,69 @@ let type_token env lexbuf =
    * Number literals
    *)
 
-  | Opt neg, binnumber, (letter | '2'..'9'), Star alphanumeric ->
+  | Opt neg, binbigint ->
+    let num = lexeme lexbuf in
+    Token (env, mk_bignum_singleton BINARY num)
+  | Opt neg, (binnumber | binbigint), (letter | '2'..'9'), Star alphanumeric ->
```
gabelevi: Needs to be split up

src/parser/flow_lexer.ml
goodmind: Wouldn't it break because `(* Numbers cannot be immediately followed by words *)` case would match `n` character at the end earlier if I move this to top?

src/parser/flow_lexer.ml
goodmind: If I do something like this 
```ocaml
| binbigint, Plus alphanumeric ->
  recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
  | _ -> failwith "unreachable bigint"
  )
| binnumber, (letter | '2'..'9'), Star alphanumeric ->
  recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
  | _ -> failwith "unreachable number"
  )
```
It always fails with "unreachable number" on `0b101011101n` literal. Why first matcher doesn't work?

gabelevi: In your example, `0b101011101n` does NOT match the first pattern. Remember, `binbigint` includes the `n`. So while `0b101011101nx` would match the first pattern, `0b101011101n` would not.

I would instead recommend writing 

```ocaml
| binbigint, Plus alphanumeric ->
  recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
  | _ -> failwith "unreachable bigint"
  )
| binbigint ->
   Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
| binnumber, (letter | '2'..'9'), Star alphanumeric ->
  recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
  | _ -> failwith "unreachable number"
  )
| binnumber ->
   Token (env, T_NUMBER { kind = BINARY; raw = lexeme lexbuf })
```

* `0b101nx` would match the first pattern
* `0b101n` would match the second pattern
* `0b101x` would match the third pattern
* `0b101` would match the fourth pattern

src/parser/flow_lexer.ml
```diff
@@ -58,6 +58,14 @@ let scinumber = [%sedlex.regexp?
 let wholenumber = [%sedlex.regexp? underscored_digit, Opt '.']
 let floatnumber = [%sedlex.regexp? Opt underscored_digit, '.', underscored_decimal]
 
+let binbigint = [%sedlex.regexp? binnumber, 'n']
+let octbigint = [%sedlex.regexp? octnumber, 'n']
+let legacyoctbigint = [%sedlex.regexp? legacyoctnumber, 'n']
+let hexbigint = [%sedlex.regexp? hexnumber, 'n']
+let scibigint = [%sedlex.regexp? decintlit, ('e' | 'E'), Opt ('-' | '+'), underscored_digit, 'n']
+let wholebigint = [%sedlex.regexp? underscored_digit, 'n']
```
gabelevi: Make sure to add a test for something like `1.n`. It doesn't look supported by the bigint proposal but `floatbigint` won't allow it.

src/parser/flow_lexer.ml
```diff
@@ -58,6 +58,14 @@ let scinumber = [%sedlex.regexp?
 let wholenumber = [%sedlex.regexp? underscored_digit, Opt '.']
 let floatnumber = [%sedlex.regexp? Opt underscored_digit, '.', underscored_decimal]
 
+let binbigint = [%sedlex.regexp? binnumber, 'n']
+let octbigint = [%sedlex.regexp? octnumber, 'n']
+let legacyoctbigint = [%sedlex.regexp? legacyoctnumber, 'n']
+let hexbigint = [%sedlex.regexp? hexnumber, 'n']
+let scibigint = [%sedlex.regexp? decintlit, ('e' | 'E'), Opt ('-' | '+'), underscored_digit, 'n']
```
gabelevi: Why not `[%sedlex.regexp? scinumber, 'n']`? Since `scibigint` isn't supported by the proposal, why does this differ from `scinumber`?

src/parser/flow_lexer.ml
```diff
@@ -58,6 +58,14 @@ let scinumber = [%sedlex.regexp?
...
+let wholebigint = [%sedlex.regexp? underscored_digit, 'n']
+let floatbigint = [%sedlex.regexp? Opt underscored_digit, '.', underscored_decimal, 'n']
```
gabelevi: Can this be `[%sedlex.regexp? floatnumber, 'n']`?

Or maybe `[%sedlex.regexp? (floatnumber | ('.', underscored_decimal)), 'n']` to outlaw `.1n`?

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,105 @@ let token (env: Lex_env.t) lexbuf : result =
       is_tail
     ))
 
+
+  (* TODO: This matcher doesn't work *)
```
gabelevi: Responded to your question with how I think this could work.

src/parser/estree_translator.ml
```diff
@@ -966,6 +966,7 @@ end with type t = Impl.t) = struct
       | Literal.Boolean b -> bool b
       | Literal.Null -> null
       | Literal.Number f -> number f
+      | Literal.BigInt f -> number f
```
gabelevi: So I looked a little into estree and it looks like they're working on how to represent bigints in the AST (https://github.com/estree/estree/pull/179).

One downside, however, is that the spec seems to want to set `value` to be a bigint, and I'm not sure if js_of_ocaml can support that.

src/parser/flow_lexer.ml
goodmind: Doesn't Opt already outlaw this?

src/parser/estree_translator.ml
goodmind: JSON doesn't support it either

goodmind: https://github.com/estree/estree/pull/179#discussion_r158063172 `value` property also supports null, so translator can just return null

src/parser/flow_lexer.ml
goodmind: Maybe this should be in separate PR, or I can add comment about Windows?

src/parser/flow_lexer.ml
goodmind: What about `1..n`?

src/parser/flow_lexer.ml
goodmind: I think it is never empty string, but should I check for `size == 0` just in case?

src/parser/flow_lexer.ml
gabelevi: Sorry, typo. I meant outlaw `1.n`

src/parser/flow_lexer.ml
gabelevi: So the real question here is "what invariants can `bigint_strip_n` assume about its input?"

The basic choices are:

1. Make no assumptions. Handle empty strings and strings without `n` at the end. Maybe throw in those cases, since the lexer should never generate these bad strings?
2. Assume a well-formed bigint string. No need to check for empty string and can assume it ends with `n`.

Right now it's kind of half & half. I prefer (1), but either is probably fine.

src/parser/estree_translator.ml
```diff
@@ -966,6 +966,7 @@ end with type t = Impl.t) = struct
       | Literal.Boolean b -> bool b
       | Literal.Null -> null
       | Literal.Number f -> number f
+      | Literal.BigInt _ -> null
```
gabelevi: So this isn't quite what I mean. I think you need to add a new node for bigint literals. You can either try and follow the estree proposal or you can copy what babylon does.

[Babylon outputs a node like](https://github.com/babel/babel/blob/master/packages/babel-parser/test/fixtures/experimental/bigint/valid-hex/output.json)

```
        "expression": {
          "type": "BigIntLiteral",
          "start": 0,
          "end": 9,
          "loc": {
            "start": {
              "line": 1,
              "column": 0
            },
            "end": {
              "line": 1,
              "column": 9
            }
          },
          "extra": {
            "rawValue": "0xFFF123",
            "raw": "0xFFF123n"
          },
          "value": "0xFFF123"
        }
```

Whereas estree would be something like

```
"expression": {
  "type": "BigIntLiteral",
  "start": 0,
  "end": 9,
  "loc": {
  "start": {
    "line": 1,
    "column": 0
  },
  "end": {
    "line": 1,
    "column": 9
  },
  "value": 0xFFF123n;
  "bigint": "0xFFF123n";
}, 
```

goodmind: Then how should be named node for type literals?

src/parser/flow_lexer.ml
goodmind: Empty string would error in `Int64.of_string` 😄then
```ocaml
utop # Int64.of_string "";;
Exception: Failure "int_of_string".
utop # Pervasives.float_of_string "";;
Exception: Failure "float_of_string".
```

goodmind:
@nmote would rebase and force push to fix conflicts remove the review from pull request?

nmote:
I don't know

goodmind:
@nmote ok, thanks

src/parser/estree_translator.ml
```diff
@@ -510,6 +510,7 @@ end with type t = Impl.t) = struct
             "filter", option expression filter;
           ]
       | _loc, Identifier id -> identifier id
+      | loc, Literal ({ Literal.value = Ast.Literal.BigInt _; _ } as lit) -> bigint_literal (loc, lit)
```
goodmind: I didn't add new BigIntLiteral constructor, because it started to error across the project about missing cases, not only in parser

gabelevi: To be clear, you don't need to add a `BigIntLiteral` constructor to the OCaml AST that the Flow parser generates. I'm talking about the JSON estree AST that `estree_translator` outputs. This is the JSON that shows up in the `.tree.json` test files. Looking at your PR, it looks like you've correctly updated this. For example, in `valid-binary.tree.json` I see

```
      "expression":{
        "type":"BigIntLiteral",
        "loc":{"source":null,"start":{"line":1,"column":0},"end":{"line":1,"column":12}},
        "range":[0,12],
        "value":null,
        "bigint":"0b101011101n"
      },
```

which conforms to the estree spec. So nice job!

src/parser/estree_translator.ml
```diff
@@ -966,6 +967,7 @@ end with type t = Impl.t) = struct
       | Literal.Boolean b -> bool b
       | Literal.Null -> null
       | Literal.Number f -> number f
+      | Literal.BigInt _ -> null
```
gabelevi: It's my impression that this pattern match will never be hit. So it might be a little safer to instead write something like

  | Literal.BigInt _ -> failwith "We should not create Literal nodes for bigints"

src/parser/expression_parser.ml
```diff
@@ -860,6 +860,43 @@ module Expression
     Expect.token env (T_NUMBER { kind; raw });
     value
 
+  (* TODO: don't know if this function should be here *)
```
gabelevi: Yeah, it's a little non-ideal that this is duplicated between here and the lexer. That said, `Flow_lexer.mk_num_singleton` and `Expression_parser.number` seem awfully similar. So what you're doing is no worse than what already exists. And the code is slightly different, I suppose :P

src/parser/flow_ast.ml
```diff
@@ -73,6 +74,14 @@ and NumberLiteral : sig
   [@@deriving show]
 end = NumberLiteral
 
+and BigIntLiteral : sig
+  type t = {
+    value: float;
```
gabelevi: Strictly speaking, I suppose we really should use some sort of OCaml bigint implementation here. A float might lose precision. But I think this is totally fine for now. Maybe in the future we can import Big_int from Core and use that :)

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,129 @@ let token (env: Lex_env.t) lexbuf : result =
...
+    )
+  | binbigint ->
+    Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
   | binnumber, (letter | '2'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
-    recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
+    recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
     | binnumber ->
-      Token (env, T_NUMBER { kind = BINARY; raw = lexeme lexbuf })
+      Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
```
gabelevi: This should be `T_NUMBER`

gabelevi: I suppose we must be missing a test that looks for `0b123hello` or something like that. Would you mind adding one?

src/parser/flow_lexer.ml
gabelevi: ¯\_(ツ)_/¯ don't feel strongly. A comment would suffice!

src/parser/flow_ast.ml
gabelevi: Actually, would you mind renaming this field? Maybe something like 

```
approx_value: float; (* Warning! Might lose precision! *)
```

? AFAICT it looks like you're correctly using the `bigint` field everywhere, but anyone who uses the `value` field should be aware that it's approximate.

src/parser/token.ml
```diff
@@ -136,7 +137,9 @@ type t =
   | T_EMPTY_TYPE
   | T_BOOLEAN_TYPE of bool_or_boolean
   | T_NUMBER_TYPE
+  | T_BIGINT_TYPE
   | T_NUMBER_SINGLETON_TYPE of { kind: number_type; value: float; raw: string }
+  | T_BIGINT_SINGLETON_TYPE of { kind: number_type; value: float; raw: string }
```
gabelevi: Maybe this should also `s/value/approx_value/`? I feel less strongly here than in the ast, since the lexer data structures are far less commonly used.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,129 @@ let token (env: Lex_env.t) lexbuf : result =
       is_tail
     ))
 
+
+  | binbigint, word ->
+    (* Numbers cannot be immediately followed by words *)
+    recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
```
gabelevi: I was wondering why, in your new tests, the `0b101011101nhello` example didn't have an `Unexpected token ILLEGAL` error. It's because you're ignoring the new env and lexbuf being passed into this function. This should be

```
recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
```

Since you're using the `env` and `lexbuf` that were in scope before `recover` was called, the `env` doesn't include the error that `recover` automatically adds.

src/parser/flow_lexer.ml
```diff
@@ -724,65 +767,129 @@ let token (env: Lex_env.t) lexbuf : result =
...
+    | binbigint -> 
+      Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
+    | _ -> failwith "unreachable"
+    )
+  | binbigint ->
+    Token (env, T_BIGINT { kind = BINARY; raw = lexeme lexbuf })
   | binnumber, (letter | '2'..'9'), Star alphanumeric ->
     (* Numbers cannot be immediately followed by words *)
-    recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
+    recover env lexbuf ~f:(fun _ _ -> match%sedlex lexbuf with
```
gabelevi: Ditto

src/parser/flow_lexer.ml
```diff
@@ -1613,16 +1751,39 @@ let type_token env lexbuf =
...
+    (* Numbers cannot be immediately followed by words *)
+    recover env lexbuf ~f:(fun env lexbuf -> match%sedlex lexbuf with
+    | Opt neg, hexbigint ->
+      let num = lexeme lexbuf in
+      (* Reconsider Windows support of FloatOfString *)
+      begin try Token (env, mk_bignum_singleton NORMAL num)
+      with _ when Sys.win32 ->
+        let loc = loc_of_lexbuf env lexbuf in
+        let env = lex_error env loc Parse_error.WindowsFloatOfString in
+        Token (env, T_BIGINT_SINGLETON_TYPE { kind = NORMAL; approx_value = 789.0; raw = "789" })
```
gabelevi: Nit - I think `raw` should be a valid bigint, so it should probably end with `n`. Extra points to making it an authentic `hexbigint`. So maybe 

```
Token (env, T_BIGINT_SINGLETON_TYPE { kind = NORMAL; approx_value = 291.0; raw = "0x123n" })
```

src/parser/flow_lexer.ml
```diff
@@ -1613,16 +1751,39 @@ let type_token env lexbuf =
...
+    | _ -> failwith "unreachable"
+    )
+  | Opt neg, hexbigint ->
+    let num = lexeme lexbuf in
+    (* Reconsider Windows support of FloatOfString *)
+    begin try Token (env, mk_bignum_singleton NORMAL num)
+    with _ when Sys.win32 ->
+      let loc = loc_of_lexbuf env lexbuf in
+      let env = lex_error env loc Parse_error.WindowsFloatOfString in
+      Token (env, T_BIGINT_SINGLETON_TYPE { kind = NORMAL; approx_value = 789.0; raw = "789" })
```
gabelevi: Maybe here too?

src/parser/flow_lexer.ml
goodmind: wow, good catch

goodmind:
@gabelevi can you resolve review?

gabelevi:
@goodmind it's in! Should be in v0.97!

PR #7478, opened 2019-02-15
Add type definitions for service worker PushManager

atabel:
Library definitions for service worker `PushManager` and some fixes in `ServiceWorkerRegistration` and `ServiceWorkerContainer` definitions.

https://w3c.github.io/push-api
https://w3c.github.io/ServiceWorker

### DIFF
--- a/lib/serviceworkers.js
+++ b/lib/serviceworkers.js
@@ -97,16 +97,16 @@ declare class PushSubscriptionJSON {
 }
 
 declare class PushSubscription {
-  endpoint: string,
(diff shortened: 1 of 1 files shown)


dsainati1:
Can you do another pass through this and properly mark the variance of your properties? At least according to the documentation you linked `endpoint` and `expirationTime` in `PushSubscription` are read-only, so they should be marked as covariant.

atabel:
Ok, @dsainati1 I've pushed some changes defining some properties as read-only

Issue #7362, opened 2019-01-15
createImageBitmap is not defined in browsers


bsmith-cycorp:
Spec here:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap

ollelauribostrom:
Gave this a shot in #7483 🙂

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

PR #7483, opened 2019-02-17
Add type definitions for createImageBitmap

ollelauribostrom:
#7362. Not 100% sure which parts should go in the `bom.js` vs `dom.js`. Feedback very much welcome 🙂 

### Relevant documentation
- [createImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap)
- [DOMPoint/DOMMatrix](https://www.w3.org/TR/geometry-1/)
- [SVGElement/SVGImageElement](https://www.w3.org/TR/SVG11/types.html#InterfaceSVGElement)
- [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas)


dsainati1:
This has type and parse errors in `dom.js` and `bom.js`. Can you fix these?

SamChou19815:

PR #7485, opened 2019-02-18
Allow event name symbols in events$EventEmitter.

mirek:
`events$EventEmitter` methods support Symbol as eventName https://nodejs.org/api/events.html#events_emitter_addlistener_eventname_listener - make sure flow library definition does match nodejs.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -804,26 +804,25 @@ declare module "dns" {
 
 declare class events$EventEmitter {
   // deprecated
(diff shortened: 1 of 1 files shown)


mvitousek:
This change looks correct, thank you for the PR! This does change the output of some tests, would you mind regenerating the expected test outputs for node_tests (with ./runtests -r) and adding that to the PR?

SamChou19815:

PR #7486, opened 2019-02-19
[Library definitions] Add optional `propTypes` to `React.AbstractComponentStatics`

### LINKED ISSUE #7484: propTypes no longer part of React.AbstractComponentStatics
One cannot reference `propTypes` on a `ComponentType<*>` since 2e7d5a26babe69f2b17cba25f6f62a524534c635.
Is this intended? If yes, do you have a suggestion for handling this?

Example code:
```

function testHoc<-Props>(Component: React.ComponentType<Props>, props: Props) {
	return class Wrapper extends React.Component<Props> {
      static propTypes = Component.propTypes; // Error here: property `propTypes` is missing in 
`React.AbstractComponentStatics`
      
      render() {
      	return <Component {...props} />
      }
    };
}
```

[Flow try](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgWwA5wE4AuyYAhgM5gBKApiQMZFR5yZgDketDbA3KqlACuAOwbo4wsAWpkCACTh0APAFoACs2xkAfAAoAwi1zDqwggC4qXAgDpDOCaYIAVAJ7ZqSjXC3aANGDYmmSW3loAlGAA3qgAkJwEgniSdDDkFADqeCTYHnhg1AAe0sIAJhQ09Lb2xk5ewdrRqGAtrbIkBOh0gZpuHhQAvGA1jmY2QT59Mnytrc2zYJxl1Hi6kTELYPHUiclgSiMmZtE2pxNaAL5gwNrzsxd3YBd8F0A)

cprodescu:
- fixes #7484
- add test accessing `Component.propTypes`
- fix line numbers for other tests due to shift in `lib/react.js`

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -134,7 +134,7 @@ declare type React$AbstractComponentStatics = {
   // This is only on function components, but trying to access name when
   // displayName is undefined is a common pattern.
   name?: ?string,
-  propTypes?: any,
+  propTypes?: {},
 };
 
 /**
--- a/lib/react.js
+++ b/lib/react.js
@@ -134,6 +134,7 @@ declare type React$AbstractComponentStatics = {
   // This is only on function components, but trying to access name when
   // displayName is undefined is a common pattern.
   name?: ?string,
+  propTypes?: any,
 };
 
 /**
--- a/tests/getters_and_setters/getters_and_setters.exp
+++ b/tests/getters_and_setters/getters_and_setters.exp
@@ -465,8 +465,8 @@ References:
    react.js:17:13
     17| (<Example a="bad" />); // error: number ~> string
                     ^^^^^ [1]
-   <BUILTINS>/react.js:438:36
-   438|   number: React$PropType$Primitive<number>;
+   <BUILTINS>/react.js:439:36
+   439|   number: React$PropType$Primitive<number>;
                                            ^^^^^^ [2]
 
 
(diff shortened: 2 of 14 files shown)


lib/react.js
```diff
@@ -134,6 +134,7 @@ declare type React$AbstractComponentStatics = {
   // This is only on function components, but trying to access name when
   // displayName is undefined is a common pattern.
   name?: ?string,
+  propTypes?: any,
```
TrySound: @jbrown215 This will not let to enable dynamic-export lint ever

jbrown215: Thanks for bringing that up. I am discussing with @dsainati1 and we'll figure out a way to move forward.

cprodescu: Thank you for following up on this :)

cprodescu:
Assuming this can't merge as is due to dynamic-export concerns, I see a few options:
- drop this completely (suggestion on how to handle access to `propTypes` of an abstract component without $FlowIgnore would be helpful)
- try to type the propTypes fully (however this will need to stay in sync with `facebook/prop-types`)
- use a fairly abstract type like `propTypes: {}` (might also be banned by dynamic-export lint later)

Do you see any of these as viable?

jbrown215:
> propTypes: {}

This sounds like a reasonable way to move forward, assuming it works (I don't see why it wouldn't at first glance). Though, I would make it `propTypes?: {}`.

jbrown215:
@cprodescu: Were you able to satisfy your use case with this definition? Would you mind adding a test to tests/react_abstract_component with your use case?

jbrown215:
Also, I can handle re-recording the tests for you.

cprodescu:
I had added the use-case to `tests/react/hoc.js` (since in our xp that's where it occurs), but will move it to `tests/react_abstract_component`. Also still need to rebase.
Thank you for the prompt reply!

jbrown215:
@cprodescu: Class components already allow propTypes, so a test that sets it directly on an AbstractComponent is what we need. Something like:
```
declare var x: React$AbstractComponent<{}, mixed>;
x.propTypes = {foo: 'something'};
```

PR #7494, opened 2019-02-21
Model indexer keys as contravariant (update)

### LINKED ISSUE #5566: Covariant indexers are broken on interfaces

popham:
Currently Flow applies indexer variance to both indexer key and indexer value. Indexer keys should be modelled contravariant regardless of the indexer variance. Fixes #5566.

### DIFF
--- a/src/typing/class_sig.ml
+++ b/src/typing/class_sig.ml
@@ -134,7 +134,7 @@ let add_field ~static name loc polarity field x =
 let add_indexer ~static polarity ~key ~value x =
   let kloc, k = key in
   let vloc, v = value in
-  x |> add_field ~static "$key" kloc polarity (Annot k)
+  x |> add_field ~static "$key" kloc Type.Negative (Annot k)
     |> add_field ~static "$value" vloc polarity (Annot v)
 
 let add_name_field x =
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -6248,13 +6248,10 @@ and flow_obj_to_obj cx trace ~use_op (lreason, l_obj) (ureason, u_obj) =
   (match ldict, udict with
     | Some {key = lk; value = lv; dict_polarity = lpolarity; _},
       Some {key = uk; value = uv; dict_polarity = upolarity; _} ->
-      (* Don't report polarity errors when checking the indexer key. We would
-       * report these errors again a second time when checking values. *)
-      rec_flow_p cx trace ~report_polarity:false ~use_op:(Frame (IndexerKeyCompatibility {
+      rec_flow_t cx trace ~use_op:(Frame (IndexerKeyCompatibility {
         lower = lreason;
         upper = ureason;
-      }, use_op)) lreason ureason (Computed uk)
-        (Field (None, lk, lpolarity), Field (None, uk, upolarity));
+      }, use_op)) (uk, lk);
       rec_flow_p cx trace ~use_op:(Frame (PropertyCompatibility {
         prop = None;
         lower = lreason;
@@ -7396,7 +7393,7 @@ and check_polarity cx ?trace polarity = function
     check_polarity_propmap cx ?trace polarity obj.props_tmap;
     (match obj.dict_t with
     | Some { key; value; dict_polarity; _ } ->
-      check_polarity cx ?trace Neutral key;
+      check_polarity cx ?trace (Polarity.inv polarity) key;
       check_polarity cx ?trace (Polarity.mult (polarity, dict_polarity)) value
     | None -> ())
 
@@ -10706,7 +10703,7 @@ and rec_flow cx trace (t1, t2) =
 and rec_flow_t cx trace ?(use_op=unknown_use) (t1, t2) =
   rec_flow cx trace (t1, UseT (use_op, t2))
 
-and flow_opt_p cx ?trace ~use_op ~report_polarity lreason ureason propref =
+and flow_opt_p cx ?trace ~use_op lreason ureason propref =
   function
   (* unification cases *)
   | Field (_, lt, Neutral),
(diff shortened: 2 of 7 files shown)


nmote:
As discussed in #5565, I'll import this and see how many errors it causes in our main codebases, as a way to gauge the rollout pain. Once I have that information, we'll go from there.

nmote:
Sorry for the delay, I'm planning to spend time on this today

nmote:
I have some good news and some bad news.

The good news is that the rollout pain for this change as implemented is not too bad. There were quite a few errors in our internal codebase, but the vast majority of them traced back to a single library function.

The bad news is that while experimenting with this change, I found an issue and I'm not sure what to do about it.

Given this code:

```
// @flow

type Foo = {+[empty]: string};
type Bar = {[string]: string};

declare var bar: Bar;
const x = {foo:""};

(bar: Foo);
(x: Foo); // error? why?
((x: Bar): Foo); // no error
```

I get the following error:

```
Error ----------------------------------------------------------------------------------------------------- test.js:10:2

Cannot cast `x` to `Foo` because string `foo` [1] is incompatible with empty [2] in the indexer property's key.

   test.js:10:2
   10| (x: Foo); // error? why?
        ^ [1]

References:
   test.js:3:15
    3| type Foo = {+[empty]: string};
                     ^^^^^ [2]



Found 1 error
```

I would expect Flow to give no errors here. Since we are treating keys as contravariant, an indexer property with a key type of `empty` (the bottom type) should accept any key types. Furthermore, this breaks transitivity, which is problematic.

This is related to the conversions we have in place between ordinary record-like objects and objects with indexer properties. @samwgoldman pointed me towards this code snippet: https://github.com/facebook/flow/blob/a2a3fd845621b1f20a09d648721d4e0f11f369e3/src/typing/flow_js.ml#L6425-L6428

I verified that it's the piece of code that controls this behavior.

Unfortunately I'm not sure what the solution is here. If we simply reverse that covariant check and make it contravariant, then we don't allow e.g. `({foo: 'bar'}: {+[string]: mixed})` which doesn't seem right. Do you have any thoughts on this issue?

nmote:
Maybe that is right, following the intuition of the rest of the PR. But it looks like it will result in some extensive rollout pain on our codebases, and probably others. I'm pretty sure that the conversion from objects-as-records to objects with indexer properties is unsound to begin with, so it's hard to find a strong intuition for what should happen here.

samwgoldman:
Forbidding `({foo: 'bar'}: {+[mixed]: mixed})` is the expected behavior, I believe, although a version of this change with that (correct, expected) behavior might entail more new errors if types are written that way.

popham:
I don't have much of an intuition for how variance should apply to indexers. I just map it to a `get` and/or `set` as noted over at https://github.com/facebook/flow/pull/5565. Mapping @nmote's use case from above, I get [flow.org/try/...](https://flow.org/try/#0PQKgUALgngDgpgAgGIHsUILwIN4GoDacAtjNALoBcCAzhAE4CWAdgOYC+A3JLIgEICGdTDny1GrSjXrN2XMABM4AYwA2gxADdBCAEaCqAulyUomtBAA9h2AGZoKAIgecwYABR66VVCgCUXNwtvNH8EYGAEODo6FDoAfgQAdwALKDj3QINBX2C-DjCIpnQomLowEGBXaHhkNGswBAQWOAg3AGsqYlIoHKlxFjA2bhrDesbm1o6+mV6xGQAaBpoW9qo51nmELRUAVzg16VZejRQGeUG5RVV1Le1PLKMwEzMISzGmlamHOxQHXxwEHQWjs6EwEE58kMXO57rU8u4gnD-BlEYZej5kUA), where the PR's current behavior deviates from current `get`/`set` behavior. Mapping @samwgoldman's expectation, I get [flow.org/try/...](https://flow.org/try/#0C4TwDgpgBAKlC8UDeUDmFgAoDWAuKAzsAE4CWAdqgJT5FmUA0hGOtJFqTAbm-dflwD2pACZQAvgG4AUCIgBjADYBDYtC6qowfDBmZtyNCzxQAtqQAeEETTOXrEqjKA) which busts as expected. Does my `get`/`set` heuristic have any obvious flaws that I've been ignoring? My instinct is to rework to implement this heuristic.

nmote:
Yeah, I think your instinct is correct. Unfortunately when I tried that out myself, it led to over 12,000 type errors on one of our internal codebases. I'm not sure that the rollout pain for us and others will be worth it, especially considering that this entire mechanism is unsound to begin with.

With that said, if you update the PR I am happy to take another look. If we're lucky, many of those type errors will trace back to a single type definition. Or, maybe I made the change incorrectly and caused more type errors than necessary.

nmote:
I suppose what we really want is to take the union of all the keys and do a contravariant check against the key type of the indexer:

```
const x = {foo: 4, bar: 5};
(x: {['foo'|'bar']: number}); // passes
(x: {['foo']: number}); // passes
(x: {['foo'|'bar'|'baz']: number}); // errors
(x: {[string]: number}); // errors
```

This would be a step towards making the conversion sound, but the rollout would be really hard. We probably shouldn't do that right now.

nmote:
I've been thinking about this a bit more and I think the check that I implemented is too strict. I simply reversed the direction of the check that I linked above, therefore requiring that *every* key type is a supertype of the indexer property. I think we actually want the more relaxed constraint, that the union of all the key types is a supertype of the indexer property. Hopefully this will cut down on the error diff if you implement it this way.

popham:
Prodding at this, the unavoidable unsoundness of indexers started to really bother me. I've changed my best practices to avoid indexers like plague (except maybe on exact objects). As I'm pessimistic about this PR landing and no longer have a use case, I'm abandoning it.

nmote:
I'm disappointed we couldn't come to a workable solution, but thanks anyway for your effort.

samwgoldman:
I've been thinking about this more, and it seems like indexer keys are actually invariant, since objects provide an API to read the keys of an object as well -- e.g., Object.keys.

PR #7500, opened 2019-02-22
Support LSP 'Kind' for better icons.

vicapow:
screen shot 2019-02-22 at 2 43 50 am


vicapow:
Previously, everything was either Kind=Function or nothing.

nmote:
Is the screenshot before or after?

nmote:
This looks pretty reasonable, but could you add some tests for it?

vicapow:
@nmote the screenshot is after. Their difference isn't much in the screenshot since most of the properties are functions which are unchanged.

@nmote happy too I'm just not sure how. any examples you could point me too? I'm also not sure how to run specific tests.

nmote:
Okay, it would be nice to have a screenshot that shows more of the new icons, but it's not a big deal.

As for tests, I think the best way is to tack some additional tests onto the existing LSP autocomplete tests. Take a look at `newtests/lsp/queries/test.js` and add things as needed to the `completion` test. There's a README in the `newtests` directory. IIRC the correct command to run just the LSP tests would be `./tool test lsp`. You might also find `./tool record` useful -- it runs the tests and changes the test expectations to match the actual output.

Feel free to mention me in the discord if you have more questions and want a quicker response.

vicapow:
@nmote I've added a test for `kind`. I also moved the completion tests to their own suite.

src/services/autocomplete/autocompleteService_js.ml
```diff
@@ -64,7 +64,25 @@ let parameter_name is_opt name =
...
+  | Ty.ClassDecl _ -> Some Class
+  | Ty.StrLit _
+  | Ty.NumLit _
+  | Ty.BoolLit _ -> Some Value
+  | Ty.Fun _ -> Some Function
+  | Ty.TypeAlias _
+  | Ty.Tup _ -> Some Variable
+  | Ty.Union _ -> Some Enum
+  | Ty.Module _ -> Some Module
+  | _ -> Some Variable
```
panagosg7: Hi @vicapow! Can you expand the rest of the cases here (instead of the catch-all), so that we don't miss something when the Ty.t is updated?

PR #7507, opened 2019-02-26
querySelector[All] is difficult to use effectively.

### LINKED ISSUE #4559: querySelector[All] is difficult to use effectively
Trying to use `document.querySelector` or `document.querySelectorAll` can be quite tiresome, as the typings for these are not especially advanced, and so very simple _correct_ code that has no chance of failing in the browser, fails in Flow:

[example](https://flow.org/try/#0MYewdgzgLgBApgGxgXhgExMArgWzmKAOgEcs4AnATwGVE5goRyAKAcgEswAHLIgMxAhWASgBQ7PjGaJhMAN6iY8BIQBuAQwRkUMVgCN15VjFEBfIA)
```js
const el = document.querySelector('input.foo')
if (el) {
  el.value = 'bar' 
}
```

There are several ways to try to assuage the type system; for example casting to any, then casting to the right type:

[example](https://flow.org/try/#0MYewdgzgLgBApgGxgXhgCjQExMArgWzjCgDoBHXOAJwE8BlROYKEKtAcgEswAHXUgGYgQ7AJQAuGAEMwNCTDC4ECAD4AJACoBZADIBJXvwCiCOIWKiAUJwHpEomAG9LMeAhIA3KQkooY7ACMpKnYYSwBfIA)
```js
const el = ((document.querySelector('input.foo'): any): null|HTMLInputElement)
if (el) {
  el.value = 'bar' 
}
```

or adding invariant checks (which in this case are basically noops) everywhere:

[example](https://flow.org/try/#0MYewdgzgLgBAlmAbgQwE52WWBeGAKaZKAUwFtisAuGAIxBABtjMBKGbAPhgG94AzfAEJCJcljZQAFqhAB3GGGLyAoqhmoYAXwBQ20JFjEG7GABMQwAK5ioAOgCOl4qgCeAZSPFgUEKjwByBAAHSzs+en8WbTgBPCM2bm0YeCQ0DCw44wRCMGBiEAEACQAVAFkAGQBJMBCoZSYbKOSjWxQGJxN-GjR-GG1NIA)
```js
const invariant = (statement: boolean) => { if (!statement) throw new Error }

const el = document.querySelector('input.foo')
if (el) {
  invariant(el instanceof HTMLInputElement)
  el.value = 'bar' 
}
```

or just changing every null check to an instanceof check instead:

[example](https://flow.org/try/#0MYewdgzgLgBApgGxgXhgExMArgWzmKAOgEcs4AnATwGVE5goRyAKAcgEswAHLIgMxAhWASgBQ7PjGaIYnaAEMwwOCEkAJACoBZADIBJbrwCiCOHgLCYAb1Ex4CQgDd5CMihisARvPKsYogF8gA)
```js
const el = document.querySelector('input.foo')
if (el instanceof HTMLInputElement) {
  el.value = 'bar' 
}
```

The problem with all of these, is that we're _having to assuage the type system_ rather than actually protecting ourselves against bugs. In other words, this is a case where Flow is wrong, and it isn't helping us to correct it; we either have to add more runtime code, or try to convince flow using `any` casts which feels hacky and is a slippery slope to unilaterally opting out of flow checks.

### Proposals

An ideal proposal, would be able to have flow type on substrings, and subsequently type all querySelector functions to allow for selectors with element names return the correct element type; so perhaps introducing a `$StringMatch<regexp>` meta type? Something like:

```js
declare class Document extends Node {
    ...
    querySelector(selector: 'input'): HTMLInputElement | null;
    querySelector(selector: $StringMatch</^input[^, ]+$/>): HTMLInputElement | null;
    ...
    querySelector(selector: string): HTMLElement | null;
    ...

    ...
    querySelectorAll(selector: 'input'): NodeList<HTMLInputElement>;
    querySelectorAll(selector: $StringMatch</^input[^, ]+$/>): NodeList<HTMLInputElement>;
    ...
    querySelectorAll(selector: string): NodeList<HTMElement>;
    ...
}
```
While this doesn't _completely_ solve the problem, it does go a _very long way_ to doing so.  The benefit of this proposal is that the usability is ace - the developer doesn't have to put any more effort into typing their querySelector calls, and things "just work"

-- 

A less ideal proposal is to introduce a `$Cast` meta type which does the `((foo: any): other)` shorthand, so one can do the following:

```js
document.querySelector('input.something'): $Cast<HTMLInputElement>
```

While this puts more work in the hands of the developer, it would be a more succinct way of dealing with it. 

I'm also happy to hear other ideas from folks.

anthony-redFox:
uses Casting Expressions Fixed #4559

The same things have in TypeScript:
```
querySelector<K extends keyof HTMLElementTagNameMap>(selectors: K): HTMLElementTagNameMap[K] | null;
    querySelector<K extends keyof SVGElementTagNameMap>(selectors: K): SVGElementTagNameMap[K] | null;
    querySelector<E extends Element = Element>(selectors: string): E | null;
    /**
     * Returns all element descendants of node that
     * match selectors.
     */
    querySelectorAll<K extends keyof HTMLElementTagNameMap>(selectors: K): NodeListOf<HTMLElementTagNameMap[K]>;
    querySelectorAll<K extends keyof SVGElementTagNameMap>(selectors: K): NodeListOf<SVGElementTagNameMap[K]>;
    querySelectorAll<E extends Element = Element>(selectors: string): NodeListOf<E>;
```

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1021,6 +1021,7 @@ declare class Document extends Node {
   querySelector(selector: 'template'): HTMLTemplateElement | null;
   querySelector(selector: 'ul'): HTMLUListElement | null;
   querySelector(selector: string): HTMLElement | null;
+  querySelector<T: HTMLElement>(selector: string): T | null;
 
   querySelectorAll(selector: 'a'): NodeList<HTMLAnchorElement>;
   querySelectorAll(selector: 'area'): NodeList<HTMLAreaElement>;
@@ -1085,6 +1086,7 @@ declare class Document extends Node {
   querySelectorAll(selector: 'template'): NodeList<HTMLTemplateElement>;
   querySelectorAll(selector: 'ul'): NodeList<HTMLUListElement>;
   querySelectorAll(selector: string): NodeList<HTMLElement>;
+  querySelectorAll<T: HTMLElement>(selector: string): NodeList<T>;
 
   // Interface DocumentTraversal
   // http://www.w3.org/TR/2000/REC-DOM-Level-2-Traversal-Range-20001113/traversal.html#Traversal-Document
@@ -1211,7 +1213,9 @@ declare class DocumentFragment extends Node {
   prepend(...nodes: Array<string | Node>): void;
 
   querySelector(selector: string): HTMLElement | null;
+  querySelector<T: HTMLElement>(selector: string): T | null;
   querySelectorAll(selector: string): NodeList<HTMLElement>;
+  querySelectorAll<T: HTMLElement>(selector: string): NodeList<T>;
 }
 
 declare class Selection {
@@ -1530,6 +1534,7 @@ declare class Element extends Node {
   querySelector(selector: 'template'): HTMLTemplateElement | null;
   querySelector(selector: 'ul'): HTMLUListElement | null;
   querySelector(selector: string): HTMLElement | null;
+  querySelector<T: HTMLElement>(selector: string): T | null;
 
   querySelectorAll(selector: 'a'): NodeList<HTMLAnchorElement>;
   querySelectorAll(selector: 'area'): NodeList<HTMLAreaElement>;
@@ -1594,6 +1599,7 @@ declare class Element extends Node {
   querySelectorAll(selector: 'template'): NodeList<HTMLTemplateElement>;
   querySelectorAll(selector: 'ul'): NodeList<HTMLUListElement>;
   querySelectorAll(selector: string): NodeList<HTMLElement>;
+  querySelectorAll<T: HTMLElement>(selector: string): NodeList<T>;
 
   // from ChildNode interface
   after(...nodes: Array<string | Node>): void;


anthony-redFox:
@keithamus please take a look.

keithamus:
Thanks for pinging me @anthony-redFox. Sadly I'm not entirely sure this will do what you expect it to. How do you expect this to be used? Could you please show an example?

anthony-redFox:
@keithamus  it will be similar the what we have in TypeScript.
it is mean you define what you are expected.
```
const images = document.querySelectorAll<HTMLImageElement>('img.special_class'); // return NodeList(HTMLImageElement)
const image = document.querySelector<HTMLImageElement>('img.special_class'); // return HTMLImageElement or null
```
The same thing @jakewtaylor wrote in comment https://github.com/facebook/flow/issues/4559#issuecomment-339642966

keithamus:
@anthony-redFox as @jakewtaylor mentions in their comment - that code is invalid. 

[Here's a reduced use case in the Flow repl](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwCIHswFcC2IAdgC7wDeAUAJACOOIMAngMogTjGYwA8AKgFzxC+AEYMAfAApkbDl0HJiMAJaEA5gEpBveAB8hOCBADcFAL4UwmQovjBs+IqQC8QkAHcMDgiQqXrtmzwrva4PsQAdHQMLLJgnDyKKupSAOQAZpiYqRpAA):

```ts
declare class Document {
	querySelector<T: number>(selector: string): T | null;
}
const document = new Document

const el = document.querySelector<string>('foo')
```
```
 const el = document.querySelector<string>('foo')
                                     ^ Cannot call `document.querySelector` because string [1] is incompatible with number [2] in type argument `T`.
References:
6: const el = document.querySelector<string>('foo')
                                     ^ [1]
2: 	querySelector<T: number>(selector: string): T | null;
                     ^ [2]
```

keithamus:
@anthony-redFox playing around, I think this might work:

```ts
declare class Document {
    querySelector('input'): HTMLInputElement;
    //...
    querySelector(selector: string): HTMLElement | null;
    querySelector<T>(selector: string): T | null;
}
```

[Here it is in the repl](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwBIBUCyAZAohCALYgB2ALvAN4C+AUKEnItKhjrgJJkAOArhUIlyVEAA8K5YGix5hpSjXrx4ANygR+IAFzxkFGAEsyAc3oMm0FknZzcAMQD2ThaPgSpZGR3lFFVNQq8ABmME4ARgAUAJR6ak5GwBb0jODWCLZoACJOYPwByqqqAI7aMACeAMogRGAUTjBRAOQmAhTNcb7cfIJulADcwaXl1bXgDU3I4-WNegbGZl32-VQAPvBk-BAQQwCQZSCVNXWTADyYAHxR06dz+oYmpl2Y8BtbO0MMYE5kBvDAPIFdwAXk2IAA7vBcvkAqkiFQ2oJavAwYDYaIAHSHY4zSYtJEdGL0QkogBkZPgpIgmI0WhA8JAVBCLhRaKBAWxoxOE0aZ3szlc-lE12aLKcnXo4vJlOlNLCkViQA):

```ts
declare class HTMLElement {}
declare class HTMLInputElement extends HTMLElement {
  value: string
}
declare class HTMLFooElement extends HTMLElement {
  frob(): void
}

declare class Document {
    querySelector('input'): HTMLInputElement;
    querySelector(selector: string): HTMLElement | null;
    querySelector<T>(selector: string): T | null;
}
const document = new Document

let inputel = document.querySelector('input')
inputel && inputel.value

let fooel = document.querySelector<HTMLFooElement>('foo')
fooel && fooel.frob()
```
```
No errors!
```

The key being rather than having `querySelector<T: HTMLElement>` you have `querySelector<T>` and `querySelector(string): HTMLElement | null`

anthony-redFox:
@keithamus Yes, it is works, but by MDN querySelector(All) should return base HTMLElement elements
```
declare class HTMLElement {}
declare class HTMLInputElement extends HTMLElement {
  value: string
}
declare class FooElement {
  frob(): void
}

declare class Document {
    querySelector('input'): HTMLInputElement;
    querySelector(selector: string): HTMLElement | null;
	querySelector<T>(selector: string): T | null;
}
const document = new Document

let inputel = document.querySelector('input')
inputel && inputel.value

let fooel = document.querySelector<FooElement>('foo')//should be Error because FooElement is not base on HTMLElement
fooel && fooel.frob()
```
[edit your example](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwBIBUCyAZAohCALYgB2ALvAN4C+AUKEnItKhjrgJJkAOArhUIlyVEAA8K5YGix5hpSjXrx4ANygR+IAFzxkFGAEsyAc3oMm0FknYAxAPYOFo5aoBmMBwCMAFAEo9NQcjYAt6RnBrBFs0ABEHMH5FKmoVVXgAR20YAE8AZRAiMAoHGF8AchMBCgrAjjweGpdKAG501WyQPMLi0vLkIvB+vQNjM3q5AiIU+AAfeDJ+CAh2gEgunqGSsoAeTAA+X0G+stHDE1N6zHnF5dXwsAcyA3hgROTXAF5FkAB3eAJJIpCJEKjVQRFeA-d7A0QAOk2BW2-UqENq-no6KhADIcfBsRB4RotCBQSAqO4nFCYR8UoicsjTjBdo5nDNREcKlSHHUAPR85AACwcy2A8G8CHwMC8MAl4Cg-EG8DZLXBaDIDio3hQCGeDWmIko9B5uPxpqJnh8ASAA) and I should get Error! but no error

keithamus:
Ahh I understand now. Yes the PR is correct then, and LGTM 👍

nmote:
Unfortunately this is unsafe. Type parameters which appear only in output positions are underconstrained. In this case, you could call `document.querySelector(...)` and Flow would give you whatever subtype of `HTMLElement` that the downstream code expects, even though there is no guarantee that you will actually have that type at runtime.

I think the best thing to do here is to continue to cast through `any` (e.g. `((document.querySelector(...): any): HTMLImageElement)` which makes it obvious to the author and maintainers that Flow is not providing typechecking there, or to refine the result using `instanceof`. In the original issue, someone also proposed defining an unsafe utility function with this signature, which might be a good option for somebody who wanted to make a different tradeoff between safety and ease-of-use.

I don't think that weakening the guarantees that Flow can provide to avoid slightly more verbose code is a good tradeoff.

anthony-redFox:
@nmote  I don't understand your point about a cast through `any` because [like example](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwBIBUCyAZAohCALYgB2ALvAN4C+AUPaEnItKvAGID23N98eADMY3AEYAKAJQAueADduAS2D0GTcNFZIOAEW5gArqUr9BggI6GQMAJ4BlEETAVuMAIIQIE5E-CuYOWQKGCUyAHNZDBwCIhMqAB94MkMvAG41ejBuMmD4YANjcioAXmSQAHd4fSN4xiIqIV4neDKJCQLa4oA6KxsHPxc3T28AciVicNGoqDJbKJ5uKXgAehXkAAtuVOB4MQR8GFEYPfAoQ18uXkISYvglNDJuKjEUBBzovBu6pu4WgDJ-sJmhBuiJxNIgA)
```TypeScript
declare class HTMLElement {}

declare class Foo {
  frob(): void
}

declare class Document {
    querySelectorAll(selector: string): HTMLElement | null;
}
const document = new Document

let fooel = ((document.querySelectorAll('img'): any): Foo) //should be Error because Foo is not base on HTMLElement
fooel && fooel.frob()
```
I can convert result of querySelector(All)  thought cast `any` to any Object and you say it is more safe?

> Type parameters which appear only in output positions are underconstrained. In this case, you could call `document.querySelector(...)` and Flow would give you whatever subtype of `HTMLElement` that the downstream code expects, even though there is no guarantee that you will actually have that type at runtime.

It is true because I can use a Custom Element  which extended HTMLElement. People which uses cast `any` also not protected that they are get a Object which are set.

The issue #4559 we have since 2017 and cast `any` doesn't resolve the issue and this solution to use the same practices which are works (as [example for TypeScript](https://github.com/Microsoft/TypeScript/issues/12568)).

keithamus:
We've ended up with [utility functions for this](https://github.com/github/query-selector). It's extra code served to the user which is not ideal. It would be nice if flow had a solution for this without resorting to extremes like the double casting trick. IMO it encourages bad future practices.

goodmind:
@nmote
> Flow would give you whatever subtype of `HTMLElement` that the downstream code expects, even though there is no guarantee that you will actually have that type at runtime.

What you mean?
[This](https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwBIBUCyAZAohCALYgB2ALvAN4C+AUPaEnItKhjrgJLFQDmIQiXJUQADwrlgaLHmGlKNevHhQIFAFzxkFGAEsy-ANwrEAe2IAHIlO0Ajc+aJQyx+AHoP8OFGDmyCABPMzAYc1QAeQN+Q20Afl0DI1NVMABXGDhKAGUYMG0kwxNPb19-QJDVAAsQfX5qrXgydOJ7EBhU+H1kbCgrBycXNzMyKApM9XQ6hqaWto73Lx8QPwDg0fHJiAB1fWAKau159s7SlbXKs2R9AC8QZEK9Yq7kfKfkk2v85BAmopSZnSvz6Ax0z0BqgA7vtDsdWqdTAwmOBoKwkBwAGJOZSqABm4XsAAoAJTaABu5n29GRzDRCAxsi43DIVnSFAUongEikZBknHkREUVGoZnJ6nSIA+xRpKJYDPYTLw2PMnKUPOkSoIQq5ovxhNJFKpwFlcvpbBQaAAIuYMsLcap4ABHSUwII5EBEMAUcwwIkAckMbIo-rJAp4rPZaooXVULo67s94B9fsDxH4oe0ch4fEE0djztdia9KaJvxLvulRjD2ej8AAPs10hAIAX426PRWYAAeTBZrjRgB8ZaT3sr4M+YcwDabLaR9DAAV08H8dq5AF5miAofAbWvKIxF2Rl3j4Juifp0-28Czg3XGy0WySz4OaMi8UTV61RAA6dvF5NfQDS8MxJEkgA) clearly errors
```ts
declare class Document {
    querySelector('input'): HTMLInputElement;
    querySelector('img'): HTMLImageElement;
    querySelector(selector: string): HTMLElement | null;
    querySelector<T: HTMLElement>(selector: string): T | null;
}
const document = new Document

const f = (img: HTMLInputElement | null) => {}

f(document.querySelector('img'))
```

nmote:
@anthony-redFox:

> I can convert result of querySelector(All) thought cast `any` to any Object and you say it is more safe?

My point is that it's unsafe either way. I would prefer to have the unsafe cast made explicit, so that it is obvious to both the author and future maintainers that the code in question deserves special attention. I've used the explicit cast through any for other dynamic sources of data in the past and it has worked well for me.

@keithamus those utilities look great. They are both safe and easy to use -- seems like a big win to me. I understand the concern that it will increase code size, but it seems like a reasonable tradeoff to me. And the casting through `any` trick could always be used instead, if code size is the top priority.

@goodmind that's not exactly the case I was interested in, since there is an explicit overload for `'img'`. However, your example prompted me to take a closer look at how overloads get resolved here. It turns out that Flow won't select the overload with the type parameter unless a type parameter is explicitly passed. I verified with some other members of the team that this is behavior we can rely on.

This fact makes me feel less sure about the PR than I did previously. The main danger that I was worried about was somebody calling the function without a type parameter, not realizing that Flow would just infer whatever type parameter was convenient. That worry was unfounded, but even if you pass a type parameter explicitly, it's still unsafe without being obvious that it's unsafe. That makes me uncomfortable, and I don't think we want to support that kind of behavior in Flow.

goodmind:
@nmote maybe something like (this)[https://flow.org/try/#0CYUwxgNghgTiAEkoGdnwBIBUCyAZAohCALYgB2ALvAN4C+AUPaEnItKhjrgJLFQDmIQiXJUQADwrlgaLHmGlKNevHhQIFAFzxkFGAEsy-ANwrEAe2IAHIlO0Ajc+aJQyx+AHoP8OFGDmyCABPMzAYc1QAeQN+Q20Afl0DI1NVMABXGDhKAGUYMG0kwxNPb19-QJDVAAsQfX5qrXgydOJ7EBhU+H1kbCgrBycXNzMyKApM9XQ6hqaWto73Lx8QPwDg0fHJiAB1fWAKau159s7SlbXKs2R9AC8QZEK9Yq7kfKfkk2v85BAmopSZnSvz6Ax0z0BqgA7vtDsdWqdTAwmOBoKwkBwAGJOZSqABm4XsAAoAJTaABu5n29GRzDRCAxsi43DIVnSFAUongEikZBknHkREUVGoZnJ6nSIA+xRpKJYDPYTLw2PMnKUPOkSoIQq5ovxhNJFKpwFlcvpbBQaAAIuYMsLcap4ABHSUwII5EBEMAUcwwAA8mAAfETfl6fTBtAByQxsiiRskCnis9lqqgAH2a6QgEC6qhdHXdnvA4YDwdDxd9Uf0xH48e0ch4fEEqfgGZa2dzztdhbDvtLIaL3sr4M+CYbLbbWZzZjz3Y9vf9mHrXFTZcH4elRgTmFbmY7srAAV08H8dq5AF5miAofAbWfKIxD2Rj3j4JeidX+Mu8CzYxO9xAJJvoGNDIniRKnq0ogAHT5m684VjARLRjW8YkvQ4GQcKsFzuufYNr+KY6pQwYobWJLoZhtpQZQOEFghQ5IWRaFAA] would work?
```js
declare class Document {
    querySelector<T>(selector: 'input'): HTMLInputElement | null;
    querySelector<T>(selector: 'img'): HTMLImageElement | null;
    querySelector<T>(selector: string): HTMLElement | null;
    querySelector<T: HTMLElement>(selector: string): T | null;
}
const document = new Document

const f = (img: HTMLInputElement | null) => {}

f(document.querySelector('img')) // error
f(document.querySelector<HTMLInputElement>('img')) // error
```

PR #7516, opened 2019-02-27
[docs] Add second argument to $Call utility type

kassens:
I think the second argument is required unless $Call has varargs semantics and should be something like `$Call<F, T...>`?


samwgoldman:
`$Call<F, T...>` is the actual form here, since `F` could be a function requiring 0+ arguments.

kassens:
Updated to `T...` hope this makes sense now!

PR #7518, opened 2019-02-28
Support LSP Function parameter completion

vicapow:
Before

function-completion-before

After

function-complete-ater

### DIFF
--- a/newtests/lsp/completion/test.js
+++ b/newtests/lsp/completion/test.js
@@ -9,6 +9,9 @@ import {suite, test} from 'flow-dev-tools/src/test/Tester';
 export default suite(
   ({
     ideStartAndConnect,
+    ideStart,
+    ideRequest,
+    lspInitializeParams,
     ideRequestAndWaitUntilResponse,
     addFile,
     lspIgnoreStatusAndCancellation,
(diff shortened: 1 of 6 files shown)


vicapow:
this doesn't need to be stacked. will update

nmote:
This looks pretty good. I'll have to double-check it against the LSP spec, but your screencast is compelling evidence that it works. I think I'll wait to review until (1) #7500 has landed (since it looks like you'll have to solve merge conflicts) and (2) you've added tests (which I assume you're waiting on #7500 for).

vicapow:
the CI failure seems unrelated?

````
pushd : Cannot find path 'C:\projects\flow\bin' because it does not exist.
At C:\projects\flow\resources\appveyor\after_build.ps1:2 char:3
+   pushd "C:\projects\flow\bin"
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\projects\flow\bin:String) [Push-Location], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.PushLocationCommand
 
    Directory: C:\projects\flow

````

**Comment by nmote** [comment]
The real error is higher up:

```
File "hack/heap/sharedMem.ml", line 105, characters 8-19:
Error: Unbound module EventLogger
```

The blame appears to be 3c80858d1b2c0f3bab9bb24e7460ca93f7dfb162, so yeah, unrelated.

File: src/common/flow_lsp_conversions.ml [hunk 1]
```diff
@@ -18,17 +18,21 @@ let flow_completion_to_lsp
     let params = Core_list.map ~f:(fun p -> p.param_name ^ ": " ^ p.param_ty) params in
     "(" ^ (String.concat ", " params) ^ ")"
   in
-  let itemType, inlineDetail, detail = match item.func_details with
+  let flow_params_to_lsp_snippet name params =
+    let params = Core_list.mapi ~f:(fun i p -> "${" ^ string_of_int (i +1 ) ^ ":" ^ p.param_name ^ "}") params in
+      name ^ "(" ^ (String.concat ", " params) ^ ")"
```
nmote: nit: indentation is off from the previous line

vicapow: fixed

vicapow:
@nmote I've updated this PR to avoid the deprecated API. It was a little more work but I think this makes sense.

nmote:
I have a few comments:
* Unless I'm mistaken, you haven't updated the tests with the new changes. Could you update them?
* It looks like it still provides `insertTextFormat` even though it's no longer providing `insertText` -- was that an oversight, or does the spec say that `insertTextFormat` applies to `textEdits` as well?
  * Along the same lines, it looks like it uses the snippet format for `textEdits`.
* Could you verify that you've tested this iteration manually with VSCode?

vicapow:
> I have a few comments:
> 
> * Unless I'm mistaken, you haven't updated the tests with the new changes. Could you update them?

Good catch. Updated.

> * It looks like it still provides `insertTextFormat` even though it's no longer providing `insertText` -- was that an oversight, or does the spec say that `insertTextFormat` applies to `textEdits` as well?

The spec seems to say insertTextFormat applies to `textEdit` (and `textEdits`) as well. From the spec:

````
	/**
	 * The format of the insert text. The format applies to both the `insertText` property
	 * and the `newText` property of a provided `textEdit`.
	 */
	insertTextFormat?: InsertTextFormat;
````

>   
>   * Along the same lines, it looks like it uses the snippet format for `textEdits`.

We use the snippet format with `insertTextFormat = 2 (Snippet)`.

> * Could you verify that you've tested this iteration manually with VSCode?

I have tested this iteration manually in VSCode, as well.

One more thing, I just realized, we should check for `snippetSupport` from the client capabilities response.

**Comment by nmote** [comment]
This looks good, thanks. The manual position manipulation makes me a little bit nervous, so I'm going to take a closer look at that in a bit. First though, I'll import it and spend some time playing around with it to make sure it behaves the way I expect.

Unfortunately we're having some technical difficulties, and I can't import PRs right now. Hopefully that will be resolved soon, and I'll be able to move forward.

**Comment by nmote** [comment]
The `lsp/completion` tests failed for me when I tried running them locally (both on OS X and Linux). Could you verify that the tests pass for you?

File: newtests/lsp/completion/test.js [hunk 2]
```diff
@@ -176,5 +176,70 @@ export default suite(
...
     ]),
-  ],
+    test('textDocument/completion', [
+      addFile('params.js'),
+      ideStartAndConnect(),
+      ideRequestAndWaitUntilResponse('textDocument/completion', {
+        textDocument: {uri: '<PLACEHOLDER_PROJECT_URL_SLASH>params.js'},
+        position: {line: 9, character: 15},
+        context: { triggerKind: 1 }
+      }).verifyAllIDEMessagesInStep(
```
gabelevi: I suspect this test is asserting that, for each completion item, these fields are correctly set. However, it doesn't assert that these are ALL the fields. So even though this test is passing, some of these completion items have other fields too.

I suspect this because the relevant fields are missing here. Like I think there should at least be a `textEdits` field which specifies what the IDE should autofill in. And since that's the thesis of this diff, the test should probably check it.

Can you update the test? Or let me know if I'm wrong? Thanks!

vicapow: it turns out I really wasn't sending the textEdit response. but that was because in the test, I didn't set `snippetSupport: true`. I fixed that and have now updated the tests.

PR #7532, opened 2019-03-06
Improve types of static functions on Object.

nmn:
The types were too generic so far, this makes it so that functions such as `Object.keys` and `Object.entries` are usable and don't just get rid of type information.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -45,8 +45,8 @@ declare class Object {
     static create(o: any, properties?: PropertyDescriptorMap): any; // compiler magic
     static defineProperties(o: any, properties: PropertyDescriptorMap): any;
     static defineProperty<T>(o: any, p: any, attributes: PropertyDescriptor<T>): any;
-    static entries(object: any): Array<[string, mixed]>;
-    static freeze<T>(o: T): T;
+    static entries<TObj: {}>(object: TObj): Array<[$Keys<TObj>, $Values<TObj>]>;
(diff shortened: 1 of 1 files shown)


goodmind:
Doesn't it breaks with inexact objects?

nmn:
Yeah. I’m checking a bunch of edge-cases and there might be some problems I need to fix.

gkz:
`Object.key` returns own properties, but object types currently can represent both own and non-own properties - so the value returned by `Object.keys` might not necessarily be the type of `$Keys<...>`, e.g.

https://flow.org/try/#0MYGwhgzhAECC0G8BQ1oA8Bc0B2BXAtgEYCmATgNwrTAD22EALqbsAzaQBSY4EmkCUiKqgYALAJYQAdGmgBedJVQBfKgDMOg5KlSliDXKWzQxkmUuirVSBgE8ADsWgB5eYgDU3PETIAaaO5qWJryAHw8PqTKlLT0DCbEjFiuCtjEAO5wHABM-JQcAORoBdBYACQA0sS2EAA8zqF50AD0zdA06cb2pDSOpHZIhWol5VU19Y3kLW3YNPEdXT19A0ixEDQgxFIgNADmHM6EAFbErFIA1tUQHAyJDPxNrdAA2gBEaK8AukA

gkz:
(we are not moving forward with this)

PR #7533, opened 2019-03-06
Change readonly collection types to co-variant.

nmn:
I'm not sure why these weren't co-variant to begin with.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -534,7 +534,7 @@ declare function $asyncIterator<T>(p: AsyncIterable<T>): T;
 
 /* Maps and Sets */
 
-declare class $ReadOnlyMap<K, V> {
+declare class $ReadOnlyMap<+K, +V> {
     @@iterator(): Iterator<[K, V]>;
     entries(): Iterator<[K, V]>;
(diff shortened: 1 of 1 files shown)


TrySound:
https://github.com/facebook/flow/releases/tag/v0.86.0
> Added core types $ReadOnlyMap, $ReadOnlyWeakMap, $ReadOnlySet, and $ReadOnlyWeakSet. These types and their mutable counterparts mostly follow the pattern of $ReadOnlyArray and Array: the read-only parent class provides non-mutating methods, and the usual mutable class extends it. Unlike $ReadOnlyArray, the new types are invariant in their type parameters.

jbrown215:
@nmn: `V` can be covariant in the maps, but both `K` in the maps and `T` in the sets appear contravariantly in the signatures of the data structures. Flow doesn't check variance annotations in libdefs (though we should), which is why this isn't an error.

There's still is a way to make the type parameters covariant, though. @wchargin suggested that we can make `get` and `has` accept `mixed` as an argument so that `K` and `T` don't appear in contravariant positions.

goodmind:
I consider making only `V` covariant, this is what Kotlin does

SamChou19815:
We already made all of the `V` tparams covariant

PR #7537, opened 2019-03-07
[DOCS] Add mention about benefits of covariance and pitfalls of destructuring unions

villesau:
Common misconception is that you should be able to pass an array of "stricter" types to an array that accepts wider range of types. This is not true, and oftentimes it causes confusion why it's not possible. It's fairly easy to fix by changing types to be read-only.

Other thing often causing confusion is destructuring object unions. 

Feel free to fix potential grammar errors and enhance the explanation!


website/en/docs/faq/index.md
```diff
@@ -143,6 +143,76 @@ const isNumber = (valueToRefine: ?number): %checks => typeof valueToRefine === '
...
+
+fn(arr); // Error!
+```
+
+but with $ReadOnlyArray you can achieve what you were looking for:
+
+```js
+// @flow
+const fn = (arr: $ReadOnlyArray<string | number>) => {
+    // arr.push(321) //note! Array<string> passed in and after this it would also contain numbers
```
AugustinLF: I would update this comment, to say that since you're using a `$ReadOnlyArray`, we cannot push.

website/en/docs/faq/index.md
```diff
@@ -143,6 +143,76 @@ const isNumber = (valueToRefine: ?number): %checks => typeof valueToRefine === '
...
+
+fn(object); // Error!
+```
+
+but with covariant property you can achieve what you were looking for:
+
+```js
+// @flow
+const fn = (obj: {| +a: string | number |}) => {
+  	// obj.a = 123 //note! {| a: string |} passed in and after this it would also contain numbers
```
AugustinLF: Here to, it's interesting to point out in the first example why it's an error, and here why the types prevent us from doing this error!

villesau: Good points!

villesau:
I think the build failure is unrelated.

villesau:
@gkz rebased and the tests passes now.

villesau:
A good example of a developer struggling with covariance in Twitter: https://twitter.com/sstur_/status/1107484822119698432

This might be one of the most misunderstood parts of Flow.

PR #7544, opened 2019-03-09
[DOCS] Update creation.md

nazimjamil:
I've replaced the alert TODO with a clear indication that the documentation is simply not ready yet. Defining modules that are accessed via a relative require/import path is not a well covered area and to find some mention of it in the docs but be presented with an alert box isn't great.


website/en/docs/libdefs/creation.md
```diff
@@ -89,8 +89,7 @@ declare module "some-third-party-library" {
 The name specified in quotes after `declare module` can be any string, but it
 should correspond to the same string you'd use to `require` or `import` the
 third-party module into your project. For defining modules that are accessed via
-a relative `require`/`import` path, check out the docs on
-[`.js.flow`](javascript:alert("TODO")) files.
```
goodmind: 😀

PR #7554, opened 2019-03-13
[DOCS] Update celebrity name in website en/utilities

brookemitchell:
Was looking up `$Diff` and thought the celebrity family name used could do with an update in light of controversy around the Jackson family. Plus who doesn't like Toni Braxton 🚀


jbrown215:
Haha, thanks for this! I'll get it merged

PR #7558, opened 2019-03-15
Fix return type for _write

MatrixFrog:
The documentation doesn't say anything about what it should return, so I assume it should return void.

https://nodejs.org/dist/latest-v10.x/docs/api/stream.html#stream_writable_write_chunk_encoding_callback_1


MatrixFrog:
CLA is signed

MatrixFrog:
Taking a look at the test failures now.

mvitousek:
Thanks a lot for the PR! It looks great but unfortunately I can't merge it because 5a988d12 already made that change a few days ago. Really sorry we didn't get to it sooner, but thanks again for the contribution.

PR #7559, opened 2019-03-15
Add promisify and callbackify types

goodmind:
https://nodejs.org/api/util.html#util_util_promisify_original
https://nodejs.org/api/util.html#util_util_callbackify_original


goodmind:
This is wrong definitions, doesn't infer multiple arguments

PR #7567, opened 2019-03-18
docs: label user reported bugs with "needs triage" label

FezVrasta:
This change will help to identify verified/triaged bugs. Right now anyone can create a new issue labelled as bug and it's difficult to help triaging issues.


jbrown215:
Have you tested this? I'm unfamiliar with how tags with spaces work. Do they need to be quoted?

goodmind:
@jbrown215 just tested it in my repo, it works, but flow doesn't have "needs triage" label

jbrown215:
Now it does. Thanks!

PR #7568, opened 2019-03-18
Add BeforeUnloadEvent

btmills:
Spec: https://www.w3.org/TR/html50/browsers.html#beforeunloadevent

#6119 was a previous pull request to add this event, but it was closed without merging after the author didn't respond to a [request](https://github.com/facebook/flow/pull/6119#discussion_r251652611) to change the type for `returnValue` from `mixed` to `string` per the spec.


panagosg7:
Thanks @btmills !

Issue #1242, opened 2016-01-04
`EventHandler` should have `mixed` return type


nmote:
Currently it is `void` which means that, for example, an async function cannot be an `EventHandler` because its return type is `Promise<void>`.

I'll put up a diff shortly.

PR #7569, opened 2019-03-18
Stronger react dom refreshed

villesau:
This is revived https://github.com/facebook/flow/pull/6727 (cc. @motiz88 ) so basically just test cases updated. I don't think it makes sense to cover everything in this PR (missing todos from the original pull request). Relevant docs (copied over from previous PR):

* [`DOMProperty.js`](https://github.com/facebook/react/blob/master/packages/react-dom/src/shared/DOMProperty.js) in `react-dom`
* `react-dom` event system [implementation](https://github.com/facebook/react/tree/master/packages/react-dom/src/events) and [docs](https://reactjs.org/docs/events.html)
* [The HTML living standard](https://html.spec.whatwg.org/) and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML), naturally

Some insights from running this against our code base: 
- Should ReactDOM$GlobalEventHandlers functions return type be any/mixed instead of void?
  - e.g async functions always returns promise.
  - currently it's void (e.g [`onAbort?: (SyntheticEvent<E>) => void,`](https://github.com/facebook/flow/compare/master...villesau:stronger-react-dom-2?expand=1#diff-ce626ccb6b8eb922c17388db8d33cbabR415)) but does it matter what the function returns?
- <input /> `onChange` is SyntheticEvent, should it be SyntheticInputEvent?
  - [`onChange?: (SyntheticEvent<E>) => void,`](https://github.com/facebook/flow/compare/master...villesau:stronger-react-dom-2?expand=1#diff-ce626ccb6b8eb922c17388db8d33cbabR435)
- dangerouslySetInnerHTML expects string, should it be optional `?string`? 
  - We often pass nullable values. Should these be fixed or is it fine?
  - [`dangerouslySetInnerHTML?: { __html: string },`](https://github.com/facebook/flow/compare/master...villesau:stronger-react-dom-2?expand=1#diff-ce626ccb6b8eb922c17388db8d33cbabR586)
- <input /> and textArea onChange `event.target.value` is missing from event.target, should it?
  - example: `onChange={event => this.props.onChange({value: event.target.value})}` <- event.target.value missing
  - [`+target: EventTarget;`](https://github.com/facebook/flow/compare/master...villesau:stronger-react-dom-2?expand=1#diff-ce626ccb6b8eb922c17388db8d33cbabR137)
- ~~className only takes string or undefined, should it also allow null?~~
  - ~~example: `className={listStyle ? null : tableScrollWrapper`~~
  - ~~[className?: string,](https://github.com/facebook/flow/compare/master...villesau:stronger-react-dom-2?expand=1#diff-ce626ccb6b8eb922c17388db8d33cbabR585)~~
  - E: Now allows null. Can be made stricter later if deemed necessary.
- onClick only allows function. Should it also allow falsy values?
  - example: `onClick={busy ? null : onClick}`

The biggest cause of issues for us is inputs returning SyntheticEvent instead of SyntheticInputEvent


goodmind:
@villesau there shouldn't be `value` on `event.target`, use `event.currentTarget`

villesau:
@panagosg7 any news or comments about my findings?

villesau:
@panagosg7 has anyone had time to take a look at it or think about the findings from the first message? Would it make sense to weight in someone from React team to provide context?

lib/react-dom.js
```diff
@@ -404,19 +404,281 @@ declare type $JSXIntrinsics = {
...
+  onBeforeInputCapture?: (SyntheticInputEvent<E>) => void,
+  onBlur?: (SyntheticFocusEvent<E>) => void,
+  onBlurCapture?: (SyntheticFocusEvent<E>) => void,
+  onCancel?: (SyntheticEvent<E>) => void,
+  onCancelCapture?: (SyntheticEvent<E>) => void,
+  onCanPlay?: (SyntheticEvent<E>) => void,
+  onCanPlayCapture?: (SyntheticEvent<E>) => void,
+  onCanPlayThrough?: (SyntheticEvent<E>) => void,
+  onCanPlayThroughCapture?: (SyntheticEvent<E>) => void,
+  onChange?: (SyntheticEvent<E>) => void,
```
panagosg7: `SyntheticEvent` has a `target` property of type `EventTarget` which is missing `value`. This, however, seems to be common in examples. See https://reactjs.org/docs/forms.html#handling-multiple-inputs

lib/react-dom.js
```diff
@@ -404,19 +404,281 @@ declare type $JSXIntrinsics = {
...
+
+// TODO: Can we get an actual Flow $CaseInsensitive builtin for singleton
+// strings and enums?
+type $TODO$CaseInsensitive<Value: string> = Value;
+
+type ReactDOM$Style = Object; // TODO
+
+// NOTE: In principle this can be `number | string`, but constraining values to
+// numbers might be a useful design choice.
+type ReactDOM$Number = number;
```
panagosg7: This causes quite a few errors might wanna consider making it a `number | string`.

lib/react-dom.js
```diff
@@ -404,19 +404,281 @@ declare type $JSXIntrinsics = {
...
+  onVolumeChangeCapture?: (SyntheticEvent<E>) => void,
+  onWaiting?: (SyntheticEvent<E>) => void,
+  onWaitingCapture?: (SyntheticEvent<E>) => void,
+  onWheel?: (SyntheticWheelEvent<E>) => void,
+  onWheelCapture?: (SyntheticWheelEvent<E>) => void,
+};
+
+type ReactDOM$ElementProps = {
+  children?: React$Node,
+  className?: string,
```
panagosg7: This often appears to be `null`, `void` or `String`.

lib/react-dom.js
panagosg7: Same holds for properties like `checked`, and less frequently for `files`, `id`, `blur, etc.

panagosg7: This probably has to do with the typing of `target` in `SyntheticEvent`:

https://github.com/facebook/flow/pull/7569/files#diff-ce626ccb6b8eb922c17388db8d33cbabR137

goodmind: `target` is fine as is, `currentTarget` should be used

pascalduez: > `currentTarget` should be used

Not on input or select elements, where we most likely don't need/want event propagation.

pascalduez: I usually type `onChange` handlers as `onChange?: (SyntheticInputEvent<>) => void`.   
Does it means it will break when this lands?

villesau: Do we have some sort of consensus here? :D

lib/react-dom.js
villesau: true. But is `null` correct? Though, the defs can can always be made stricter later if decided so.

lib/react-dom.js
goodmind: @pascaldues, what you mean by event propagation? it IS `target` that propagates events

MichaReiser:
Hi everyone. I stumbled upon this PR because I was interested in how event handlers in react should be typed (returning `void` or `mixed`?).

I looked at some other PR's and it seems that flow allows `mixed` return type for [React lifecycle](https://github.com/facebook/flow/pull/4886) methods as well as [DOM event handlers](https://github.com/facebook/flow/issues/1242). Might it, therefore, make sense to type the event handlers with `mixed` as return type since it doesn't matter for react whatever is returned? This would also reduce the amount of breaking changes introduced by this PR since all existing callbacks will have the right return type.

lib/react-dom.js
villesau: @panagosg7 @pascalduez @goodmind so what should this type actually be?

lib/react-dom.js
villesau: made null possible.

lib/react-dom.js
villesau: made it `number | string`

villesau:
@panagosg7 addressed all the comments except the one which has some confusion around it.

lib/react-dom.js
villesau: @panagosg7, @goodmind pointed to this in Discord: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11508#issuecomment-256045682

Does this make sense to you?

goodmind:
I would love to see SyntheticEvent exported in `react` aka `import {SyntheticEvent} from 'react'` and not global

villesau:
pint @panagosg7 any news?

panagosg7:
Sorry for the delay. I tried these changes internally, but I'm afraid the overhead is more than we can afford at this point. In addition to the error diff, which we could probably work around, the merge time increased by more than 20%. This makes it really hard to move forward with this.

villesau:
@panagosg7 what do you mean by merge time?

This obviously is a big change: Right now React is totally untyped around this area which is super risky and bug prone. Not only once or twice types around this would have caught bugs that we had later noticed too late.

In addition to added safety this would impact dev experience since after this Flow would start to provide better auto complete for plenty of React specific functions I believe. In overall I'm quite confident this would make developers more confident with Flow.

TS is currently doing very well around this area: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2499

panagosg7:
By merge time I mean the main Flow's main checking procedure.

I am sold with and very appreciative of the improvement in checking that this offers, but we've invested about a year on performance gains, so have to be extra careful in making sure we deliver on that account. (Realistically, even if I landed this change as is, it would possibly get reverted come release time.)

In the meantime, until we figure out why this lib change causes this perf regression, can these typings be imported to your project so that you can maintain the checking benefits? (In TS these typings seem to live in Definitely typed rather than the main TS repo.)

goodmind:
@panagosg7 this would require per-lib ignore as in 

https://github.com/facebook/flow/pull/7732, so that I can ignore only `react` from core libs

villesau:
Oh, the perf issue sounds significant :/

In case of TS I think it does not have (almost) any builtin library definitions. Everything tends to live in definitelytyped, even Node definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node

Hypnosphi:
Will it fix the following false negative?

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H31hgwAEgCiADxw1Gw-C3wAuMAAsuUMOUVwArjAAmYAEY01hfBgB2Ac1QAeUkpUBeAN4BGAEwBmAL5hgAPlQgA

```
/* @flow */

// $ExpectError: href should be string
<a href={123} />
```

goodmind:
Yes, it should fix it

SamChou19815:
Should be replaced by https://github.com/facebook/flow/commit/ed61eb78b88748a2e80d4aa7ac2603590725d69a

PR #7570, opened 2019-03-19
Avoid using `any`'s in common Date class

nvie:
This causes any JS library exporting `Date` instances to blow up on the `dynamic-export` linter flag. Disabling the flag is not an option for library authors, since any project using that library will also have to disable that linter flag in turn.


mroch:
thanks!

Issue #5134, opened 2017-10-18
Convert Optional Type Properties to Required


arahansen:
I have a fairly common use case of typing server responses (specifically graphql responses) that contain many optional properties.

For example:
```js
type Animal = {
  kind?: string,
  age?: number
}
```

Where kind and age may or may not be defined for the response. I'd like to use this same type definition for my React components, the difference being `kind` and `age` are required properties that I would cover with defaultProps. So my React component might look like this:

```js
type Props = {
  kind: string,
  age: number
}

class Animal extends React.Component<Props> { 
  static defaultProps = {
    kind: 'dog',
    age: 2
  }
  ...
}
```

is my only option to rewrite the entire type definition of Animal, minus the `?` syntax? Or is there a way to tell flow that I want to use the Animal type, but make all properties required properties?

apsavin:
You do an opposite thing

[try](https://flow.org/try/#0C4TwDgpgBAggdgSwLYEMA2UC8UDeAoKKAawTgBMAuKAZ2ACdSBzAGgKhUYirgFckAjCHTwBfPHlCRYiVGgDKQgG5CAShGpgA9nGrRsAEjkALFJAA88ZOgB8AbnGKUdKHRT9+CYFUuysuYqSUUADkru6ewczsnFQAjFAi9o7OYR7ACnTKdGoa2rreMugZWTlaOnr+JORUoW5pwQn2eAD0zVAAonR0mnTUFHjJLnWexarqZfnSVvJKY7nlfjhQAO7dcIwACt1gVPQ80IkDTkPhXlO+2EtVQbWnDYlAA)

skyrpex:
`$Shape` works great! Why isn't it documented?

apsavin:
@mrkev Do you know why no one from [PRs with docs for $Shape](https://github.com/facebook/flow/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+%24shape) do not merged?

mrkev:
no idea, maybe it fell of the radar ¯\\_(ツ)_/¯

FlavienBusseuil:
Any progress on the original issue by any chance?

gkz:
> Any progress on the original issue by any chance?

New type utilities like this are blocked by another project which we are actively working on.

SamChou19815:
We now have `Required`

PR #7571, opened 2019-03-19
$Required utility type

goodmind:
Fixes https://github.com/facebook/flow/issues/5134

It doesn't touch `T | void` unions


dsainati1:
This looks like a good approach, but I have a couple of concerns.

I'm not entirely convinced by how this utility type behaves with respect to class and instance types. You are able to cast something of type `PointC` to `$Required<PointB>`, which seems to violate the nominality classes are supposed to have. I think because you are producing an object type in the code where you handle instances, your check ignores the class hierarchy which I don't think is what we'd want. The code as written also only handles own props. Is there a good reason we'd want this utility to work on instances in the first place? It seems like this would be much simpler if we restrict it to object types. 

Your provided tests also have some discrepancies in them: you add an error comment on 
`(pc: $Required<PointB>); // error`, `(ps: $Required<{ x?: number }>); // error`, and `(pi: $Required<{ x?: number }>); // error` but no error is reported in these places. Did you intend for there to be an error on these casts?

I'd also like to see a few tests that ensure that property variance is properly maintained by this utility.

goodmind:
@dsainati1 it converts classes to objects, just like `$ReadOnly` and `$Rest` does

src/typing/flow_js.ml
```diff
@@ -7351,6 +7407,7 @@ and eval_destructor cx ~trace use_op reason t d tout = match t with
   | ReadOnlyType ->
     let open Object in
     ObjKitT (use_op, reason, Resolve (Next), ReadOnly, tout)
+  | RequiredType -> RequiredT (reason, tout)
```
jbrown215: Instead of adding a new use type, can you follow the way ReadOnly is implemented in object_kit and do the same? It will significantly simplify your code by handling a ton of edge cases for you.

For example, your code will error on mixed/empty, which the object_kit code can handle more elegantly.

You also won't need the IncompatibleRequiredT because object kit handles errors with types that can't be converted into an object.

You also shouldn't need to distinguish between instances and objects, since object_kit will take care of that for you.

newtests/Required/test.js
```diff
@@ -0,0 +1,376 @@
+/*
```
jbrown215: Please add tests that use polymorphic type parameters + $Required

jbrown215: Please also add tests that work on unions + $Required

src/typing/flow_js.ml
goodmind: But object_kit doesn't have polarity

goodmind: If I change `and prop = TypeTerm.t * bool (* own *)` to `and prop = TypeTerm.t * bool (* own *) * TypeTerm.polarity` in type.ml, would it be okay in object_kit?

goodmind: Also there is issue in object_kit with statics, maybe you know how to fix it?
https://github.com/facebook/flow/issues/7572

goodmind:
I don't understand why contravariant properties get replaced with mixed in object_kit
https://github.com/facebook/flow/blob/master/src/typing/flow_js.ml#L10884-L10888
```ocaml
| None ->
      let reason = replace_reason_const (RUnknownProperty (Some x)) r in
      let t = DefT (reason, bogus_trust (), MixedT Mixed_everything) in
      t
    in
```

src/typing/type.ml
```diff
@@ -1981,7 +1982,7 @@ and Object : sig
   (* A union type resolves to a resolved spread with more than one element *)
   and resolved = slice Nel.t
 
-  and slice = reason * props * dict * TypeTerm.flags
+  and slice = reason * props * dict * TypeTerm.flags * Properties.t
```
goodmind: I added original props (raw_props) because I needed polarity from field structure, and also doesn't wanted to get `mixed` on write-only properties

src/typing/flow_js.ml
```diff
@@ -11276,6 +11280,36 @@ and object_kit =
...
 
+  let object_required =
+    let mk_required_object cx reason slice =
+      let (r, _, dict, flags, props) = slice in
+
+      let props = SMap.map (fun p -> 
+        match p with
+          | Field (_, DefT (_, _, OptionalT t), polarity) -> Field (None, t, polarity)
+          | Field (_, t, polarity) -> Field (None, t, polarity)
+          | Get (_, t) -> Get (None, t)
```
goodmind: This handles getters/setters in object, but not in classes, not sure why

src/typing/flow_js.ml
```diff
@@ -10985,15 +10988,16 @@ and object_kit =
           Obj_type.sealed_in_op reason flags1.sealed &&
           Obj_type.sealed_in_op reason flags2.sealed;
       } in
-      reason, props, dict, flags
+      let raw_props = raw_props1 in
```
goodmind: Probably should merge original props also? But they don't get used here

goodmind:
Pick implemented with `$Required`

```js
type Pick<T1, T2> = $Rest<
  T1,
  $ObjMap<$Required<$Rest<T1, T2>>, <V>(V) => any>
>;

declare var picked: Pick<{| a?: number, b?: string, c: string |}, {| a: any |}>;

picked.a;
picked.b;
picked.c;
```

goodmind:
@jbrown215 friendly ping here

jbrown215:
I can't make time for this yet. I'm spending all my time on a big hurdle in my spreads project. Once I get past that, this will be the first thing I review.

goodmind:
The more i think of this, probably there should be more general solution, i.e. some mapper or even $ObjMap itself should be allowed to configure property type descriptor (similar to js, but not identical)

```js
type D = {|+hello: number|}
type M = {|hello?: number|}
type S = {|-hello: number|}

type Descriptor<T> = {|
   optional: boolean,
   enumerable: boolean, // ???
   value: T,
   polarity: 
     | '+' // positive 
     | '-'  // negative
     | 'N'  // neutral
|}

type $ReadOnly<T> = $ObjModify<
  T, 
  <Descriptor>(Descriptor) => {|...Descriptor, polarity: '+'|}
>
type $WriteOnly<T> = $ObjModify<
   T, 
   <Descriptor>(Descriptor) => {|...Descriptor, polarity: '-'|}
>
type $ReadWrite<T>  = $ObjModify<
  T, 
  <Descriptor>(Descriptor) => {|...Descriptor, polarity: 'N'|}
>
or
type $Invariant<T> = $ObjModify<
  T, 
  <Descriptor>(Descriptor) => {|...Descriptor, polarity: 'N'|}
>

type $Required<T> = $ObjModify<
  T, 
  <Descriptor>(Descriptor) => {|...Descriptor, optional: false|}
>
```

dsainati1:
So after chatting to some people on the team I think we would like to decrease the number of `$` types in Flow's system, so I am going to close this since it is moving us in slightly different direction that we'd like. This can be equivalently implemented with `$ObjMap` in any case.

goodmind:
@dsainati1 this can't be implemented with `$ObjMap`

PR #7575, opened 2019-03-20
Avoid using `Object` in server options

HsuTing:
If `handle` is `Object`, this will cause this error:

```js
server.listen(true); // true is considered as Object.
```

Docs: [net_server_listen_options_callback](https://nodejs.org/api/net.html#net_server_listen_options_callback).


goodmind:
@HsuTing please post link to the docs about this argument

HsuTing:
I added.

mroch:
cool!

PR #7576, opened 2019-03-20
Fix getters and setters printing

goodmind:
Fixes getters and setters printing
```
set a(number): void
get a(number): number
```

### DIFF
--- a/src/common/ty/ty_printer.ml
+++ b/src/common/ty/ty_printer.ml
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
       | Get t -> fuse [
           Atom "get"; space;
           to_key key;
+          Atom "(";
+          Atom ")";
+          Atom ":";
           type_ ~depth t;
         ]
 
       | Set t -> fuse [
           Atom "set"; space;
           to_key key;
+          Atom "(";
+          type_ ~depth t;
+          Atom ")";
+          Atom ":";
           type_ ~depth t;
         ]
       end
--- a/src/common/ty/ty_printer.ml
+++ b/src/common/ty/ty_printer.ml
@@ -220,23 +220,23 @@ let type_ ?(size=5000) ?(with_comments=true) t =
           type_function ~depth ~sep:(Atom ":") func;
         ]
 
-      | Get t -> fuse [
+      | Get t -> group [
           Atom "get"; space;
           to_key key;
           Atom "(";
+          softline;
           Atom ")";
           Atom ":";
+          pretty_space;
           type_ ~depth t;
         ]
 
-      | Set t -> fuse [
+      | Set t -> group [
           Atom "set"; space;
           to_key key;
-          Atom "(";
-          type_ ~depth t;
-          Atom ")";
+          wrap_and_indent (Atom "(", Atom ")") [type_ ~depth t];
           Atom ":";
-          type_ ~depth t;
+          type_ ~depth Void;
         ]
       end
 


src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
+          Atom ")";
+          Atom ":";
           type_ ~depth t;
         ]
 
       | Set t -> fuse [
           Atom "set"; space;
           to_key key;
+          Atom "(";
+          type_ ~depth t;
```
mroch: printing `t` twice seems odd to me. should the return type be void instead?

goodmind:
```
const m = {set foo(value) {}}
let d = m.foo = 1
d // 1
```

goodmind: Also, I think this should be properties with variance annotations instead in types, but not sure if this is intended

goodmind: You also can declare any type in return, but it doesn't have effect, so perhaps setters should error on specifying return type entirely
```js
const bar = {
  set a(value: string): number {
    return 1;
  }
};
let d = (bar.a = '')
d // ''
```

samwgoldman: in `let d = m.foo = 1`, `d` is `1` because an expression statement evaluates to the right-hand side, not because the return type of the setter.

goodmind: @samwgoldman I think it still makes sense because you can't get `void` from setter

goodmind: But maybe setters and getters should be displayed as covariant and contravariant properties, and regular property if it is both

samwgoldman: IMO it's fine to just print `void` for the return type. We can either (1) ban return type annotations / enforce void return or (2) allow the annotations and just store the return type along with the setter property, even though it is ignored.

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
       | Get t -> fuse [
```
mroch: change to `group`

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
           Atom "get"; space;
           to_key key;
+          Atom "(";
```
mroch: either add `softline` between the parens, or change to `wrap_and_indent (Atom "(", Atom ")") []`, to allow the line to wrap between the parens.

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
+          Atom ")";
+          Atom ":";
```
mroch: add `pretty_space` after the colon

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
           type_ ~depth t;
         ]
 
       | Set t -> fuse [
```
mroch: change `fuse` to `group`

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
           Atom "set"; space;
           to_key key;
+          Atom "(";
```
mroch: use `wrap_and_indent (Atom "(", Atom ")") [type_ ~depth t]`

src/common/ty/ty_printer.ml
```diff
@@ -223,12 +223,19 @@ let type_ ?(size=5000) ?(with_comments=true) t =
...
+          type_ ~depth t;
+          Atom ")";
+          Atom ":";
```
mroch: let's drop the return type. it adds noise and I think we'll either make it a parse error to specify one or only allow `void`. but either way you shouldn't/can't return a value.

mroch:
thank you!

Issue #5705, opened 2018-01-23
Can not use String.raw with Array<string>


HsuTing:
https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBjOA7AzgFzHwFMCB9AgJwEtsBzALjCtrrAF4wBlfG+gOkoBDBAAoA3qjBhhCJgHISBeagC+ASgDc6LHkJL8ZIZWEBPJi3oduvVoJESpMkUwDai0vnkBdNVqA

It works when I run the code. How can I fix this bug?

idiostruct:
Convert [raw](https://github.com/facebook/flow/blob/master/lib/core.js#L327)'s value to a string:

```typescript
const test_array: string = String.raw({
  raw: ['test'].join("")
});
```

HsuTing:
This is a solution, but `String.raw` is supported to use `Array` in `{raw}`.
Can not change the flowType of the String?

HsuTing:
I find the declare of `String`, [here](https://github.com/facebook/flow/blob/master/lib/core.js#L327).
I try to change the `callSite: $Shape<{raw: string}>` to `callSite: $Shape<{raw: string | Array<string>}>`.
However, it does not work.
How can I change this declare and test in local?

jcready:
Use this:

```js
callSite: $Shape<{raw: string | $ReadOnlyArray<string>}>
```

HsuTing:
But I does not know how to test in local.
I did not find the document about the development.

Issue #2616, opened 2016-10-11
Incomplete support for tagged template literals: String.raw doesn't work


Macil:
It seems like Flow's type definition for String.raw is broken because String.raw appears to always fail:

``` js
/* @flow */
const s = String.raw `abc\ndef ${'aaa'} \n`;
console.log(s);
```

```
$ flow version
Flow, a static type checker for JavaScript, version 0.33.0
$ flow
test.js:2
  2: const s = String.raw `abc\ndef ${'aaa'} \n`;
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ encaps tag. Function cannot be called on any member of intersection type
  2: const s = String.raw `abc\ndef ${'aaa'} \n`;
               ^^^^^^^^^^ intersection
  Member 1:
  274:     static raw(templateString: string): string;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: ../flow/flowlib_2457d6cd/core.js:274
  Error:
    2: const s = String.raw `abc\ndef ${'aaa'} \n`;
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ array. This type is incompatible with
  274:     static raw(templateString: string): string;
                                      ^^^^^^ string. See lib: ../flow/flowlib_2457d6cd/core.js:274
  Member 2:
  275:     static raw(callSite: $Shape<{raw: string}>, ...substitutions: any[]): string;
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See lib: ../flow/flowlib_2457d6cd/core.js:275
  Error:
  275:     static raw(callSite: $Shape<{raw: string}>, ...substitutions: any[]): string;
                                       ^^^^^^^^^^^^^ object type. Expected object instead of. See lib: ../flow/flowlib_2457d6cd/core.js:275
    2: const s = String.raw `abc\ndef ${'aaa'} \n`;
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ array


Found 1 error
```

Here's Flow's type definition for String.raw:

```
    static raw(templateString: string): string;
    static raw(callSite: $Shape<{raw: string}>, ...substitutions: any[]): string;
```

[...]

PR #7580, opened 2019-03-21
Template strings array

goodmind:
Fixes https://github.com/facebook/flow/issues/5705, https://github.com/facebook/flow/issues/2616

I'm not sure if core feature should rely on library definitions, and perhaps this would break already written code with tagged template strings

https://tc39.github.io/ecma262/#sec-gettemplateobject

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -239,6 +239,10 @@ declare class $ReadOnlyArray<+T> {
     +length: number;
 }
 
+declare class TemplateStringsArray extends $ReadOnlyArray<string> {
+  +raw: $ReadOnlyArray<string>;
+}
+
 declare class Array<T> extends $ReadOnlyArray<T> {
     copyWithin(target: number, start: number, end?: number): T[];
     every(callbackfn: (value: T, index: number, array: Array<T>) => any, thisArg?: any): boolean;
@@ -334,7 +338,7 @@ declare class String {
     static fromCharCode(...codes: Array<number>): string;
     static fromCodePoint(...codes: Array<number>): string;
     static raw(templateString: string): string;
-    static raw(callSite: $Shape<{raw: string}>, ...substitutions: any[]): string;
+    static raw(callSite: TemplateStringsArray, ...substitutions: any[]): string;
 }
 
 declare class RegExp {
--- a/lib/core.js
+++ b/lib/core.js
@@ -239,7 +239,7 @@ declare class $ReadOnlyArray<+T> {
     +length: number;
 }
 
-declare class TemplateStringsArray extends $ReadOnlyArray<string> {
+type TemplateStringsArray = interface extends $ReadOnlyArray<string> {
   +raw: $ReadOnlyArray<string>;
 }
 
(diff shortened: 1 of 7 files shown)


lib/core.js
```diff
@@ -334,7 +338,7 @@ declare class String {
     static fromCharCode(...codes: Array<number>): string;
     static fromCodePoint(...codes: Array<number>): string;
     static raw(templateString: string): string;
-    static raw(callSite: $Shape<{raw: string}>, ...substitutions: any[]): string;
+    static raw(callSite: TemplateStringsArray, ...substitutions: any[]): string;
```
Bannerets: `static raw(templateString: string): string;` is invalid (`String.raw('str')` throws an error).

Probably `raw` declaration should be like this:

```js
  static raw(callSite: { +raw: $ReadOnlyArray<string>; ... }, ...substitutions: any[]): string;
```

By this It also accepts a `TemplateStringsArray` instance and objects like `{ raw: ['ab', 'cd'] }`.

Test cases:

```js
String.raw`str` // ok
String.raw`a${2}b` // ok
String.raw({ raw: ['ab', 'cd'] }, 1) // ok
String.raw('str') // error
String.raw({ k: 'v' }) // error
```

goodmind:
This is not exactly library definitions, it is also typing part

prelude/prelude.js
```diff
@@ -16,6 +16,10 @@ declare class $ReadOnlyArray<+T> {
   @@iterator(): Iterator<T>;
 }
 
+type TemplateStringsArray = interface extends $ReadOnlyArray<string> {
```
lydell: The items of the array can be undefined.

```
$ node
> (x => x)`\u`
[ undefined ]
> (x => x.raw)`\u`
[ '\\u' ]
```

¯\\\_(ツ)_/¯

goodmind: Why would you use tags with `undefined` array

lydell: Because somebody accidentally put an "invalid escape" in there.

goodmind: Then it would error?

lydell: Yeah, if the author of the tag expects all items to be string it'll probably crash with `TypeError: Cannot read property 'slice' of undefined` or something like that.

goodmind: I mean typechecker would error

lydell: Oh, cool, I didn't even think of that the typechecker could statically know if there will be `undefined`s in the array or not!

I tried to see what's it like today on [Try Flow](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVUCuA7AxgFwEs5swo44AKQ-AUwFsBnALjAEEAnDgQwE8AeRvg6FsAcwB8ASlZCR4sAG9UYMB1r5MHUjQaMAdPW4AHSozABeCWAP44AVWPHaHAMLdGtSlKn6AVnCi3gDcqAC+6ORwAAbcADqY0ahAA) but it gave me a parse error.

```
/* @flow */

function foo(items: Array<string>): string {
  return items.map(s => s.toUpperCase()).join();
}

foo`a\u`
```

```
 foo`a\u`
         ^ Unexpected token ILLEGAL
```	

It would be awesome though if Flow could detect that I'm trying to call `foo` with `Array<string | void>` which is incompatible.

goodmind: How can I get `undefined` without using `\u`?

lydell: > How can I get `undefined` without using `\u`?

This article seems to list all the cases:

http://2ality.com/2016/09/template-literal-revision.html#problem-some-text-is-illegal-after-backslashes

> TypeScript has same parse error `Hexadecimal digit expected.`

Yeah, maybe they haven't made updates for this ES2018 feature yet. I'll try and see if I can find a github issue about it.

PR #7586, opened 2019-03-25
[DOCS] Update tuples.md

segmentationfaulter:
If a tuple is defined as follows:

```js
let tuple: [number, boolean, string] = [1, true, "three"];
```

and an access to one of its elements is attempted like this:

```js
let none: void = tuple[3];
```

Access attempt will result in an error rather than returning a value of type `void`, so suggesting this change.


panagosg7:
Thanks @segmentationfaulter !

PR #7587, opened 2019-03-25
fix native event annotations for mouse based events

Kiwka:
Current annotation of the `SyntheticDragEvent`, `SyntheticWheelEvent`, `SyntheticPointerEvent` leads to wrong annotations for the nativeEvents (in all these three events annotation for native event will be having type `MouseEvent`). 

Flow docs themselves contain information, how SyntheticEvent<T> types that React provides and the DOM event are relate for 2 of 3 changed annotations: https://flow.org/en/docs/react/events/

### DIFF
--- a/lib/react-dom.js
+++ b/lib/react-dom.js
@@ -198,7 +198,7 @@ declare class SyntheticKeyboardEvent<
   which: number;
 }
 
-type $SyntheticMouseEvent = {|
+class $SyntheticMouseEvent {
   altKey: boolean;
   button: number;
   buttons: number;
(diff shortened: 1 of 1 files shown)


lib/react-dom.js
```diff
@@ -215,17 +213,25 @@ declare class SyntheticMouseEvent<
   screenX: number;
   screenY: number;
   shiftKey: boolean;
+|}
+
+declare class SyntheticMouseEvent<
+  +T: EventTarget = EventTarget,
+> extends SyntheticUIEvent<T, MouseEvent> {
+  ...$SyntheticMouseEvent,
```
goodmind: Spread in classes?

Kiwka: 🤦‍♀️ good point. Alternatively, I was thinking about introducing second param to `SyntheticMouseEvent` that would accept `Event`, and will default to `MouseEvent`

goodmind: maybe `mixins`?
```
declare class A {
  a: string
}
declare class C {
  b: string
}
declare class B extends C mixins A {}

const b = new B()

b.a;
b.b;
```

lib/react-dom.js
```diff
@@ -198,9 +198,7 @@ declare class SyntheticKeyboardEvent<
   which: number;
 }
 
-declare class SyntheticMouseEvent<
-  +T: EventTarget = EventTarget,
-> extends SyntheticUIEvent<T, MouseEvent> {
+class $SyntheticMouseEvent {
```
nmote: How about making this extend `SyntheticUIEvent`, take two type parameters, and just pass them through?

e.g.:

```
declare class $SyntheticMouseEvent<+T: EventTarget, NativeEvent> extends SyntheticUIEvent<T, MouseEvent> { ... }
```

and then below:

```
declare class SyntheticMouseEvent<+T: EventTarget = EventTarget> extends $SyntheticMouseEvent<T, MouseEvent> { ... }
```

And similar for the others below. I don't think `mixins` is really necessary here.

Kiwka: I made changes very nearly to what was suggested. Please, take a second look whenever you'll have free time

Kiwka:
Is there anything I can do to move this PR forward?

nmote:
Sorry, I've been meaning to take a second look but haven't gotten to it. I'll look now.

nmote:
The changes here look good to me. I imported it to test against our codebases to make sure it doesn't have any unexpected effects, but we are having some issues with tooling currently so I will try again tomorrow. Assuming everything looks good I will re-record the tests and get it reviewed.

Thanks for your patience, and sorry again for the delay.

nmote:
All the results look good, I'll re-record the tests tomorrow

PR #7593, opened 2019-03-26
Number literal object keys typechecking

### LINKED ISSUE #380: "non-string literal property keys not supported"

goodmind:
Fixes #380 

Doesn't allow in interfaces
Doesn't allow in class declarations

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -4573,6 +4573,30 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
           rec_flow cx trace (DefT (r1, trust, ArrT (TupleAT (t1, ts1))), u)
       end
 
+    (* Tuples can flow to numeric objects *)
+    | DefT (r1, trust, ArrT (TupleAT (_, ts1))),
+      UseT (use_op, DefT (r2, _, ObjT o)) ->
+      let fresh = (desc_of_reason r1) = RArrayLit in
+      let fields = Core_list.foldi ts1 ~f:(fun i map value -> 
+        let key = (Dtoa.ecma_string_of_float (float_of_int i)) in
+        let prop = Context.get_prop cx o.props_tmap key in
+        let use_op = Frame (TupleElementCompatibility {
+          n = i + 1;
+          lower = r1;
+          upper = r2;
+        }, use_op) in
+        let value = match prop with 
+          | Some field ->
+            (match Property.read_t field with
+              | Some t2 -> flow_to_mutable_child cx trace use_op fresh value t2;
+              | None -> ());
+            field
+          | None -> Field (None, value, Neutral) in
+        SMap.add key value map
+      ) ~init:SMap.empty in
+      let props_tmap = Context.make_property_map cx fields in 
+      rec_flow cx trace (DefT (r1, trust, ObjT {o with props_tmap;}), u)
+
     (* Read only arrays are the super type of all tuples and arrays *)
     | DefT (r1, _, ArrT (ArrayAT (t1, _) | TupleAT (t1, _) | ROArrayAT (t1))),
       UseT (use_op, DefT (r2, _, ArrT (ROArrayAT (t2)))) ->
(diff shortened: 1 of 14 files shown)


goodmind:
Thanks @mroch for this library https://github.com/flowtype/ocaml-dtoa

nmote:
Not sure who should review this. @dsainati1, @jbrown215, or @mvitousek, any of you have some time to spare for this?

dsainati1:
Given that @mroch had raised concerns on #380 I'd like to get his opinion on this, but at the very least I think we'd want to see more tests here. It would be good to test number keys with different variance, as well as cases where objects are declared with number/string keys that are in conflict. I.e., what should happen here?
```
let a = {1 : "A", "1" : 3};
(a[1] : number)
```

goodmind:
@dsainati1 it does the same as `{'1': 'A', '1': 3}` i.e. uses latter definition, added tests

goodmind:
Parser test for contravariant number keys fails, probably it parses it as negative number

goodmind:
Fixed contravariant properties parsing

goodmind:
image
This probably should display numbers too

goodmind:
@dsainati1 do you think tuples should be affected by this?
```js
//@flow

declare var e: { 0: number, 1: number, length: 2 };
declare var b: [number, number];

(e[0]: number);
(e[1]: number);

(b[0]: number);
(b[1]: number);

const m: { 0: number, 1: number, length: 2 } = b;
const m2: [number, number] = e;
```

src/typing/flow_js.ml
```diff
@@ -5297,8 +5297,13 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
...
       let propref = match l with
       | DefT (reason_x, _, StrT (Literal (_, x))) ->
-          let reason_prop = replace_reason_const (RProperty (Some x)) reason_x in
-          Named (reason_prop, x)
+        let reason_prop = replace_reason_const (RProperty (Some x)) reason_x in
+        Named (reason_prop, x)
+      | DefT (reason_x, _, NumT (Literal (_, (x, _)))) ->
+        let x = Dtoa.ecma_string_of_float x in
+        let reason_prop = replace_reason_const (RProperty (Some x)) reason_x in
+        (* Probably should be different type constructor *)
```
goodmind: Looks like this broke `{[key: number]: any}` reads,
```js
declare var e: { [key: number]: any };

e[0]; // string `0` [1] is incompatible with  number [2].Flow(InferError)
```

goodmind: I think this https://github.com/facebook/flow/pull/2293#issuecomment-261344263 applies to this also, so probably I fixed it wrong

goodmind:
This PR probably shouldn't mix number keys with string keys

goodmind:
😃I forgot about this PR completely. I think to split parsing from typechecking in two PRs
@mroch do you have anything to say about number keys?

goodmind:
Moved parsing here #7798

nnmrts:
@goodmind Wow, thanks for your work on this! Do you have any updates for us?

goodmind:
@nnmrts I don't think this would get in as-is, I hope at least parsing would get reviewed

nnmrts:
@goodmind So it seems you're just getting ignored here. Nice! 👍👍👍

goodmind:
@nnmrts not really, I have tons of PR merged already

nnmrts:
Any update on this? 😆

nnmrts:
Really? It's almost 9 months now!? This is ridiculous. Maintainers, then open up flowlib (and js types) for developers at least, if you don't want to fix major bugs like this one.

Flow is officially an unusable typechecker for 2014 javascript now. Numerical keys are standard since JUNE 2015!???

How is this even a typechecker when half of the types aren't following any standard?

PR #4175, opened 2017-06-14
Use SingletonStrT in case of const literal declarations

agentcooper:
```
// @flow

const x = 'X';
const y: typeof x = 'Y';
```

Before: no errors.

After:
```
Cannot assign 'Y' to y because string [1] is incompatible with typeof x [2].

        1│ // @flow
        2│
        3│ const x = 'X';
 [2][1] 4│ const y: typeof x = 'Y';
        5│
```

Fixes https://github.com/facebook/flow/issues/2639.

This is my first attempt to dive into Flow codebase, not sure if the implementation is sane. I would like to get some comments on this. Will add tests if everything is ok.

PR #7607, opened 2019-04-02
Use SingletonStrT, SingletonNumT, SingletonBoolT in case of const literal declarations

### LINKED ISSUE #2639: `typeof` string literal
This is more of a question, than a bug report. Flow allows this:

``` js
const x = 'X';
const y: typeof x = 'Y';
```

It implies that inferred either `typeof` doesn't return the lowest possible, or that type of `x` is `string`, not `'X'`, although literal value of `x` is preserved:

``` js
const x = 'X';
const z: 'X' = x;
```

The question is: why it is implemented this way? It makes impossible to define literal type and instance of that type without duplication:

``` js
const Foo = 'Foo';
type FooT = typeof Foo; // doesn't do what you'd expect
```

goodmind:
Revived https://github.com/facebook/flow/pull/4175
Fixes #2639


goodmind:
@mrkev probably

mrkev:
Let's check this out

mrkev:
I'm seeing issues similar to this:

```
const foo = {
  bar: 10,
  baz(x: number) {
    foo.bar = x; // Error, number incompatible with number
  },
}
```

Also tagging @nmote so this PR get some visibility from someone in the team.

goodmind:
@mrkev I'm not sure how this is related, because it only uses affects const primitive literals, not object literals properties, I'll check

goodmind:
Maybe you mean something like this
```js
const bar = 10;
const foo = {
  bar,
  baz(x: number) {
    foo.bar = x; // Cannot assign `x` to `foo.bar` because  number [1] is incompatible with  number
  },
}
```

goodmind:
Yeah, totally bug, I should replace `reason` of `DefT`

goodmind:
Now this errors correctly
```
Cannot assign `x` to `foo.bar` because  number [1] is incompatible with  number literal `10` [2]
```

goodmind:
@mrkev do you think this shouldn't error at all?

goodmind:
It's interesting that in TypeScript if type is inferred then it's widened, but if it's specified by user it isn't

```ts
const str = 'string';
const num = 200;
const bool = true;
const str1: 'string' = 'string';
const num1: 200 = 200;
const bool1: true = true;

const foo = {
  str,
  num,
  bool,
  str1,
  num1,
  bool1,
  method() {
    foo.str = 'any string' // no error
    foo.num = 300 // no error
    foo.bool = false // no error
    foo.str1 = 'any string' // error
    foo.num1 = 300 // error
    foo.bool1 = false // error
  }
}
```

goodmind:
I think it depends if this code is bug or not
https://flow.org/try/#0MYewdgzgLgBAYgeQQLhgIkQtMC87NoDcAUMADYCGEEMAQhQE4wDexMMUAngA4Cmu8JCXYALXmTIgAFAEoWbdhxEBLCADoufAWjDLgvIgoC+xE6EiwARowGt2m3qkwAaBWInS5dxdYYae-Hg6egbCMCZGhEA

goodmind:
Would it make sense to lose literal type in invariant and contravariant positions?

goodmind:
I'm now not sure what the cases for this are, refinements and switch seems to work fine with just regular const

mrkev:
> I think it depends if this code is bug or not
> https://flow.org/try/#0MYewdgzgLgBAYgeQQLhgIkQtMC87NoDcAUMADYCGEEMAQhQE4wDexMMUAngA4Cmu8JCXYALXmTIgAFAEoWbdhxEBLCADoufAWjDLgvIgoC+xE6EiwARowGt2m3qkwAaBWInS5dxdYYae-Hg6egbCMCZGhEA


I think what typescript does falls in the right direction, and would personally err on the side of widening, though this is a good question to bring up on the Discord channel for discussion IMO.

goodmind:
So this is stalled because we didn't found out how to widen literals to `string` in mutable locations. This would definitely need another kind of `SingletonStrT`, and more handling related to variance?

goodmind:
@mrkev what you think about implementing const assertions instead?
https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/#const-assertions

goodmind:
Related issue https://github.com/facebook/flow/issues/7961

goodmind:
@mrkev 

```js
declare var foo: {+a: 'str'}
declare var foo_a: 'str';

const bar = {
  a: foo.a,
  b: foo_a,
  method() {
    bar.b = 'bar' // error
    bar.a = 'bar' // no error
  }
}
```


https://flow.org/try/#0CYUwxgNghgTiAEA3W8BmB7dAueBvA1FDgOQDOALjMQL4BQoksCyMamA+kfGZcQNy1aYdADsK8AEYoAvHlrx4XDOgB0UADTzJOZZ00KAtiHIALdMAAUASjkKFUmConxZxB8S33Yal93da6aiA

Kinda strange, how can we get `bar.b` work the same as `bar.a`?

goodmind:
@mkrev I solved this problem, should I add more tests?

goodmind:
This actually makes it inconsistent with `let`, hmm

PR #7620, opened 2019-04-06
Implement Object.freeze for arrays/tuples

### LINKED ISSUE #961: Union from array literal
Let's say a value can be one of a finite number of string values:

``` js
type State = 'disconnected' | 'connecting' | 'connected';
let state: State = ...;
```

If I already have an array of all valid states in my code somewhere, it'd be nice if it could be used to create a type union from that, e.g.:

``` js
const STATES = ['disconnected', 'connecting', 'connected'];
type State = $Values<STATES>;
let state: State = ...;
```

I'm proposing `$Values` here analogously to `$Keys`. Alternatively, `$Either` could be extended to not just accept a list of types but also an array.

goodmind:
Fixes #961, #7733

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -5152,6 +5152,26 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       let new_obj = DefT (reason, trust, ObjT {objtype with flags}) in
       rec_flow_t cx trace (new_obj, t)
 
+    (*************************)
+    (* arrays can be frozen *)
+    (*************************)
+
+    | DefT (reason_o, trust, ArrT arrtype), ObjFreezeT (reason_op, t) ->
+      (* make the reason describe the result (e.g. a frozen array literal),
+         but point at the entire Object.freeze call. *)
+      let desc = RFrozen (desc_of_reason reason_o) in
+      let reason = replace_reason_const desc reason_op in
+
+      let new_arrtype = match arrtype with
+        | ArrayAT (value, _) -> ROArrayAT value
+        (* TODO: Tuples aren't covariant, so Object.freeze actually doesn't do anything to them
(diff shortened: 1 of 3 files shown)


goodmind:
Should preserve string literals probably, but only after tuple elements can be covariant

goodmind:
Also I'm not sure if this is supposed to refine to frozen type, because Object.freeze mutates original object
```
const x = {foo: 1}
Object.freeze(x)
x.foo = 2 // no error
```

mvitousek:
Can you add some tests?

goodmind:
@mvitousek looks like it is broken :\
```js
const bar = Object.freeze([('12345': '12345')]) // why type becomes $ReadOnlyArray<number | "12345">
bar[0] = 1; // error
```

goodmind:
I'm not sure, but probably this would work better as libdef change

mvitousek:
That actually looks like a separate bug to me: my feeling is that if we've raised a type error from trying to write a string to a ROArray, we shouldn't propagate the string into the ROArray but instead maybe propagate `any` instead. If we made that change then your libdef approach might work better. @dsainati1 thoughts?

goodmind:
@mvitousek I'm not sure this is about `$ReadOnlyArray`, because this one works fine
```
const m: $ReadOnlyArray<number> = [1, 2, 3]
m[0] = 'wat'
m[0] // number
```

dsainati1:
I don't really see there being an issue in this case. We correctly report an error when you try to write to an read-only array, so as long as we properly taint the resulting type I wouldn't say anything incorrect is happening. I'd be much more concerned if you were able to write to a frozen array without an error.

goodmind:
The use case for this is to then do `$ElementType<typeof bar, number>` and get correct types, I think writes shouldn't break it

PR #7621, opened 2019-04-06
Implement $Values destructor for arrays/tuples

goodmind:
Related #7620


goodmind:
Probably this isn't needed because `$ElementType<ArrayLike, number>` does the same thing

PR #7627, opened 2019-04-08
[DOCS] deprecate `*`  (existential) type

lyleunderwood:
According to discussion in discord, `*` has been deprecated and added to the `deprecated-utility` lint. This changes the docs correspondingly.

### DIFF
--- a/website/en/docs/linting/rule-reference.md
+++ b/website/en/docs/linting/rule-reference.md
@@ -216,6 +216,11 @@ Triggers when you use the `$Supertype` or `$Subtype` utility types, as these typ
 unsafe and usually just equivalent to `any`. If the utilities were being used in a sound manner, the
 desired behavior can usually be recovered through the [`$Shape`](../../types/utilities/#toc-shape) utility or [bounded generics](../../types/generics/#toc-generic-types-act-as-bounds).
 
+#### `deprecated-type` <a class="toc" id="toc-deprecated-type" href="#toc-deprecated-type"></a>
+Triggers when you use the `*` (existential) utility type, as this type is
+unsafe and usually just equivalent to `any`.
+The effect of `*` can generally be achieved by simply not providing a type annotation.
+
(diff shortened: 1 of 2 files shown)


website/en/docs/linting/rule-reference.md
```diff
@@ -212,9 +212,10 @@ Triggers when you use `invariant` to check a condition which we know must be tru
 Note that this lint does not trigger when we know a condition is always `false`. It is a common idiom to use `invariant()` or `invariant(false, ...)` to throw in code that should be unreachable.
 
 #### `deprecated-utility` <a class="toc" id="toc-deprecated-utility" href="#toc-deprecated-utility"></a>
-Triggers when you use the `$Supertype` or `$Subtype` utility types, as these types are
-unsafe and usually just equivalent to `any`. If the utilities were being used in a sound manner, the
+Triggers when you use the `$Supertype`, `$Subtype`, or `*` (existential) utility types, as these types are
```
jbrown215: I think `deprecated-type` is the lint that does this, but it isn't documented at all. From looking at the code, it looks like it only fires when it encounters `*`

lyleunderwood: I added documentation for `deprecated-type` and cross-linked from the deprecation notice.

goodmind:
>usually just equivalent to `any`.

Is this true?

lyleunderwood:
@goodmind Discussion revealed that if flow can't find a suitable candidate for `*` then it will just return `any`. Additionally, apparently `*` becomes `any` at module barriers.

lyleunderwood:
@jbrown215 I couldn't get `make js` working because I was getting some error, but leaving that out I was able to get the screenshots below. I'm assuming `make js` is probably not relevant in the case of these markdown changes.

image
image

lyleunderwood:
Hmm, it seems like all builds are currently failing in appveyor.

jbrown215:
Yea, don't worry about the failing appveyor build. It's pretty flaky, and won't be affected my markdown changes

PR #7642, opened 2019-04-13
Fix $ObjMap and $ObjMapi exactness

goodmind:
Does similar behaviour as `$ReadOnly` type, tho I'm not sure why flags.exact is different from ExactT.
Also added `trust`


goodmind:
Or maybe this is intended behaviour?

Tuti1808:
Hdhfookndujfo

Tuti1808:
Gebcjsnkxlmxbhdjc

PR #7646, opened 2019-04-17
Make flow-dev-tools public

karanjthakkar:
Inspired from: https://twitter.com/threepointone/status/1118442563046989824

This PR makes `flow-dev-tools` public which allows devs to use this tool easily.

Previously:

- Clone flow repo
- install deps
- cd into product repo
- Use the tool like so: `../../<flow-repo>/packages/bin/tool ...`

Now:

- `npx flow-dev-tools ...`

I removed the `private: true` from the `package.json`. I'm unsure if I need to do anything more in order to allow this to be published to npm.

### DIFF
--- a/packages/flow-dev-tools/package.json
+++ b/packages/flow-dev-tools/package.json
@@ -1,8 +1,12 @@
 {
   "name": "flow-dev-tools",
-  "private": true,
   "license": "MIT",
   "version": "0.1.0",
+  "bin": "./bin/tool",
+  "files": [
+    "bin",
+    "src"
+  ],
   "dependencies": {
     "babel-jest": "^21.0.2",
     "babel-plugin-syntax-async-functions": "^6.8.0",


packages/flow-dev-tools/package.json
```diff
@@ -49,5 +53,8 @@
   },
   "scripts": {
     "test": "make -C ../.. test"
+  },
+  "devDependencies": {
+    "babel-cli": "^6.26.0"
```
threepointone: why do you need babel-cli here? alternately, why did you use the dist version in `tool`?

karanjthakkar: @threepointone tbh, I followed what was done in the flow-upgrade package. I imagine we use some modern syntax that might not be supported out of the box. What do you think is the right way?

threepointone: I dunno :) I was just curious. `babel-node` isn't called anywhere, so wanted to understand why you added this here

karanjthakkar: Oh right. Good catch. I forgot to add the build step 🤦🏻‍♂️Will do.

threepointone: would it be simpler to just run `src/main.js`?

karanjthakkar: We use es6 imports in the package. So I'm not sure how using the source directly would work if we plan on distributing the package.

threepointone: It already uses `require("babel-register")(babelConfig);` so I imagine it would just work? I don't think anyone will `import Something from 'flow-dev-tools'` in their code.

Don't mind me, I'm just doing driveby commenting :)

karanjthakkar: You're right. I misunderstood your initial comment.

namuol:
Any updates on this? Still seems like pretty low hanging fruit 🍎 ❤️ 🙏

oonsamyi:
`flow-dev-tools` is a very important package for our team.
We use one for upgrade flow version (add-comment, remove-comment, upgrade codemodes, etc.)
Unfortunately, we had to fork this repo and publish one in the privet npm registry.
Please, give any reasons why this package can't be published!

liuyangcharlie:
Noticed some test failed and the PR failed to merge. What's our plan now? Do we still want to make it public or have any alternative ways?

PR #7656, opened 2019-04-21
Change return type of 'process.exit' from 'void' to 'empty'

Bannerets:
`process.exit` never returns, therefore it is safe.

https://nodejs.org/docs/latest-v10.x/api/process.html#process_process_exit_code

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2337,7 +2337,7 @@ declare class Process extends events$EventEmitter {
   ): void;
   execArgv : Array<string>;
   execPath : string;
-  exit(code? : number) : void;
+  exit(code? : number) : empty;
   exitCode? : number;
   getegid? : () => number;
   geteuid? : () => number;


goodmind:
```js
function f(m: string = process.exit(0)) { // error, but it is impossible
  m
}
```
[Flow try] (https://flow.org/try/#0CYUwxgNghgTiAEA3W8AOMD2YQGccC54BvAKHnhAA8BLAFwAowNQB+eQgOwFcBbAIxAwAlOyQZqwANwkAviRIAzLhzC1qGDvAX0ehHLRjUOAc3gBeNJmx4AdFTr0ADEJGlyPWUA)

goodmind:
@mroch can you import this?

mroch:
fixed in 44da3223fb2d8399c30546f6731f68d43d21930a, thanks!

PR #7660, opened 2019-04-22
Add strict mode assert typedefs

### LINKED ISSUE #7217: Missing Node assert strict mode: require('assert').strict
In Node 10 and 8.13.0 you can do (abridged for clarity):
```
$ node
> const assert = require('assert').strict
> assert.equal(1, '1')
{ [AssertionError [ERR_ASSERTION]: Input A expected to strictly equal input B:
+ expected - actual

- 1
+ '1'] }
> assert.strictEqual(1, '1')
{ [AssertionError [ERR_ASSERTION]: Input A expected to strictly equal input B:
+ expected - actual

- 1
+ '1'] }
```

but the current definition is missing it: https://github.com/facebook/flow/blob/a2ed5108713c75bf7fbb2d3bbeb2c1281290609e/lib/node.js#L2057

docs: https://nodejs.org/docs/latest-v10.x/api/assert.html#assert_strict_mode

Since `assert.strict` has the same type signatures as `assert`, I have an idea of how to do it, but I'm sure it's not the best way.

goodmind:
Closes #7217
Revived https://github.com/facebook/flow/pull/7401

https://nodejs.org/api/assert.html#assert_strict_mode

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2216,6 +2216,25 @@ declare module "zlib" {
 
 declare module "assert" {
   declare class AssertionError extends Error {}
+  declare type AssertStrict = {
+    (value: any, message?: string): void;
+    ok(value: any, message?: string): void;
+    fail(message?: string | Error): void;
+    // deprecated since v10.15
+    fail(actual: any, expected: any, message: string, operator: string): void;
+    equal(actual: any, expected: any, message?: string): void;
+    notEqual(actual: any, expected: any, message?: string): void;
+    deepEqual(actual: any, expected: any, message?: string): void;
+    notDeepEqual(actual: any, expected: any, message?: string): void;
+    throws(
+      block: Function,
+      error?: Function | RegExp | (err: any) => boolean,
+      message?: string
+    ): void;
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -2216,6 +2216,26 @@ declare module "zlib" {
 
 declare module "assert" {
   declare class AssertionError extends Error {}
+  declare type AssertStrict = {
+    (value: any, message?: string): void;
+    ok(value: any, message?: string): void;
+    fail(message?: string | Error): void;
+    // deprecated since v10.15
+    fail(actual: any, expected: any, message: string, operator: string): void;
```
AlexeyVykhrystyuk-GB: Should we update `assert.fail` the same way it's done there https://github.com/facebook/flow/pull/7915?

AVykhrystyuk: Should we update `assert.fail` the same way it's done there https://github.com/facebook/flow/pull/7915?

goodmind: probably

nmote:
I had to add the inexact type syntax and record the tests after rebasing, but otherwise the rebased version was unchanged

PR #7667, opened 2019-04-24
Slight correction in refinements.md (any -> many)

### LINKED ISSUE #7645: Refinements frequently used?
https://flow.org/en/docs/lang/refinements/ starts out "Refinements are a frequently used aspect of any type system."  My understanding is that many (most?) type systems don't have refinements, in the sense used here.  Reporting so that it can be fixed or I can learn something.

dlthomas:
Fixes #7645.

### DIFF
--- a/website/en/docs/lang/refinements.md
+++ b/website/en/docs/lang/refinements.md
@@ -2,7 +2,7 @@
 layout: guide
 ---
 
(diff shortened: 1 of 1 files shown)


dlthomas:
@mrkev Per discussion on the ticket :)

mrkev:
👍

Issue #7672, opened 2019-04-26
Explicit type specification for component rendering <MyComponent<T> />


retyui:
## Proposal
Now there is no possibility to explicitly specify the generic type for the component.

This syntax can solve the problem:
```js
<MyComponent<T> />;
```

I can be rea

## Use case
```js
// @flow
import React from 'react';

class MyComponent<T> extends React.Component<{| +item: T |}>{  }

<MyComponent<number> item={1} />; // work good
<MyComponent<string> item={1} />; // error
```

jamesisaac:
Yeah agreed this could be useful, I've run into it before (although for my use-case I've now just started using Hooks, which are easier to type).  In the end I worked around it like this:

```js
class ConcreteComponent extends MyComponent<number> {}

<ConcreteComponent item={1} />
```

...although that solution only works for class components, not functional components.

Supported in TypeScript since 2.9: https://github.com/Microsoft/TypeScript/pull/22415

goodmind:
I have implemented parsing side, would submit PR soon

jbrown215:
This is something that's been on my list for a while, but I was never sure I would get to it. Thanks for looking into this, @goodmind! I will review as soon as I can.

villesau:
@goodmind will your PR also add support for something like:

```
type LinkProps<T> = {a: T};
declare var AAA<T>: React$StatelessFunctionalComponent<{children: (props: LinkProps<T>) => void}>;
```

[try flow](https://flow.org/try/#0JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AoGATzCTgBlgA7AawAUcwBnAHgBUAfHAC8cAN4oAXHD4BfCgBMkaADYoicAG7q4AQX38B05OhgASAMowUMJCqRcuAMQCuTDMAhMUKgMK5IJiQmGB4xNAALYBUFIiZpAAowTi5pRlYOCG5DAEoRIU0IYAVZAQpyHn1dAS4YKGYAcx4AeiqyoA)

see: https://github.com/flow-typed/flow-typed/pull/3281/files#diff-f6fab52153e41297b233b7e3a1975af6R86

goodmind:
@villesau, no, why?

villesau:
Would be useful with type definitions.

goodmind:
Why not generic function?

villesau:
How would you express this:

```
type LinkProps<T> = {a: T};
[...]

PR #7674, opened 2019-04-26
Fix comment syntax inside jsx

goodmind:
Temporary workaround for #7672

```js
function Component<T>(props: {|type: T|}) {
  return null
}

const el = <Component /*::type={1}*/ />
```


gabelevi:
What's the use case for this? Has this been requested?

goodmind:
@gabelevi isn't it more consistent for comment syntax to work everywhere?
This is workaround for providing generic type to component
```js
function Component<T>(props: {|type: T|}) {}
const el = <Component /*::type={1}*/ />
```

goodmind:
@gabelevi I updated main post

src/parser/flow_lexer.ml
```diff
@@ -1501,6 +1501,40 @@ let jsx_tag env lexbuf =
...
+        Continue env
+
+  | "*/" ->
+    if is_in_comment_syntax env then
+      let env = in_comment_syntax false env in
+      Continue env
+    else begin
+      Sedlexing.rollback lexbuf;
+      match%sedlex lexbuf with
+      | "*" -> Token (env, T_MULT)
```
gabelevi: This isn't right. Outside of a JSX tag, `*` would be the `T_MULT` token (for multiplication). But inside of a JSX tag I think it's an unexpected token. You can probably just emit an error. Something like

```
| "*" -> Token (env, T_ERROR (lexeme lexbuf))
```

goodmind:
I can't test it now, but I'm not sure if this would be allowed too?
```js
</*::Test*/ />
```

goodmind:
@gabelevi I'm going to close it because I'm not really want to introduce some unexpected behaviour here with jsx, also I think this is not worth it

PR #7675, opened 2019-04-27
Parse type instantiation in JSX

goodmind:
Discussion: #7672 

Example:
```jsx
const foo = <Bar<number> />
```


goodmind:
Not sure how this desugars into `React.createElement`

```js

declare class M<T> extends React.Component<{callback: T => void}> {}
declare function H<T>(props: {callback: T => void}): void;

declare var A: H<number> // Can't apply function type arguments before-hand
declare var B: Class<M<number>>

// Error: Cannot use function type as a type because function type is a value. To get the type of a value use `typeof`
React.createElement(A, {callback: data => { 
   data // empty
}})

React.createElement(B, {callback: data => {
   data // number
}})
```

PR #7677, opened 2019-04-28
Remove Object type (any) from WeakSet and $ReadOnlyWeakSet

goodmind:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#Parameters


goodmind:
@mvitousek can you look at this one?

PR #7681, opened 2019-04-29
add "value" property to DOMTokenList

shawnbot:
The [DOMTokenList](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList) `value` property has [broad browser support](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/value#Browser_compatibility), and should be included in Flow's type definition.


goodmind:
@shawnbot no one responded from Flow team

Issue #7603, opened 2019-04-01
Missing three-argument function declarations for Node http and https modules


nickshanks:
Disclaimer: Every time I've reported a bug here before it's been either my mis-understanding of what Flow is intended to be, or a mis-understanding of how flow is intended to work. This time however, I really think I have a genuine bug!

These two modules:
https://github.com/facebook/flow/blob/master/lib/node.js#L1375
are supposed to type for these four function calls, which take up to three arguments:
https://nodejs.org/api/http.html#http_http_get_url_options_callback
https://nodejs.org/api/http.html#http_http_request_url_options_callback
https://nodejs.org/api/https.html#https_https_get_url_options_callback
https://nodejs.org/api/https.html#https_https_request_url_options_callback

However flow only types for two arguments. The Node documentation states: "If both url and options are specified, the objects are merged, with the options properties taking precedence."

Flow version:
0.93.0

## Expected behaviour
Function type definition would exist for http/https.request/get(url, options, callback) API

## Actual behaviour
Error ┈┈┈
Cannot call protocol.get because:
 • no more than 2 arguments are expected by function type [1].
return protocol.get(href, {}, cb);

Error ┈┈┈
Cannot call protocol.get with object literal bound to callback because:
 • a callable signature is missing in object literal [1] but exists in function type [2].
return protocol.get(href, {}, cb);

* Link to Try-Flow or Github repo:

I cannot show a flow.org/try error because it doesn't support Node's http & https modules:

https://flow.org/try/#0MYewdgzgLgBA5gUygVQE4BsYF4YAoAWqCAZgJTYB8ewAhuugEY3ADW5WVA3gFAwyiRY+KFAAO2GEQCOAVwCWRXAHJhYpaQDcvfuGgxVoiBOnzFKkYfVa+AvaNQgoIUJhyESAbQAsAXWxYcJQglGAB+fQsjAC4IsWsYe0dnEHQAOkQoAiJiABoYTgBfPNp6JlZNbgKNIA

(is that another bug, or intended?)

PR #7687, opened 2019-05-01
Update http(s).get and http(s).request libdefs

mrtnzlml:
Closes: https://github.com/facebook/flow/issues/7603

See:

- https://nodejs.org/api/http.html#http_http_get_url_options_callback
- https://nodejs.org/api/http.html#http_http_request_url_options_callback
- https://nodejs.org/api/https.html#https_https_get_url_options_callback
- https://nodejs.org/api/https.html#https_https_request_url_options_callback

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1372,6 +1372,29 @@ declare class https$Server extends tls$Server {
   timeout: number;
 }
 
+type requestOptions = {|
+  auth?: string,
+  defaultPort?: number,
+  family?: number,
+  headers?: {[key: string] : mixed},
+  host?: string,
+  hostname?: string,
+  localAddress?: string,
+  method?: string,
+  path?: string,
+  port?: number,
+  protocol?: string,
+  setHost?: boolean,
+  socketPath?: string,
+  timeout?: number,
+|}
+
+type http$requestOptions = {
+  ...requestOptions,
+  agent?: boolean | http$Agent<net$Socket>,
+  createConnection?: (options: net$connectOptions, callback?: Function) => net$Socket,
+};
+
 declare module "http" {
   declare class Server extends http$Server {}
   declare class Agent extends http$Agent<net$Socket> {
@@ -1385,11 +1408,21 @@ declare module "http" {
     requestListener?: (request: IncomingMessage, response: ServerResponse) => void
   ): Server;
   declare function request(
-    options: Object | string,
+    options: http$requestOptions,
+    callback?: (response: IncomingMessage) => void
+  ): ClientRequest;
+  declare function request(
+    url: string,
+    options?: http$requestOptions,
     callback?: (response: IncomingMessage) => void
   ): ClientRequest;
   declare function get(
-    options: Object | string,
+    options: http$requestOptions,
+    callback?: (response: IncomingMessage) => void
+  ): ClientRequest;
+  declare function get(
+    url: string,
+    options?: http$requestOptions,
     callback?: (response: IncomingMessage) => void
   ): ClientRequest;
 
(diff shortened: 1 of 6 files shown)


mvitousek:
Sorry it took a bit to get to this, it looks good. Is it possible though to get a more precise type for `options` than just `Object` (which is unsafe and generally should be avoided)?

mrtnzlml:
@mvitousek Sure, will do. I was just extending what was already there a little bit. I'll let you know when it's ready to be checked again. :)

mrtnzlml:
@mvitousek Updated and rebased. There are two things I am unsure about:

1) `http$requestOptions` and `https$requestOptions` are very similar with some small changes for HTTS - not sure if this is the best way how to do it. I would appreciate some guidance here if not.
2) I noticed that `createConnection` methods  in http(s) `Agent` are not compatible. I didn't touch this part at all in this PR but should probably be fixed. It's kind of related because there is also `createConnection` property in the request options I created.

Please, have a look and let me know what to improve.

lib/node.js
```diff
@@ -1372,6 +1372,25 @@ declare class https$Server extends tls$Server {
   timeout: number;
 }
 
+type http$requestOptions = {
+  agent?: boolean | http$Agent<net$Socket>,
+  auth?: string,
+  createConnection?: (options: net$connectOptions, callback?: Function) => net$Socket,
+  defaultPort?: number,
+  family?: number,
+  headers?: {[key: string] : string},
```
mvitousek: Unless I'm missing it the docs don't seem to specify that `headers` only maps strings to strings--seems like you could have at least numbers too. Maybe `{[key: string] : mixed}`?

mvitousek:
Looks really good, thank you for taking this on! Made one suggestion, and since the request object is different only in its default values (which aren't visible by the caller), could you just use the same object for both?

Would you mind filing an issue regarding the `createConnection` definition?

mrtnzlml:
> ... since the request object is different only in its default values (which aren't visible by the caller), could you just use the same object for both?

I tried to follow the differences between  HTTP and HTTPS module so there are additional differences in `http$Agent<net$Socket>` vs. `http$Agent<tls$TLSSocket>` and similar for return type of `createConnection`. Should it really be the same?

lib/node.js
mrtnzlml: ✅ Mixed it is.

mvitousek:
Oops great catch, thank you. What I would do then--when most attributes are the same, but some are different--is to put the attributes in common in a single type, and then use spreading to copy those attributes into distinct types that also capture the differences. See here for an example: 
https://flow.org/try/#0C4TwDgpgBMULxQN4B8oEMBcUB2BXAtgEYQBOUyAvlAPTVQDKYJEaAJgJbYDmUA7gPYkA1gGcoxEbF7tgACygQAHmgDGsfoQBWENTHAQRAKFCR08JADorwADTiskkpx4Vj+8ecRWLt+zgLEZK6GrDoANmjMUABukVCKWGgA3IaxZNhYeESk5ooWyalxIg7ATty5FoRJQA

That said, if you'd prefer to not go down this route I'm happy to merge this as-is, as long as it passes all our tests.

mrtnzlml:
@mvitousek Updated. Please have a look.

mrtnzlml:
@mvitousek Hi, is there anything I can do to make it mergeable? :)

mvitousek:
Hi @mrtnzlml, sorry for taking so long to finish this! The PR is landing right now :) Thanks a lot for your work on this PR, it's a great contribution and we appreciate it.

PR #7698, opened 2019-05-04
Add flowconfig parser js_of_ocaml package

goodmind:
image

### DIFF
--- a/js/str.js
+++ b/js/str.js
@@ -1,14 +1,217 @@
-//Provides: re_search_forward
-function re_search_forward() {
-  throw new Error("re_search_forward: not implemented in JS");
-}
+// Partially from
+//  https://github.com/ejgallego/jscoq/blob/2718f9caf31398704c2d84ff089e3f6f0321eada/coq-js/js_stub/str.js
 
-//Provides: re_replacement_text
-function re_replacement_text() {
-  throw new Error('re_replacement_text: not implemented in JS');
-}
+//Provides: str_ll
+function str_ll(s, args) { if (str_ll.log) joo_global_object.console.warn(s, args); }
+str_ll.log = false;
 
 //Provides: re_string_match
-function re_string_match() {
-  throw new Error('re_string_match: not implemented in JS');
+//Requires: str_ll, re_match
+function re_string_match(re, s, pos) {
+  // external re_string_match : regexp -> string -> int -> int array
+  //str_ll('re_string_match', arguments);
+  var res = re_match(re, s, pos, 0);
+  return (res === 0) ? [0] : res;
+}
+
+//Provides: re_search_forward
+//Requires: str_ll, re_search_forward_naive
+function re_search_forward(re, s, pos) {
+  // external re_search_forward: regexp -> string -> int -> int array
+  //str_ll('re_search_forward', arguments);
+  return re_search_forward_naive(re, s, pos);
+}
+
+//Provides: re_partial_match
+//Requires: str_ll
+// external re_partial_match: regexp -> string -> int -> int array
+function re_partial_match() { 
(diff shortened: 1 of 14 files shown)


js/str.js
```diff
@@ -1,14 +1,217 @@
+// Partially from
+//  https://github.com/ejgallego/jscoq/blob/2718f9caf31398704c2d84ff089e3f6f0321eada/coq-js/js_stub/str.js
```
mroch: this is AGPL-3+, and the part based on strstubs.c is LGPL2 + static linking exception. we will have to review license compatibility...

without looking into how hard it'd be first, i'd be more inclined to remove the uses of `Str` instead. might even make sense to use a real parser (we already depend on sedlex and ocamllex) but that's a bigger project.

src/commands/config/js/stdlib.js
```diff
@@ -0,0 +1,21 @@
+// Stdlib overrides until js_of_ocaml fixes it
```
mroch: is there an upstream github issue we can reference?

js/str.js
goodmind: Thanks, I'll remove it for now, I think LGPL2 is compatible with MIT, not sure about AGPL3

goodmind: https://github.com/ejgallego/jscoq/issues/97

src/commands/config/js/stdlib.js
goodmind: https://github.com/ejgallego/jscoq/issues/97

goodmind: > At any rate, we plan to push these changes upstream to JSOO as they are not specific to jsCoq.

goodmind:
I'm not sure about specific goal, I heard that babel needed flowconfig parsing for `all` option, also this can incorporate `flow init` but accept js objects instead of CLI

js/str.js
goodmind: `Str` also would be useful for flow.js overall, not just flowconfig parser

src/commands/config/js/stdlib.js
goodmind: https://github.com/ocsigen/js_of_ocaml/pull/791

goodmind: It was fixed, but I know that Flow uses old js_of_ocaml

goodmind:
@mroch I think LICENSE was updated
```
Files: coq-js/js_stub/str.js
License: LGPL2.1
Comment: Derived from OCaml (https://github.com/ocaml/ocaml).
```

Is this still doesn't work with MIT?

goodmind:
My goal is to allow any js tool to write or read flowconfig, related #7771

js/str.js
hhugo: Js_of_ocaml (3.6) now include all full support for Str

src/commands/config/js/stdlib.js
```diff
@@ -0,0 +1,21 @@
...
+      case 45: i++; sign = -1; break;
+      case 43: i++; sign = 1; break;
+    }
+  }
+  if (i + 1 < len && caml_string_unsafe_get(s, i) == 48)
+    switch (caml_string_unsafe_get(s, i + 1)) {
+      case 120: case 88: base = 16; i += 2; break;
+      case 111: case 79: base = 8; i += 2; break;
+      case 98: case 66: base = 2; i += 2; break;
+      case 117: case 85: sign = 0; i += 2; break;
```
hhugo, line 18: `sign = 0` is not correct.

SamChou19815:
Closing since there are license issues, and we already have support for flowconfig in try-flow

PR #7699, opened 2019-05-06
[circle] use opam2 and non-EOL debian

mroch:
debian 8 is EOL, so switch to building with debian 9.

this bumps the version of glibc used from [2.19](https://packages.debian.org/source/jessie/glibc) to [2.24](https://packages.debian.org/source/stretch/glibc). I did not confirm whether this affects compatibility of our published binaries, but debian 8 and similarly-old distros are so old that I think it's reasonable to drop pre-built support for them. You can still build from source on those platforms.

TODO before merge:
- [ ] `dmesg: Cannot allocate memory` error in the linux tests appears to be new and is probably related

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,7 +5,7 @@ aliases:
 
   - &docker_opam
     docker:
-      - image: ocaml/opam:debian-8_ocaml-4.05.0
+      - image: ocaml/opam2:debian-9-ocaml-4.05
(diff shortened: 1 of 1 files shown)


gkz:
ctrl-f finds no mentions of "debian" in https://github.com/facebook/flow/blob/main/.circleci/config.yml so I assume this PR is no longer relevant

PR #7708, opened 2019-05-08
[DOCS] Recommend using ES6 destructuring syntax for default props in functional components

cyan33:
As our internal docs suggests.

https://fburl.com/wiki/vnzyb9h7


Brianzchen:
This is not exactly correct, It's not supported in `React.forwardRef` because it returns `AbstractComponent` which includes things like `React.memo` also.

gkz:
We've updated our React component docs to suggest destructuring with default props:
https://flow.org/en/docs/react/components/

Issue #7651, opened 2019-04-18
HTMLElement.focus missing options object


gaastonsr:
## Missing/Incorrect APIs

The focus method of HTMLElement is [missing](https://github.com/facebook/flow/blob/e0b7df928800d54df4f66c450ac436e5a1419b3f/lib/dom.js#L1621) the [options](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) object. 

## Relevant documentation

[Typedef](https://github.com/facebook/flow/blob/e0b7df928800d54df4f66c450ac436e5a1419b3f/lib/dom.js#L1621).
[Focus docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus).

cindyyu:
Hi,

I am attempting to address this issue in PR #7711 as my first contribution. :smile:

Thanks,
Cindy

cindyyu:
This should be addressed by #7711.

PR #7711, opened 2019-05-09
HTMLElement: add options parameter for focus method

cindyyu:
Issue #7651: The HTMLElement focus method is missing the optional options parameter.
[MDN documentation on focusOptions parameter](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#Syntax) 

I added a new type for the focus options and added an optional parameter to the focus method on HTMLElement.

*HTMLInputElement*
I noticed that HTMLInputElement also has a focus() method on it, but I am unsure if HTMLInputElement's focus also accepts options. When I looked at the HTMLInputElement MDN docs, I was under the impression that it would also accept the same options since the focus() hyperlink directs you to the docs on HTMLElement.focus. I did not dig too much into this since I wanted to focus (😄) on a small change and understand how the contribution process works. 

Please let me know if I am missing any steps or if anything is sorely incorrect... this is my first attempt at contributing. 😄

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1289,6 +1289,9 @@ declare class Range { // extension
 declare var document: Document;
 
 // TODO: HTMLDocument
+type FocusOptions = {
+  preventScroll?: boolean
+}
 
 declare class DOMTokenList {
   @@iterator(): Iterator<string>;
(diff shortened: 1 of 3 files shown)


dsainati1:
This looks mostly fine to me; can you include a link to the documentation you referenced in the summary? I can import it after you do that.

cindyyu:
@dsainati1 Added a link to the MDN documentation describing the optional options parameter to the summary!

PR #7713, opened 2019-05-10
lib/node: Add util.TextDecoder/TextEncoder

cakoose:
Library docs:
- [TextDecoder](https://nodejs.org/api/util.html#util_class_util_textdecoder)
- [TextEncoder](https://nodejs.org/api/util.html#util_class_util_textencoder)

Node 8.3 added them to 'util'.

Node 11.0 added them to the global object, but I didn't do that part.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2023,6 +2023,25 @@ declare module "util" {
   declare function deprecate(f: Function, string: string): Function;
   declare function promisify(f: Function): Function;
   declare function callbackify(f: Function): Function;
+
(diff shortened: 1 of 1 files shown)


goodmind:
This is actually just duplicates same thing from `bom.js`

cakoose:
Oh, interesting.  I wonder if there's a way to share some of the definitions.

One tricky thing is that the Node API is slightly different.  For example, the `TextEncoder` constructor doesn't accept an argument, since it only supports UTF-8.

Issue #6927, opened 2018-09-22
What’s going on with `Object.keys`?


wchargin:
The core libdef for `Object.keys` is clear as can be:

```js
declare class Object {
    // [snip]
    static keys(o: any): Array<string>;
    // [snip]
}
```

(<https://github.com/facebook/flow/blob/172d28f542f49bbc1e765131c9dfb9e31780f3a2/lib/core.js#L58>)

This suggests that the following code should have a type error:

```js
// @flow
type K = "A" | "B";
const x: {[K]: number} = {};
(Object.keys(x): Array<K>);  // `string` should be incompatible with `K`
```

And yet the above code typechecks. Why?

This was not always the case. The above code did not typecheck in Flow v0.59.0, but has typechecked since Flow v0.60.0. But the release notes for v0.60.0 don’t mention such a change.

Also, this does _not_ work if the key type is opaque (_possibly_ related to #6569 and friends):

```js
// @flow
declare opaque type O: string;
const y: {[O]: number} = {};
(Object.keys(y): Array<O>);  // fails (sad)
```

Having the refined type is quite useful, though the grievances in #5210 (from before the change) are legitimate. What is going on? Whence the inconsistency between the libdef, the working example with a transparent subtype of `string`, and the failing example with an opaque subtype of `string`?

([Full Flow-Try here.][try])

[try]: https://flow.org/try/#0PTAEAEDMBsHsHcBQAXAngBwKagNKgLygBEAgkaAD7EBCRA3IgMawB2AzsqAB4BcoA3gG0cAXT4sArgFsARpgBOAXwIDFDABQB5GQCtMjZADoA1plRt1XAJR8S8+QENUAHhwA+K3VCgQoeLHljNlB1AHNYWAATABpQKVgOaFQAflA2TGwAYgBWACYARgAGK0RESP1oB3lsWHQHAEcJbDQsUE0+DnkASxZQhmZ2TlQ+IU0xUElZBWVCfjVELV19I1NzdVQbUDtHF00PLx8wSAcu6GD1NgdIkqA

dsainati1:
Object.keys is actually implemented internally using a special type constructor that basically ignores the libdef definitions. Why it's done this way, I don't know, but I'd bet the definition can be removed.

jcready:
> Object.keys is actually implemented internally using a special type constructor that basically ignores the libdef definitions.

[...]

PR #7717, opened 2019-05-12
Add generic to Object.entries

Dimitreee:
In my humble opinion there is one strong problem, in ```Object.entries()``` typings .
Me and many other developers cannot use this perfect API because we should "fight" with this ```mixed``` Tuple second value type, and as the result, flow wins, this method keeps being unused.

[Here is simple implementation of the problem](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoAlgOwC4FMBOUAhgMa5gDKAFnLgM5gDeqYYd6AXrgFxiYCuAWwBGBADQswJOPHy862fFgDmEgL4BuVKmwBPAA7kACvlxQCpgCbVadCp3pgAvGACC+fEV0AeANoKlTFU+IVF8AF0APh0DcgBhfgU4QQIAESJsImcmXwBrXF1uAJVw3gERAjVtPUNKGno4mRhcEmx0OExs908fG3pomNq+ugBZIn1DfGyACjp6ul5hgEpnSJCK-GrYsAA1Ihh0SwzcYdcANyJ0GCJhK-Q9GckSROxktIyiXj2Do7xTi6uNzuegkrCSJFyw0aMGarXamEW82hsLaHQkKyca2EcBkuCImG00kwCjYSKaLVRCLqtmRFPh2V8kmYrDBDl4AEYAAyglnSWS8ABEwmuEIFPLUPOZLPYXA5ACYeaw+XA5GABQgqPdcGLJBKmZJWbKwABORVSGQqwW6XAwxA61h61hSw08MBcgB0AFYzcrVUKRbl7WA9eEtKgiSTBLoAGJEQRXXQmMwWXDWeb2LgLMAJJIpfDpTLZKUCkbofICjncyQCgBCVxgFbd7J5Atcc2aukbAA51GGI9hSbYxhMCIih+NJjNGGwHMGMWsZbgw8BgObiQOLj9jv9Ltdbgc9F99odt-NzrugQfdDNnrn3pkxGxXhCoeS4R150xwx0SZuT38zwBPdgWvFxplvV48wLIhH3BSEyRhOkP1WL8lR-AciCAm5mgzeheG6LxvHKMI1hcODX0Q99MHdQRxlmeZh0mJYtDQ9cwH0UxzA8VNcLoAAVfh9GaWkqN4JMuKsYZeOyAB5YQACsKXdXAcCUehwJeN58w+ZjtFYFcwFMbB+HwToOOTbjrAcfjBOEt9KXdKx+DIaZZkyYys2xXF8UfXwwEfcyJJ42dwk-KUDNYdAoDAaYAEIFAyRIViMkzOgSjyWLAFdJAisAopizCL2FE5rPdLASBgfhLHUwKUysrglhWcLVxZQzcGM0ywEUfhcBylrgz64NH263AllQTRtAMiNcXdeBlGmP9fhOQCipA6Yo1jeMYETTi6qk6zYIQlF4UarQgA)

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -45,7 +45,7 @@ declare class Object {
     static create(o: any, properties?: PropertyDescriptorMap): any; // compiler magic
     static defineProperties(o: any, properties: PropertyDescriptorMap): any;
     static defineProperty<T>(o: any, p: any, attributes: PropertyDescriptor<T>): any;
-    static entries(object: mixed): Array<[string, mixed]>;
+    static entries<T>(object: { [key: string]: T }): Array<[string, T]>;
     static freeze<T>(o: T): T;
     static fromEntries<K, V>(entries: Iterable<[K, V] | {'0': K, '1': V}>): {[K]: V};
 


lyleunderwood:
This is unsound [given inexact objects](https://flow.org/try/#0MYewdgzgLgBArpAhgMwKYFExQE4EtUQBcMAFADwAqAfCSAEYBWqwUxA3jANoDWqAnsWh4wAcwC6xCjAC+AShgBeKjACC2bIj5lOQ3KIA0MCmKryFpAPKNmUAHSoseAsURg+sgNwAoGL5heoPgAHVBgASTBUAA9EFismFgpg0PM2Hz9fAFtUTLpUbEEcPRF9L2lvL1BIWCgCKHibYgjo2PrrROTFGDSMrJy8gpgAclroIdLemAATVxF8kDgIAFl+-OIARlLyr0rwaBgAG1xoYjUNLR0igxhdURMuhAgUDEd8CBJRtoSoTx2qiBAB1QtgOIBEJCO0FsmUQQRIJE4AH1DAA3RAHOCoQrCcRmZRojHAoGiKAAC1knhgAHoqTAAO6kxCwRB0BawAAGM1E80WK1y+XZAH5-BkaTBcFAhhAYIgYGA4PzsIYJdMQAQwCMYIyUaFZeziSIyeyYEFsCAQthArZrUA).

goodmind:
How many PRs adding this were made? Same for Object.keys, maybe docs should explain soundness better?

lyleunderwood:
It would be nice if `Object.values` and `Object.entries` could be re-addressed when the switch to exact-by-default happens.

Dimitreee:
@lyleunderwood @goodmind 
I prefer to control the "Types" with my hands and work with that "Type" I need, and not put up with those "mixed". Its sounds horrible that i should make theese "Types" of this default API to be reassigned to "any" by me. 

By the way, @goodmind, why dont you [use "Exact" ](https://flow.org/try/#0MYewdgzgLgBArpAhgMwKYFExQE4EtUQBcMAFADwAqAfCSAEYBWqwUxA3jANoDWqAnsWh4wAcwC6xCjAC+AShgBeKjACC2bIj5lOQ3KIA0MCmKryFpAPKNmUAHSoseAsURg+sgNwAoGL5heoPgAHVBgASTBUAA9EFismFgpg0PM2AB8fP18AW1RsulRsQRw9EX0vNOlvL1BIWCgCKHibYgjo2KbrROTFGDZMrNz8wuIAcgboUfKs3wATVxFCkDgIAFk8gqKYAEZyqq8a8GgYABtcaGI1DS0dEoMYXVETXoQIFAxHfAgSCc6EqE8B1qEBAJ1QthOIBEJDO0Fs2UQQRIJE4AH1DAA3RAnOCoYrCcRmZRYnHgsGiKAAC1knhgAHo6TAAO6UxCwRB0ZawAAG81ESxW62G2G5AH5-FkGTBcFBRhAYIgYGA4MLDDKYLMQAQwOMYKyMaFFdzySIqdyYEFsCAQthArZ7UA) ?

nnmrts:
@goodmind Just be honest and say you never use Object.values/.entries/.keys together with flow and that's why you just don't see a problem with that.

`Object.entries` resulting in `Array<[string, mixed]>` every time, no matter what, is a **bug**.

Same thing with `Object.values` btw...

Again:
This. Is. A. Bug.

I don't know why you refuse to understand that. You always bring up the same argument, "this breaks with inexact objects", well, yeah, thank you very much, we know that too.

There are a lot of good suggestions by a lot of people, a lot of pull requests and issues were opened, but they all get closed for no reason, one after another. For more than 3 years now. And now the most popular typechecking library isn't able to properly typecheck objects. Congrats.

And we can't even override flowlib definitions? What the hell?

TIHI.

goodmind:
@nnmrts you don't need to be so negative here, I understand your frustration, I'm just a user too. This just goes against Flow goal to be as sound as possible, Flow never had completeness in mind.
It's totally possible to make sound Object.keys/values/entries but my PR wasn't accepted too because "exact constraint is flawed idea"

nnmrts:
Ye, sorry. I don't want to attack anyone personally here, but the fact that flow is missing this feature is just mind-boggling to me.
So, "exact constraint is flawed idea"? Why? Did anyone actually explained why? And/Or provided another solution?

Because just not fixing it **and** not letting others try to fix it, is not a solution, Facebook.

Every other library I know would give you an option to opt-in for a feature, when the developers think the idea of it is "flawed" but the feature still "just works" for most of the users.

> Flow never had completeness in mind

That's not a completeness issue tho, sorry. Flow is described as "a static typechecker for JavaScript." and this is a direct quote from the website:

> Flow is designed to understand idiomatic JavaScript. It understands common JavaScript patterns and many of the weird things we JavaScript developers love to do.

Turns out it doesn't understand common JavaScript patterns.

I would understand the completeness argument if we are talking about async/await stuff, experimental es7 stuff, generator stuff, even class stuff. No sane person would complain about that, because it's acceptable that a typechecker can't understand language syntax that was basically released yesterday. But:

- The `Object.values` and `Object.entries` definitions were finalized in the specification for ECMAScript2017 (!).
- The `Object.keys` and `Array.prototype.map` definitions were finalized in the specification for ECMAScript 5.1 (June 2011!!!)
- `Object.values(obj)` is just a shorthand for `Object.keys(obj).map((key) => obj[key])`
- `Object.entries(obj)` is just a shorthand for `Object.keys(obj).map((key) => [key, obj[key]])`

A typechecker that doesn't support basic, 8 years old features of a language isn't "incomplete", it's useless.

goodmind:
What I'm referring to https://github.com/facebook/flow/issues/6927#issuecomment-515685530

nnmrts:
@goodmind Thanks, I've already read those comments, however we got no reaction to them.

This whole problem is ignored since 2016, at least here on GitHub.

PR #7735, opened 2019-05-18
Method predicates

goodmind:
Enables parsing and typechecking stuff like this

```js
class Hello {
  test(x: mixed): boolean %checks { return typeof x === 'string' }
}
const m = new Hello()
const foo: string | number = ''
if (m.test(foo)) {
  foo // string
}
```

```js
declare function isString(x: mixed): boolean %checks(typeof x === 'string')
declare var is: {|
   string: typeof isString
|}
const foo: string | number = ''
if (is.string(foo)) {
  foo // string
}
```

### DIFF
--- a/src/parser/object_parser.ml
+++ b/src/parser/object_parser.ml
@@ -553,7 +553,7 @@ module Object
             env |> with_allow_super Super_prop
         in
         let value = with_loc (fun env ->
-          let sig_loc, (tparams, params, return) = with_loc (fun env ->
+          let sig_loc, (tparams, params, return, predicate) = with_loc (fun env ->
             let tparams = Type.type_parameter_declaration env in
             let params =
               let yield, await = match async, generator with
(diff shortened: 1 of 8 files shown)


goodmind:
@panagosg7 is there something I'm missed?

goodmind:
@panagosg7 anything you can review here?

goodmind:
@panagosg7 is this still on your radar?

mischnic:
Any updates?

mischnic:
I've rebased this PR: https://github.com/facebook/flow/pull/8525

STRML:
Would love to see this merged, it also relates to https://github.com/babel/rfcs/pull/7.

panagosg7:
I'm closing this in favor of #8525 which is essentially the same change, but updated more recently. Unfortunately, this is still blocked on some non-trivial extension that we'll have to address once we get some free cycles.

PR #7736, opened 2019-05-18
Add mixins and implements printers

goodmind:

### DIFF
--- a/src/parser_utils/output/__tests__/js_layout_generator_test.ml
+++ b/src/parser_utils/output/__tests__/js_layout_generator_test.ml
@@ -2113,6 +2113,9 @@ let tests = "js_layout_generator" >::: [
     begin fun ctxt ->
       assert_statement_string ~ctxt "declare class a{}";
       assert_statement_string ~ctxt "declare class a extends b{}";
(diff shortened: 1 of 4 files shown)


panagosg7:
Thanks for filling in these cases @goodmind !

Would you mind adding your tests in https://github.com/facebook/flow/blob/master/src/parser_utils/output/__tests__/js_layout_generator_test.ml instead of the suggest command. While `flow suggest uses this generator its output is not closely monitored and will probably soon move to a diff-based output.

goodmind:
How to run them?

panagosg7:
Just checked and this currently not possible. Will be possible soon though. I'll import and add the tests myself.

mroch:
it's possible to run _some_ of the tests now with `make ounit-tests`, or to run just this test:

```
make _build/src/parser_utils/output/__tests__/parser_utils_output_tests.native
_build/src/parser_utils/output/__tests__/parser_utils_output_tests.native
```

I also had to manually install ounit so I guess we need a `"ounit" {test}` or whatever in `opam`.

this also affects the signature generator, which is tested by src/typing/__tests__/, which can't run the tests externally yet. still need to do some fixing to unblock this...

PR #7737, opened 2019-05-18
Media BOM update

goodmind:
https://html.spec.whatwg.org/multipage/system-state.html#the-navigator-object
https://html.spec.whatwg.org/multipage/history.html#history-3
https://html.spec.whatwg.org/multipage/web-sockets.html#binarytype
https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket-send

https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API
https://developer.mozilla.org/en-US/docs/Web/API/Worker/Worker#Parameters

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -638,10 +638,10 @@ declare class WorkerLocation {
     hash: string;
 }
 
-declare class WorkerNavigator mixins 
-  NavigatorID, 
-  NavigatorLanguage, 
-  NavigatorOnLine, 
+declare class WorkerNavigator mixins
+  NavigatorID,
+  NavigatorLanguage,
+  NavigatorOnLine,
   NavigatorConcurrentHardware {
     permissions: Permissions;
   }
@@ -770,7 +770,7 @@ type PositionOptions = {
 
 type AudioContextState = 'suspended' | 'running' | 'closed';
 
-declare class BaseAudioContext {
+declare class BaseAudioContext extends EventTarget {
   currentTime: number;
   destination: AudioDestinationNode;
   listener: AudioListener;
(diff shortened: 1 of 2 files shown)


dsainati1:
This has an error diff; do you think you could break this up into a couple smaller, more specific PRs, to make it easier to review?

goodmind:
What errors?

goodmind:
MIDI Access https://github.com/facebook/flow/pull/7804

lib/bom.js
```diff
@@ -67,57 +66,116 @@ type Gamepad = {
...
-    title?: string;
-    text?: string;
-    url?: string;
+  title?: string;
+  text?: string;
+  url?: string;
 }
 
-declare class NavigatorCommon {
+type PermissionName = "geolocation" | "notifications" | "push" | "midi" | "camera" | "microphone" | "speaker" | "device-info" | "background-sync" | "bluetooth" | "persistent-storage" | "ambient-light-sensor" | "accelerometer" | "gyroscope" | "magnetometer" | "clipboard";
```
mroch: "clipboard" should be "clipboard-read" | "clipboard-write" according to https://w3c.github.io/clipboard-apis/#clipboard-permissions and https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API. and its own descriptor, I suppose

goodmind: Maybe I'm missing something, but in https://w3c.github.io/permissions/#enumdef-permissionname there isn't "clipboard-read" and "clipboard-write"

goodmind: Yeah, I tried using this API, and there isn't "clipboard"

PR #7740, opened 2019-05-19
Add SpeechSynthesis definitions

goodmind:
https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -1490,3 +1490,86 @@ declare interface CredMgmtCredentialsContainer {
   ): Promise<?CredMgmtCredential>;
   preventSilentAccess(): Promise<void>;
 }
+
+type SpeechSynthesisErrorCode = 
+  | "canceled" 
+  | "interrupted" 
+  | "audio-busy" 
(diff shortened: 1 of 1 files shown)


lib/bom.js
```diff
@@ -1490,3 +1490,86 @@ declare interface CredMgmtCredentialsContainer {
...
+  onend: ?((ev: SpeechSynthesisEvent) => mixed);
+  onerror: ?((ev: SpeechSynthesisErrorEvent) => mixed);
+  onpause: ?((ev: SpeechSynthesisEvent) => mixed);
+  onresume: ?((ev: SpeechSynthesisEvent) => mixed);
+  onmark: ?((ev: SpeechSynthesisEvent) => mixed);
+  onboundary: ?((ev: SpeechSynthesisEvent) => mixed);
+}
+
+type SpeechSynthesisEvent$Init = Event$Init & {
+  utterance: SpeechSynthesisUtterance;
```
panagosg7: why is this not optional?

PR #7747, opened 2019-05-21
[core] Make String#split() return Array<?string>

STRML:
It is otherwise quite easy to write unsound code here that passes Flow.

For instance, consider the following:

```js
function removeCommaSeparators(x: string) {
  const parts = x.split('.');
  parts[0] = parts[0].replace(',', ''); 
  return parts.join(".");
}

removeCommaSeparators('1,000.09'); // fine
removeCommaSeparators('1,000'); // crashes!
```

It is necessary to manually refine the items in the array returned by String#split as you cannot know for sure if they will be present when arbitrary input is allowed.


STRML:
Build error is before tests even run:

```
<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build core_kernel v0.11.1
```

goodmind:
How your example crashes? It returns `1000` for me

STRML:
Apologize, that's a bad example. Imagine taking the decimal portion at
index 1.

TrySound:
All array access are unsound in flow. I think the point is simple usage with interators like forEach, map or for of.
```js
array.forEach(item => {
  item // not nullable
})
``
Direct access is unsafe everywhere. Fixing it like this in every definition will be a big breaking change. Ideally flow should change direct access output to nullable `array[0]` but keep the type non nullable for cases with iterators.

For this case I have this utility.
```js
get<T>(array: $ReadOnlyArray<T>, index: number): ?T
```

jamesisaac:
There's ongoing discussion about a similar change in #6823

STRML:
Compared to indexed access, the DX around this is much less obnoxious. I would argue that `.pop()` and `.shift()` actually make sense to move to nullable types as they are often used in loops where you are checking for undefined anyway, like in [this example](https://flow.org/try/#0MYewdgzgLgBANgS2jAvDA2gRgDQCZsDMAugNwBQAZgK5jBQLgxhVxwCGARnAKYAKIABwA8AFQB8ACjYAnaQC4YAQVlsAnqLEBKBQH4RMAN5kYMadyhVpYGDOkA6AYImbyAXzI9YCKNwC25AHcACwQeGAlvP1QmFnYuPidEaE1NQ3DI3wVmXw5uaRcYVyA).

```js
const list = [1,2,3];
function nullablePop<T>(arr: Array<T>): ?T {
  return arr.pop();
}
let item;
while (item = nullablePop(list)) { (item: number); }
```

My hunch is that there isn't much use of `String#split()` that would be a chore to convert.

PR #7748, opened 2019-05-22
Ignore opam switch folder

goodmind:
This ignores `_opam` folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *TAGS
 _build
+_opam
 /dist
 /_obuild
 /.ocp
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 *TAGS
 _build
+_opam
```
mroch: can it appear in nested folders? if not, `/_opam` will make it only apply at the root, like with `/_obuild` (which i'm gonna go delete, it was for ocp-build). we probably could change to `/_build` too

goodmind: I think it can't

goodmind: _build can appear in nested folders tho

PR #7749, opened 2019-05-22
Add explainers about Object.keys not matching libdefs

goodmind:

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -58,12 +58,12 @@ declare class Object {
 
     static getOwnPropertyDescriptor<T>(o: $NotNullOrVoid, p: any): PropertyDescriptor<T> | void;
     static getOwnPropertyDescriptors(o: {}): PropertyDescriptorMap;
-    /* 
+    /*
       This is no-op.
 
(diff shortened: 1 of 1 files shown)


lib/core.js
```diff
@@ -58,13 +58,27 @@ declare class Object {
...
     static is<T>(a: T, b: T): boolean;
     static isExtensible(o: $NotNullOrVoid): boolean;
     static isFrozen(o: $NotNullOrVoid): boolean;
     static isSealed(o: $NotNullOrVoid): boolean;
+    /* 
+      This is no-op.
+
+      Object.keys is actually defined in OCaml code.
+
+      https://github.com/facebook/flow/blob/master/src/typing/statement.ml#L6308
```
jcready: Should use the permalink otherwise the line numbers won't match in the future: https://github.com/facebook/flow/blob/8ac01bc604a6827e6ee9a71b197bb974f8080049/src/typing/statement.ml#L6308

goodmind: You mean with commit hash?

jcready: Yes

PR #7751, opened 2019-05-22
[DOCS] bug in explicit createContext type example

redbmk:
I ran into a syntax error with the current docs' suggestion of:

```
const Theme = React.createContext<"light" | "dark">("light");
```

I was able to fix it by instead doing:

```
const Theme: React.Context<"light" | "dark"> = React.createContext("light");
```

Also, it seems to recommend [here](https://flow.org/en/docs/react/types/) to use `import * as React` instead of `import React` (especially if you're going to use some types) so I changed that here as well.

### DIFF
--- a/website/en/docs/react/context.md
+++ b/website/en/docs/react/context.md
@@ -13,7 +13,7 @@ Flow will infer types from the way you use a context's `{ Provider, Consumer }`
 pair:
 
 ```js
-import React from "react";
+import * as React from "react";
 
 const Theme = React.createContext();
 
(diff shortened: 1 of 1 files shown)


FireyFly:
Generics in function calls is a fairly recent syntactic addition, so you probably have to update tooling (like babel/its plugins or flow itself) for the documented syntax to work. It certainly does if you try it in flow.org/try.

redbmk:
Ah, ok I just saw that typing context should be supported as of flow 0.70, and I'm on 0.94, but yeah babel and eslint are a bit out of date, so that explains it.

Issue #7750, opened 2019-05-22
New library definitions approach


goodmind:
## Proposal

Support `.js.flow` files as new library definition format

### Main Proposal

https://github.com/goodmind/flow-new-libdef-approach/

### Alternative? Not sure about multiple files then

Enable `@providesModule` for select directories with section (doesn't matter which resolver is choosen `node` or `haste`)
Support `@scoped/package` for `@providesModule`

```
[provides_modules]
node_modules/@flowtyped
```

## Use case

flow-typed migration to NPM

nmote:
We've discussed this in Discord and I think this sounds like a good approach. I am particularly interested in the `*.js.flow` file portion, which will have the following advantages:

* Easy to generate (we basically have the infrastructure for this now, it's just a matter of hooking it up to a CLI command)
* Flow treats them as implementation files, so when they change they don't require a restart, unlike libdef files (and this would be tricky to fix for libdef files)
* They have fewer weird pitfalls, since they are just ordinary implementation files. IMO they are easier to write manually. You just write your normal JS except you can use declare function etc. to avoid providing the bodies of things. For example, everything you declare in a libdef file is considered exported, which can be kind of annoying if you want helper types or something that aren't actually exported.
* You can more accurately model requires because `.js.flow` files follow ordinary node module resolution while libdefs just make a module globally available.

The necessary changes to Flow itself are:

* A feature added to support the `flow.name` field in `package.json` (possibly restricted to only `node_modules/@flow-typed` directories). This is only necessary for the `*.js.flow` portion of the proposal.
* A bug fix to Flow to avoid issuing parse errors when `.json` files are encountered in `libs` directories (#7746). This is only necessary for the libdefs portion of the proposal.

chrisdothtml:
I know this is an old PR, but what are the implications of this for a pnp system (e.g. yarn v2)? Currently the only way to get flow to work with pnp is via libdefs, so if libdefs were moved from `flow-typed` to `node_modules` that would remove all possibility of using flow with pnp

nmote:
If something like this does move forward, I would expect it to be possible for the `flow-typed` CLI tool to pull down the necessary libdefs. Flow can be configured to look at places besides `node_modules` so perhaps we could standardize on putting libdefs if `node_modules_libdefs` or something. Then, PnP could proceed as normal. At runtime, the `node_modules_libdefs` folder would be ignored, but Flow would be able to use it when looking up the types of modules.

[...]

Issue #4917, opened 2017-09-18
RFC: Versioned .js.flow files


willmruzek:
# The Problem

Imagine you have a shared library of (React) components shared across your company's projects. 

This library is on its own version of flow, while all consumers are also on different versions of flow.

Sometimes this results in different flow errors in each project. For example, a comment suppression may be needed and valid in the library, but the consuming app's version of flow throws an `Unused suppression` error. Another case might be the use of opaque types in the library, but the app's flow doesn't support them yet. 

Another side effect is that it's difficult to make major upgrades (e.g. 0.53.x regarding React).

# Proposed Solution

Currently flow consumes *.js.flow files found in a published (public or private) package.

To solve the problem, flow could also consume version-tagged files.

For example, type definitions intended for 0.52 would be available with a *.js.flow-0.52.x file

Note: Generating the different versions of types would be a separate issue

# Why not...

## Force consumers to upgrade

This is a reasonable solution with minor changes between versions, but quickly becomes unreasonable with larger changes (again, 0.53.x regarding React).

## Keep with the lowest common denominator

Flow keeps getting better with each version. It would be a shame to miss out on all the improvements. This would greatly slow the adoption of a new version.

## Submit types to flow-typed

In many cases, companies do not want to release or open source their code. There needs to be a way to share (versioned) types privately.

## Build and maintain a private flow-typed repository

This would be a viable option. The flow-typed CLI already contains the logic to resolve libdefs with different versions of flow. However, the company would need to set up and maintain their own repository and tools. It's possible the resources aren't there to do this, or it's not worth it for one library.

I'm happy to submit a PR, but I wanted to refine the idea first with the community before starting.

jedwards1211:
@mruzekw since `flow` itself and tools like `flow-upgrade` rely on finding files ending in `.flow`, maybe it would be better to for the defs for `src/foo.js` to live in `src/flow-typed/<flow version>/foo.js.flow` instead.

mruzekw:
@jedwards1211 Thanks for your input.

I thought .js.flow files were usually automatically copied from the original source. Why would flow-upgrade care about derivative files?

jedwards1211:
@mruzekw what do you mean automatically copied?  Are you talking about `gen-flow-files` (which last I checked, doesn't always work?)  Or does `cp file.js file.js.flow` typically work well enough?  I've been hand-writing `.js.flow` files with only typedefs in them...

[...]

PR #7758, opened 2019-05-24
Add @flowtyped scope resolution before actual Node resolution

goodmind:
Proof of concept for #7750 

## Issues

- Depends on flow-typed NPM migration

- Clashes with (in particular if there would be migration period from old libdefs to .js.flow)
```
[libs]
node_modules/@flowtyped
```


goodmind:
@nmote this doesn't actually work if this and [libs] .json fix both get merged, because @flowtyped is still using libdefs

goodmind:
I tried to implement there this feature request #4917. Not sure how slower this makes dependency resolving

goodmind:
@jedwards1211 I think you would be interested to check this out?

jedwards1211:
Thanks, though I've never found time to learn much OCaml so it's hard for me to tell exactly what this is doing...what do you mean by versioned .js.flow?  I didn't see anything in your proposal repo about it

goodmind:
@jedwards1211 check tests in this PR

goodmind:
https://github.com/facebook/flow/tree/d4176e0a9b0487bd1abee050d0db2fba94521bfd/tests/declaration_files_node_versions

PR #7760, opened 2019-05-24
Fix flow-upgrade

jbrown215:
This PR does a few things:
1. Upgrades the jscodeshift version (still using my custom fork)
2. fixes the tests
3. bumps the version number

After this is merged, I'll deploy this to NPM. Note that I used this version of flow-upgrade to create D15503065

### DIFF
--- a/packages/flow-upgrade/package.json
+++ b/packages/flow-upgrade/package.json
@@ -1,6 +1,6 @@
 {
   "name": "flow-upgrade",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "A utility for upgrading your codebase to the latest version of Flow.",
   "engines": {
     "node": ">=6"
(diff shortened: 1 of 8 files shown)


osmestad:
I do not think this (version 1.1.1) has been released to NPM? 
Or is this not recommended to be used any more? I did try running flow-upgrade (checked out from the master branch) on a rather big project which is currently on Flow 0.109.0, that did introduce quite a few errors.

jbrown215:
What kind of errors did it introduce? What command did you run? This is still the tool that should be used to upgrade codebases to exact-by-default, but there are some cases where it misses something.

osmestad:
I tried a bit more now, and if I split up the version range in two, it seems the older migration is what messed it up for me. Running:
`node ../flow/packages/flow-upgrade/dist/index.js 0.50.0 0.60.0`

Gives changes like:
From:
`class WindowsSystemMediaTransportControls extends React.Component<Props> {`
To:
```
class WindowsSystemMediaTransportControls extends React.Component<> {
  static defaultProps: Props;
```

So I'm guessing that (parts of) that codemod can not be run on (partly) more up to date codebases?
Running it again also seems to eat more of the signature, for cases where both Props and State are in use.

If I run:
`node ../flow/packages/flow-upgrade/dist/index.js 0.60.0 0.109.0`

Then that seems to work as expected! (Adding explicit inexact)

osmestad:
And for completeness, if I run it the documented way (from NPM):
`yarn create flow-upgrade`

These are the versions it tries to update between:
```
The Flow version you're upgrading from: 0.52.0
The Flow version you're upgrading to:   0.53.0
```

PR #7771, opened 2019-05-30
Support all options in `flow init --options`

### LINKED ISSUE #7769: `flow init --options` is broken
Flow version: 0.99.1

`flow init --options "esproposal.class_instance_fields=enable;"`

## Expected behavior

Generates config with `esproposal.class_instance_fields`
```
[ignore]

[include]

[libs]

[lints]

[options]
esproposal.class_instance_fields=enable

[strict]
```

## Actual behavior

```
[ignore]

[include]

[libs]

[lints]

[options]

[strict]
```

goodmind:
Fixes #7769

Doesn't support `module.name_mappers`

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,4 +1,4 @@
-true: package(core_kernel), package(ppx_deriving), package(dtoa), package(wtf8)
+true: package(core_kernel), package(ppx_fields_conv), package(ppx_deriving), package(dtoa), package(wtf8)
 <**/*.ml*>: ocaml, warn_A, warn(-4-6-29-35-44-48-50), warn_error_A, safe_string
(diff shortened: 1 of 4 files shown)


mroch:
this is going to take me some time to merge because our internal build system (buck) does not like the new ppx. it would seem that ppx_fields_conv doesn't have a way to build its own ppx executable nor a self-contained deriving plugin (ppx_fields_conv.cmxs links against Base). I think we might have to build our own driver similar to ppx_jane to get this to work with buck.

goodmind:
It's mostly for typechecking when new fields would get added

gkz:
Our options have changed a lot since this diff.

PR #7781, opened 2019-06-03
Fix flow-remove-types for input with multi-byte characters

### LINKED ISSUE #7779: flow-remove-types produces invalid JS after unicode characters in comments
Flow version: 0.100.0

## Expected behavior

Given the following input:

```js
// @flow
// ≈
function foo(bar: number) {}
```

`flow-remove-types` should produce:

```js
//      
// ≈
function foo(bar        ) {}
```

## Actual behavior

It produces invalid JS:

```js
//      
// ≈
function foo(bar:         {}
```

cc @mroch @samwgoldman

mourner:
Closes #7779 by implementing the approach described by @mroch in https://github.com/facebook/flow/issues/7779#issuecomment-498290377.

Remaining work:
- [ ] Try eliminating all string manipulation and `buffer.write` calls in favor of purely byte-based operations for performance.
- [ ] Benchmark the change to make sure performance didn't regress.
- [ ] Address any review suggestions.


mourner:
@mroch this is ready for review. While the performance didn't regress, there's no point in optimizing the code much because almost all time is spent in `flow-parser`. Preliminary benchmarks show that it's about 4–5 times slower than the previous `flow-remove-type` version based on `babylon`. 😭

mourner:
The build failed for some reasons unrelated to the PR changes it seems.

packages/flow-remove-types/index.js
```diff
@@ -397,6 +407,7 @@ function getTrailingLineNode(context, node) {
 // Creates a zero-width "node" with a value to splice at that position.
 // WARNING: This is only safe to use when source maps are off!
 function getSpliceNodeAtPos(context, pos, loc, value) {
+  context.bytesAdded += value.length;
```
motiz88: This is _technically_ mixing JS string length (UTF-16 code units) and UTF-8 bytes. Not a problem yet because the only call to `getSpliceNodeAtPos` uses a plain ASCII string (`' =>'`), but maybe we should future-proof this?

packages/flow-remove-types/index.js
```diff
@@ -98,32 +101,34 @@ function resultPrinter(options, source, removedNodes) {
...
           }
         }
       }
 
-      return (result += source.slice(lastPos));
+      offset += sourceBuffer.copy(buf, offset, lastPos, sourceBuffer.length);
+
+      return buf.toString('utf8', 0, offset);
     },
     generateMap: function() {
```
motiz88: How does this issue translate to source maps? I'm suddenly worried that the source map spec is silent on what "columns" mean, but I'd imagine it's meant to be interpreted more like "code points" than "bytes". So we might need to do some Unicode bookkeeping in `generateSourceMappings` as well.

packages/flow-remove-types/test/source.js
```diff
@@ -1,6 +1,8 @@
 /* @flow */
 // @nolint
 
+// multi-byte chars: Гарного дня, котики!
```
motiz88: Can we add some non-BMP (multi UTF-16 code unit) characters to the test? e.g. emoji: `'🐈'.length == 2`, `Buffer.from('🐈').length == 4`

Also, I'd add a test case with non-ASCII characters in the actual code that we process, e.g.
```js
var lambda: λ = (α: number): number => α;
```

thehogfather:
@mourner have you still got time to look into this and address the comments?

mourner:
@thehogfather apologies — can't get to it at the moment but hopefully I'll find some time in a week or two. The only comment left unaddressed is this one:

> How does this issue translate to source maps? I'm suddenly worried that the source map spec is silent on what "columns" mean, but I'd imagine it's meant to be interpreted more like "code points" than "bytes". So we might need to do some Unicode bookkeeping in `generateSourceMappings` as well.

If anyone's up to helping out on the source maps part, I'd appreciate it!

Issue #2207, opened 2016-08-07
"delete" ignored


also:
Flow 0.30.0 allows `delete a.a` in locations where `a.a = undefined` would trigger an error:

``` javascript
/* @flow */

type A = {a: string};

function deletes(a: A): A {
  delete a.a;
  return a;
}

console.log(deletes({a: 'a'}).a.length);
```

I would expect this to result in an error like:

```
6:   delete a.a;
              ^  undefined. This type is incompatible with
3: type A = {a: string};
                ^ string
```

samwgoldman:
Thanks for the report!

vkurchatkin:
@samwgoldman 
I think the tricky part here is that `delete` works with reference (per spec), not with values. So it's not as simple as making delete behave like a function.

It seems that at the moment `delete` is roughly a function with type `(t: mixed) => boolean`, so when you have something like `delete a.a` Flow treats `a.a` as property lookup, while in fact it's not. For example, this should work:

```js
type A = { -a?: string };

declare var a: A;

const x = delete a.a;
```

But instead we get this:

```
7: const x = delete a.a;
                    ^ object type. Contravariant property `a` incompatible with covariant use in
7: const x = delete a.a;
                    ^ property `a`
```

samwgoldman:
That looks like a variance bug to me, for sure. `delete a.a` is a contravariant use, afaict.

vkurchatkin:
@samwgoldman yep, it just confirms that none of "write" semantics of `delete` is implemented

PR #7782, opened 2019-06-04
Implement `delete` operator write semantics, but without refinements

goodmind:
Can't help with #2207, but disallows deleting read-only properties
Only allow `MemberExpression`, `Identifier`

### DIFF
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -149,6 +149,7 @@ type 'loc virtual_reason_desc =
   | RRestParameter of string option
   | RIdentifier of string
   | RIdentifierAssignment of string
+  | RPropertyDelete of string option
   | RPropertyAssignment of string option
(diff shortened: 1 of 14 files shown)


goodmind:
Error message is bad one tho

goodmind:
Error message should be something like this if read-only would be supported
```
Cannot perform delete operation on `e.a` because property `a` is not writable.
```

goodmind:
@mvitousek wow, can I close this now, does refinements work with master?

goodmind:
@mvitousek `delete super.something` shouldn't work
```
class Foo {
      bar() {
        delete super.name;
      }
    }
```
```
VM929:1 Uncaught ReferenceError: Unsupported reference to 'super'
    at Foo.bar (<anonymous>:1:21)
    at <anonymous>:1:11
```

Issue #7777, opened 2019-06-03
Add documentation around function statics and callable objects


wgao19:
I've learned about function statics and callable objects after the release of 0.99. They're really cool, have been there for a while, but are not very well documented in the docs.

Would love work on this if this issue is accepted, and would like some guidance on what should be documented and / or where they should go.

gkz:
Callable objects docs: https://flow.org/en/docs/types/functions/#toc-callable-objects

Issue #106, opened 2014-11-22
Arbitrary properties of function object have type any


dmitry-vsl:
Search keywords: function, statics

Flow finds no error in the following code:

```
/* @flow */
function noop(){}
console.log(noop.foo/noop.bar);
```

I think it is a bug because both noop.foo and noop.bar are undefined.

facebook-github-bot-2:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

savvopoulos:
Would be curious to hear if there's an update on this.

StyMaar:
Any news on this ?

nmn:
@avikchaudhuri Do you think defaulting to `?mixed` instead of `any` will be easy to do for this? That seems to solve the problem.

danwang:
bump

joelochlann:
Just re-activating this thread, by pointing out that this weakness in Flow's type system does lead to real-world bugs, particularly around the use of promises.

For instance:

```javascript
function getPromise(): Promise<boolean> {
  return Promise.resolve(true)
}

// type-checks, works at runtime
getPromise().then(() => true);

// type-checks, fails at runtime
getPromise.then(() => true);
```
https://flow.org/try/#0GYVwdgxgLglg9mABAcwKZQAoCc4FsYDOqAFAJQBci2ehqAPAEZxwA2qAhmAHyIDeAUIkRZ0ILEmr4iAOhEFWANxJQsIVKX4BffvzSYcUkqWlQAFqjDEyiALw8Va0gG5d6SbRPnL1u4gfqnIA

I understand that functions are ultimately objects in JavaScript, but so are arrays, and they have stricter type-checking in Flow. Why not functions too?

```javascript
// type-checks, fails at runtime
({}).blah(() => 'hey')

// type-checks, fails at runtime
(() => {}).blah(() => 'hey')

// doesn't type-check, fails at runtime
[].blah(() => 'hey')
```

Any insights you can give into the decisions around this would be much appreciated!

kangax:
@avikchaudhuri I see you mentioned 5 years ago that this will be fixed. Just pinging to see if there's any progress :D

gkz:
Fixed 0.187

PR #7790, opened 2019-06-05
[DOCS] Improve docs about callable property

wgao19:
Addressing #7777, this PR hopes to improve docs around callable property.

The content of these edits credits to the discussions at niieani/typescript-vs-flowtype#55, especially insights from @omninonsense and @goodmind. Would really appreciate your inputs on this revision to the docs as well.

### DIFF
--- a/website/en/docs/types/functions.md
+++ b/website/en/docs/types/functions.md
@@ -332,7 +332,7 @@ const g: F = (x?: number | string) => { // $ExpectError
 }
 ```
 
-Callable objects can be viewed equivalently as functions with static fields. Function statics are initially typed to `{}`. Flow will notice when you try to access function statics without annotation.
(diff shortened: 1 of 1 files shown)


goodmind:
Function statics isn't `{}`
https://flow.org/try/#0GYVwdgxgLglg9mABMAFASkQbwL4ChfAB0wccQA
https://github.com/facebook/flow/issues/106

wgao19:
Is it about function _type_ statics? Because I read [in this commit](https://github.com/facebook/flow/commit/b868f61026b9b34cb22c3f59e1bcbb65a614c71b) that _function type statics_ has been changed to "an empty, inexact object type". I'm wondering if rephrasing it as follows will be more accurate and clearer:

> Callable objects can be viewed equivalently as functions with static fields. Function _type_ statics are initially `{}`. Flow will notice when you annotate a variable as function type and try to access its statics without annotation.

```js
type F = () => void;

const f: F = () => {
  f.foo = 'bar'; // $ExpectError "missing in statics of function type"
}
```

PR #7798, opened 2019-06-07
Number literal object keys variance annotation parsing

goodmind:
Related #7593 

It's impossible to have negative number key, so it can only mean contravariant annotation

```js
const bar = {-1: 'foo'}

VM153:1 Uncaught SyntaxError: Unexpected token -
```

```js
type X = {-1: T}
type X = {+1: T}
type X = {1: T}
```


goodmind:
@mroch I think this is related to number keys and #380

goodmind:
@mroch can you look into this?

PR #7800, opened 2019-06-07
Make operator assignment aware of property variance

### LINKED ISSUE #7195: `+=`, `*=`, etc not caught by read only modifier
```
/* @flow */

class ReadOnlyNum {
  +val: number;

  inc() {
    // this.val = 1; // causes error
    
    // does not cause error
    this.val += 1;
    this.val -= 1;
    this.val *= 1;
    this.val /= 1;
    this.val %= 1;
    this.val &= 1;
    this.val <<= 1;
    this.val >>= 1;
    this.val >>>= 1;
    this.val ^= 1;
    this.val |= 1;
    this.val **= 1;
  }
}
```

goodmind:
Fixes #7195

### DIFF
--- a/src/typing/statement.ml
+++ b/src/typing/statement.ml
@@ -5134,7 +5134,7 @@ and op_assignment cx loc lhs op rhs =
   | Assignment.BitAndAssign
     ->
       (* lhs (numop)= rhs *)
-      let reason = mk_reason (RCustom "(numop)=") loc in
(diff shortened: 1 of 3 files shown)


goodmind:
I just noticed that error messages doesn't match with operator

goodmind:
@mvitousek would this be fixed too?

PR #7802, opened 2019-06-07
Resolve imported types

vicapow:
Changes this:

Screen Shot 2019-06-10 at 1 36 50 PM

to be this:

Screen Shot 2019-06-10 at 1 36 16 PM


src/common/ty/ty_printer.ml
```diff
@@ -140,12 +140,6 @@ let type_ ?(size=5000) ?(with_comments=true) t =
 
   and type_alias { ta_name = { provenance; name; _ }; ta_tparams; ta_type } =
     match provenance with
```
panagosg7: You don't need to match against provenance here. Just always return `type ...`.

vicapow: okay! should be updated now

Issue #7461, opened 2019-02-08
Unexpected reserved type in typeof _ 


goodmind:
Flow version: 0.92.1

```js
declare var _: number
type A = typeof _
```

## Expected behavior

No errors

## Actual behavior

```js
2: type A = typeof _ 
                   ^ Unexpected reserved type
```

* Link to Try-Flow or Github repo: https://flow.org/try/#0CYUwxgNghgTiAEA3W8D6AueA7ArgWwCMQYAoAFwE8AHBAQXgF55KaB7AMzRKA

Bannerets:
Regression in v0.84.0.

goodmind:
Same error with any other reserved type
```
declare var PropTypes: {any: string}
type S = typeof PropTypes.any
```
```
2: type S = typeof PropTypes.any                             
                              ^ Unexpected reserved type
```
https://flow.org/try/#0CYUwxgNghgTiAEA3W8AKMD2AHAKgTyxAGcAueAbygDs8yiAXGASyoHMBfAKHoIQGV4AXng9CGAGZpMuXkQB01PJyA

yuyao17:
Is there a workaround for this?

goodmind:
@yuyao17 renaming variable

goodmind:
Better example if Flow implements https://github.com/tc39/proposal-promise-any
```js
declare class Promise<+R> {
  static any<Elem, T: Iterable<Elem>>(promises: T): Promise<Elem>;
}

type F = typeof Promise.any
```
```js
5: type D = typeof Promise.any                           ^ Unexpected reserved type
```

PR #7809, opened 2019-06-09
Remove reserved type check in typeof

### LINKED ISSUE #2682: typeof annotation seems to parse incorrectly
It seems that flow parses annotations like `T: typeof string` incorrectly, it treats `string` as if it were a `StringTypeAnnotation`, but presumably it should be an identifier or `GenericTypeAnnotation`?

``` js
const string = "Hello World";
const foo: typeof string = string;
```

### LINKED ISSUE #3055: `typeof bool` error: unexpected typeof expression
```js
let bool = true;
let test: typeof bool = false;
// Error: ^ Unexpected typeof expression
```

> @mroch: the lexer sees `bool` as a T_BOOL, sounds like the parser doesn't allow for reserved words in typeof

goodmind:
Temporarily fixes #3055, #7461

Doesn't #2682 

I tried to replace this parsing with `MemberExpression` but this breaks too many things in typechecking part, so I decided to workaround this in parser. In long term it should parse it as expression, and should be fixed in Babel too

### DIFF
--- a/src/parser/parser_env.ml
+++ b/src/parser/parser_env.ml
@@ -178,6 +178,7 @@ type env = {
   no_in                 : bool;
   no_call               : bool;
   no_let                : bool;
+  no_reserved_type      : bool;
   no_anon_function_type : bool;
   no_new                : bool;
   allow_yield           : bool;
@@ -230,6 +231,7 @@ let init_env ?(token_sink=None) ?(parse_options=None) source content =
     no_in = false;
     no_call = false;
     no_let = false;
+    no_reserved_type = true;
     no_anon_function_type = false;
     no_new = false;
     allow_yield = false;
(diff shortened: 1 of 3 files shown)


tanhauhau:
What will should it be parsed as?
UnaryExpression with typeof operator?
Let me fix it on the babel side

goodmind:
@tanhauhau no, I think babel doesn't need anything now, I mean if we would make it `MemberExpression` in Flow, but currently it breaks many things

tanhauhau:
@goodmind apparently babel has [reserved type check as well](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEAEDMBsHsHcBQAXAngBwKagIagLyhpaySgD6A3EA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=flow&prettier=false&targets=&version=7.4.5&externalPlugins=%40babel%2Fplugin-transform-flow-strip-types%407.4.4)

goodmind:
@tanhauhau it works with `type a = typeof boolean;` when Flow doesn't, so looks like it is only `_`

PR #7811, opened 2019-06-09
Add Shared Memory definitions

goodmind:
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Atomics

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -908,10 +908,10 @@ declare var console: {
 };
 
 declare class SharedArrayBuffer {
-  constructor(byteLength: number): SharedArrayBuffer;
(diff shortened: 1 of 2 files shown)


devongovett:
Awesome! Could you also add `SharedArrayBuffer` support to `Buffer.from` as well? https://github.com/facebook/flow/blob/master/lib/node.js#L119

goodmind:
@devongovett wow, I thought everyone already moved to TS

Issue #6427, opened 2018-06-07
Flow cannot resolve 'http2'


budarin:
http2 is out in node but it is not typed in Flow.

SamChou19815:
The libdef has been moved to flow-typed. If you want to make improvements, please make PR over there.

Issue #7719, opened 2019-05-13
Add additional type safety for unhandledRejection


KevinGrandon:
Flow version: 0.98.1

## Expected behavior
Process methods which listen for unhandledRejection should have additional type information. The argument should be `mixed` or similar, forcing the user to refine it.

## Actual behavior
Anything can pass through the callback to unhandledRejection, including non-Error objects. Perhaps by default the type could be `mixed`?

* Reproducible code:
```
// This code produces no errors in Flow, can it?
process.once('unhandledRejection', async e => {
  return e + 1;
});
```

Hello, not sure if something like this is possible, but it would definitely help prevent bugs if so.

dsainati1:
This can probably be done via overloading the definition for `once` in the libdefs to be more specific when the argument is `'unhandledRejection'`. I would be happy to accept a PR for this.

goodmind:
I think there was some bug if you overload string literals

dsainati1:
We do this all the time in the libdefs (see dom.js for a few examples) so I think it should definitely be possible here.

KevinGrandon:
I threw together something in #7761, but not sure if this is what we had in mind, also not able to run tests yet locally. I'll try again in the future, or someone can feel free to steal it. Thanks for chiming in.

goodmind:
@dsainati1 I tried it in Try Flow, but it doesn't select right overload...
[...]

PR #7814, opened 2019-06-10
Add http2 node module definitions

goodmind:
https://nodejs.org/dist/latest/docs/api/http2.html

Fixes https://github.com/facebook/flow/issues/6427

This doesn't overload event emitters because flow can't decide what to select from previous issues
#7719


SamChou19815:

PR #7816, opened 2019-06-10
[LSP] fix range in completion response

mroch:
the range in a TextEdit needs to point to the whole range to replace. when we send back a TextEdit in a completion response, it was just the cursor position (ish?). So if you're completing `foo.ba|` and we return `bar()`, the range needs to cover all of `ba` so that it gets replaced.

### DIFF
--- a/newtests/lsp/completion/test.js
+++ b/newtests/lsp/completion/test.js
@@ -130,9 +130,9 @@ export default suite(
                 },
                 {
                   label: 'aUnion',
-                  kind: 6,
-                  detail: 'number',
-                  inlineDetail: 'number',
+                  kind: 13,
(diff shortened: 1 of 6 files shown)


newtests/lsp/completion/test.js
```diff
@@ -130,9 +130,9 @@ export default suite(
                 },
                 {
                   label: 'aUnion',
-                  kind: 6,
-                  detail: 'number',
-                  inlineDetail: 'number',
+                  kind: 13,
```
mroch: this is unrelated, but the test runner doesn't catch this. it is doing a very fuzzy string match and `kind: 6` and `detail: 'number'` exist elsewhere in the response; and it doesn't check that the fragments in the actual response exist in the expected response.

PR #7817, opened 2019-06-11
fix: allow postMessage take array of transferables

Gozala:


goodmind:
@Gozala isn't array already iterable?

Gozala:
> @Gozala isn't array already iterable?

It is `Iterable` but not an `Iterator` which is what current typedefs expect.

goodmind:
Maybe Iterable was intended, w3c defines it as `sequence<T>`, maybe Transferable also can be implemented

nmote:
Closing as this isn't backed up by docs. Feel free to reopen/create a new one if this is still important.

PR #7827, opened 2019-06-13
Add Element and Window scroll methods

lyleunderwood:
All of this is just copied from https://github.com/facebook/flow/pull/7462, it's unclear why that PR was closed.

>`(Window|Element).scroll(To|By)?` are a set of functions for managing browser window and element scroll.
>
>MDN documentation -
>
>ScrollToOptions interface - https://developer.mozilla.org/en-US/docs/Web/API/ScrollToOptions
>`Window.scroll()` - https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll
>`Window.scrollBy()` - https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollBy
>`Element.scroll()` - https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll
>`Element.scrollBy()` - https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollBy
>From testing -
>
>With `(x: number, y: number)`, both params are required.
>With `(options: ScrollToOptions)`, none of the properties are required.
>Any number in both cases can be null or undefined or even a string, in which case the browser will attempt to cast it to a number, although I don't think typing it that way is a good idea

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -121,6 +121,12 @@ declare type ShadowRootInit = {
        ...
 }
 
+declare type ScrollToOptions = {
+  top?: number;
+  left?: number;
+  behavior?: 'auto' | 'smooth';
+}
+
 type EventHandler = (event: Event) => mixed
 type EventListener = { handleEvent: EventHandler, ... } | EventHandler
 type MouseEventHandler = (event: MouseEvent) => mixed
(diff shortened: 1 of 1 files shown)


lyleunderwood:
nice

PR #7829, opened 2019-06-14
fix: Typo in react abstract component test comment

wgao19:
I was trying to understand React abstract component and realized that it might have been a typo here?


panagosg7:
I think the comment is correct here, ie. the definition of `AbstractComponent` would be something like
```
React$AbstractComponent<-Config, +Instance>
```
so because:
```
Subcomponent <: Component
```
we get 
```
React$AbstractComponent<{}, Subcomponent> <: React$AbstractComponent<{}, Component>
```

wgao19:
Oh, I see. Thanks for the explanation

Issue #7491, opened 2019-02-20
[>= v0.89.0] $ObjMap<...> always compatible with empty objects?


namuol:
Flow version: **0.89.0 - 0.93.0**

[**Try on Flow.org**](https://flow.org/try/#0C4TwDgpgBAggQgYSgXigbwFBSgQwFxQDOwATgJYB2A5gDRZQBGBx51d2Axs6ZbRgL4BuDBlCRYiAOplgACwByAVwC2DCCUIooAEgDyDAFYBZHGAA88BDSgAKHBRAoAfFAoq1JAJRPhGDgHsKYihgCGIARgJLaTklVXVNVExsfCgABnZGAnDMrigAJjohEQB6EqgASQoANxwAGzIAE2t1En8NIll-AHcKXE0IAA9IDlDG22UyQkJeKAADDjnPPD9A4NDifKipGQV3BK1k3AIM+iYoHPoyqDzCgV9rqtqG5sZFYFd-KFb2zUIu3pQShQJzINIAOgAHABOFYBIIfDbAADM2wQMT28Q6SSu5VSp2w13Ol0J5VuRQe5Se9SaLRIbQ6-x6fWBoIhMNsDHerggEHGwC+3RIpig3V2OgAooMcKNPKsESEwsAACwEbRSmXACw7WL7DQuHGk47pTJE7KmskEO5CIA)

![screen recording 2019-02-20 at 03 17 pm](https://user-images.githubusercontent.com/444629/53132083-6835e200-3523-11e9-92d8-74f87c2d58c5.gif)

## Expected behavior

Given this snippet:

```js
const foo = {
  a: 'test',
  b: 'hello',
  c: 'okay',
};

type FooWithNumbers = $ObjMap<typeof foo, (any => number)>;

const test1: FooWithNumbers = {};
```

I expect to see a warning about missing `a`, `b`, and `c`:

```
9: const test1: FooWithNumbers = {};
                                 ^ Cannot assign object literal to `test1` because property `a` is missing in object literal [1] but exists in object type [2].
References:
9: const test1: FooWithNumbers = {};
                                 ^ [1]
9: const test1: FooWithNumbers = {};
                ^ [2]
9: const test1: FooWithNumbers = {};
                                 ^ Cannot assign object literal to `test1` because property `b` is missing in object literal [1] but exists in object type [2].
References:
9: const test1: FooWithNumbers = {};
                                 ^ [1]
9: const test1: FooWithNumbers = {};
                ^ [2]
[...]

Issue #7463, opened 2019-02-10
Regression since 0.89.0 (Maybe types for objects)


debel27:
Flow version: 0.89.0 and above

## Expected behavior

The doc on [_Maybe Types_](https://flow.org/en/docs/types/maybe/) states the following

> In the case of objects, a **missing** property is not the same thing as an explicitly `undefined` property.
> ```js
> // @flow
> function acceptsMaybeProp({ value }: { value: ?number }) {
>  // ...
> }
>
> acceptsMaybeProp({ value: undefined }); // Works!
> acceptsMaybeProp({});                   // Error!
> ```

I would therefore expect an error on `acceptsMaybeProp({});`

## Actual behavior

From [v0.89.0](https://github.com/facebook/flow/releases/tag/v0.89.0), Flow doesn't trigger an error for missing properties anymore.

```js
function acceptsMaybeProp({ value }: { value: ?number }) {
  // ...
}

// until v0.88.0 : KO (as expected)
// from v0.89.0 : OK (possible regression)
acceptsMaybeProp({});
```

[Try it out here](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoKBXAdgYwC4CWcWYAhjjgKYAOeAzgLKkCeARpQAoBOc1AFAG8wAN1IwMlMAF8AXGCGjxlOQH4sGALbsu0gJTzUYMMGBgAdBdRTUqclVqMW7br0EixEudgAmlKASxKbz0AbmNTAHU4LgBrOgBCWwoaeiY2Th5+ASldMKMjIA).

It seems that React elements [are not affected](https://flow.org/try/#0PTAEAEDMBsHsHcBQBLAtgB1gJwC6gFSgCGAzqAEoCmRAxnpFrKqAORbV0sDcikArgDs6yWANABZAJ4BhJpgGUBOABTpG6EgC5QAbxJNKABXXaA-AL6oARpSwBfAJS7EoUOxx8sYi9Gg87iCCgANIA8og0oiR4qDJyooo4AKLQoAC8oAA8UrIYCUqgwAB8XEA).

vtereshyn:
`someProp?: number` - says that you can not have your `someProp`, so it can be undefined
`someProp: ?number` - says that your `someProp` can be null. 

Maybe you need first case?

[...]

Issue #7456, opened 2019-02-07
No error assigning {} to object type with required properties


jtbandes:
Flow version: `master`, v0.92.1

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoALgTwA4FMwAaYAvGAN5gCGAXGAM7oBOAlgHYDmYAvgNyoy50YOACMAVrSKkyvfoPpNaDFhxLDxAOko8gA
```js
/* @flow */
type X = { a: string };
let obj: X = {};
let str: string = obj.a;
```

## Expected behavior

Behaves as expected in Flow v0.88.0:
```
3: let obj: X = {};
                ^ Cannot assign object literal to `obj` because property `a` is missing in object literal [1] but exists in `X` [2].
References:
3: let obj: X = {};
                ^ [1]
3: let obj: X = {};
            ^ [2]
```

## Actual behavior

No error in Flow v0.89.0 and newer.

dsainati1:
This is intentional. The empty object `{}` is treated as unsealed. If you want to prevent this behavior, mark `X` as exact as so: 
```
type X = {| a: string |};
let obj: X = {};
let str: string = obj.a;
```

jtbandes:
Still, within a declaration-with-initializer statement, Flow should be able to know that it doesn’t satisfy the requirements of the destination type, right?

dsainati1:
This is an unfortunately side-effect of the way that unsealed objects behave. If you want to avoid this behavior, I strongly recommend not using them, and making all your object types exact in order to get the guarantees that you want.

jtbandes:
That makes sense, but even if it's intentional that `{}` is treated as unsealed, I would still consider this a pretty dangerous behavior. Here's another example: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoALgTwA4FMwAaYAvGAN5gCGAXGAM7oBOAlgHYDmYAvgNyqpQArqwDG6ZnFZh2udAEEAFHABGAK1oEAlOVRgwjWYMZSVqgHSU+XfjFn0mtBiw4lpsxWS6a+QA

```js
type X = { a: string };
function getA(obj: X) {
  return obj.a;  // looks good, any object of type X should have the .a property
}

// elsewhere...
let str: string = getA({});  // no error
```

The recommendation to use exact object types makes sense, and I generally try do to that, but this behavior seems to show that someone using a non-exact type can't even rely on the properties they expected being present at all 😕

[...]

PR #7839, opened 2019-06-20
[DOCS] Specify new behavior on {} starting in Flow 0.89

SebastienGllmt:
This behavior surprised our team and we're not the only ones (see #7491 #7456 #7463)

This change in behavior should be mentioned in the changelog at the least.


goodmind:
Can you explain how this works with Flow Try, to me looks like a bug

SebastienGllmt:
You can see some examples in the issues I linked in this PR. There seems to be some disagreement about whether or not this is a bug. @dsainati1 says this is an intentional change and @jbrown215 that it is something that they hope to fix. Regardless of the result, we should at least mention that this change happened in the changelog (what this PR is for) and the discussion can continue somewhere else if needed

Issue #2675, opened 2016-10-24
Support flow:main in manifest


steelbrain:
(I searched for an existing issue but couldn't find one so sorry if this is a duplicate)
### The problem

The problem is that we want flow to understand the types of flow-typed modules correctly when requiring them. 
### The workaround

We currently have to keep `.flow` files along with `.js` files after they are babel compiled and that adds an extra build step and a ton of headache.
### The solution

If Flow has support for `flow:main` (like `jsnext:main` [for rollup](https://github.com/rollup/rollup/wiki/jsnext:main)), we would not have to add anything new to our build step. We can just leave the `src` dir in our npm package and set `flow:main` to `src/`.

This is a follow up to https://github.com/facebook/flow/pull/1459

natew:
Any update here? This prevents our entire stack from being flow typed properly!

steelbrain:
@STRML About your comment on the PR, `AgentMe/flow-copy-source` doesn't integrate well with babel so I've forked the babel cli and created a dropin replacement [babel-flow-cli](https://github.com/steelbrain/babel-cli-flow), it works well even with the watcher and updates `$.flow` copies instantly

TrySound:
You might don't need to look for `module` field. Use this way
```js
{
  "main": "./lib/index.js",
  "module": "./esm/index.js",
  "files": ["lib", "esm", "src"],
  "scripts": {
    "build:flow": "echo \"// @flow\\n\\nexport * from '../src';\" > lib/index.js.flow"
  }
}
```

wtgtybhertgeghgtwtg:
I think that's what he was addressing under "The workaround".  It's certainly an option some libraries are taking, but it feels hacky and potentially increases bundle size.

TrySound:
@wtgtybhertgeghgtwtg How this increases bundle size? `lib/index.js.flow` file is used only by flow. Bundlers do not care about it. Also with this you don't need to duplicate types anywhere. Sust place one `.js.flow` reference near entry point.

TrySound:
Also there is #5729 with similar feature, so this can be closed

/cc @vkurchatkin

wtgtybhertgeghgtwtg:
Sorry, I meant package size.
I don't think #5729 solves this issue.  Ideally, Flow types should work without an extra build step and without the consumer having to alter their configuration.

TrySound:
Package size? With a few bytes? Who cares? My way is pretty good for this without any new fields in package.json.

[...]

PR #7842, opened 2019-06-21
Implement versioned .js.flow

goodmind:
Proof of concept for #4917 
<s>Has bad performance on all resolutions</s>

## Performance checks

Current measurements: https://gist.github.com/goodmind/65695c3b7a3aa0766812b15779a984dc

1. Build flow binary
2. Run `flow check --profile` on codebase
3. Record results
4. Repeat with regular flow binary and this PR, properly stopping all servers before running new binary


src/parser_utils/package_json.ml
```diff
@@ -7,18 +7,27 @@
 
 module Ast = Flow_ast
 
+type flow_t = {
+  typesVersions: string SMap.t;
```
nmote: Let's name this `types_version` in keeping with typical OCaml naming idioms

src/parser_utils/package_json.ml
```diff
@@ -7,18 +7,27 @@
...
 let (>>=) = Core_result.(>>=)
 
-let empty = { name = None; main = None; }
+let empty = { name = None; main = None; flow = None; }
+
 let name package = package.name
 let main package = package.main
+let flow_types_versions package = match package.flow with
+  | Some t -> t.typesVersions
+  | None -> SMap.empty
```
nmote: Is there any case where a client may want to distinguish between a package with no `flow` entry, a package with a `flow` entry that does not contain a `typesVersion` property, and a package with a `flow` entry containing an empty `typesVersion` property?

goodmind:
@nmote how can I test server restart?

src/parser_utils/package_json.ml
goodmind: I think there isn't any difference, but packages with empty `types_versions` shouldn't be added to cache and skip checking entirely, I think I missed this

goodmind: It also depends if we add more fields to `flow`, like #2675

nmote:
I took a quick look and didn't see any existing examples of testing for server restart. With that in mind, I would be happy if you manually test that the server restarts and update the PR description with the steps you took.

nmote:
It appears that we don't have Core's String module available for our internal build, so the build here is failing. I would have mentioned this sooner but this was masked by another build issue that was easy to fix.

@mroch, do you know if it would be reasonable to pull in Core's String? @goodmind, if not, could you rewrite this without using that library?

Sorry, I know it's not a good developer experience to have a separate build system that you can't see or use. Unfortunately there's not a whole lot we can do about that currently.

goodmind:
Yeah, it just needs `substr_index`, couldn't find this in jane street core

goodmind:
@nmote I removed `Base`

nmote:
Unfortunately the test you've added in this commit fails on Linux. Do you have access to a Linux development environment to look into this failure?

goodmind:
@nmote no, I'm on Mac, what fails?

goodmind:
@nmote any news?

nmote:
Yes, I've been trying to get to this. Sorry again for the delay 😞 

I'm planning to rebase it past the ocamlformat change for you, since I'm not sure whether we have ocamlformat properly set up on GitHub yet. I was hoping to get to that yesterday, but ran out of time. I should be able to do that and give you a review today.

nmote:
I've rebased this past the ocamlformat update and I put up #8135 with the updated code. Feel free to take what's there and push to your branch here.

I'll do a round of review over there, so that I don't have to wait for you to update this PR.

nmote:
I've done a round of reviewing over there. Let me know when you get a chance to respond.

nmote:
I'm still concerned about the code in the `types_versions_candidates` function and `node_paths_part.ml` -- see my comment in #8135

nmote:
Closing since there hasn't been any movement on this for some time

PR #7843, opened 2019-06-21
Update docs on `Object` and `Function` to call out alias to `any`.

vicapow:


website/en/docs/types/objects.md
```diff
@@ -304,6 +304,10 @@ function add(id: number, name: string) {
 
 ### `Object` Type <a class="toc" id="toc-object-type" href="#toc-object-type"></a>
 
+** NOTE: `Object` has become an alias to `any`. For new code, either use `any` or
+`{ [key: string]: any}` but there really is no reason to use `Object` and will
```
kangax:
```suggestion
`{ [key: string]: any}` but there really is no reason to use `Object` and it will
```

website/en/docs/types/functions.md
```diff
@@ -321,6 +321,10 @@ add.bar = "hello world";
 
 ### `Function` Type <a class="toc" id="toc-function-type" href="#toc-function-type"></a>
 
+** NOTE: `Function` has become an alias to `any`. For new code, either use `any` or
+`(...args: Array<any>) => any` but there really is no reason to use `Function` and will
```
kangax:
```suggestion
`(...args: Array<any>) => any` but there really is no reason to use `Function` and it will
```

website/en/docs/types/objects.md
```diff
@@ -314,17 +318,29 @@ function method(obj: {}) {
 ```
 
 However, if you need to opt-out of the type checker, and don't want to go all
-the way to `any`, you can instead use `Object`. **`Object` is unsafe and
-should be avoided.**
+the way to `any`, you could use `{ [key: string]: any}`. ** NOTE [`any`](../any/) is unsafe and
+should be avoided.** For historical reasons, the `Object` keyword is still avaiable
+but is an alias to [`any`](../any/). In previous versions of Flow, `Object` was the same
+as `{ [key: string]: any}`.
```
kangax: maybe say "was roughly the same" since there are subtle behaviour differences, e.g.:

``` 
function f(): Object {
  return {}
}
f() + 1
```
won't throw but:

```
function f(): { [key:string]: any } {
  return {}
}

f() + 1
```
will

FireyFly: The former does seem to [produce an error in v0.88.0 and below][1], which I believe is what the paragraph is referring to about previous versions.

[1]: https://flow.org/try/#0PTAEAEDMBsHsHcBQjIFcB2BjALgS1uqJABQCUAXKAPIBGAVgKY6gDeiooATg9qp4SwC+iYSjKgA1KACMiIA

website/README.md
```diff
@@ -47,6 +47,7 @@ Use Jekyll to serve the website locally (by default, at
 `http://localhost:8080`):
 
 ```sh
+make js
```
lyleunderwood:
```suggestion
$ make js
```

villesau: Out of curiosity: Why it's common practice to add $ before each line in shell examples? It just makes copy & pasting harder :D

lyleunderwood: it's because most terminal prompt strings end in "$". the $ generally indicates a normal user where # is used to indicate root, for a quick visual indication of whether or not you are root. the $ in the lines of the example are an indication that these are lines to be run in the terminal, otherwise it might be ambiguous what the purpose of these example lines is. in this case, if you're familiar with `make` and `cd` then it's probably pretty clear but if you saw something like:

```
find '.'
```

then it might be significantly more ambiguous depending on the context that you should run this line at your terminal.

also [here's](https://superuser.com/questions/57575/what-is-the-origin-of-the-unix-dollar-prompt) some quick history of $, apparently it dates at least as far back as unix 7.

website/README.md
```diff
@@ -47,6 +47,7 @@ Use Jekyll to serve the website locally (by default, at
 `http://localhost:8080`):
 
 ```sh
+$ make js
```
panagosg7: Can you remove this? It's unrelated and also implied by the `make` two lines below.

website/en/docs/types/functions.md
```diff
@@ -348,5 +351,16 @@ method(function(a: number, b: number) {
...
+
+```js
+function method(obj: Function) {
+  obj = 10;
+}
+
+method(function(a: number, b: number) {
+  // ...
+});
+
```
panagosg7: ### 
```suggestion
```
```

Comment (panagosg7) [inline website/en/docs/types/functions.md]: Close the triple backticks here

File: website/en/docs/types/objects.md [hunk 7]
```diff
@@ -304,6 +304,10 @@ function add(id: number, name: string) {
 
 ### `Object` Type <a class="toc" id="toc-object-type" href="#toc-object-type"></a>
 
+** NOTE: `Object` has become an alias to `any`. For new code, either use `any` or
+`{ [key: string]: any}` but there really is no reason to use `Object` and it will
```
panagosg7: Similar change in tone like I suggested for `Function`.

website/en/docs/types/functions.md
```diff
@@ -321,6 +321,10 @@ add.bar = "hello world";
 
 ### `Function` Type <a class="toc" id="toc-function-type" href="#toc-function-type"></a>
 
+** NOTE: `Function` has become an alias to `any`. For new code, either use `any` or
```
panagosg7: I don't feel very strongly about this, but you could tone it down a bit:
"For new code prefer `any` or `(...args: Array<any>) => any`. `Function` will be deprecated and removed in a future version of Flow."

website/en/docs/types/functions.md
vicapow: good catch! sorry I wasn't actually able to run the site generation locally.

PR #7849, opened 2019-06-25
Add $ObjSingleton destructor

goodmind:
Allows to create objects from unions, but I don't know how to implement type-at-pos for it.
I tried making it real destructor, but it doesn't work entirely, so I made it work like `$Exact`

```js
type S = $ObjSingleton<"a" | "b", mixed> // {| a: mixed, b: mixed |}
```

```js
declare function onAction<T, Result>(myEnum: T, pattern: $ObjSingleton<T, mixed>): T;

declare var action: "Action1" | "Action2" | "Action3";

const result = onAction(action, {
  // no error
  Action1: () => "",
  Action2: () => "",
  Action3: () => ""
});

const result2 = onAction(action, {
  // error
  Action1: () => "",
  Action2: () => ""
});
```

```js
declare function foo<T, U>(
  key0: T,
  key1: U
): {|
  ...$ObjSingleton<T, number>,
  ...$ObjSingleton<U, string>
|};

foo("b", "a").a; // string
foo("b", "a").b; // number
foo("b", "a").c; // error
```


goodmind:
Should be superseded by #7913

gkz:
We are working on implementing a version of mapped types instead of going with the approach of adding more such utility type functionality.

Issue #7397, opened 2019-01-26
flat/flatMap not available


gajus:
There is currently no open issue describing the lack of [`flatMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap).

As this is a [stage 3](https://github.com/tc39/proposals#stage-3) proposal, it should be supported by Flow.

lll000111:
Reopened #6602 which has more info.

gajus:
> Reopened #6602 which has more info.

#6602 asks how to implement such types in user-land instead of raising the issue that types for Array#flat/ Arrray#flatMap are missing in the standard Flow type definitions. The two are related, but different issues.

lll000111:
They are not different, Flow types are Flow types. It's really obvious that it should be in the Flow lib since the question is for an official JS function.

wchargin:
See #6948 for an attempt at this.

amiralies:
anyone progress on `flat`?

amiralies:
as a workaroung im using identity function as mapper in flatMap 
```js
[[1], [2], [3]].flatMap(x =>x);
```

gkz:
I added support for common patterns (that are possible to type) for `.flat`:

https://flow.org/try/#0MYewdgzgLgBGCuBbCAuGBBATpghgTwB4FEAjAU0wD4YBeGAbXoEYBdAGgYCZ2GBmFlgDoAZgBscUABQBKANxA

callumgare:
This doesn't seem to deal with arrays of different levels of nesting https://flow.org/try/#0MYewdgzgLgBGCuBbCAuGBBATpghgTwB4FEAjAU0wD4YBeGAbXoEYBdAGgYCZ2GBmHgCwsAdADMANjigAKAJQBuIA

That seems like a bit of a major issue to me since a key use case of .flat() is to take input with some varied level of array nesting and flatten it out so all elements are consistent. It's the very first example given in the MDN entry https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

PR #7854, opened 2019-06-26
Add Array#flatMap method

goodmind:
Fixes part of https://github.com/facebook/flow/issues/7397

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -226,6 +226,7 @@ declare class $ReadOnlyArray<+T> {
     keys(): Iterator<number>;
     lastIndexOf(searchElement: mixed, fromIndex?: number): number;
     map<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => U, thisArg?: any): Array<U>;
+    flatMap<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => $ReadOnlyArray<U> | U, thisArg?: any): Array<U>;
 
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => T,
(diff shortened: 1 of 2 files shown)


goodmind:
@nmote

tests/arraylib/array_lib.js
```diff
@@ -62,3 +62,53 @@ function of_test() {
...
+  }
+  function case4() {
+    let a = [5, 2, 3, 4];
+    let b: Array<number | string> = a.flatMap(n => (n < 0 ? [1, 2, 3] : "ok"));
+    let c: $ReadOnlyArray<number> = [5, 2, 3, 4];
+    let d: Array<number | string> = a.flatMap(n => (n < 0 ? [1, 2, 3] : "ok"));
+  }
+  function case5() {
+    let c: $ReadOnlyArray<number> = [5, 2, 3, 4];
+    let d: Array<number> = a.flatMap(n => {
```
chicoxyzzy: Seems that `a` is missing here

lib/core.js
```diff
@@ -226,6 +226,7 @@ declare class $ReadOnlyArray<+T> {
     keys(): Iterator<number>;
     lastIndexOf(searchElement: mixed, fromIndex?: number): number;
     map<U>(callbackfn: (value: T, index: number, array: $ReadOnlyArray<T>) => U, thisArg?: any): Array<U>;
```
kangax: why does this return `$ReadOnlyArray<U>` while map returns just `U`?

goodmind: https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

samwgoldman:
Why does the mapper function return `$ReadOnlyArray<U> | U`, instead of just `$ReadOnlyArray<U>`?

goodmind:
@samwgoldman flatMap can work like `map`
```js
[ [1, 2, 3] ].flatMap(() => 1) // [1]
```

samwgoldman:
Oof. Annoying, but certainly in line with the gratuitously overloaded nature of existing APIs.

goodmind:
Well, it's easier to use with conditionals
```
[ [1, 2, 3] ].flatMap((a) => a[0] === 2 ? a : 1)
```
this would be much bigger if `flatMap` supported only arrays

samwgoldman:
It seems easy enough to just write this:

```
[ [1, 2, 3] ].flatMap((a) => a[0] === 2 ? a : [1])
```

goodmind:
oops

PR #7860, opened 2019-06-28
Add $TupleReduce destructor

goodmind:
```js
// @flow

declare var union: $TupleReduce<
  [1, 2, 3],
  <Acc, Value, Key>(Acc, Value, Key) => Acc | Value
>;

declare var accumulate: $TupleReduce<
  [1, 2, 3],
  <Acc, Value, Key>(Acc, Value, Key) => [Acc, Value],
  []
>;

union; // 1 | 2 | 3
accumulate; // [[[[], 1], 2], 3]

```

### DIFF
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -132,7 +132,8 @@ type 'loc virtual_reason_desc =
   | RNoSuper
   | RDummyPrototype
   | RDummyThis
-  | RReduce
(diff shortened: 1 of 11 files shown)


SamChou19815:
We are unlikely to pursue this, since we don't want more special builtin types. For highly dynamic functions, a lot of overloads might be better.

PR #7864, opened 2019-06-28
Add $ObjReduce destructor

goodmind:
```js
$ObjReduce<
  {| a: 1, b: 2 |}, 
  <Acc, Key, Value>(Acc, Key, Value) => Acc | Key, 
  empty
> // 'a' | 'b'

$ObjReduce<
  {| a: 1, b: 2 |}, 
  <Acc, Key, Value>(Acc, Key, Value) => {| 
     ...Acc, 
     [K in Key]: Value // hypothetical computed key
  |},
  empty
> // {| a: 1, b: 2 |}
```

### DIFF
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -133,6 +133,7 @@ type 'loc virtual_reason_desc =
   | RNoSuper
   | RDummyPrototype
   | RDummyThis
+  | RObjectReduce
   | RTupleMap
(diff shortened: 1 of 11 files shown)


gkz:
We're moving away from adding more such utility types. I think some of these use cases can be accomplished with our upcoming version of mapped types.

PR #7867, opened 2019-06-30
Add crash handling to Try Flow

goodmind:
image
This looks very error-prone itself... maybe something can be done from js_of_ocaml side to convert errors to more convenient structure


SamChou19815:
This is already handled in the new website.

PR #7869, opened 2019-07-01
Config docs: add info about ignore exclusion logic

mrtnzlml:
See: https://github.com/facebook/flow/commit/6acd727e4493af59d12e79b714de83b023321638

### DIFF
--- a/website/en/docs/config/ignore.md
+++ b/website/en/docs/config/ignore.md
@@ -13,7 +13,7 @@ default, nothing is ignored.
 
 Things to keep in mind:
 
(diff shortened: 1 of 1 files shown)


mrtnzlml:
Hi @mroch! I see this PR was imported but never landed in master. Is there anything I can improve?

kevinbarabash:
Having this in the docs would be helpful.  I didn't realize that negation patterns were allowed since the current docs only state that OCaml regexes are allowed and those don't support negation.

mrtnzlml:
Already in docs: https://flow.org/en/docs/config/ignore/#toc-ignore-exclusions

PR #7871, opened 2019-07-02
Add coverage to Try Flow

goodmind:


src/flow_dot_js.ml
```diff
@@ -338,6 +395,21 @@ let dump_types js_file js_content =
 
       js_of_json types_json
 
+let coverage js_file js_content =
+    let filename = File_key.SourceFile (Js.to_string js_file) in
```
mroch: nit: 2 spaces

src/flow_dot_js.ml
```diff
@@ -338,6 +395,21 @@ let dump_types js_file js_content =
...
+    let root = Path.dummy_path in
+    let content = Js.to_string js_content in
+    match parse_content filename content with
+    | Error _ -> failwith "parse error"
```
mroch: this seems too easy to hit. I'd say make it return a JSON error instead

website/_assets/js/tryFlow.js.es6.liquid
```diff
@@ -106,12 +106,25 @@ function removeChildren(node) {
...
 
 function getAnnotations(text, callback, options, editor) {
-  Promise.resolve(editor.getOption('flow'))
-  .then(flowProxy => flowProxy.checkContent('-', text))
-  .then(errors => {
-    CodeMirror.signal(editor, 'flowErrors', errors);
+  const flow = Promise.resolve(editor.getOption('flow'));
+  const sources = Promise.all([
+    flow
+      .then(flowProxy => flowProxy.coverage('-', text))
```
mroch: this code has to run on older versions of flow. I guess the `catch` will handle that, but swallowing errors always comes back to bite me. if it returns a JSON error for parse errors instead of throwing, then you could return a similar error if `flow.coverage == null`.

goodmind:
Where would settings be?

src/flow_dot_js.ml
goodmind: maybe `raise`?

website/_assets/js/tryFlow.js.es6.liquid
goodmind: How is this different from catch?

goodmind: not catching inside Promise.all, would kill whole sequence, checkContent wouldn't be called then

goodmind: > swallowing errors always comes back to bite me.

type-at-pos bites back, if you add `console.error` to type-at-pos catch it would complain about `re_string_match` not implemented 😬

goodmind: https://github.com/facebook/flow/issues/7939

goodmind:
@mroch I added coverage checkbox

Issue #5681, opened 2018-01-18
Can't bind the apply function


marcusdarmstrong:
Hi there! I've been trying to get rid of a $FlowFixMe in our codebase and have isolated it down to this:

https://flow.org/try/#0PTAEAEDMBsHsHcBQiCmAPADrATgF1ACYqQCGArtPgBSQBcoAYmQHYDGuAlrMwJSgC8APlCQAdCQwZoAT1EAjDswI0eAbkRA

That is,

```javascript
// @flow

export default (f: Function) => f.apply.bind(f);
```

yields

```
3: export default (f: Function) => f.apply.bind(f);
                                   ^ call of method `bind`. Callable signature not found in
3: export default (f: Function) => f.apply.bind(f);
                                   ^ function type
```

Even though...

```javascript
const foo = () => console.log('Works');
const bar = f => f.apply.bind(f);
bar(foo)(); // Works
```

oriSomething:
From curiosity, why binding `apply`?

marcusdarmstrong:
Not my code so I can't be 100% sure, but I think it might have just been so that we could use an args array.

SamChou19815:
This now works

PR #7879, opened 2019-07-02
Fix fn.apply.bind

goodmind:

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -5586,8 +5586,68 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
     (*******************************************)
 
     (* resolves the arguments... *)
-    | FunProtoApplyT lreason,
+    | FunProtoApplyT (lreason, None),
         CallT (use_op, reason_op, ({call_this_t = func; call_args_tlist; _} as funtype)) ->
+      (* Drop the specific AST derived argument reasons. Our new arguments come
+       * from arbitrary positions in the array. *)
+      let use_op = match use_op with
+      | Op FunCall {op; fn; args = _; local} -> Op (FunCall {op; fn; args = []; local})
+      | Op FunCallMethod {op; fn; prop; args = _; local} ->
+          Op (FunCallMethod {op; fn; prop; args = []; local})
+      | _ -> use_op
+      in
+
+      begin match call_args_tlist with
+      (* func.apply() *)
+      | [] ->
+         let funtype = { funtype with
+           call_this_t = VoidT.why reason_op |> with_trust bogus_trust;
+           call_args_tlist = [];
+         } in
+         rec_flow cx trace (func, CallT (use_op, reason_op, funtype))
+
+       (* func.apply(this_arg) *)
+      | (Arg this_arg)::[] ->
+         let funtype = { funtype with call_this_t = this_arg; call_args_tlist = [] } in
+         rec_flow cx trace (func, CallT (use_op, reason_op, funtype))
+
+       (* func.apply(this_arg, ts) *)
+      | first_arg::(Arg ts)::[] ->
+         let call_this_t = extract_non_spread cx ~trace first_arg in
+         let call_args_tlist = [ SpreadArg ts ] in
+         let funtype = { funtype with call_this_t; call_args_tlist; } in
+         (* Ignoring `this_arg`, we're basically doing func(...ts). Normally
+          * spread arguments are resolved for the multiflow application, however
+          * there are a bunch of special-cased functions like bind(), call(),
+          * apply, etc which look at the arguments a little earlier. If we delay
+          * resolving the spread argument, then we sabotage them. So we resolve
+          * it early *)
+         let t = Tvar.mk_where cx reason_op (fun t ->
+           let resolve_to = ResolveSpreadsToCallT (funtype, t) in
+           resolve_call_list cx ~trace ~use_op reason_op call_args_tlist resolve_to
+         ) in
+         rec_flow_t cx trace (func, t)
+
+      | (SpreadArg t1)::(SpreadArg t2)::[] ->
+           add_output cx ~trace
+             (Error_message.(EUnsupportedSyntax (loc_of_t t1, SpreadArgument)));
+           add_output cx ~trace
+             (Error_message.(EUnsupportedSyntax (loc_of_t t2, SpreadArgument)))
+      | (SpreadArg t)::[]
+      | (Arg _)::(SpreadArg t)::[] ->
+          add_output cx ~trace
+             (Error_message.(EUnsupportedSyntax (loc_of_t t, SpreadArgument)))
+      | _::_::_::_ ->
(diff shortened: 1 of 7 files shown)


src/typing/type_annotation.ml
```diff
@@ -852,9 +852,12 @@ let rec convert cx tparams_map = Ast.Type.(function
     )
 
   | "Function$Prototype$Apply" ->
-    check_type_arg_arity cx loc t_ast targs 0 (fun () ->
-      let reason = mk_reason RFunctionType loc in
-      reconstruct_ast (FunProtoApplyT reason) None
+    let reason = mk_reason RFunctionType loc in
+    (match convert_type_params () with
```
mroch: i think this is the only place we'd allow a sometimes-parameterized generic. the most similar concept is having a default, but in that case you'd have to use an empty `<>` (https://flow.org/en/docs/types/generics/#toc-adding-defaults-to-parameterized-generics).

this feels like a new behavior that we should consider carefully. also, we should consider how it interacts with annotating `this`, because it might be closer to `Function$Prototype$Apply<this=T>` or whatever syntax we end up with for that -- we've gone back and forth about annotating `this` for functions as either `function f(this: T, ...)` or `function f<this = T>(...)` (cc @gabelevi )

goodmind: I like first one because it mirrors TypeScript

goodmind: How can I make this support `<>`?

samwgoldman:
A silly question: What significance does binding the apply function itself have? Can you share an example where the observable effect of checking is different?

If it is significant, I think the solution of adding a this type to the internal representation of ApplyT is reasonable. However, I don't think we should add a type parameter to the annotation syntax, or at least not until the overall `this`-type story is sorted.

goodmind:
@samwgoldman it would be possible to apply `$Compose` or other custom functions on type-level

samwgoldman:
Can you please share an example? :)

mroch:
```
function spread(fn) {
  return Function.apply.bind( fn, null );
}

spread(fn, [1, 2, 3]) // fn(1, 2, 3)
```

goodmind:
https://github.com/lttb/flown/blob/master/src/Reduce/index.js

goodmind:
```js
declare function reduce<Fn, A, T>(Fn, A, T):
  $Call<Function$Prototype$Apply<$ComposeReverse>,
    $TupleMap<T, Fn>
  >

export type _Reduce = typeof reduce

export type Reduce<Fn, A, T> = $Call<_Reduce, Fn, A, T>

type MergeAll<T> = Reduce<<X, V>(V) => X => { ...$Exact<X>, ...$Exact<V> }, {}, T>

declare function mergeAll<A>(A): MergeAll<A>

const x = mergeAll([{ a: 1 }, { b: 1 }, { c: 1 }, { a: 2 }])

;(x: { a: 2, b: 1, c: 1 })

/**
 * $ExpectError
 *
 * the type of 'c' is incompatible with number
 */
;(x: { a: 2, b: 1, c: '1' })
```

samwgoldman:
@goodmind I don't see how these examples are related to the linked issue, or Function.prototype.apply/bind. I would much prefer to focus on the problem at hand, since there's enough work involved in supporting the use case from Marshall's example.

1. I still don't think we should add a type argument to Function$Prototype$Apply. Even if it makes some things expressible, it's putting the cart before the horse.
2. Per Marshall's example, adding the this type to the internal FunApplyT repr isn't enough, we also need to add the partial argument list.
3. The implementation here does not look correct (will comment inline) so some tests are definitely needed.

src/typing/flow_js.ml
```diff
@@ -5710,6 +5714,9 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       ) call_args_tlist;
       rec_flow_t cx trace (l, call_tout)
 
+    | FunProtoApplyT (lreason, _), BindT (_, _, { call_this_t; call_tout; _ }, _) ->
+      rec_flow_t cx trace (FunProtoApplyT (lreason, Some call_this_t), call_tout)
```
samwgoldman: This doesn't look right. Look at the `FunT _, BindT _` rule for guidance.

1. The `call_this_t` field of the `BindT` use is the this type corresponding to the call of bind. E.g., `x` in `x.bind(...)`. You want the first argument instead.
2. Looking at `FunT _, BindT _` as a guide, you'll notice some complexity missing from here. Specifically, the constraints involving `ResolveSpreadsToMultiflowPartial`. This is needed to figure out the full argument list in the presence of spread arguments.

goodmind: What first argument? This is forwarding From `x.bind(...)` to `$Function$Prototype$Apply<typeof x>`

samwgoldman: If you write a test, I think you will see what I mean. You want the first argument passed to the bind call, which is not `call_this_t`.

samwgoldman: Ah, sorry, my mistake -- I was confusing BindT with FunProtoBindT. You're right that `call_this_t` is the first parameter.

You do still need to do something with `call_args_tlist`, though.

goodmind: You mean something like this?
```js
const appliedFn = Function.apply.bind(x); // ok
const reappliedFn = appliedFn.bind(y); // wrong
reappliedFn(null, [""]); // no error
```

goodmind: @samwgoldman I think you was right 😀

goodmind: @samwgoldman would this mean adding second argument with applied parameters? I think one argument is already too much

src/typing/type_annotation.ml
goodmind: @mroch you can mimic second one with first one
`type F<T> = (this: T) => void`

goodmind:
@samwgoldman now it should work, I didn't exposed partial apply to arguments itself, only to `use_t`

goodmind:
@samwgoldman how it would work then for this case
```js
type Apply<Fn, Args> = $Call<Function$Prototype$Apply<Fn>, null, Args>
```

samwgoldman:
We don't need it to fix #5681. If you want to support another case, I think that you should write up a more thorough proposal. For example, why should we support that?

goodmind:
@samwgoldman maybe I can remove argument if you don't want it and decide later how to make my case work? (or write proposal)

samwgoldman:
Yeah. I would want to have a full understanding of your use case before adding something. It's possible that another solution would be better. I'm not opposed to adding something later, but we need to follow a process to make sure that we're adding the right things for the right reasons.

goodmind:
@samwgoldman removed it

goodmind:
Probably related issue, methods doesn't have function prototype
https://flow.org/try/#0MYGwhgzhAECC0G8BQ1oQC5nQS2NAZgBRgBc0AdgK4C2ARgKYBOAlIitAL5JdKwB0+PmAAOwkAE9CVECAA00ANoByJQF1mQA

Issue #5443, opened 2017-12-01
Support Pipeline Operator


k15a:
The pipeline `|>` operator is a proposal which is currently in stage 1. It would be great if flow could support this proposal.

I could imagine that this is not too hard because it's basically only syntactic sugar for function calls IIRC.

https://github.com/tc39/proposal-pipeline-operator

Hypnosphi:
I would like to help implementing that. Where should I start?

gajus:
It has landed in Firefox under `--enable-pipeline-operator` compile flag.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Pipeline_operator

mAAdhaTTah:
I've commented this elsewhere, but I strongly suggest avoiding implementing the pipeline operator until we've at least selected which of the 3 proposals we're going with. It's been in Firefox for some time behind a flag, but my current understanding is the FF implement doesn't currently conform with any of the 3 proposals' current semantics.

sartaj:
@mAAdhaTTah It seems [Proposal 0](https://github.com/tc39/proposal-pipeline-operator/wiki#proposal0-original-minimal-proposal) fits specs for all other proposals. Couldn't flow at least implement this minimal behavior to start? This would benefit anyone who would love to use piped unary functions with typed data structures.

Hypnosphi:
Flow is still 0.x so I personally wouldn't mind if this would be supported in some form now and then changed later

johnsonw:
I would really like to see this move forward. After spending quite some time in F# I find myself using the pipeline operator routinely. https://babeljs.io/docs/en/babel-plugin-proposal-pipeline-operator shows that we can use the "minimal" proposal at the moment. Is there any way to suppress the following error:

```
 21:   const c = xs |> fn1 |> fn2 |> fn3;
                            ^ Unexpected token >
```
// $FlowFixMe doesn't seem to ignore the error.

pnevyk:
> // $FlowFixMe doesn't seem to ignore the error.

That's because $FlowFixMe is for suppressing type errors. However, this is "syntax error" for flow parser so it does not get into the type checking phase at all.

bsmith-cycorp:
Please please please fix this. The basic spec (1-arity functions) seems very unlikely to change and relatively easy to parse, and it's pretty much all that people are asking for.

mAAdhaTTah:
> The "minimal" spec (1-arity functions) seems very unlikely to change

[...]

PR #7898, opened 2019-07-06
F# pipeline operator

goodmind:
Discussion: #5443 

- [x] `await` parsing
- [x] arrow body parsing
- [x] `.flowconfig` option


goodmind:
Well, it's actually looks like I implemented `minimal`, not `F#`

goodmind:
image

goodmind:
I ended up implementing typechecking as well 😀

goodmind:
Babel PR https://github.com/babel/babel/pull/9450 for reference

goodmind:
Well, there's already was function body parsing in `assignment_cover`, so I just copy-pasted it

goodmind:
I don't think it's possible to implement all three in Flow, because there isn't such thing as parser plugins. Flow just parses everything and then errors on wrong things with `PseudoParseError`

goodmind:
@mAAdhaTTah What you think? Is this still too early?

mAAdhaTTah:
Probably. I don't know the Flow team's policy around this sort of thing. I know TS won't implement anything until stage 3, especially after the pain they've had with decorators. 

I'll point out that the committee isn't necessarily sold on the operator at all and there is a significant contingent who prefer the bind operator. Not only are we risking implementing the wrong version of the operator, we're risking the syntax not advancing and needing to rip it out.

goodmind:
Well, Flow supports nullish coalescing, and optional chaining, but they're still stage 2

goodmind:
@mAAdhaTTah I thought bind operator was dropped in 2015 and always was stage 0?

mAAdhaTTah:
AFAIK, Bind is still Stage 1, but it's stalled out. However, we have gotten suggestions, including from committee members, about bind: https://github.com/tc39/proposal-pipeline-operator/issues/110 and https://github.com/tc39/proposal-pipeline-operator/issues/107 so I'm still viewing it as a competing proposal.

SamChou19815:
Closing since the PR has too many merge conflicts, and we do not plan to implement proposals before stage 3

PR #7902, opened 2019-07-07
Fix Array#reduce and Array#reduceRight

### LINKED ISSUE #5182: type comparing is wrong of usage of Array.reduce with 2nd argument
Using Array.reduce with 2nd argument and annotating return type for 1st callback argument, Flow shows error message.

[try flow](https://flow.org/try/#0PTAEAEDMBsHsHcBQAXAngBwKagEqeQK4BOAdgCobYC8oA3gL4DciiIJsyo8sRA1ogG0AugDoimACYEAxpgAUiUKEgES05AEtYJBUoCQ6AFy58xcpQA0ipaEzGAzsiIaSAc2sBKY3kKkKWOlAQUABDEnZkEORscV8SUDQsayVYszoma3oLdMQPZlZgZAALDXsuHl57QVFxKVldZVV1LR1rA29TP0tk2wcnF3clD0Dg+A1i2AJOMIiolp7U0nTmJSycvMQgA)

According to them, Flow detected its usage is `callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: Array<T>) => U`.
Its correct !!

But Flow said `This parameter is incompatible`.
This might be wrong.

In the detail of error message, Flow compared return type `ReturnType` to 2nd parameter `e: string`.
```
This parameter is incompatible:
9:   ): ReturnType { // annotate return type
        ^ object type. This type is incompatible with
8:     e: string          ^ string
```

But in this case, `ReturnType` should be compared to only 1st parameter, because of declaration is `callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: Array<T>) => U`

This is from 0.56.0 and 0.57.3 is the same. 

```js
// @flow
type ReturnType = {};

//not work
[].reduce(
  function(
  	p: ReturnType,
    e: string
  ): ReturnType { // annotate return type
    return {};
  }, {}
);

//this works
[].reduce(
  function(
  	p: ReturnType,
    e: string
  ) { // without annotation
    return {};
  }, {}
);
```

### LINKED ISSUE #6656: Unexpected and incorrect error message when passing `any` type as initial value to `Array.prototype.reduce`.
### Background
`Array.prototype.reduce` has [two signatures](https://github.com/facebook/flow/blob/27d4ac68457e097061f99fb22411244108dbae36/lib/core.js#L216-L223). When the second argument (`initialValue`) is omitted/undefined, the first element of the `this` array is used as the initial value, so the type of the accumulator argument (`previousValue`) to the first argument (`callbackfn`) is `T`, the type of array elements (also the type of `currentValue` and the `reduce` return type). When the second argument (`initialValue`) is defined and of type `U`, the accumulator argument (`previousValue`) to the first argument (`callbackfn`) is `U`, which is also the `reduce` return type in this case but different than the type of `this` array elements and `currentValue`.

### Bug
The types all appear to be correct as far as I can tell, but in one particular case I would not expect an error _and_ the error message is factually wrong. In [this code](https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMVrKALygDeAPpAQFygDOKATgJYB2A5uQL4DciAxrFYNQVWHVoBBRowCGaADz5YAPmKgA2qTG0A5LIBG-XVwC6fVJhwAhWY3UUDd2rgCurfimZDufQcJRQJ3sSUmDaAAoASmI1Ui5eZBBQAElWZi9ZaFAANyzXHFhIUHRsUAADW0Zy0ABaUFZCLBlYRkQxOgA6RiwAE1d+LAjEUFAI2X5+ABpRAhiiONAJ-k7g6J5QHpRXRlYlyY2uKZGx8NAqqMQoi2S0jOYs3PzC4tKcctlWNBr69ywAD2wnj6oEkADkACKgNiCGRYTygZqMVrtAhdHr9QbDUbjSYzMTzRbLVZ2dabLDbXb7fiHY44s6fNCXa5JMBgwhse6PPLQAozZgAWww0GY-Ay0DQs3EGgADKZQEUSlYKsofojAfCUCCkSiOt0+gMhidcdMpYSyNSSYwyVsdntlocrjwgA), you can see that when using an initial value with a `Bar` type, flow understands that the accumulator (first argument to `reduce`'s first argument) is of type `Bar`, and when omitting an initial value, it understands that the accumulator is of type `Foo`. But when passing in an initial value typed as `any`, not only does it error, which I do not expect (that's the whole point of any – you can call `.bar()` on it even if it doesn't know that it's there...), but more egregiously it outputs an incorrect and misleading error message:
```
Cannot call `acc.bar` because property `bar` is missing in `Foo` [1].
```

The property `bar` is not missing in `Foo`. `acc` is of type `any`, not `Foo`!

The only explanation I can think of is that because it's typed as `any`, it could be `undefined`, which would make the accumulator's type `Foo`. But even in this case flow is not painting anywhere near the full picture.

### Related issues(?)

These issues may be related or the same? I can't quite tell:
- https://github.com/facebook/flow/issues/5339
- https://github.com/facebook/flow/issues/5182
- https://github.com/facebook/flow/issues/4214

### LINKED ISSUE #7680: [Regression] Parameter type inference in `Array.reduce`
0.98.0 appears to have broken type inference around `Array.reduce`, specifically when the initial accumulator is a different type that the array values.

Flow version:
0.98.0

## Expected behavior
Infer the accumulator parameter type based on the initial accumulator value (when present).

## Actual behavior
Infers the accumulator parameter type to always be the array's type.

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVUCuA7AxgFwEs5swE4AnAa0OwHMBxAU2yYsNwCUmATTXJgB4AKgBowANQB8AClRgwcAEYArAFxgA3gEMKFbQE8NAQT2GRUgL6j5YbblyYAtphjb8lDRJsKn2gA7+bBoy3mDCAJRgALxSkjYRXlq2FEz4mBSkyioAdLr6BjmpfAJyCr4BQRQ+5faOLm4eFLYRANyoluhYeEQkYEoUcFQszKzsXLz8QmKSsrbZGjpmRmCmBRbWtnXOru6e8bZ+gcFgoeKRMXHeqImSyQqp6ZmKqnnLRZOltgrAwGAAogBHTCEABu2hgLHwYA8diUcFBTDATAAHgJ-ND8AALQgAZzAeLsYG6BGIpAQhGxdlItCgbGKYH8um0TjSbBhBiCxMoYAARNpeadsYTyNR8bQwAAGHIATgA7BFvqdtOJQVFYmAjlUZCqwGqagptg09s0FG0OqggA

### LINKED ISSUE #7901: Incorrect behaviour of Array#reduce with undefined as 2nd argument
Currently in Flow `arr.reduce(fn, undefined)` is the same as `arr.reduce(fn)`, which is wrong.

```js
var und1: void = [0].reduce(acc => acc, undefined); // unexpected error
var und2: void = [0].reduceRight(acc => acc, undefined); // unexpected error
```

[try flow](https://flow.org/try/#0FANwhgTgBArgdgEwIwC4ogPYEsFQLxQDaADALoB0EApgjAMZUAUYdd+AfFC3QDSyJUAZljg0AlAG4oAemn8qADwAOVOgBcaUKhAgYIoSPwQAmNJhz4iZSjXpUASlgDmACzXNWHLqz7wEQkXEpWXllVQ1cbV0IIA)

image

goodmind:
Fixes #5182
Fixes #6656
Fixes #7680
Fixes #7901

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -230,7 +230,6 @@ declare class $ReadOnlyArray<+T> {
 
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => T,
-      initialValue: void
     ): T;
     reduce<U>(
       callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => U,
@@ -238,7 +237,6 @@ declare class $ReadOnlyArray<+T> {
     ): U;
     reduceRight(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => T,
-      initialValue: void
     ): T;
     reduceRight<U>(
       callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: $ReadOnlyArray<T>) => U,
@@ -266,7 +264,6 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     push(...items: Array<T>): number;
     reduce(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: Array<T>) => T,
-      initialValue: void
     ): T;
     reduce<U>(
       callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: Array<T>) => U,
@@ -274,7 +271,6 @@ declare class Array<T> extends $ReadOnlyArray<T> {
     ): U;
     reduceRight(
       callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: Array<T>) => T,
-      initialValue: void
     ): T;
     reduceRight<U>(
       callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: Array<T>) => U,
--- a/tests/arraylib/issue_5182.js
+++ b/tests/arraylib/issue_5182.js
@@ -1,0 +1,26 @@
+// @flow
+
+type ReturnType = {|
+  a: string
+|};
+
+// works
+const a = [].reduce(
+  (p: ReturnType, e: string): ReturnType => {
+    // annotate return type
+    return { a: "" };
+  },
+  { a: "" }
+);
+
+// works
+const b = [].reduce(
+  (p: ReturnType, e: string) => {
+    // without annotation
+    return { a: "" };
+  },
+  { a: "" }
+);
+
+(a: ReturnType); // ok
+(b: ReturnType); // ok
--- a/tests/arraylib/issue_6656.js
+++ b/tests/arraylib/issue_6656.js
@@ -1,0 +1,4 @@
+// @flow
+
+const a = ["a", "b"].reduce(acc => acc * 2, 1.0); // works
+const b = ["a", "b"].reduce(acc => acc * 2, (1.0: any)); // works
--- a/tests/arraylib/issue_7680.js
+++ b/tests/arraylib/issue_7680.js
@@ -1,0 +1,21 @@
+// @flow
+
+function workingGenericReduce<T, V>(
+  obj: { array: Array<T> },
+  accumulator: V,
+  mapper: (V, T) => V
+): V {
+  return obj.array.reduce(mapper, accumulator);
+}
+
+function brokenGenericReduce<T, V>(
+  obj: { array: Array<T> },
+  accumulator: V,
+  mapper: (V, T) => V
+): V {
+  return obj.array.reduce(
+    // works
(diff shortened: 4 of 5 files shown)


nmote:
I believe that the lines you've deleted were added before we were strict about function arity. We would allow you to call functions with too many arguments, which meant that Flow had too much flexibility when choosing which overload to use. These were added as a workaround, but as the issues demonstrate, they caused their own problems.

Now that we are strict about function arity, these lines should be safe to remove. Thanks!

PR #7903, opened 2019-07-07
Add missing esproposal.nullish_coalescing warning

goodmind:
Add missing warning

### DIFF
--- a/src/typing/debug_js.ml
+++ b/src/typing/debug_js.ml
@@ -2868,6 +2868,8 @@ let dump_error_message =
         spf "EInvalidPrototype (%s)" (dump_reason cx reason)
     | EExperimentalOptionalChaining loc ->
         spf "EExperimentalOptionalChaining (%s)" (string_of_aloc loc)
+    | EExperimentalNullishCoalescing loc ->
+        spf "EExperimentalNullishCoalescing (%s)" (string_of_aloc loc)
     | EOptionalChainingMethods loc ->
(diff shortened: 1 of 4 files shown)


mroch:
was this not being caught by the parser here? 
https://github.com/facebook/flow/blob/c48b92cbaa7225d5843f162b763e0517b415552f/src/parser/expression_parser.ml#L315-L316

i wonder why optional chaining is erroring in statement.ml instead of the parser.

goodmind:
@mroch there's comments about it, it is for nicer error messages, also parser options is always enabled

mroch:
you mean nicer because one has `code` formatting and the other doesn't?

it parses the same way with the option enabled or not, so if we made the errors prettier and disabled the parser option along with the typechecker option, then would it behave the way you want?

I don't love the duplication this adds, but unifying them could be future work.

goodmind:
@mroch Yeah, it is probably because of code formatting

goodmind:
@mroch could be it is also related to how location of error is determined if it is halted in the middle of parsing?

goodmind:
@mroch so would you merge or maybe I should add some comments that this is temporary before parser errors are unified with type errors?

goodmind:
Looks like this proposal going into stage 3
https://onedrive.live.com/view.aspx?resid=5D3264BDC1CB4F5B!5285&ithint=file%2Cpptx&authkey=!AC0xl44ouDVFxEM

goodmind:
I'm going to close this, since it is stage 3

Issue #396, opened 2015-04-17
way to extend/override flow/lib types


blax:
Built-in node.js type declarations are still far from being complete, and filling the missing parts is very cumbersome as Flow doesn't allow to overwrite single modules.

For instance, if I want to fill in some gaps in HTTP types, I have to use flow with `--no-flowlib` flag, which forces me to copy all the remaining built-in interfaces to my project. 

Although such workaround works for now, it's unmaintainable in the longer term, and not obvious to newcomers.

Would it be a problem to make custom declarations override the default ones?

popham:
Why not edit the lib files in place and rebuild Flow?  You could pull request your modifications eventually.

gyzerok:
@popham why not to check types using your eyes?

+1 for topic

nmn:
I propose the following syntax for overriding flow types:

Firstly, we need a way to extend types that are not classes. The syntax is obviously Object spread:
```
type A = {a: boolean, b: number, c: string}
type B = {...A, d: (...rest: any)=> any}
```

Next, it should be possible to reassign a declaration:
```
declare class Document extends Document {
  newProp: boolean;
}
```

popham:
What's wrong with `class MyDocument extends Document { newProp: boolean }`?  If your environment has a different `Document` than Flow's, then why did you use Flow's to begin with?

In these situations, my gut says that you should put the Flow libraries in your project somewhere under version control.  When you need to override something, edit it.  Any external libraries that you've incorporated into your project will be validated against your environment specification (the version controlled library files).  (Corollary: If you're making a library, work against Flow's unaltered libraries.)  I guess this could be facilitated by moving Flow's libraries to a separate repo so that users could shallow clone them into projects?

The counterargument goes:  "I want my project to see updates to Flow's libraries."  I say, "You want your environment specification to change at the whim of Flow's maintainers?  You shouldn't."  Sure Flow's libraries are still immature, but I think that the proper tact is to help improve them instead of implementing override support (this is low-hanging fruit for non-maintainers).  Eventually they'll stabilize and track some ideal environment as it evolves with JS, popularity of Node, popularity of IE8, etc.

(I love the spread syntax, though.  Although I would describe it as a non-disjoint union, not as an override.)

samwgoldman:
@nmn For your A/B example, you should be able to use intersection types: `type B = A & { d: Function }` (btw, `Function` is identical to your `(...rest: any) => any` type).

popham:
[...]

PR #7919, opened 2019-07-11
[WIP] Builtin declaration merging

goodmind:
_Somewhat_ (~~sometimes~~) working fix for https://github.com/facebook/flow/issues/396

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -908,3 +908,5 @@ declare var console: {
     warn(...data: Array<any>): void,
     ...
 };
+
+declare var globalThis: $GlobalThis;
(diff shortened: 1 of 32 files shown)


tonyn4444:
Nice!

goodmind:
Not really, I'm not sure this is right way to do this, but it works at least

kevinbarabash:
@goodmind any plans to eventually add declaration merging for `declare module` eventually?

goodmind:
@kevinbarabash I'm not even sure if I'm implementing it right way, still wait on review from Flow team

goodmind:
@kevinbarabash just implemented `declare module` it was pretty easy actually, but it supports only ES `declare export` from modules, not private interfaces

kevinbarabash:
I noticed you added `GlobalThisT` in this PR.  It seems independent of the declaration merging.  `globalThis` seems really hard to type properly given that it can have different properties depending on execution context.

goodmind:
Ah, because I was working on top of my `globalThis` PR, I noticed that I can do `globalThis.HTMLElement = 1;` and it would make union of `class HTMLElement | number`

SamChou19815:
closing since we are now implementing it ourselves

PR #7923, opened 2019-07-13
Add --evaluate-type-destructors and --expand-type-aliases option to dump-types and type-at-pos

goodmind:


goodmind:
/сс @panagosg7 looks like this got tons of conflicts :\

PR #7049, opened 2018-10-19
Update babel instructions to babel 7

KidkArolis:

PR #6396, opened 2018-05-31
Changed instructions to use @babel/preset-flow

soska:
Past instructions caused babel to throw an error:

```
Error: Plugin/Preset files are not allowed to export objects, only functions.
```

I don't know why this happens, I only know that this fixes it. Took the tip from this other issue: https://github.com/babel/babel-loader/issues/540.

I hope this is useful.

PR #7933, opened 2019-07-16
Fix typo in installation instructions

mgmarlow:
Installation instructions for babel + NPM instructed installing the incorrect babel plugin.

### DIFF
--- a/website/en/docs/_install/compiler-babel.md
+++ b/website/en/docs/_install/compiler-babel.md
@@ -2,15 +2,15 @@
 support for Flow. Babel will take your Flow code and strip out any type
 annotations.
 
-First install `babel-cli` and `@babel/preset-flow` with either
-[Yarn](https://yarnpkg.com/) or [npm](https://www.npmjs.com/).
+First install `@babel/core`, `@babel/cli`, and `@babel/preset-flow` with
+either [Yarn](https://yarnpkg.com/) or [npm](https://www.npmjs.com/).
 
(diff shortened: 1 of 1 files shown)


goodmind:
@mgmarlow I thought there already was ton of PRs fixing this, but each next PR finds old babel plugin... Can you sign CLA?

mroch:
thanks!

@goodmind: there are at least 3 other PRs doing various conflicting things, in various needs of rebasing.

#6396 makes a similar change to a different part of the docs, which seems to have already been fixed
#6952 is similar to this PR, but it also updates `babel-cli` to `@babel/cli` and replaces some other references, which I think mostly have already been fixed by other PRs.
#7049 also switches to `@babel/cli` but additionally installs `@babel/core`. I have no idea if this is necessary.

since this one is current, it seems like the only one that can be merged without conflicts. I think we should probably update to `@babel/cli` and maybe add `@babel/core`. since we don't use this workflow internally, maybe you guys can help test that?

mgmarlow:
@mroch sure I'll take a look. I set my project up using rollup so I didn't need the CLI but I can run through a similar flow with those tools.

mgmarlow:
@mroch You're 100% correct, you also need `@babel/core` when running `@babel/cli`. I've updated the docs accordingly.

mroch:
awesome, thank you! (and also @KidkArolis and @jeffvandyke for #7049)

PR #7945, opened 2019-07-19
[DOCS] Documentation for $Exports

SebastienGllmt:
I added documentation for the previously undocumented utility function (see #2464)

To be honest, maybe we should consider deprecating `$Exports` because the `import typeof` feature is now supported and mentioned here: https://flow.org/en/docs/types/modules/

### DIFF
--- a/website/en/docs/types/utilities.md
+++ b/website/en/docs/types/utilities.md
@@ -21,6 +21,7 @@ Table of contents:
 - [`$Call<F, T...>`](#toc-call)
 - [`Class<T>`](#toc-class)
 - [`$Shape<T>`](#toc-shape)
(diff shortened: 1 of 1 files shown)


goodmind:
@SebastienGllmt `import typeof` wouldn't really work for dynamic functions, or custom requires

goodmind:
Ah, no, it's impossible to use it like this
```
declare function myImport<T: string>(a: T): $Exports<T>; // error
```

SebastienGllmt:
Yeah I gave it some thought and couldn't really come up with a scenario where it's different than `import type` beyond the one I mentioned, but if you can think of anything else feel free to add it to this PR!

gabelevi:
Thanks for the docs! I'll get this merged!

PR #7948, opened 2019-07-20
Fix $ElementType union property key

### LINKED ISSUE #4211: $ElementType works incorrectly with unions
[Example](https://flow.org/try/#0CYUwxgNghgTiAEA3W8AuAueBtAdgVwFsAjEGAGngGdUYBLHAcwF0BuAKFQE8AHBAFXgBeeABIAohBAEQOVHx4gAPF14B7AGZoKABngAfeAEYAfOzahIsBMhhJMfMwApEmfMVL6qNegwCULeAB6QPgAeQBpCjBVGDgwVDY2RwBydVVVZPt-IJCxWJiKemjY8ATzcGg4JBQAD0xqOkZ4ADJ4NxIYJzr4Pmzg+AB3GIBrSkSgA):

```js
declare var t: [number, string];
type T = $ElementType<typeof t, 0 | 1>;
declare var v: T;
(v: number | string); // OK, correct

('foo': T); // Error, incorrect

declare var x: string & number;

(x: T); // works
```

`T` is expected to be equivalent to `string | number`. Instead it works as `string & number` in upper bound position.

goodmind:
Similar issue was fixed in de994b684b3b0561df4e596e0a46b63a93d70a33.

When union types are flowing into `ElemT` it splits into two different constraints. Consider the following example:

```js
type T = $ElementType<[number, string], 0 | 1>;
("string": T);
```

When evaluating this declaration, we emit the following constraints:

1. EvalT (ElementType, ...) -> ...
2. 0 | 1 -> ElemT ([number, string], ReadElem T)

Constraint (2) is processed by splitting, which turns into two
separate constraints `0 -> ElemT (...)` and `1 -> ElemT (...)` then we get two constraints for our result tvar after:

1. number -> T
2. string -> T

Since `AnnotT` can't hold tvars with multiple types (see comment below) we see a spurious error that `string` is
not compatible with `number`.

> Note on usage: AnnotT can be used as a general wrapper for tvars as long
        as the wrapped tvars are 0->1. If instead the possible types of a
        wrapped tvar are T1 and T2, then the current rules would flow T1 | T2 to
        upper bounds, and would flow lower bounds to T1 & T2.

This diff deals with the above by handling union types specially in the ElemT use type.

Fixes #4211

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -2971,11 +2971,21 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
            degenerate to a singleton) *)
         rec_flow_t cx trace (l, result)
 
+    | UnionT (_, rep),
+      ElemT (use_op, reason, tin, ReadElem tout) ->
+      let t0, (t1, ts) = UnionRep.members_nel rep in
+      let f t =
+        AnnotT (reason, Tvar.mk_where cx reason (fun tvar ->
+          rec_flow cx trace (t, ElemT (use_op, reason, tin, ReadElem tvar))
+        ), false)
+      in
+      let rep = UnionRep.make (f t0) (f t1) (Core_list.map ts ~f) in
+      rec_unify cx trace ~use_op (UnionT (reason, rep)) tout
+
     | UnionT (_, rep), _
       when (match u with
         (* For l.key !== sentinel when sentinel has a union type, don't split the union. This
            prevents a drastic blowup of cases which can cause perf problems. *)
-        | ElemT (_, _, _, ReadElem _)
         | PredicateT (RightP (SentinelProp _, _), _)
         | PredicateT (NotP (RightP (SentinelProp _, _)), _) -> false
         | _ -> true
@@ -5351,18 +5361,6 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       CallElemT (reason_call, reason_lookup, key, ft) ->
       let action = CallElem (reason_call, ft) in
       rec_flow cx trace (key, ElemT (unknown_use, reason_lookup, l, action))
-
-
-    | UnionT (_, rep),
-      ElemT (use_op, reason, tin, ReadElem tout) ->
-      let t0, (t1, ts) = UnionRep.members_nel rep in
-      let f t =
-        AnnotT (reason, Tvar.mk_where cx reason (fun tvar ->
-          rec_flow cx trace (t, ElemT (use_op, reason, tin, ReadElem tvar))
-        ), false)
-      in
-      let rep = UnionRep.make (f t0) (f t1) (Core_list.map ts ~f) in
-      rec_unify cx trace ~use_op (UnionT (reason, rep)) tout
 
     | _, ElemT (use_op, reason_op, (DefT (_, _, ObjT _) as obj), action) ->
       let propref = match l with
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -5343,6 +5343,17 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       rec_flow cx trace (key, ElemT (use_op, reason_op, l, WriteElem (tin, tout)))
 
     | (DefT (_, _, (ObjT _ | ArrT _)) | AnyT _),
+      GetElemT (use_op, reason_op, UnionT (_, rep), tout) ->
+      let t0, (t1, ts) = UnionRep.members_nel rep in
+      let f t =
+        AnnotT (reason_op, Tvar.mk_where cx reason_op (fun tvar ->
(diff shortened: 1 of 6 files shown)


gabelevi:
@jbrown215 will take a look when he has a minute!

src/typing/flow_js.ml
```diff
@@ -5352,6 +5353,17 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       rec_flow cx trace (key, ElemT (unknown_use, reason_lookup, l, action))
 
 
+    | UnionT (_, rep),
+      ElemT (use_op, reason, tin, ReadElem tout) ->
+      let t0, (t1, ts) = UnionRep.members_nel rep in
+      let f t =
+        AnnotT (reason, Tvar.mk_where cx reason (fun tvar ->
```
jbrown215: @samwgoldman: Removing this AnnotT still passes the tests, [including the one that surfaced internally](https://flow.org/try/#0C4TwDgpgBAYglhANgEwM5QLxQN4B8BQUUAhgFxSrABOcAdgOYA0hUARubQK4C2rEVzXAF8A3PnwBjAPa1KUZMWBlYCFOizYWygOQAzKVO3Mi7KABYATM1HjdnWhOBwZUehGDwkyADwBpcgAkvhAgqN6eagB8kQAUANYh5L4AlIEAoogQ3BC0wAAq4BDhqmiMUL6ROCxU7pxUtPKKxADaCSAAumJC4tKywCQAasSInNBYbh4lMdrE2slivXKsQyNjru4RyNOscyJAA)

Going back to the conversation we had two days ago, I think that the UnionT's tvars should "in theory" all be 0->1 anyway, so distributing ElemT through the union should preserve the 0->1-ness. On the other hand, we know that doesn't hold in practice, so maybe goodmind should add a field like DestructAnnot vs. DestructInfer to ElemT and only use this special case on `Annot`.

goodmind: Should I wait for @samwgoldman answer or add this field now?

jbrown215: Wait a few days first, but let's say if Sam doesn't answer by Monday then yes! I just want to give him a fair amount of time to take a look here first.

goodmind:
Similar bug, my fix doesn't seem to work with it
https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMQJZbQAmAzqALygDeAPoqKAIYBcoAFAJSUB8opKAE74AdgHMANA1AAjNl16gRAVwC2MrIMS0AvgG5kAY1giBoYkxSs8hEuSrVp1hRT4AiSLFhupjOR25XUAAWACZEfWRiLENoJkEcADd40ABrLDQ2ABIAaQzSAB4CIjIeA2jY+KSUyGyAUWgsVSwRFAAVTCwi2zIJUFz87pLSHjKjEzMAC0pQSC4DY1MUczZ0bFhIUGmqOc4FieXSbIBhJmhoArWsDdm+Ha5EIA

jbrown215:
not clear if that's a bug to me. uncertain how the semantics of Call with no arguments should be

samwgoldman:
Please provide some more context about your change here. In particular, please provide some insight into your analysis of the underlying bug. What is the root cause? Separately, why does this change address that root cause?

goodmind:
Root cause is same as destructuring bug

samwgoldman:
I would be surprised if that's the case, but regardless note that de994b684b3b0561df4e596e0a46b63a93d70a33 has a quite substantial analysis which both explains the underlying mechanism that causes the buggy behavior as well as an explanation of the fix. This is the kind of analysis I am looking for here.

goodmind:
Yeah, it basically turns unions into two separate constraints like destructuring did, so it behaves like union in lower bound and intersection in upper bound, it's explained it issue itself and error looks similar how destructuring error looked

goodmind:
@samwgoldman I updated main post

goodmind:
If you think this is not the same bug, then how this should be fixed?

panagosg7:
Wasn't aware of this PR. https://github.com/facebook/flow/commit/f322d37b1a209587449f73642b13d866e8bd9eb8 added something similar.

> Similar bug, my fix doesn't seem to work with it EDIT: better example https://flow.org/try/#0PTAEAEDMBsHsHcBQiAuBPADgU1AMQJZbQAmAzqALygDeAPoqKAIYBcoAFAJSUB8opKAE74AdgHMANA1AAjNl16gRAVwC2MrIMS0AvgG5k6bKAAqlUABIAwk2jQAPBYCi0LKqwiUJzFnsEiZBKWANJYaKR+hCSkPLEGxFgAxtBMgjgAbqmg6WwmBuw5SmoagqC0-EKiYpz5AIxsKuqaZRXC4jWI7PWmHYhAA

This was fixed more recently with https://github.com/facebook/flow/commit/79a1f56596446269aa2392951fdfbecc77318517

PR #7951, opened 2019-07-21
[DOCS] Documentation for `this` constraint

SebastienGllmt:
This is based on my discussion with on Discord followed-up by https://github.com/facebook/flow/issues/452#issuecomment-513224097 

I thought this was important and common enough to be worth documenting. Note that it may be quickly deprecated depending on what happens with https://github.com/facebook/flow/pull/7807


gabelevi:
Thanks for taking the time to write this up! However, I talked with the team, and I'm not sure we'd like to put this workaround in our official documentation. Using an unstable part of the type system to hack around an unimplemented part just seems unwise. 

We do intend at some point to build the ability to annotate the `this` type. Until then, you're welcome to use your workaround but I'm afraid we'd rather avoid putting this in the docs.

PR #7952, opened 2019-07-22
Improve FileReader types

nwoltman:
Update the `FileReader` types to follow the [spec](https://w3c.github.io/FileAPI/#dfn-filereader) more closely.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -20,23 +20,23 @@ declare class Blob {
 }
 
 declare class FileReader extends EventTarget {
+  +DONE: 2;
+  +EMPTY: 0;
+  +LOADING: 1;
+  +error: null | DOMError;
+  +readyState: 0 | 1 | 2;
+  +result: null | string | ArrayBuffer;
   abort(): void;
-  DONE: number;
-  EMPTY: number;
-  error: DOMError;
-  LOADING: number;
-  onabort: (ev: any) => any;
-  onerror: (ev: any) => any;
-  onload: (ev: any) => any;
-  onloadend: (ev: any) => any;
-  onloadstart: (ev: any) => any;
-  onprogress: (ev: any) => any;
+  onabort: null | (ev: ProgressEvent) => any;
+  onerror: null | (ev: ProgressEvent) => any;
+  onload: null | (ev: ProgressEvent) => any;
+  onloadend: null | (ev: ProgressEvent) => any;
+  onloadstart: null | (ev: ProgressEvent) => any;
+  onprogress: null | (ev: ProgressEvent) => any;
   readAsArrayBuffer(blob: Blob): void;
   readAsBinaryString(blob: Blob): void;
   readAsDataURL(blob: Blob): void;
   readAsText(blob: Blob, encoding?: string): void;
-  readyState: 0 | 1 | 2;
-  result: string | ArrayBuffer;
 }
 
 declare type FilePropertyBag = {


gabelevi:
Looks good to me! Will get this reviewed and merged ASAP!

villesau:
@nwoltman looks like MDN has examples that does not work after these fixes:

https://developer.mozilla.org/en-US/docs/Web/API/FileReader/onload

Also following piece is not typechecking anymore:

```
    const reader = new FileReader();
    reader.onload = readerEvent => {
      readerEvent.target.result.trim();
    };
```

nwoltman:
@villesau I don't see the problem. What exactly is your issue?

goodmind:
@villesau you can just use `reader` or `ProgressEvent` should be updated to be generic on `target` property, I'm not sure if it is right, because `target` can change

goodmind:
@nwoltman `target` isn't generic

nwoltman:
Yes, unfortunately the [`target`](https://github.com/facebook/flow/blob/fe6c84a152e040700e9f6710c8ce2bd0642175fd/lib/dom.js#L277) on events is stuck as a plain `EventTarget` and can't be easily specified as something else. I do wish that Flow would use generics for the `Event` class (like how React does with the `SyntheticEvent` class), however, I have my doubts that a PR to add generics to `Event` would be accepted (as it might cause a lot of breakage).

villesau:
@nwoltman I'm sorry that I was unclear. Here are the errors and snippets:

```
// @flow

const reader = new FileReader();
reader.onload = readerEvent => {
  readerEvent.target.result.trim();
};

```

gives

`Error:(5, 22) Cannot get `readerEvent.target.result` because property `result` is missing in  `EventTarget` [1].`

So `result` is missing.

nwoltman:
@villesau Yes, that is unfortunate, but expected. `target` is just an `EventTarget`, not a `FileReader`. As I sort of explained in my previous comment, Flow would need to implement generics on the `Event` class to be able to make `readerEvent.currentTarget` a `FileReader` instance.

Just so you know, there wasn't an error before because the `onload` event was typed with `any` as the callback's first argument, so the previous typing was mostly useless. If you want to make the error go away, you can use `reader` instead of `readerEvent.target` (though you'll get a different type error on `result` if you do this). You can also skip type-checking altogether (like it was before this PR was merged) by forcing `readerEvent` to have the type `any`.

goodmind:
@nwoltman 
> as it might cause a lot of breakage

You would just need to specify default generic, and change `Event` to `Event<>` everywhere, it's annoying that `Event` doesn't already mean `Event<>`

nwoltman:
@goodmind Yes, that's exactly the kind of breakage I found while testing out `Event` with generics. I was hoping that `Event` would be the same as `Event<>`, but unfortunately it's not 😞. I wonder if that could be changed though 🤔.

wgao19:
Encountered the same thing as @villesau, so this cannot be properly handled?

goodmind:
I guess it's possible to add generic only to `ProgressEvent` without affecting `Event`

wojdyr:
> Yes, that is unfortunate, but expected. `target` is just an `EventTarget`, not a `FileReader`. As I sort of explained in my previous comment, Flow would need to implement generics on the `Event` class to be able to make `readerEvent.currentTarget` a `FileReader` instance.

Given that currently it's not possible to avoid false positives, I'm wondering if this change is worth it? In my little project it almost doubled the number of `// $FlowFixMe` lines.

nwoltman:
You can get the old functionality back by using `any` in your function signature:

```js
fileReader.onprogress = (ev: any) => {
   // Use ev here without any need for $FlowFixMe
};
```

PR #7959, opened 2019-07-22
[DOCS] Update docs for Symbols

SebastienGllmt:
Not sure what's the current state for Symbols in Flow so somebody who knows more can add to this PR. At any rate, the current documentation is outdated so I fixed it.


gkz:
The docs were updated when I added improved support for symbols: https://flow.org/en/docs/types/primitives/#toc-symbols

PR #7968, opened 2019-07-25
Document `typeof` for generics

SebastienGllmt:


SebastienGllmt:
This trick only works if the generic type you want to bind to is part of the argument list so it's not really an ideal trick.

SebastienGllmt:
I'm just going to close this since after more thought this trick has too many limitations for it to be useful for most people. Probably a language feature is needed to make this more practical.

PR #7969, opened 2019-07-26
Fix: IDBDatabase.objectStoreNames should be of type DOMStringList

bnelo12:
I was working with IDBDatabases and found that this object property has the wrong type. It should have the type DOMStringList instead of string[].

So I have added an interface for DOMStringList and the appropriate type definition for the property  IDBDatabase.objectStoreNames. 

The documentation for DOMStringList can be found here: https://developer.mozilla.org/en-US/docs/Web/API/DOMStringList

And the documentation for IDBDatabase.objectStoreNames can be found here: https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/objectStoreNames

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -96,6 +96,13 @@ declare class DataTransferItem {
 
 declare type DOMStringMap = { [key:string]: string, ... }
 
+declare class DOMStringList {
+  +[key:number]: string;
+  +length: number;
+  item: (number) => string | null;
+  contains: (string) => boolean;
+}
+
 declare class DOMError {
   name: string;
 }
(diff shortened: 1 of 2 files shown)


goodmind:
This probably should be class, not interface

lib/dom.js
```diff
@@ -96,6 +96,12 @@ declare class DataTransferItem {
 
 declare type DOMStringMap = { [key:string]: string, ... }
 
+declare interface DOMStringList {
```
goodmind: `declare class DOMStringList`

lib/dom.js
```diff
@@ -96,6 +96,12 @@ declare class DataTransferItem {
...
+     length: number;
```
goodmind: This is read-only field according to docs

bnelo12: I think I've corrected it by adding a + symbol.

bnelo12:
All fixed! Let me know if you see anything else.

goodmind:
There's also should be `+[key: number]: string`, like on read-only arrays

bnelo12:
Agreed! It's in.

PR #5048, opened 2017-10-06
Add support for sentinel and length refinement of tuples

popham:
Fixes #4050, fixes #4377, fixes #4556, and fixes #4873.

PR #7597, opened 2019-03-28
Fix bracket refinements

goodmind:
Fixes #7475 
Fixes #2990

PR #7974, opened 2019-07-27
Support length refinement of tuples

ilya-bobyr:
I have rebased changes from https://github.com/facebook/flow/pull/5048 on top of the current master branch. (It was ~4000 changes behind %)
Addressed the only comment that the review had.

I must admit that I do not understand the flow source code as well as the original author.
But I will do my best to make sure this gets in.

I can see the original pull request was marked "interesting" here: https://github.com/facebook/flow/projects/4#card-20788575

It also feels that the test coverage is incomplete.  There are no tests that cover error messages.
Only the happy path is tested.  I did not figure out yet how to add this kind of tests.  Any kind of guidance is very welcome.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -6334,6 +6334,14 @@ struct
         | ( DefT (reason, _, ArrT (ArrayAT (t, _))),
             (GetPropT _ | SetPropT _ | MethodT _ | LookupT _) ) ->
           rec_flow cx trace (get_builtin_typeapp cx ~trace reason "Array" [t], u)
+        (*************************)
+        (* Tuple "length" access *)
+        (*************************)
+        | (DefT (reason, trust, ArrT (TupleAT (_, ts))), GetPropT (_, _, Named (_, "length"), tout))
+          ->
+          (* Use definition as the reason for the length, as this is
+           * the actual location where the length is in fact set. *)
+          rec_flow_t cx trace (tuple_length reason trust ts, tout)
         | ( DefT (reason, _, ArrT ((TupleAT _ | ROArrayAT _) as arrtype)),
             (GetPropT _ | SetPropT _ | MethodT _ | LookupT _) ) ->
           let t = elemt_of_arrtype arrtype in
@@ -9389,6 +9397,15 @@ struct
     sentinel_prop_test_generic key cx trace result obj (sense, obj, t)
 
   and sentinel_prop_test_generic key cx trace result orig_obj =
+    let desc_of_sentinel sentinel =
+      match sentinel with
+      | UnionEnum.(One (Str s)) -> RStringLit s
+      | UnionEnum.(One (Num (_, n))) -> RNumberLit n
+      | UnionEnum.(One (Bool b)) -> RBooleanLit b
+      | UnionEnum.(One Null) -> RNull
+      | UnionEnum.(One Void) -> RVoid
+      | UnionEnum.(Many _enums) -> RUnionEnum
+    in
     (* Evaluate a refinement predicate of the form
 
       obj.key eq value
@@ -9424,18 +9441,10 @@ struct
       | Some p ->
         (match Property.read_t p with
         | Some t ->
-          let desc =
-            RMatchingProp
-              ( key,
-                match sentinel with
-                | UnionEnum.(One (Str s)) -> RStringLit s
-                | UnionEnum.(One (Num (_, n))) -> RNumberLit n
-                | UnionEnum.(One (Bool b)) -> RBooleanLit b
-                | UnionEnum.(One Null) -> RNull
-                | UnionEnum.(One Void) -> RVoid
-                | UnionEnum.(Many _enums) -> RUnionEnum )
+          let reason =
+            let desc = RMatchingProp (key, desc_of_sentinel sentinel) in
+            replace_desc_reason desc (reason_of_t result)
           in
-          let reason = replace_desc_reason desc (reason_of_t result) in
           let test = SentinelPropTestT (reason, orig_obj, key, sense, sentinel, result) in
           rec_flow cx trace (t, test)
         | None ->
@@ -9486,6 +9495,14 @@ struct
           | DefT (_, _, InstanceT (_, _, _, { own_props; _ })) ->
             (* TODO: add test for sentinel test on implements *)
             flow_sentinel sense own_props obj s
+          (* tuple.length ===/!== literal value *)
+          | DefT (reason, trust, ArrT (TupleAT (_, ts))) when key = "length" ->
+            let test =
+              let desc = RMatchingProp (key, desc_of_sentinel s) in
+              let r = replace_desc_reason desc (reason_of_t result) in
+              SentinelPropTestT (r, orig_obj, key, sense, s, result)
+            in
+            rec_flow cx trace (tuple_length reason trust ts, test)
           | IntersectionT (_, rep) ->
             (* For an intersection of object types, try the test for each object
            type in turn, while recording the original intersection so that we
--- a/src/typing/type.ml
+++ b/src/typing/type.ml
@@ -3698,6 +3698,19 @@ and extract_getter_type = function
   | DefT (_, _, FunT (_, _, { return_t; _ })) -> return_t
   | _ -> failwith "Getter property with unexpected type"
 
+and tuple_length reason trust ts =
+  let r =
+    let desc = RTupleLength (List.length ts) in
+    replace_desc_reason desc reason
+  in
+  let t =
+    let n = List.length ts in
+    let float = Pervasives.float_of_int n in
+    let string = Pervasives.string_of_int n in
+    SingletonNumT (float, string)
+  in
+  DefT (r, trust, t)
+
 and elemt_of_arrtype = function
   | ArrayAT (elemt, _)
   | ROArrayAT elemt
--- a/newtests/tuples/test.js
+++ b/newtests/tuples/test.js
@@ -75,10 +75,10 @@ export default suite(({addFile, addFiles, addCode}) => [
        `
          test.js:3
            3: function foo(x: [1,2]): string { return x.length; }
-                                                      ^^^^^^^^ Cannot return \`x.length\` because number [1] is incompatible with string [2].
+                                                      ^ Cannot return \`x.length\` because tuple length of 2 [1] is incompatible with string [2].
            References:
-           255:     +length: number;
-                             ^^^^^^ [1]. See lib: [LIB] core.js:255
+             3: function foo(x: [1,2]): string { return x.length; }
+                                ^^^^^ [1]
              3: function foo(x: [1,2]): string { return x.length; }
                                         ^^^^^^ [2]
        `,
--- a/src/common/reason.ml
+++ b/src/common/reason.ml
@@ -91,6 +91,7 @@ type 'loc virtual_reason_desc =
   | RROArrayType
   | RTupleType
   | RTupleElement
+  | RTupleLength of int
   | RTupleOutOfBoundsAccess
   | RFunction of reason_desc_function
   | RFunctionType
@@ -240,8 +241,8 @@ let rec map_desc_locs f = function
     | RVoid | RNull | RSymbol | RExports | RNullOrVoid | RLongStringLit _ | RStringLit _
     | RNumberLit _ | RBigIntLit _ | RBooleanLit _ | RObject | RObjectLit | RObjectType
     | RObjectClassName | RInterfaceType | RArray | RArrayLit | REmptyArrayLit | RArrayType
-    | RROArrayType | RTupleType | RTupleElement | RTupleOutOfBoundsAccess | RFunction _
-    | RFunctionType | RFunctionBody | RFunctionCallType | RFunctionUnusedArgument
+    | RROArrayType | RTupleType | RTupleElement | RTupleLength _ | RTupleOutOfBoundsAccess
+    | RFunction _ | RFunctionType | RFunctionBody | RFunctionCallType | RFunctionUnusedArgument
     | RJSXFunctionCall _ | RJSXIdentifier _ | RJSXElementProps _ | RJSXElement _ | RJSXText | RFbt
       ) as r ->
     r
@@ -539,6 +540,7 @@ let rec string_of_desc = function
   | RTupleType -> "tuple type"
   | RTupleElement -> "tuple element"
   | RTupleOutOfBoundsAccess -> "undefined (out of bounds tuple access)"
+  | RTupleLength i -> spf "tuple length of %d" i
   | RFunction func -> spf "%sfunction" (function_desc_prefix func)
   | RFunctionType -> "function type"
   | RFunctionBody -> "function body"
@@ -1295,6 +1297,7 @@ let classification_of_reason r =
   | RObjectClassName
   | RInterfaceType
   | RTupleElement
+  | RTupleLength _
   | RTupleOutOfBoundsAccess
   | RFunction _
   | RFunctionType
--- a/src/common/reason.mli
+++ b/src/common/reason.mli
@@ -41,6 +41,7 @@ type 'loc virtual_reason_desc =
   | RROArrayType
   | RTupleType
   | RTupleElement
+  | RTupleLength of int
   | RTupleOutOfBoundsAccess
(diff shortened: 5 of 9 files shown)


goodmind:
@ilya-bobyr can you sign CLA?

goodmind:
@ilya-bobyr this is just my personal issues list, regarding to "interesting"

ilya-bobyr:
I do not think I need to sign the CLA.  Github does not show that as a necessary action.
Or am I just not seeing it?

goodmind:
@ilya-bobyr https://code.facebook.com/cla

ilya-bobyr:
I know where the CLA is.

I work for Google and our OSS documentation says that I can not sign any CLAs without approval.  If I actually need to sign a CLA, I will need to take it through a certain process.
At the same time the documentation says that there is a preexisting CLA for the Facebook Flow project, though maybe I am misreading it.

I manage only one GitHub project, but there is automatically says when a CLA signature is required for a change to be merged.  And I do not see this check triggered here.  I see "All checks have passed".  Maybe I am just not seeing results of this particular check, or maybe it is not integrated into this repository.  Or the facebook-github-bot is how it works here?

ilya-bobyr:
After looking at other poll requests I see that it is the bot responsibility to add a `CLA Signed` label.
I will go through the process of signing the CLA.

Maybe someone can review the change meanwhile? :)

src/typing/type.ml
```diff
@@ -3358,9 +3364,13 @@ let string_of_use_ctor = function
   | DestructuringT _ -> "DestructuringT"
   | ModuleExportsAssignT _ -> "ModuleExportsAssignT"
 
+let string_of_key = function
+  | StringKey s -> s
+  | NumberKey n -> Pervasives.string_of_float n
```
goodmind: `Dtoa.ecma_string_of_float` ?

src/typing/refinement.ml
```diff
@@ -39,17 +39,30 @@ and key_of_identifier (_, { Ast.Identifier.name; comments = _ }) =
...
+  | PropertyExpression (_, Ast.Expression.Literal {
+      Ast.Literal.value = Ast.Literal.Number f; _; }) -> (
+    (* The `raw` string is no good as a refinement key because of equivalence
+       classes like "1" = 1 = 0.9... = 1.0 = .... We just leave 0.9... to OCaml
+       randomness because the ES spec leaves its handling ambiguous. *)
+    let name =
+      let i = Pervasives.int_of_float f in
+      if Pervasives.float_of_int i = f
+      then Pervasives.string_of_int i
+      else Pervasives.string_of_float f
```
goodmind: `Dtoa.ecma_string_of_float`?

goodmind:
I'm not sure this would be merged given that no-one replied to old PR :\

goodmind:
Related commit https://github.com/facebook/flow/commit/64ab73fd5bd04d2486a7bddff215198b9421abde

goodmind:
I think it would be nice to split sentinel refinement to own PR, and leave this one with only `length` refinement

ilya-bobyr:
> I'm not sure this would be merged given that no-one replied to old PR :\

Can you elaborate on this one?
There was a discussion in the original PR.  And there are bugs that it references: https://github.com/facebook/flow/pull/5048#issue-145201732
My understanding is the original author has lost interest, but maybe I am wrong.

goodmind:
@ilya-bobyr I mean there wasn't replies from Flow team

ilya-bobyr:
> I think it would be nice to split sentinel refinement to own PR, and leave this one with only `length` refinement

I was thinking the same.
After https://github.com/facebook/flow/commit/64ab73fd5bd04d2486a7bddff215198b9421abde the floating-point indices are not allowed anymore, right?

I will try to split the refinement and will remove the floating-point rounding.
This PR is different from the original one, so maybe just reuse it?

> @ilya-bobyr I mean there wasn't replies from Flow team
> I'm not on Flow team, I just have rights to add labels to PRs and issues

Is there something you would recommend here?
I can try fixing other issues to get some attention?
I saw you marked one as https://github.com/facebook/flow/labels/good%20first%20issue

goodmind:
Well, I still have [bracket refinement PR](https://github.com/facebook/flow/pull/7597) open even with Flow team response

> Is there something you would recommend here?

You can try to join Discord in contributor-support channel

> This PR is different from the original one, so maybe just reuse it?

What you mean?

popham:
My 2 cents: You're pissing in the wind.

ilya-bobyr:
Sorry I did not post any updates.
I was moving with no Internet access.

I will try to get back to this during next week.
Will split the refinement from the rounding and will post to the discord channel.

panagosg7:
@ilya-bobyr : I second @goodmind 's suggestion:
> I think it would be nice to split sentinel refinement to own PR, and leave this one with only length refinement

This will make it easier for me to review. Mostly so you don't misinterpret me importing the PR as is.

ilya-bobyr:
Sorry it took me this long.

Anyways.  I have removed the rounding logic.  Considering that at run-time no rounding occurs, it seems wrong to introduce refinements based on rounded values.
Also, rebased against the latest `facebook/master`.

@panagosg7 Please, take a look.  Does this make sence?

ilya-bobyr:
Replaced `Pervasives.string_of_float` with `Dtoa.ecma_string_of_float`, as @goodmind have suggested.

ilya-bobyr:
Updated the tests to use the `newtests` format.
Rebased on top of the latest `master` state.

goodmind:
newtests is for LSP only, it isn't new format :\

goodmind:
Can you split PR into two like I suggested?

ilya-bobyr:
What is LSP?  I assumed that as the directory is called "newtests", it is the new format.

Will split the length refinement - looking into it.
Sorry, I missed the suggestion.

ilya-bobyr:
Found an answer to the "What is LSP?" question in the discord channel:

>  Its a common interface to talk to ide so that we can integerate to different ide. You can read more about it here  https://microsoft.github.io/language-server-protocol/specification

ilya-bobyr:
Removed the sentinel refinement form this PR.
Switched back to `tests/` and added a few more tests.

I am confused as to why the last two new tests pass.
I would expect the refinement to fail, but it seems to succeed in a way I do not understand completely.
How to debug incorrect refinement?

goodmind:
https://github.com/facebook/flow/issues/4181 how to debug

ilya-bobyr:
@goodmind Thanks!
`bin/flow check --verbose-indent` prints a lot of details.
And using an `ocamldebug` seems like an even more challenging approach for someone not too familiar with the flow internals.  I will try to understand what is going on.

@popham Maybe you would have an idea as to what is going on?
The first test that I have added that I would expect to fail looks like this:
```
function why_does_this_pass(a: [number] | [string, number]): void {
  function is_number(v: number) {}
  if (a.length === 7) {
    is_number(a[1]);
    (a[1]: number);
  }
}
```

goodmind:
@ilya-bobyr it's because refinement is `empty`?
image

goodmind:
What you think about making `length` number literal type instead of regular number type?

src/typing/flow_js.ml
```diff
@@ -6558,6 +6558,18 @@ let rec __flow cx ((l: Type.t), (u: Type.use_t)) trace =
       (GetPropT _ | SetPropT _ | MethodT _ | LookupT _) ->
       rec_flow cx trace (get_builtin_typeapp cx ~trace reason "Array" [t], u)
 
+    (*************************)
+    (* Tuple "length" access *)
+    (*************************)
+
+    | DefT (_, trust, ArrT (TupleAT (_, ts))),
+      LookupT (reason, _, _, Named (_, "length"), ReadProp {
```
goodmind: Can you add `GetPropT` to right hand match?

ilya-bobyr: I can definitely do it, but at the moment I do not understand what it would mean.
I will spend some time reading the code, to understand why would it be better.

ilya-bobyr:
> What you think about making `length` number literal type instead of regular number type?

It seems quite reasonable.  I actually thought that this is how it already is.
I still need to spend some time on it, to understand it better.  But it looks like flow knows that `length` can never reach `7` here, as `a` has an empty type.
I guess, I am confused why is it still allowed to issue indexed access.
Maybe tuples are missing some checks in this case?

src/typing/flow_js.ml
goodmind: Currently `x.length` doesn't work on tuples outside of `if`, it is obvious if you change number to number literal in `tuple_length`

goodmind:
@ilya-bobyr it's how `empty` works I think

goodmind:
There's some conflicting files

ilya-bobyr:
> There's some conflicting files

I will resolve the conflicts and will make other changes you suggested today or tomorrow.

src/typing/type.ml
```diff
@@ -3435,6 +3435,19 @@ and extract_getter_type = function
...
+and tuple_length reason trust ts =
+  let r =
+    let desc = RTupleLength (List.length ts) in
+    replace_desc_reason desc reason
+  in
+  let t =
+    let n = List.length ts in
+    let float = Pervasives.float_of_int n in
+    let string = Pervasives.string_of_int n in
+    NumT (Literal (None, (float, string)))
```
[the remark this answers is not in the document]
ilya-bobyr: Thanks!

src/typing/flow_js.ml
ilya-bobyr: Replaced `LookupT` with `GetPropT`.

ilya-bobyr: And then replaced back with both versions :)

ilya-bobyr:
Updated the tests to check for `empty`, changed `tuple_length` to emit `SingletonNumT` and added `GetPropT` in addition to the `LookupT` when matching for the "length" access.

ilya-bobyr:
A negative test in the LSP test suite is broken, as the type of `[1,2].length` have changed from `number` to `tuple length of 2`.
I can easily fix the test, but it also exposed a bug (at least I think it is a bug) in the error reporting.
Test runs this code:
```
function foo(x: [1,2]): string { return x.length; }
```
And currently the error looks like this:
```
test.js:3
  3: function foo(x: [1,2]): string { return x.length; }
                                             ^^^^^^^^ Cannot return \`x.length\` because tuple length of 2 [1] is incompatible with string [2].
  References:
    3: function foo(x: [1,2]): string { return x.length; }
                                                 ^^^^^^^^ [1]
    3: function foo(x: [1,2]): string { return x.length; }
                               ^^^^^^ [2]
```
The error needs to refer to the source of the `tuple length of 2` type, and it is referring to the `.length` expression (though the positions is somewhat off - 2 extra characters are included).

One questions is: do we want it to refer to the tuple itself, instead of the `.length` access?
Or is this the expected form and I just need to fix those extra characters?

goodmind:
@ilya-bobyr I think it should point to tuple itself

goodmind:
@ilya-bobyr any news?

ilya-bobyr:
Seems like the problem is not with this change, but with the way the "reasons" are serialized for the output used by the LSP tests.

Given this expression:
```
function foo(x: [1,2]): string { return x.length; }
```
When I run `flow check` in the console I see this:

image

When I run `flow check | less` (to output to a "dumb" terminal) I see this:

image

But when the same expression is checked in an LSP test, the output is like this:

```
/home/ilya/works/flow/newtests/tuples/test.js line 74 col 7
Actual new errors (+) didn't match expected new errors (-)
           test.js:3
             3: function foo(x: [1,2]): string { return x.length; }
                                                        ^ Cannot return `x.length` because tuple length of 2 [1] is incompatible with string [2].
             References:
               3: function foo(x: [1,2]): string { return x.length; }
                                  ^^^^^ [1]
               3: function foo(x: [1,2]): string { return x.length; }
                                          ^^^^^^ [2]
```

Notice that the error text is next to the error location here, while `flow` outputs directly to a terminal the error text is output separately as the first line.  In the first case, `flow` provides more details as `x` and `.length` are colored using different colors.  But when the same message is output to a "dumb" terminal, these details are lost.
I assume there is some discrepancy between the way `flow` outputs to a "dumb" terminal, and the output that is used by the LSP tests.  I have looked a bit at the code, but it is non-trivial, so I was not able to find a confirmation and/or a specific block responsible for this.

goodmind:
@panagosg7 can you review?

src/typing/flow_js.ml
```diff
@@ -6647,6 +6647,15 @@ struct
         | ( DefT (reason, _, ArrT (ArrayAT (t, _))),
             (GetPropT _ | SetPropT _ | MethodT _ | LookupT _) ) ->
           rec_flow cx trace (get_builtin_typeapp cx ~trace reason "Array" [t], u)
+        (*************************)
+        (* Tuple "length" access *)
+        (*************************)
+        | ( DefT (reason, trust, ArrT (TupleAT (_, ts))),
+            ( LookupT (_, _, _, Named (_, "length"), ReadProp { use_op = _; obj_t = _; tout })
```
panagosg7: Is the `LookupT` match necessary? It's typically used for traversing prototype chains. (None of the tests are affected by removing it.)

ilya-bobyr: Yep, I was also not sure if it needs to stay.  You probably have a better understand on how it works.
There was a discussion in `#contributor-support` and @goodmind suggested we keep it.
@jbrown215 also commented, but I am not sure the conclusion was very solid.

I can easily remove it if it does not matter.  And if it does, it would be good to add a test.  But, unfortunately, I do not know what the code should look like for `LookupT` to be triggered and for the `GetPropT` not to.

Here is a discussion:
```
ilya-bobyr09/16/2019
Actually, if I replace LookupT with GetPropT, the tests pass %)
would you recommend a function, module, and/or a comment to look at to better understand how all these types are used?

goodmind09/16/2019
no idea, but I would keep LookupT because it's more general than GetPropT or ask when someone from flow team is online
there isn't much description of differences between them

[...]

jbrown21509/16/2019
@goodmind: I'm 99% certain that GetPropT eventually delegates to LookupT
but LookupT assumes certain properties about the left-hand-side of the flow, so GetPropT first does some work on that

goodmind09/16/2019
@jbrown215 on tuples it is just handle-all case
so idk how LookupT, GetPropT works there
I thought LookupT is only for prototype chain
And tuples aren't implemented like this

jbrown21509/16/2019
LookupT definitely traverses the prototype chain. I'm pretty unfamiliar with the tuple implementation

goodmind09/16/2019
that's all
so it looks like you have to duplicate matchers for LookupT and GetPropT if you want to overwrite some property, objects are more complex than this
```

Any idea what can be a candidate for a test for `LookupT`?

panagosg7: Okay, I took a closer look, and I think we should lose the `LookupT` check. The only case where I managed to see a difference is with the `--no-flowlib` flag. Take the following code for example
```
declare var a: [string];
(a.length === 1) ? true : false;

declare var b: Array<string>;
(b.length === 1) ? true : false;
```
With your changes, I get 
```
$ flow check tuples/ --no-flowlib
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ tuples/file.js:7:2

Property length is missing in array type [1].

      4│ (a.length === 1) ? true : false;
      5│
 [1]  6│ declare var b: Array<string>;
      7│ (b.length === 1) ? true : false;
      8│

Found 1 error
```
In line 4, the `ArrT (TupleAT _), LookupT ("length")` rule fires and prevents the error even though we are not importing the `Array` defs from `lib/core.js`. For consistency with what we do for arrays, we probably should error in this case as well.

Without the flow `--no-flowlib` and without the `LookupT` case, the `length` property gets looked up in the `Array` prototype, so we don't report any errors.

ilya-bobyr: Thanks for looking into it.  Removed `LookupT`.

ilya-bobyr:
@panagosg7 Updated.

ilya-bobyr:
> 
> 
> @panagosg7 merged this pull request in [460ba7f](https://github.com/facebook/flow/commit/460ba7f44da0ee7a65005c84c3113305383f56c1).

Thanks!  And thanks for fixing the offsets in the tests!

Issue #7189, opened 2018-11-17
Indeed it looks like Flow currently special-cases `value` properties in the descriptor:


m0n1ter:
https://github.com/facebook/flow/blob/master/src/typing/type_inference_js.ml#L3651

It&#39;s definitely worth looking into support for more of the descriptor features than just `value`.

_Originally posted by @jeffmo in https://github.com/facebook/flow/issues/285#issuecomment-76244248_

PR #7981, opened 2019-07-30
Do nothing when value property isn't found in Object.defineProperty

goodmind:
Fixes https://github.com/facebook/flow/issues/285
Fixes https://github.com/facebook/flow/issues/7189

Doesn't support `get` or `set`


panagosg7:
Sorry for the delay @goodmind. I just rebased and put it up for internal review.

goodmind:
Should I close this with recent commit?

PR #7990, opened 2019-08-05
Add destructuring to catch clause

### LINKED ISSUE #6636: destructuring catch parameter throws Error
```javascript
function test() {
    try {
      //
    } catch ({ message }) {  // Unsupported catch parameter declaration.
      //
    }
}
```

https://flow.org/try/#0GYVwdgxgLglg9mABFApgZygCgJSIN4BQixyATgJ75EnED0t1xAvohAIZQQAWimeiAW3Ro2AcxSImuQjToMaTAkyA

not sure if this is related:

```javascript
function test() {
    try {
      //
    } catch (err: { message: string}) {  // Type annotations for catch parameters are not yet supported
      //
    }
}

https://flow.org/try/#0GYVwdgxgLglg9mABFApgZygCgJSIN4BQixyATgJ75EnED0t1xAvohAIZQQAWimKppAFz5EAW3Ro2AcxTCMpGGClNchGnQY0mBJkA

### PR DESCRIPTION
**Pull request by goodmind**
Fixes #6636

I also moved location of error to be more accurate when there is type annotation

### CODE REVIEW

**Comment by goodmind** [comment]
Not sure if this is considered type annotation?
```js
try {
} catch ({message = 'foo'}) {
  message // type is string
}
```

gkz:
We will not be adding this as the only "safe" type for a catch argument is `unknown` (with `use_unknown_in_catch_variables=true`) - https://flow.org/en/docs/config/options/#toc-use-unknown-in-catch-variables
and you can't destructure `unknown` safely.
I would instead use [`match`](https://flow.org/en/docs/match/) on the argument to safely destructure

Issue #7467, opened 2019-02-12
Cannot set defaultProps with React.memo and React.forwardRef


ryanashcraft:
Flow version: 0.92.1 (flow-bin)
macOS 10.14.2

## Expected behavior

You can set `defaultProps` on a component returned from `React.memo` and `React.forwardRef`.

## Actual behavior

Flow will yield an error with the following message:

Cannot assign object literal to `App.defaultProps` because property `defaultProps` is missing in `React.AbstractComponentStatics` [1].

---

Example with React.forwardRef:

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVBLAtgBzgJwBdkwBDAZzACUBTUgY2KnzizAHJ87H2BuVYMDDY8Rat0IARAPIBZMM1ZgARFwaEAtABNWy-qkIBPHDTAAFFjkoBeMAG9UYMADtSWGgC4w5QvgzOAcwAaVABffXo4Zx8wAEEcHDBbWnUAOigCBFJ8LVooAAp8nEtyLws4KwBKJIA+e0cwLkIAV3xnMAAeLQwANxqACxoYeCD7YoryVNd3UI7gbr6wyv14nFStGihSZphCcqsk+qdpzw5SZyjDLDhm8nYQ8PRBMEjo4hY4QgBRGBp3Z2Ith09Ga-0IqQCNG+vzBACFDABJLT5VRwT7KZYCIQpRgyWSpLjODb4fIdVZgYA1UYfaF-GgAzFAA

Corresponding codesandbox showing that setting `defaultProps` should work:

https://codesandbox.io/s/xv8m3qy6jo

---

Example with React.memo:

[...]

PR #8001, opened 2019-08-08
Add defaultProps to AbstractComponent

goodmind:
Partial fix for https://github.com/facebook/flow/issues/7467

But I don't think it's faithful to how Flow models `defaultProps` ?

`$Rest<X, {}>` is basically better `$Shape` (makes fields really optional)

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -127,7 +127,8 @@ declare class LegacyReactComponent<Props, State>
   state: State;
 }
 
-declare type React$AbstractComponentStatics = {
+declare type React$AbstractComponentStatics<Config> = {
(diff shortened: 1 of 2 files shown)


jbrown215:
This is not sound:
```
class A extends React.Component<{| foo: number, bar: number |}> {
  static defaultProps = {foo: 3};

  render() {
    this.props.foo.toString();
  }
}

const AC: React.AbstractComponent<{| foo?: number, bar: number |}> = A;

AC.defaultProps = {bar: 3};

<AC bar={3} />; // foo is undefined at runtime, so the `toString()` call will error.

PR #8007, opened 2019-08-09
Throw expressions

### LINKED ISSUE #6605: Support for throw expressions
```js
const fn = () => true ? throw 'true' : throw Error('false')
```

causes `Unexpected token throw`.

https://github.com/tc39/proposal-throw-expressions (stage 2)

goodmind:
Fixes #6605


goodmind:
```js
const fn = (x?: string = throw 'x') => {
  x * 2
}
```

Kills typechecker, nice

SamChou19815:
Close since we won't implement at least stage 3 proposal, and the codebase has since changed a lot and the PR needs a lot of rework anyways.

PR #8015, opened 2019-08-12
Improve lib def for async act use case in react-dom/test-utils

leonp1991:
After attempting to upgrade to react 16.9, I noticed some flow errors when trying to use act (from `react-dom/test-utils`) with an async callback result. After some digging I noticed the type definition in https://github.com/facebook/flow/blob/master/lib/react-dom.js#L64 does not directly match https://github.com/facebook/react/blob/master/packages/react-reconciler/src/ReactFiberWorkLoop.js#L205. This PR aims to rectify this.

Is there a test suite for these lib defs somewhere where I can confirm this is working correctly.

### DIFF
--- a/lib/react-dom.js
+++ b/lib/react-dom.js
@@ -61,7 +61,7 @@ declare module 'react-dom/server' {
   declare var version: string;
 }
 
-type Thenable = { then(resolve: () => mixed, reject?: () => mixed): mixed, ... };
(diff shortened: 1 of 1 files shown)


goodmind:
Why close?

leonp1991:
Tests were not passing and this one wasn't a priority for me. Feel free to pick up where I left off @goodmind 👍

PR #8016, opened 2019-08-12
Re-sync with internal repository

mroch:
The internal and external repositories are out of sync. This attempts to brings them back in sync by patching the GitHub repository. Please carefully review this patch. You must disable ShipIt for your project in order to merge this pull request. DO NOT IMPORT this pull request. Instead, merge it directly on GitHub using the MERGE BUTTON. Re-enable ShipIt after merging.


goodmind:
Why it goes out-of-sync? Is it always with `hack` folder?

mroch:
we were syncing parts of hack that we weren't using. I stopped syncing these directories, so we have to delete them from the github repo even though they still exist in hack.

goodmind:
@mroch I think you can unsync this file too
https://github.com/facebook/flow/blob/master/hack/utils/php_escaping.ml

PR #8018, opened 2019-08-13
Add [rollouts] to .flowconfig

gabelevi:

### DIFF
--- a/src/commands/config/flowConfig.ml
+++ b/src/commands/config/flowConfig.ml
@@ -898,6 +898,110 @@ let parse_strict lines config =
   StrictModeSettings.of_lines lines >>= fun strict_mode ->
   Ok ({config with strict_mode}, [])
 
+(* Basically fold_left but with early exit when f returns an Error *)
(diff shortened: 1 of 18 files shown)


goodmind:
What is it?

gabelevi:
> What is it?

The final commit will have a much more descriptive commit message. I'll also document it.

Internal builds of Flow log a lot of perf info (we stub it out for OSS). So when we launch a new behavior (e.g. `experimental.abstract_locations=true`) we look at how the perf changes. But the repositories we test on also change over time, so it's not a great way to A/B test.

This `[rollouts]` feature of the `.flowconfig` will let us turn it on for 50% of our internal users. Then we can have cleaner data about the effects of a launched change.

I'm not sure how useful this will be for OSS users though.

goodmind:
I thought that it is env-related, like
```
[ignore]
(ignore_everything=true) *.js
```

```sh
$ export ignore_everything=true 
$ flow check
```

PR #8021, opened 2019-08-14
Fix type alias in block

### LINKED ISSUE #6961: Strange bound-name bug when declaring type within block scope
The following code throws a strange error:
```
{
  type Foo = string;
}
```
Here's the error message:
```
2:   type Foo = string;
          ^ Cannot declare `Foo` [1] because the name is already bound.
    References:
    2:   type Foo = string;
              ^ [1]
```
According to Flow, the type is being redeclared... in the same place it's being declared.

[(Try Flow link to the above code)](https://flow.org/try/#0N4KABGAuCeAOCmYBiB7FYC8YDOkBOAlgHYDmA3CAL5A)

goodmind:
Fixes #6961

### DIFF
--- a/src/typing/env.ml
+++ b/src/typing/env.ml
@@ -456,11 +456,12 @@ let bind_entry cx name entry loc =
       (* if no entry already exists, this might be our scope *)
       | None -> Entry.(
         match scope.Scope.kind, entry with
-        (* lex scopes can only hold let/const/class bindings *)
+        (* lex scopes can only hold let/const/class/type bindings *)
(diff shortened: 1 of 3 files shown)


mroch:
it's a bit weird to me that we'd allow this, but let's discuss on the issue.

if we're going ahead with this, can you add some tests please? in particular, that the same name can be used in different blocks, that it notices redeclaration in nested blocks, and that you can't use the type outside the block.

SamChou19815:
Looks like this is no longer an issue.

PR #8024, opened 2019-08-15
[prelude] Add undefined

goodmind:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined lol


goodmind:
What exactly should be in `no_flowlib` lib?

PR #8029, opened 2019-08-18
[DOCS] Add .js.flow documentation

### LINKED ISSUE #3609: Document .js.flow files
The old documentation described how you could omit types in a file called `foo.js` and declare the types in a file called `foo.js.flow` that is located in the same directory. I can't find that in the new documentation. Is that still a supported way to specify types?

### LINKED ISSUE #8026: defining modules that are accessed via a relative require/import path
This page https://flow.org/en/docs/libdefs/creation/#toc-declaring-a-module states `defining modules that are accessed via a relative require/import path will be documented soon`. Is this available yet? It seems the line has been around for 2 years now.

goodmind:
This is old page about .js.flow repurposed
 
Closes #8026
Closes #3609

### DIFF
--- a/website/_data/guides.yml
+++ b/website/_data/guides.yml
@@ -103,6 +103,14 @@
     - path: "/docs/libdefs/"
     # - path: "/docs/libdefs/usage/"
     - path: "/docs/libdefs/creation/"
+      additional_reading:
+        - guide: declarations
+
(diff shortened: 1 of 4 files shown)


goodmind:
lol, this is exactly on leap year

Issue #7673, opened 2019-04-26
Cannot resolve JSON file starting with a dot


mrtnzlml:
Flow version:

```
version 0.97.0
```

Flow cannot resolve require with JSON filename starting with a dot:

```js
const meta = require('../.meta.json');
```

It results in this error:

```text
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/apps/graphql-skymock/datasets/__tests__/MetafileConsistency.test.js:7:22

Cannot resolve module ../.meta.json.

      4│ import path from 'path';
      5│ import Ajv from 'ajv';
      6│
      7│ const meta = require('../.meta.json');
      8│
      9│ test('every dataset defined in metafile should exist', () => {
     10│   expect.hasAssertions();
```

However, when I remove the dot (and rename the file) everything is fine:

```js
const meta = require('../meta.json');
```

```text
No errors!
```

Thank you very much for checking it! :)

mrtnzlml:
This is true even for JS files (`.meta.js`).

goodmind:
Yeah, Flow ignores all dotfiles

mrtnzlml:
Can I opt-out from this? I know it's unusual to start with a dot so it's not very common but it really makes sense in my case. Why does Flow ignore dot files by default?

goodmind:
No idea, it also wouldn't work with [include]

goodmind:
Maybe you can try symlinks

gabrielrumiranda:
Hi, can I work on this ?

goodmind:
@gabrielrumiranda Sure you can, what you want to know about codebase?

gabrielrumiranda:
@goodmind is my first contribution in project, i don't know much about codebase. The module who works with imports is ` flow/src/typing/import_export.ml ` ?

goodmind:
@gabrielrumiranda 
https://github.com/facebook/flow/blob/95ef2463e608ce74d04fdccb26a673a7870d52fc/src/common/files.ml#L97-L102
This should be here I think

gabrielrumiranda:
@goodmind  the tests of this part of code is in [tests/declarations_files_node](https://github.com/facebook/flow/tree/95ef2463e608ce74d04fdccb26a673a7870d52fc/tests/declaration_files_node) ?

goodmind:
[...]

PR #8045, opened 2019-08-26
Files starting with dot

gabrielrumiranda:
This PR solve the problem of this [issue](https://github.com/facebook/flow/issues/7673). Lets you import files that start with dots


gabelevi:
I'm not sure I'm comfortable with deleting this behavior. Like there's a lot of JS for me in `.hg/`. Someone could add that to their `[ignores]`, but I'd rather that the default behavior does the right thing.

I suppose I'm fine making it an option? Like if someone wants to set `ignore_dotfiles=false` in their `.flowconfig` then they're welcome to do so? You would need to modify `flowConfig.ml` to add the option and then also add it to `type options` in `files.ml`. Then the compilation errors should direct you to all the other places you need to fix.

gabrielrumiranda:
@gabelevi I think this the better way, I go make this changes

goodmind:
@gabrielrumiranda can you add tests with nameless files?

gabrielrumiranda:
@goodmind I separate the test in another folders, when you tell nameless file you tell a file with name `.` ?

goodmind:
@gabrielrumiranda yeah, like `.flowconfig`, `.js`

gabrielrumiranda:
I'm try but don't work

gabrielrumiranda:
Other thing, files begin with dot inside a node_module cannot be import using relative path. I test this in a simple node.js project and don't work too

goodmind:
Hmm

goodmind:
@gabrielrumiranda it should work now

Issue #2696, opened 2016-10-27
Some issues with CSSOM


k15a:
Hey,

I am currently working with stylesheets in JavaScript and would love to check my code with flow.
I have some issues where I don't know if the type checking is correct. Would be lovely if someone could check if I am doing something wrong.

Why is this just a StyleSheet and not a CSSStyleSheet?

``` JS
const style: ?StyleSheet = ((): ?StyleSheet => {
    // Create the <style> tag
    const style: HTMLStyleElement = document.createElement('style')

    style.type = 'text/css'

    // Add the <style> element to the page
    document.head.appendChild(style)

    return style.sheet
})()
```

If this is the indented behaviour how could I get an CSSStyleSheet?

According to https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule the return type of insertRule is a number but the flow return type is void https://github.com/facebook/flow/blob/master/lib/cssom.js#L39. Who is wrong?

kof:
Just started to add types to [JSS](https://github.com/cssinjs/jss), types for CSSOM is partially wrong and not fully implemented.

Missing interfaces I use are for e.g. : HTMLStyleElement, CSSStyleSheet, CSSRuleList, CSSStyleRule

kof:
For e.g. I think https://github.com/facebook/flow/blob/master/lib/dom.js#L2648 sheet should be of type CSSStyleSheet, otherwise you get

```
src/backends/DomRenderer.js:211
211:     return this.element.sheet.cssRules
                                   ^^^^^^^^ property `cssRules`. Property not found in
211:     return this.element.sheet.cssRules
                ^^^^^^^^^^^^^^^^^^ StyleSheet
```

foray1010:
have same problem, wonder why `sheet` must be `StyleSheet` but not `CSSStyleSheet` which it actually is

necolas:
Other types like `CSSMediaRule` are missing too

kof:
This can be closed as well.

PR #8071, opened 2019-09-10
Add a number of CSSOM interfaces + fix HTMLStyleElement.sheet type

### LINKED ISSUE #2943: Missing CSSStyleRule which extends CSSRule
Initially I thought `CSSRule` declaration should have `style` property, but no actually we miss an entire class `CSSStyleRule` which inherits from `CSSRule` and implements `style` property.

https://github.com/facebook/flow/blob/master/lib/cssom.js#L42

kof:
Fixes #2943 and #2696
- CSSStyleSheet - https://www.w3.org/TR/cssom-1/#the-cssstylesheet-interface
- CSSStyleRule - https://www.w3.org/TR/cssom-1/#cssstylerule
- CSSMediaRule - https://www.w3.org/TR/cssom-1/#the-cssmediarule-interface
- CSSConditionRule - https://developer.mozilla.org/en-US/docs/Web/API/CSSConditionRule
- CSSGroupingRule - https://www.w3.org/TR/2016/WD-cssom-1-20160317/#the-cssgroupingrule-interface

### DIFF
--- a/lib/cssom.js
+++ b/lib/cssom.js
@@ -7,12 +7,12 @@
 
 declare class StyleSheet {
   disabled: boolean;
-  href: string;
-  media: MediaList;
-  ownerNode: Node;
-  parentStyleSheet: ?StyleSheet;
-  title: string;
-  type: string;
+  +href: string;
+  +media: MediaList;
+  +ownerNode: Node;
+  +parentStyleSheet: ?StyleSheet;
+  +title: string;
+  +type: string;
 }
 
 declare class StyleSheetList {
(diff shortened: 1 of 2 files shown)


lib/cssom.js
```diff
@@ -35,7 +35,26 @@ declare class CSSStyleSheet extends StyleSheet {
...
+  deleteRule(index: number): void;
+  insertRule(rule: string, index: number): number;
+}
+
+declare class CSSConditionRule extends CSSGroupingRule {
+  conditionText: string;
+}
+
+declare class CSSMediaRule extends CSSConditionRule {
+  media: MediaList;
```
mvitousek: It looks like this attribute should be read-only, right? https://www.w3.org/TR/cssom-1/#the-cssmediarule-interface

mvitousek: This should be read-only, right? https://www.w3.org/TR/2016/WD-cssom-1-20160317/#the-cssmediarule-interface

kof: yep, I just went through all added interfaces and some previously existing once and updated all readonly props

kof:
I wish I knew what is failing in Facebook internal builds.

kof:
🎉

mvitousek:
Thanks for the contribution, and sorry about the delay in merging!

PR #8077, opened 2019-09-12
Add HeadersInit and signal to Fetch

goodmind:
https://fetch.spec.whatwg.org/#headers-class
https://fetch.spec.whatwg.org/#request-class

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -1350,7 +1350,7 @@ declare class WaveShaperNode extends AudioNode {
 
 // this part of spec is not finished yet, apparently
 // https://stackoverflow.com/questions/35296664/can-fetch-get-object-as-headers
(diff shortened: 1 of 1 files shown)


lib/bom.js
```diff
@@ -1350,7 +1350,7 @@ declare class WaveShaperNode extends AudioNode {
 
 // this part of spec is not finished yet, apparently
 // https://stackoverflow.com/questions/35296664/can-fetch-get-object-as-headers
-type HeadersInit = Headers | { [key: string]: string, ... };
+type HeadersInit = Headers | string[][] | { [key: string]: string, ... };
```
mroch, line 1353: changing this to `Array<[string, string]>` per

Screen Shot 2020-05-14 at 9 44 04 AM

PR #7676, opened 2019-04-28
Node Crypto update

goodmind:
Most functions that accept `string | Buffer` also accept `$TypedArray | DataView`

https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_class_method_ecdh_convertkey_key_curve_inputencoding_outputencoding_format
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_cipher_setaad_buffer_options
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_sign_sign_privatekey_outputencoding
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_constants
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_privatedecrypt_privatekey_buffer
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_privateencrypt_privatekey_buffer
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_publicdecrypt_key_buffer
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_publicencrypt_key_buffer
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_scrypt_password_salt_keylen_options_callback
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_scryptsync_password_salt_keylen_options
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_generatekeypair_type_options_callback
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_generatekeypairsync_type_options
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_setengine_engine_flags
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_setfips_bool
https://nodejs.org/docs/latest-v11.x/api/crypto.html#crypto_crypto_getfips

PR #8084, opened 2019-09-16
Add crypto create(De)cipher options parameter

skeggse:
See [`crypto.createCipher`](https://nodejs.org/docs/latest-v10.x/api/crypto.html#crypto_crypto_createcipher_algorithm_password_options), [`crypto.createCipheriv`](https://nodejs.org/docs/latest-v10.x/api/crypto.html#crypto_crypto_createcipheriv_algorithm_key_iv_options), [`crypto.createDecipher`](https://nodejs.org/docs/latest-v10.x/api/crypto.html#crypto_crypto_createdecipher_algorithm_password_options) and [`crypto.createDecipheriv`](https://nodejs.org/docs/latest-v10.x/api/crypto.html#crypto_crypto_createdecipheriv_algorithm_key_iv_options) documentation. These functions supported a `duplexStreamOption` as recently as Node 8, and supported an additional `authTagLength` as recently as Node 10.

Another approach could use overloading to avoid unexpected issues with `iv`s being `null` when the `algorithm` needs an IV.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -551,26 +551,41 @@ type crypto$key = string | {
   ...
 };
 
+type cryptoCipherOptions = duplexStreamOptions & {
+  authTagLength?: number,
+  ...
+};
+
 declare module "crypto" {
   declare var DEFAULT_ENCODING: string;
 
   declare class Sign extends crypto$Sign {}
   declare class Verify extends crypto$Verify {}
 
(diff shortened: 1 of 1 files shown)


goodmind:
I rather have this merged https://github.com/facebook/flow/pull/7676

skeggse:
Closing in favor of #7676.

Issue #7853, opened 2019-06-26
typeof does not preserve exactness of an object


kangax:
```js
const initialState = { loading: false, loaded: false };
type State = typeof initialState;

// type State = $Exact<typeof initialState>; // <-- doesn't work either
// type State = {| ...typeof initialState |} // <-- doesn't work either

const f: (state: State) => State = state => {
  return {
    loading: true,
    loaded: true,
    nice: ""
  };
};
```

Flow version: 0.102.0

## Expected behavior

It should error out on an extra property return from a function. Note that this works as expected in version 0.80 but not after.

## Actual behavior

There is no error

https://flow.org/try/#0PTAEAEDMBsHsHcBQiDGsB2BnALqAluntngIbQDK2J2ApqALygDeocJAJgQOYBcokZTDQA0rWBxrs+A6ENABfANzJsATwAOdStTqM1m2JHyFiZbbWWoMOfnwAUOHX3M0AlAwB8oFw1CPansyIoKAATjTYAK6h6EEhIWyc6Lyg2KGRIsHxiZJ8aRnCWSGEKDR8AETlWUqINUA

goodmind:
Related https://github.com/facebook/flow/issues/7526

goodmind:
Still an issue in 0.105.2

SamChou19815:
Now errors

PR #8096, opened 2019-09-19
Remove literal reason on $Exact

goodmind:
Partial fix for #7853 

Prevent leaking literal object reason to `$Exact`


SamChou19815:
Closing since the implementation of $Exact has significantly changed

Issue #4509, opened 2017-08-02
Covariant member of tuple


minedeljkovic:
Is there a way to annotate covariant member of tuple, similarly to [covariant object properties](https://flow.org/en/docs/lang/depth-subtyping/)?
(In context of ADTs tuple and record are analogous types i.e. a product type)

Extended example from [docs](https://flow.org/en/docs/lang/depth-subtyping/):
```js
class Person { name: string }
class Employee extends Person { department: string }

var employee: { who: Employee } = { who: new Employee };
var person: { +who: Person } = employee;

var employeeTuple: [Employee, number] = [new Employee, 1]
var personTuple: [Person, number] = employeeTuple // Error!
```
[Try flow](https://flow.org/try/#0MYGwhgzhAEAKCmAnCB7AdtA3tNYC28AXNBAC6ICWaA5tAL4BQokMAongA4goCe880eAA9S8NABMYCZOizRx8DmESkCaUsTKUa9BgwBuywZ258icgO4ALFMXZde-etAC8lm8TTwL0e6ad0ANwGRhxIqGjE2ADU1rZw4bJ0rsYOZsEhiKn+8AAqAK5c5gDafo7wADQ4+XgARkgAuinFXj5lZlUAjA2Z0GEyaAVFxMXSEVVoNfWITW7wJuVDIAIA9Cu+iIgoiACEQA)

Analogous  way would be to mark first member as covariant like:
```js
var employeeTuple: [+Employee, number] = [new Employee, 1]
var personTuple: [Person, number] = employeeTuple // This would type check since first member of employeeTuple is covariant
```
This syntax is not currently supported in Flow.

Readonly tuple would also be enough for our use case, but I am not sure there is such a thing.

asolove:
This looks like a great request. It's odd that tuples get neither the property variance of records nor at least the read-only option like arrays.

wchargin:
Support. Tuple read/write semantics seem kind of weird: they are mutable (`t[0] = "hello"` is okay), but only support read-only array properties (`t.fill("wat")` gives "property `fill` missing in `$ReadOnlyArray`", as described in [the docs][1]). Having covariance annotations or `$ReadOnlyTuple<SomeNormalTuple>` would be nice.

[1]: https://flow.org/en/docs/types/tuples/#toc-cannot-use-mutating-array-methods-on-tuples

Bannerets:
Read-only tuples would be useful in iterators:

```js
[...]

PR #8099, opened 2019-09-21
Add read-only tuple

goodmind:
A regular tuple is still subtype of `$ReadOnlyArray` because of length-changing methods, but read-only tuple disables writing to elements

The syntax is just reusing `$ReadOnly` destructor

```js
declare var x: $ReadOnly<[number, string]>
```

Partial solution for https://github.com/facebook/flow/issues/4509


nnmrts:
Any updates on this?

PR #8100, opened 2019-09-22
Add no-floating-promises lint

goodmind:
```ini
# .flowconfig
[lints]
no-floating-promises=warn
```

```js
async function foo() { return 1 }
declare var x: Promise<number>;

foo().then(() => {}, () => {}) // fine
foo().catch(() => {}) // fine

foo().then(() => {}) // warn
foo() // warn
x; // warn
```

### DIFF
--- a/src/typing/statement.ml
+++ b/src/typing/statement.ml
@@ -1178,57 +1178,59 @@ and statement cx : 'a -> (ALoc.t, ALoc.t * Type.t) Ast.Statement.t =
             let (((_, t), _) as ast) = expression cx expr in
             (t, Some ast)
       in
+      let type_loc = loc_of_t t in
       let t =
         match Env.var_scope_kind () with
         | Scope.Async ->
           (* Convert the return expression's type T to Promise<T>. If the
            * expression type is itself a Promise<T>, ensure we still return
            * a Promise<T> via Promise.resolve. *)
-          let reason = mk_reason (RType "Promise") loc in
+          let reason_typeapp = annot_reason (mk_reason (RType "Promise") type_loc) in
           let t' =
             Flow.get_builtin_typeapp
               cx
-              reason
+              reason_typeapp
               "Promise"
               [
-                Tvar.mk_derivable_where cx reason (fun tvar ->
+                Tvar.mk_derivable_where cx (reason_of_t t) (fun tvar ->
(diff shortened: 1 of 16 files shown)


SebastienGllmt:
# Why this is useful

Promises that are never awaited can cause unexpected behavior because they may be scheduled to execute at an unexpected time.

It's easy to accidentally make this mistake. For example, suppose we have the code

```js
function writeToDb() {
  // synchronously write to DB
}
writeToDb();
```

but the code gets refactored so that writing to the database is asynchronous.

```js
async function writeToDb() {
  // synchronously write to DB
}
writeToDb(); // <- note we have no await here but probably the user intended to await on this!
```

# Why not use ESLint or a similar tool?

This kind of checking isn't really possible with ESLint for two reasons:
1) ESLint runs on a file-by-file basis (doesn't know if a function from a different file is an async function or not)
2) It's possible to have a function that returns a Promise but isn't async such as `function foo(): Promise<void> { return Promise.resolve(); };`. ESLint can't detect this without type information Flow provides

# Existing work

This check has existing in Typescript for a long time and was recently added this year to Typescript-Eslint lint also (Typescript's equivalent of ESLint). Link [here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md)

I tried to implement this rule in vanilla ESLint [here](https://github.com/SebastienGllmt/eslint-plugin-no-floating-promise) but the limitations of now having Flow inference and not working cross-file make it very limited in utility.

src/typing/statement.ml
```diff
@@ -1177,7 +1184,7 @@ and statement cx : 'a -> (ALoc.t, ALoc.t * Type.t) Ast.Statement.t =
           (* Convert the return expression's type T to Promise<T>. If the
            * expression type is itself a Promise<T>, ensure we still return
            * a Promise<T> via Promise.resolve. *)
-          let reason = mk_reason (RCustom "async return") loc in
+          let reason = mk_reason (RType "Promise") loc in
```
goodmind: This is needed to detect promises as returned type from async functions

gkz:
Check out the Flow lint `unused-promise-in-async-scope`

https://beta.flow.org/en/docs/linting/rule-reference/#toc-unused-promise-in-async-scope

Issue #7803, opened 2019-06-07
[declarations] not working as expected


nishant-dani:
Flow version:
"flow-bin": "^0.100.0",

## Expected behavior
I have the following declarations tag in .flowconfig
[declarations]
.*/node_modules/.*
.*/graphql-iso-date/.*
I am expecting to see no flow type errors in the following file:
node_modules/graphql-iso-date/dist/time/index.js.flow

## Actual behavior
Flow throws a lot of errors in that file, disregarding the declarations tag.  I was unable to setup a repro on flow try because there appeared to be no way to setup a .flowconfig file.

* Link to Try-Flow or Github repo:

nishant-dani:
I had to add a line feed after
.*/graphql-iso-date/.*

nishant-dani:
The problem is still occuring.  The line -feed did not help, it only went away temporarily but the problem has surfaced again.

FezVrasta:
I can confirm `[declarations]` seems to be completely broken on 0.100.0

mroch:
@FezVrasta does it work on previous versions?

mroch:
wondering if it could be the same as https://github.com/facebook/flow/issues/6717

FezVrasta:
No it's definitely a different issue. I didn't have time to test a previous version, I'll give it a shot tomorrow

FezVrasta:
@mroch please try this repro out:

```bash
git clone git@github.com:FezVrasta/flow-issue-repro-7803.git
cd flow-issue-repro-7803
yarn --frozen-lockfile
$(yarn bin)/flow
```

You should see errors coming from `@emotion/core` even if the Flow configuration is:

```ini
[declarations]
<PROJECT_ROOT>/node_modules/.*
```

FezVrasta:
@mroch did you ever find time to take a look at my repro?

FezVrasta:
This bug still affects Flow (version 0.108.0).

Does anyone know of any workaround to suppress these errors while maintaining the `[declarations]` behavior?

FezVrasta:
@mroch I just realized this issue is a dupe of #6717 as you suggested, I thought it was different because I didn't see `$FlowFixMe` comments in the Flow output, but actually they are just truncated from the Flow report but are there if you look at the source code that is throwing it.

I guess we can close it and focus on #6717

evbo:
I see this issue still with:

```
npm run flow --version
5.8.0
```

[...]

Issue #6717, opened 2018-08-09
trying to silence third-party errors with [declarations] seems to re-raise ignored errors instead


timri:
I try to use flow in my project and suppress errors from dependencies.
My project uses react-native-firebase, when I run flow I get 3 errors inside node_modules/react-native-firebase.
I wanted to silence them and have extended the .flowconfig with
```
[declarations]
.*/node_modules/react-native-firebase/.*
```

But now I get new errors, which are normally silenced, for example:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ node_modules/react-native-firebase/dist/modules/firestore/DocumentReference.js.flow:54:53

Cannot call CollectionReference with parentPath bound to collectionPath because null [1] is incompatible with Path [2].

     node_modules/react-native-firebase/dist/modules/firestore/DocumentReference.js.flow
     51│   get parent(): CollectionReference {
     52│     const parentPath = this._documentPath.parent();
     53│     // $FlowExpectedError: parentPath can never be null
     54│     return new CollectionReference(this._firestore, parentPath);
     55│   }
     56│
     57│   get path(): string {

     node_modules/react-native-firebase/dist/modules/firestore/CollectionReference.js.flow
 [2] 29│   constructor(firestore: Firestore, collectionPath: Path) {

     node_modules/react-native-firebase/dist/modules/firestore/Path.js.flow
 [1] 36│   parent(): Path | null {

```
The 3 original errors are also still reported.

I am running flow 0.78.0

FezVrasta:
@vjpr `[untyped]` has a different purpose, it will ignore all the types of the matching modules and replace them with `any`

FezVrasta:
@TrySound I got the same error in #7388 and I was using the default `$FlowFixMe`

murrayju:
I'm having the same problem. The desired behavior is to use any type information provided in {{node_modules}} when possible, but to *never* report errors that occur within the {{node_modules}. I only want to see errors that occur in my project's source.

The documentation reads like {{[declarations]}} should do exactly what I want, but it seems to do the opposite. Is there any way to write the configuration now to get the behavior that I want? It seems to me like this should be the default behavior. We should be encouraging projects to publish their flow type annotations in their npm packages, and by default leverage this information to improve typing in a project. But this should not result in seeing error output for a project in external dependencies that can't be fixed.

mroch:
[...]

PR #8105, opened 2019-09-27
Fix [declarations] not suppressing errors correctly & rename to [silence]

### LINKED ISSUE #6631: [declarations] doesn't work with builtins
`[declarations]` works with your own types or types defined in node modules, but does not suppress errors in flow builtins. See https://github.com/facebook/flow/pull/4916#issuecomment-404297827.

Repro in https://github.com/yamov/FlowDeclarations/.

STRML:
# Overview

In #4916, @LegNeato wrote a feature to block reporting of errors in files matching regex patterns as `[silence]` in `.flowconfig`. This sat for about 9 months as an open PR despite strong public support. Just before merging, this was renamed to `[declarations]`. 

Why do we need it? As you can read in #4916's comments, and as I wrote in https://github.com/facebook/flow/pull/4916#issuecomment-382033025, nearly every Flow version introduces breaking changes. A library written for a given Flow version is unlikely to work with any other version. This is especially true of very complex libraries like `react-native` that seem to break on a monthly basis.

Unfortunately, by default, Flow reports errors *internal to a module*, even if these internal errors don't affect any user code. This means that, if a library uses a new feature that is not in the user's current version of Flow, or uses a feature that has been changes (like object types moving to exact-by-default), internal errors will abound. These errors make it difficult to distinguish actual user type errors, and make it impossible to rely on the `flow` binary's exit code for CI.

The current alternatives are `[ignore]`, which causes import errors, and `[untyped]`, which casts to `any`. `[declarations]`has the crucial distinction of *leaving exported typedefs intact while ignoring internal errors*, making it much easier to keep typedefs active while your project and its dependencies ebb and flow across many Flow versions.

Unfortunately, #4916 did not actually implement `[silence]` correctly, and doubly unfortunate was the naming pushed through in the final hour. This PR fixes the implementation. Additionally, it renames `[declarations]` back to `[silence]`. This was not taken lightly but has been done due to the overwhelming majority of feedback preferring `[silence]` and the confusion surrounding `[declarations]` as it relates to "declaration files" (i.e. `.js.flow` files) and "type declarations" generally. This becomes even more confusing when comparing to other `.flowconfig` directives such as `[include]`.

## What Was Wrong

There are two major flaws in the original approach:

1. Silencing errors at the merge site (`Context.remove_all_errors cx`) does not work for builtins,
as those errors are computed later, and
2. Putting that suppression in an if/else where the else block
parsed comment error suppressions can cause `[declarations]` to
actually create *more* errors, which was misleading a lot of people
and masked the source of bugs in this feature.

## Commits

Fixes #6631, #6717, #7803.

https://github.com/facebook/flow/pull/8105/commits/27fa53f61826ccc7dd7bc766022d6161d472ad36 is the first commit, which fixes `[declarations]` as written.

We now suppress entire files in error collation, at the very end
of the error reporting process.

----

The second commit is https://github.com/facebook/flow/pull/8105/commits/a0aa29ebaf6906eed495ed0fbb929760870b0554.

This backs out some of the code
in #4916 related to file error suppression in the merge phase.

This has the benefit of fixing erroneous ignoring of comment
suppressions within those files, which was causing unexpected
consequences as listed in 2) above.

----

The last commit is https://github.com/facebook/flow/pull/8105/commits/a2e11e1800c0f24ee6df6127bd7b78f62e3d8226.

We rename `[declarations]` to `[silence]`
This name has not gone over well: https://github.com/facebook/flow/issues/6631#issuecomment-411285015

and is confusing in multiple places, such as:
https://github.com/facebook/flow/blob/dd93de0a3796897fe07cca8a3bdc621c992a9880/website/en/docs/declarations/index.md
and
https://github.com/facebook/flow/blob/dd93de0a3796897fe07cca8a3bdc621c992a9880/tests/lib_interfaces/.flowconfig#L5-L6

It is difficult to grep for, difficult to distinguish versus type declarations,
and is widely used interchangeably with "interfaces".

For these reasons and many more, it is best to rename this option now that it
is usable.

Pinging @mroch and @mrkev who helped get the original version of this in.

### DIFF
--- a/src/services/inference/init_js.ml
+++ b/src/services/inference/init_js.ml
@@ -64,7 +64,6 @@ let load_lib_files ~sig_cx ~options ~reader files =
          (fun (exclude_syms, results) file ->
            let lib_file = File_key.LibFile file in
            let lint_severities = options.Options.opt_lint_severities in
-           let file_options = Options.file_options options in
            let%lwt result = parse_lib_file ~reader options file in
            Lwt.return
              (match result with
@@ -95,7 +94,6 @@ let load_lib_files ~sig_cx ~options ~reader files =
                    ast
                    ~exclude_syms
                    ~lint_severities
-                   ~file_options:(Some file_options)
                    ~file_sig
                in
                let errors = Context.errors cx in
--- a/src/services/inference/merge_service.ml
+++ b/src/services/inference/merge_service.ml
@@ -125,7 +125,6 @@ let merge_context_generic ~options ~reader ~get_ast_unsafe ~get_file_sig_unsafe 
   let (master_cx, dep_cxs, file_reqs) = reqs_of_component ~reader component required in
   let metadata = Context.metadata_of_options options in
   let lint_severities = Options.lint_severities options in
-  let file_options = Some (Options.file_options options) in
   let strict_mode = Options.strict_mode options in
   let get_aloc_table_unsafe =
     Parsing_heaps.Reader_dispatcher.get_sig_ast_aloc_table_unsafe ~reader
@@ -134,7 +133,6 @@ let merge_context_generic ~options ~reader ~get_ast_unsafe ~get_file_sig_unsafe 
     Merge_js.merge_component
       ~metadata
       ~lint_severities
-      ~file_options
       ~strict_mode
       ~file_sigs
       ~phase
@@ -221,7 +219,6 @@ let merge_contents_context ~reader options file ast info file_sig =
   in
   let metadata = Context.metadata_of_options options in
   let lint_severities = Options.lint_severities options in
-  let file_options = Some (Options.file_options options) in
   let strict_mode = Options.strict_mode options in
   let get_aloc_table_unsafe =
     Parsing_heaps.Reader_dispatcher.get_sig_ast_aloc_table_unsafe ~reader
@@ -230,7 +227,6 @@ let merge_contents_context ~reader options file ast info file_sig =
     Merge_js.merge_component
       ~metadata
       ~lint_severities
-      ~file_options
       ~strict_mode
       ~file_sigs
       ~get_ast_unsafe:(fun _ -> (comments, aloc_ast))
--- a/src/typing/errors/error_suppressions.ml
+++ b/src/typing/errors/error_suppressions.ml
@@ -167,6 +167,11 @@ let in_node_modules ~root ~file_options loc =
   | None -> false
   | Some (file, options) -> Files.is_within_node_modules ~root ~options (File_key.to_string file)
 
+let in_declarations ~file_options loc =
+  match Option.both (Loc.source loc) file_options with
+  | None -> false
+  | Some (file, options) -> Files.is_declaration options (File_key.to_string file)
+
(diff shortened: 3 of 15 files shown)


STRML:
Ping @nodkz @LegNeato

FezVrasta:
Personally I would have **loved** to see two separate PRs, one with the fix, and one with the rename.

I don't want to see the fix blocked because of the time it will take to decide if the team wants to rename the feature or not...

STRML:
I go both ways on that. I see wanting to ship the feature asap, but `[declarations]` is a terrible name for describing what this feature actually does. The feature itself has been unusable since it shipped, so the day it is usable is a good day to rename it.

FezVrasta:
I get your point, but fixing the existing feature doesn't need much discussions around, it's a merge and ship.

Renaming could potentially take years if we take in consideration similar PRs in the history of this project.

So I'd be grateful to be able to use this feature with a inaccurate name, rather than not being able to use it at all.

STRML:
Let's see how they respond. This is why I contained the rename to the last
commit.

STRML:
> could potentially take years

I sure hope not. The team expressed a new commitment toward the community in https://medium.com/flow-type/what-the-flow-team-has-been-up-to-54239c62004f, let's see how that manifests.

mroch:
Thanks for this! The fix looks good to me. Beyond builtins, this should also fix it in cycles, which was broken because we check cycles in the "leader's" context, which might not be in `[declarations]`, causing those errors to still show up. 

I pretty firmly disagree with renaming it back to `[silence]`, though. the current implementation, checking these files just to throw away all the work, is not where we want to be. in [Types First](https://medium.com/flow-type/what-the-flow-team-has-been-up-to-54239c62004f) mode, which will eventually be the default/only mode, we should not even check these files -- just extract its signatures, aka declarations, aka .js.flow files. (I don't think we've made that optimization yet but we should). so, (ideally) it's not silencing anything because there are no errors, because we don't actually check it.

I think we should merge the first 2 commits (again, they look great) and can continue the naming debate in an issue.

mrtnzlml:
@mroch Could you please consider importing it without the last commit? This would be a great DX improvement. Thanks! :)

STRML:
@mroch Thanks for responding - I am more interested in getting this merged than in spending time bikeshedding the name. I am focused on the DX of using Flow, which is why this feature should exist in the first place. To be fair, the name is part of putting Flow users first and addressing their needs (i.e. "this file I don't own is throwing errors, I would like to *silence* them"). But they will get used to whatever feature is put in front of them, if it is genuinely useful.

I will remove the last commit from this PR so it can be merged wholesale.

STRML:
Last commit is removed, merge conflict fixed, assuming build goes green we are clear for landing.

mroch:
here we go!

since this PR is going to get closed and be hard to find, I'd love to continue the naming discussion in an issue.

> "this file I don't own is throwing errors, I would like to silence them"

that's a pretty good point. i think `[declarations]` could include `*/node_modules/*` by default, so hopefully you'd never run into that bad experience.

PR #8120, opened 2019-10-04
libdefs: specify process.env values as non-null (but allow void)

FireyFly:
Node's `process.env` is specified as an object mapping environment variable names to string values.  Since a requested environment variable might not be defined, the libdef has previously been specified as `{ [string]: ?string }`.

However, in practice I believe the value is never `null` (it'd be either a string, or `undefined` if you try to access a nonexistent environment variable). This causes trouble when trying to destructure `process.env`:
```js
const {
  FOO = 'some-default-value',
} = process.env;
// `FOO` has type `?string` rather than `string`
```

This PR updates the libdef to remove the possibility of `null`, so that a default value when destructuring is enough to exhaust all non-`string` possibilities.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2580,7 +2580,7 @@ declare class Process extends events$EventEmitter {
   cwd() : string;
   disconnect? : () => void;
   domain? : domain$Domain;
-  env : { [key: string] : ?string, ... };
+  env : { [key: string] : string | void, ... };
   emitWarning(warning: string | Error): void;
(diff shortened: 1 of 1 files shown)


mroch:
i think you are right based on this: https://github.com/nodejs/node/blob/566479179082154decc2af59dbb64e53f200e970/src/node_env_var.cc#L272-L284

seems like the return value is either explictly undefined if it's a symbol, definitely a string, or not set which is undefined. never null.

PR #8124, opened 2019-10-07
doc(react/components): use $Shape<Props> for defaults

jklmli:
Instead of declaring a type for the default props and composing that inside the component props, suggest using `$Shape<Props>` instead.  This increases readability of the declaration of the composed props.

### DIFF
--- a/website/en/docs/react/components.md
+++ b/website/en/docs/react/components.md
@@ -167,18 +167,11 @@ don't have to add any type annotations to use default props.
 If you would like to add a type annotation to `defaultProps` you can define the
 type as
 ```js
-type DefaultProps = {|
+type DefaultProps: $Shape<Props> = {
   foo: number,
-|}
-```
-and spread that into the `Props` type:
-```js
-type Props = {
-  ...DefaultProps,
-  bar: string,
 }
 ```
-This way you avoid duplicating the properties that happen to have a default value.
+This declares the defaults as a [subset shape](../types/utilities/#toc-shape) of the props, avoiding duplication of the properties that happen to have a default value.
 
 ## Stateless Functional Components <a class="toc" id="toc-stateless-functional-components" href="#toc-stateless-functional-components"></a>
 


panagosg7:
[`$Shape`'s semantics](https://flow.org/en/docs/types/utilities/#toc-shape) is rather unsound. Your suggestion will limit the checking that Flow currently performs. Consider for example
```js
type Props = {|
  x: number;              
  y: number;
|}

class A extends React.Component<Props, {}> {
  static defaultProps: $Shape<Props> = { x: 1 }; 
}
<A/>;
```
Flow would not error here anymore, even though `y` is not provided as a default-prop or passed explicitly. Also in [try-Flow](https://flow.org/try/#0MYewdgzgLgBASgUwIbFgXhgJwQRwK4CW2AFAOTYpSkCUA3AFD1QCeADgjAAqYisQwYA3gB96MGAA8AXDDB4AtgCMEmWuPUbNYmMxlylKhsIC+jYABskEfgEEYCCVARgAJv0SUAdAGEQ81uDOUAA83LwQADQwgsYAfNHa0EhQBMAwLggAZkh45lBhfDIAJADKABZI7KE8fPFCkjIAjDDGavSm9ME2APSxat3dMABi5iAA7ukgCPxgILAqPJgwZSocCABuzjBQZSB4AOZlMAAGzMcwBDNzMKw86wQZLjBWz+lZOXkAtLe8MCBLrCsEAQTwcrHMqQIUHMzBgjDMlmsMAAQvZHM43PBkKgfH4AmAgtVwlEYvFBIkoMlUm9srl8jV+PVpDBmq04R1gsjev1BiNxjBQJhsKgYQK8ZYCJBnoo9rBTudZrBlJL9jcgSCgA).

panagosg7:
If `$Shape` was equivalent to making all properties optional, then the above code would be equivalent to
```
class A extends React.Component<Props, {}> {
  static defaultProps: {|x?: number; y?: number|} = { x: 1 }; 
}
<A/>; // error
```
which now uncovers errors at the construction of `<A/>`.

PR #8135, opened 2019-10-16
[PR] Implement versioned .js.flow

nmote:
Summary:

Proof of concept for https://github.com/facebook/flow/issues/4917
<s>Has bad performance on all resolutions</s>

## Performance checks

Current measurements: https://gist.github.com/goodmind/65695c3b7a3aa0766812b15779a984dc

1. Build flow binary
2. Run `flow check --profile` on codebase
3. Record results
4. Repeat with regular flow binary and this PR, properly stopping all servers before running new binary
Pull Request resolved: https://github.com/facebook/flow/pull/7842

Differential Revision: D16431326


nmote:
Rebased copy of #7842. See that PR for more info.

src/services/inference/module/module_js.ml
```diff
@@ -93,6 +93,69 @@ let module_name_candidates ~options =
       in
       List.rev (name :: List.fold_left map_name [] mappers))
 
+let types_versions_package_candidates reader =
```
nmote: Looks like this function returns a pair of `(version, path)` but only the path is used by the caller. Could you change it to return only the path?

src/services/inference/module/module_js.ml
```diff
@@ -93,6 +93,69 @@ let module_name_candidates ~options =
...
+      let version_path =
+        SMap.find_first_opt
+          (fun k ->
+            try Semver.satisfies ~include_prereleases:true k Flow_version.version
+            with Semver.Parse_error _ -> false)
+          types_versions
+      in
+      version_path)
+
+let types_versions_candidates ~reader dir rel_path dirname =
```
nmote: This function and the associated code in `node_paths_part.ml` really make me nervous. I get worried whenever I see code that manipulates paths directly. There are a lot of opportunities to make mistakes, both for the original author and future maintainers.

I think it would help if I understood the motivation for this code better. What is happening here that couldn't be satisfied by reusing existing code? Fundamentally, it seems to me like we are looking in the `package.json` already for `main` to figure out what to resolve the module to, why can't we augment that logic to look at the `typesVersions` fields as well?

I have some more specific comments about this code, but I'd like to understand the context better before I dig deeper.

goodmind: Is it possible to write unit tests for this? I didn't found any runner in repository, but it has some OUnit tests

nmote: We do have some unit tests, but they are only set up to run with Buck so we haven't open sourced them. @mroch, do you know what would be involved in setting up our unit tests to run with `make`?

nmote:
Closing, since we'll actually merge the original copy

PR #8148, opened 2019-10-21
libdefs: Add missing methods to Blob

lyleunderwood:
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Blob#Methods)

Some of these are marked experimental, but it's unclear to me whether or not we care about that.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -17,6 +17,9 @@ declare class Blob {
   type: string;
   close(): void;
   slice(start?: number, end?: number, contentType?: string): Blob;
+  arrayBuffer(): Promise<ArrayBuffer>;
+  text(): Promise<string>,
+  stream(): ReadableStream,
 }
 
 declare class FileReader extends EventTarget {


lib/dom.js
```diff
@@ -17,6 +17,9 @@ declare class Blob {
   type: string;
   close(): void;
   slice(start?: number, end?: number, contentType?: string): Blob;
+  arrayBuffer(): Promise<Arraybuffer>;
```
nmote: Looks like capitalization is wrong -- it should be `ArrayBuffer` with a capital `B`, right?

lyleunderwood:
@nmote I have not tested it, I couldn't figure out where tests for `Blob` methods should actually go. I was thinking they would make sense in `tests/lib/dom` based on the file they're in, but that doesn't exist. I'm unclear on what `tests/lib/lib.exp` is, but it seems to have expected output for just a couple tests. Maybe tests should go in `tests/bom`? Though I didn't find anything super relevant there. Do tests for `Blob` methods exist anywhere currently? If not, do you have a recommendation for where I should put them?

nmote:
I'd be satisfied to hear that you built Flow and tested it manually. Some automated tests would be nice, but we don't test all of our libdefs and frankly they aren't really that error prone, so I don't think perfect test coverage is necessary.

lyleunderwood:
I haven't setup this environment for building flow yet, and now that I'm trying I'm getting errors building master :sob:

nmote:
Don't worry about it, I've at least tested that this doesn't break other things.

lyleunderwood:
Well you'll be glad to hear I finally got my build working :rofl:

PR #8163, opened 2019-10-28
allow => in objects in return types of arrow functions

mroch:
Summary:
To avoid ambiguity, Flow disallows using a function type in the return type, like

  const f = (): boolean => string => { return (b: boolean): string => "hello"; }

unless you wrap it in parens

  const f = (): (boolean => string) => { return (b: boolean): string => "hello"; }

We forgot to allow it inside objects too, like

  const f = (): { x: boolean => string } => { return { x: (b: boolean): string => "hello" }; }

Differential Revision: D18187492


thorn0:
Not a 'function type in object type' situation, but similar:
```js
const f = (): ([boolean => string]) => { return [(b: boolean): string => "hello"]; }
```
This code breaks if I remove the parens from the return type. Is this expected? If not, is this the same bug or a different one?

Issue #8167, opened 2019-10-29
Return type of `window.getComputedStyle` is incorrect.


YevhenKap:
Flow version: 0.110.1

## Missing/Incorrect APIs
I see in `lib/dom.js` (line 3966) such type definition:
```flow
declare function getComputedStyle(elt: Element, pseudoElt?: string): any;
```
Function has incorrect return type *any*. 
[MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) page describes this function to have return `CSSStyleDeclaration` object of a given *elt*. Since *elt* isn't optional parameter, `getComputedStyle` must always returns `CSSStyleDeclaration` object.

## Relevant documentation
[window.getComputedStyle](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) and [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration).

YevhenKap:
I close the issue due to merging pull request that solve it.

PR #8168, opened 2019-10-29
Change return type of window.getComputedStyle(...)

YevhenKap:
I have changed return type of `window.getComputedStyle(...)` from `any` to `CSSStyleDeclaration` according to information from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle) page.
Related [issue](https://github.com/facebook/flow/issues/8167).

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -3963,7 +3963,7 @@ declare function alert(message?: any): void;
 declare function prompt(message?: any, value?: any): string;
 declare function close(): void;
 declare function confirm(message?: string): boolean;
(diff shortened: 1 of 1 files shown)


YevhenKap:
May I ask what tests were failed? Because I reviewed them and didn't find test that use `requestAnimationFrame` function.

nmote:
I think this was based on a revision where the autocomplete tests were broken in master. I rebased the internal diff and everything passed there. I wouldn't worry about it.

PR #8191, opened 2019-11-17
Add missing label to List.sort compare function

marsam:
Otherwise the build fails with:
```
…
ocamlfind: [WARNING] Package `threads': Linking problems may arise because of the missing -thread or -vmthread switch
File "hack/heap/sharedMem.ml", line 1198, characters 21-55:
Error: This expression should not be a function, the expected type is
'a Hh_core.List.t
Command exited with code 2.
Compilation unsuccessful after building 181 targets (0 cached) in 00:00:44.
make: *** [Makefile:291: build-flow] Error 10
```

### DIFF
--- a/src/heap/sharedMem.ml
+++ b/src/heap/sharedMem.ml
@@ -1004,7 +1004,7 @@ end)
         end
         cache;
       Hashtbl.clear cache;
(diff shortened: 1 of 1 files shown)


mroch:
picked this up as part of https://github.com/facebook/flow/pull/8216, thank you for the fix!

PR #8197, opened 2019-11-21
[hack] Support src/oxidized/regen.sh in dune builds

fredemmott:
Summary:
- set `DUNE_BUILD_DIR` instead of using `--build-dir` in dune.sh as this
  interfers with some commands, such as `dune exec`
- add wrapper script to replace `buck run`
  - doesn't actually use `dune exec` as `dune exec` doesn't give us
    enough control over the working directory
- release signed-source; the tokens aren't secret; we have hack and
  rust versions public already
- add `dune` file for `hh_oxidize`
  - changed .ml file name to match executable as dune seems to want this
- improved working-dir/fbcode-root detection in regen.sh
- update generated code

Reviewed By: vsiles

Differential Revision: D18519816

### DIFF
--- a/hack/utils/signed_source.ml
+++ b/hack/utils/signed_source.ml
@@ -1,0 +1,68 @@
+(*
+ * Copyright (c) 2018-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *)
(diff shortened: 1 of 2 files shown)


fredemmott:
Whoops, this was meant to be an HHVM PR, not flow. Closing, not useful here.

Issue #8190, opened 2019-11-17
Function Generic parameter is resolved to void


moroine:
Flow version: v0.102.0

## Expected behavior

Function Generic parameter is resolved to void

- [flow try](https://flow.org/try/#0GYVwdgxgLglg9mABMMAeAKgPgBQEMBci6AlIgN4BQiATgKZQjVK4DcFAvhRQPTeIDKACzggANgBNEtatTjUKKbMRZA)
```typescript
function fn<T>(a: T) {
 return a;
}

// Should error
fn();
```

- [flow try](https://flow.org/try/#0GYVwdgxgLglg9mABMMAKAhgLkQfgM5QBOMYA5gJSIDeAUIoQKZQiFLoDcNAvjTQPR9EAZQAWcEABsAJogaFCcQjRSpy7IA)
```typescript
function fn(a: ?string) {
 return a;
}

// Should error
fn();
```

- [flow try](https://flow.org/try/#0GYVwdgxgLglg9mABMMAeAKgPgBQEMD8AXIgCQByCAsrgJ4BGApujQA4MaYCUiA3gFCIATgyghBSXAG4+AXz58A9AsQAxOABt1cAO4wwAc0QBnKLigMAtgzBRjACzgh1AE0QNBguIMR5EMI0IMAI4gMMLOnHwo2JzSfEA)
```typescript
function fn<T>(a?: $NonMaybeType<T>) {
 return a;
}

// Should error
fn();
```

- [flow try](https://flow.org/try/#0CYUwxgNghgTiAEAzArgOzAFwJYHtXwAcYcCBlKbAZ0SxEoB4AVAGngGkA+ACgCh54weYAC54XAB6iAJAFEIIALYhUGRgE8CIJq04BKeAF4O8AEY4c8qKmZ9CxAqLY3+OaQDk8AWShqTIdZpMHDy6omYWIFYA3Dw8gqiUGPBQ8MLhlvgGdiTkVDR0XBKiqMgKfjD6RvDi8MYADKwA5OKNrADe1aIAjPAAvroxPAD0Q-CkABY4yBDA8CAwxDBxeImmqemRmdlkFFjUtJSFkvAlZfOVxjX1TS0DsUA)
```typescript
declare function propSatisfies<T, K>(
  cond: (x: $ElementType<T, K>) => boolean,
  prop: K,
  o: $NonMaybeType<T>
): boolean;

const a :boolean = propSatisfies((x: number) => x > 0, 'x', { x: 1 });

[...]

PR #8201, opened 2019-11-24
[FIX] [Typing function] Make function params required by default

moroine:
**Fix:**
- #8190

**Changes:**

Function params are assumed required by default ~> will cause new errors, which can be fixed using optional `?`

I think this change will improve library definitions using function overloading with generic params.

Currently, the only way to achieve this is using rest params:

```typescript
function fn(...args: [string]) {}
```

but this is too restrictive, as it can't be passed to as a callback

```typescript
function fn1(...r: [string]): void {}
function fn2(a: string): void {}

function callFn(f: (string, number) => void) {
  f('hello', 2); 
}

// Expected error
callFn(fn1);

// No error
callFn(fn2);
```


moroine:
@jbrown215 what's your opinion about making function min-arity strict?

jbrown215:
Discussed in discord, mirroring here:

I think the crux of the issue here is that you only want to be able to omit a function argument if it is optional:
```
function f(a?: number) {}
f(); // Ok
```

But not if it accepts `undefined`-- in that case you want to explicitly force people to pass the undefined:

```
function g(a: ?number) {}
g(); // Error
g(undefined); // Ok
```

Without thinking too hard about it, this seems like a desirable change and how optional function arguments should work in the first place. I also think they should _not_ accept undefined, but that can be addressed later.

Will need to discuss with some other people on our side and come to consensus before taking a look at the code changes.

moroine:
@jbrown215 any update?

jbrown215:
Will need to import and see how hard this will be to roll out before I can give a solid answer.

moroine:
@jbrown215 I've added a codemod for explicit optional parameters. I should handle all typed functions. Only missing resolving imports & inference because it's not using FlowJS.

jbrown215:
Cross posting here for visibility:
After importing and rebasing, this PR crashes on our repos internally. This could be for many reasons, and I cannot prioritize this over my other work.

Issue #560, opened 2015-06-24
ES2015 & 2016 Features meta-issue


samwgoldman:
Feel free to open an issue for any unchecked feature that doesn't already have one, but please try to keep discussion in a single issue if possible. Use cases (and test cases) are very welcome!

To report a bug with a completed feature, please open a new issue for the bug.
### ES2015+
- [x] let/const declarations (#784)
- [x] Regular expression u flag (#2510)
- [x] Regular expression y flag
- [x] Binary literals (#566)
- [x] New octal literals (#566)
- [x] Arrow functions
- [x] Default parameters (mostly supported, except using `this`: #1234)
- [x] Rest parameters
- [x] Spread operator
- [x] super references
- [x] Classes
- [x] for-of loops
- [x] Generators (#534)
- [x] Object literal property shorthand
- [x] Object literal method shorthand
- [ ] Object literal computed properties (#252)
- [x] Template strings
- [x] Tagged templates
- [ ] Tagged template fn `strings` param has a `raw` prop
- [ ] Tagged template fn `strings` and `raw` arrays are frozen
- [x] Destructuring
- [x] Default values in destructuring (#183)
- [ ] Unicode code point escapes
- [x] Allow duplicate object literal properties (except **proto**)
- [x] Modules
- [x] [`new.target`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new.target) (#1152)
- [ ] Symbols (#810, #1015)
- [ ] `export Foo from 'foo'` (#2403)
- [x] object spread
- [x] async/await
- [x] async iterators (#1824)
- [ ] [Shared memory and atomics](https://github.com/tc39/ecmascript_sharedmem) (#7811)
- [ ] `String.prototype.matchAll` (#7812)
- [ ] `Promise.allSettled`
- [ ] [Lifting template literal restriction](https://github.com/tc39/proposal-template-literal-revision)

#### Stage 4
- [ ] [`String.prototype.replaceAll`](https://github.com/tc39/proposal-string-replaceall)

#### Stage 3
- [ ] [optional chaining](https://github.com/tc39/proposal-optional-chaining) (#4303)
- [ ] [`BigInt`](https://github.com/tc39/proposal-bigint) (#7602)
- [ ] [`globalThis`](https://github.com/tc39/proposal-global) (#7704)
- [ ] [`import.meta`](https://github.com/tc39/proposal-import-meta) (#6913)
[...]

PR #8230, opened 2019-12-14
Add type signature for Promise.allSettled()

shau-kote:
I miss the type signature for the new (ES2020) static method of the built-in Promise so I've decided to add it.

Documentation:
  - [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled);
  - [Specification](https://tc39.es/ecma262/#sec-promise.allsettled);

The change closes one of many items in https://github.com/facebook/flow/issues/560


lib/core.js
```diff
@@ -688,12 +688,16 @@ declare class Promise<+R> {
...
     static all<T: Iterable<mixed>>(promises: T): Promise<$TupleMap<T, typeof $await>>;
+    static allSettled<T: Iterable<mixed>>(promises: T): Promise<$TupleMap<T, typeof $settlement>>;
     static race<T, Elem: Promise<T> | T>(promises: Iterable<Elem>): Promise<T>;
 }
 
 // we use this signature when typing await expressions
 declare function $await<T>(p: Promise<T> | T): T;
 
+declare function $settlement<T>(p: Promise<T> | T): SettlementResult<T>;
+declare type SettlementResult<T> = { status: 'fulfilled', value: T } | { status: 'rejected', reason: any };
```
nmote: I have a few requests for this line:
* These types ought to be exact -- use `{|` and `|}` instead of `{` and `}` (this will be the default soon anyway).
* This will be more usable if the type parameter is covariant, so write `SettlementResult<+T>` instead of `SettlementResult<T>`.
  * This will also force you to make at least the `value` property covariant by adding a `+` in front of it. And you might as well just make the other fields covariant too, while you're at it.
* I'd rather see this use `mixed` for the rejection reason instead of `any`

nmote:
cc @moroine

lib/core.js
shau-kote: Thank you for the substantive feedback! I've made a changes on all the items.

shau-kote:
@nmote Thanks for the hint about tests. I am not sure I get the tests idea right, but hope I wrote them in the right way. :)

nmote:
Thanks for updating this! Just today, somebody else who wanted to use `Promise.allSettled` put together a similar change, and it looks like their change will make it in first. I'll comment here when it lands. It wouldn't hurt to have some additional tests, though. If you are interested I'd be happy to work with you to get your tests in as well.

nmote:
https://github.com/facebook/flow/commit/efc59d3ce8441bfd3f437a3e39bfa34ee0622dd6

Leaving this open in case you'd like to add your tests

shau-kote:
Well, I am always glad to help by all means. :)
 
Am I getting you right, you are suggesting me to resolve the conflict between the two implementations (in favour of the already merged one) and add my tests to the existing ones? It seems reasonable. 

Which way is preferred -- merging the `master` to my branch or rebasing my branch to the current `master`?

SamChou19815:
Closing this since it hasn't been updated for a long time.

PR #8237, opened 2019-12-17
feat: add rudimentary Array.flat

nnmrts:
ref: https://github.com/facebook/flow/pull/6948

according to https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/flat


nnmrts:
@nmote

PR #8245, opened 2019-12-26
[lib] Add brotli functions and class to NodeJS zlib

fizker:
New constants are only added in the zlib.constants object, since the NodeJS documentation states that they are deprecated in the zlib object

Also updated constants to match those available in node v12.14.0. Note that some constants were removed. It turns out that they are mentioned in the NodeJS docs, but not in the code as per current master (v 12.14.0 and https://github.com/nodejs/node/blob/5289f80/lib/zlib.js).

Official documentation: https://nodejs.org/dist/latest-v12.x/docs/api/zlib.html

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2321,15 +2321,26 @@ type zlib$brotliOptions = {
   ...
 };
 
+type zlib$brotliSyncFn = (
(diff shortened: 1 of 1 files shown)


SamChou19815:
This is already added in https://github.com/facebook/flow/commit/8fc3a08b357985361dcf9fc53d059f9f58cf460d

PR #8253, opened 2020-01-09
Add optional cb for node dgram socket.close function

davidnaas:
This optional callback seems to always have been part of the node dgram api and was probably just overlooked

https://nodejs.org/api/dgram.html#dgram_socket_close_callback


mroch:
thanks!

PR #8260, opened 2020-01-12
(Updated) Flowlint no floating promises

SebastienGllmt:
@goodmind made a PR back in Sept to add a no-floating-promises lint (https://github.com/facebook/flow/pull/8100). Unfortunately the PR now has some merge conflicts so I tried to update the PR to get it working on the latest mater branch.


SebastienGllmt:
One thing I just noticed is that probably we should figure out how to disable this for top-level async statements since you can't await on them

joshuanapoli:
> One thing I just noticed is that probably we should figure out how to disable this for top-level async statements since you can't await on them

Is there any way to condition the warning on the top-level?

PR #8273, opened 2020-01-27
Improve definitions for Node url.parse function

chicoxyzzy:
TypeScript does it right https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/url.d.ts#L20-L33

In Node every return value is monomorphic by design where it's possible.
Example of `parse` output in the worst case.
```bash
➜ node -p "require('url').parse('')"
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: ''
}
```
`parse` throws an error when is called with invalid argument types, so there is no way for it to return `undefined` inside of it's return object.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -2150,18 +2150,18 @@ declare module "url" {
     parseQueryString?: boolean,
     slashesDenoteHost?: boolean
   ): {
-    protocol?: string,
-    slashes?: boolean,
-    auth?: string,
-    host?: string,
-    port?: string,
-    hostname?: string,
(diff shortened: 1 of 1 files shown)


mroch:
this change makes `url.format(url.parse(...))` incompatible. I think you need to add `| null` to `url$UrlObject` (but keep all of those properties optional).

PR #8290, opened 2020-02-07
Fix definition of "listen" of net$Server and its subclasses

Rob--W:
The `listen` signature of `net$Server` did not match the definition of
the actual API at https://nodejs.org/api/net.html#net_server_listen

In particular, flow failed to detect invalid handles and options due to
the use of an object with all-optional properties, and the use of
`Function` (which is an alias for `any`).

This patch fixes the following issues:

- Remove repeated method declarations of `net$Server` from its subclasses, and add missing `listen` methods to `net$Server`.
- Callbacks are no longer typed `Function` (which is equivalent to `any`), but `() => mixed`.
- Rename the `listen`'s `handle` parameter to `options`.
- Require the `options` object to either contain `path` or `port`.
- The `listen` method with parameter `handle` has a type definition.

Fixed the following issues with `node_tests`:

- All existing `node_tests` that were marked with "These should fail" are now failing as expected.
- `handle` renamed to `options`.
- `fd` renamed to `path` (string values are `path`, not `fd`).
- Add new passing tests for `listen(handle, ...)`.
- Treat empty objects as a failure (because port or path is required).

The number of expected `node_tests` failures increased from 115 to 131.
12 of those are tests that were marked as "failing" but did not.
The other 4 are signatures that were marked as "passing" but should have
failed (`fd` with string value instead of number, and empty object).

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1399,25 +1399,6 @@ declare class http$ServerResponse extends stream$Writable {
 }
 
 declare class http$Server extends net$Server {
-  listen(port?: number, hostname?: string, backlog?: number, callback?: Function): this;
-  // The following signatures are added to allow omitting intermediate arguments
-  listen(port?: number, backlog?: number, callback?: Function): this;
-  listen(port?: number, hostname?: string, callback?: Function): this;
-  listen(port?: number, callback?: Function): this;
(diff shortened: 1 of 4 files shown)


SamChou19815:

PR #8303, opened 2020-02-19
Allow `export * as`, as it is in tc39 stage 4

penx:
"export ns from" is now in stage 4 so shouldn't need a warning.

https://github.com/tc39/proposal-export-ns-from


mroch:
this flag is deprecated and enabled by default now! 🎉

PR #8328, opened 2020-03-18
[PR] [docs] Eliminate duplicated string literals

thSoft:
According to the DRY principle.

### DIFF
--- a/website/en/docs/react/redux.md
+++ b/website/en/docs/react/redux.md
@@ -95,8 +95,11 @@ you'll want to split up your `Action` disjoint union into separate action
 types.
 
 ```js
-type FooAction = { type: "FOO", foo: number };
-type BarAction = { type: "BAR", bar: boolean };
+const FOO: 'FOO' = 'FOO';
+const BAR: 'BAR' = 'BAR';
+
+type FooAction = { type: typeof FOO, foo: number };
+type BarAction = { type: typeof BAR, bar: boolean };
 
 type Action =
   | FooAction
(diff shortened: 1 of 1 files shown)


thSoft:
Please ignore this PR, the used syntax is invalid, just the Flow server was not running.

jamesisaac:
How is this DRY?  You're now writing out each action type at least 3 extra times.

The way it is in the docs currently, because the action types are string literal values, Flow can ensure they're being used correctly across the entire codebase.  Importing a string literal constant, with its value having the same name as its identifier, doesn't add anything..?

jamesisaac:
This is one of the big wins of Flow+Redux IMO, that it cuts down on the boilerplate of needing constants and imports.  The approach of importing constants is useful if you have no type system, but otherwise redundant.

thSoft:
If you want to change the value of an action type (e.g. from 'BAR' to 'BAZ'):

* When using string literals everywhere, you have to change its occurrences (i.e. the type definition, the action creator code and every reducer which handles the action - at least 3 changes).
* When using constants, you have to change the identifier, the type and the value of the constant (always 3 changes), which are anyways near each other in the code.

Of course, if you don't have an IDE with Rename refactoring, the latter approach means more work, so it depends. Should I move the constant-based alternative version to a separate place (e.g. StackOverflow Q & A) and only add a link to it here?

jamesisaac:
Fair enough, I suppose I can see situations where an IDE constant rename would be more convenient than doing a project-wide find and replace on the string literal (if the same string literal is used elsewhere too... not a problem I've ever run into after many Redux projects).  But IMO that small convenience wouldn't be worth the extra verbosity of needing imports, especially on a new project that's just following a tutorial to get started with Flow.

I don't have any official say over the Flow docs, so will leave that question to someone on the Flow team.  Just wanted to share my preference for continuing to advise string literals.

nmote:
I appreciate the effort you've put in here but I agree with @jamesisaac that this change just adds complexity in this case. There are certainly some benefits to doing things the way you've described, but for the sake of documentation simplicity I think it's better as-is.

PR #8336, opened 2020-04-01
[dom.js] add ErrorEvent

kegluneq:
Summary: There is no `ErrorEvent` declaration in dom.js. This diff adds `ErrorEvent`.

MDN documentation: https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent
whatwg documentation: https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface
w3.org documentation: https://www.w3.org/TR/html52/webappapis.html#the-errorevent-interface

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -508,6 +508,25 @@ declare class AnimationEvent extends Event {
   ) => void;
 }
 
+// https://html.spec.whatwg.org/multipage/webappapis.html#the-errorevent-interface
+type ErrorEvent$Init = Event$Init & {
+  message?: string,
+  filename?: string,
+  lineno?: number,
+  colno?: number,
+  error?: Object | null,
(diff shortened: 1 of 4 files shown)


kegluneq:
It looks like there are test failures, but that these are due to offset changes in the expected test outputs.  I did some searching around but couldn't find the process for updating the expected files, if there is one. Do I just apply the diffs that get generated when I run the tests to the relevant .exp files?

Thanks!

w01fgang:
@kedromelon I couldn't find the way to update that snapshot, and I just removed old snapshot and copied raw output from the terminal after running the test on dom.

kegluneq:
I ended up using `patch` with the generated .diff files to update after manually verifying that the changes were all line numbers

PR #8363, opened 2020-05-03
Support .cjs extension

TrySound:
Node.js recently introduced .cjs extension for better dual mode support.
https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_enabling

Some packages like nanoevents already adopted it. Though flow requires
to specify it in .flowconfig
https://unpkg.com/browse/nanoevents@5.1.5/

### DIFF
--- a/packages/flow-remove-types/flow-remove-types
+++ b/packages/flow-remove-types/flow-remove-types
@@ -79,7 +79,7 @@ function mkdirp(dirpath) {
 
 // Collect arguments
 var ignore = /node_modules/;
-var extensions = [ '.js', '.mjs', '.jsx', '.flow', '.es6' ];
+var extensions = [ '.js', '.mjs', '.cjs', '.jsx', '.flow', '.es6' ];
 var outDir;
 var outFile;
 var all;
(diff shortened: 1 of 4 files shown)


mroch:
last time we added an extension it caused us some perf issues because of node's search algorithm. it adds a lot of `stat()` syscalls to see if `node_modules/foo.cjs`, `node_modules/foo/index.cjs`, etc, exists as it walks up the directory tree. ideally we would cache the directory tree in memory and not need to do that...

since our internal codebases are huge, i'll run some tests to see. for small projects it won't matter, so I think we probably should merge this and then disable it on huge repos that don't use it like Facebook's.

it'd also be great if we enforced that .mjs used `import` and .cjs used `require`, but obviously that can be future work.

TrySound:
Hi @mroch Was this change included in 0.124 release? I still get an error
```
Cannot resolve module .../node_modules/nanoevents/index.cjs.

     1│ // @flow
     2│
     3│ import * as React from 'react';
     4│ import type { Emitter } from 'nanoevents';
```

PR #8376, opened 2020-05-15
Support WebP File Types for Images

brett-puddick:
WebP is a now widely support image file extension, and it being utilised by many
companies.

This commit will add Flow support for the `.webp` file extensions, removing the
need to use `$FlowFixMe` in files using both Flow type checking and WebP images.

[WebP Support](https://caniuse.com/#feat=webp)


nmote:
Thanks, I'm going to close this in favor of #8453 though because it also adds tests.

PR #8394, opened 2020-06-03
Add exitCode to ChildProcess

isker:
https://nodejs.org/api/child_process.html#child_process_subprocess_exitcode

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -267,6 +267,7 @@ declare class child_process$ChildProcess extends events$EventEmitter {
   connected: boolean;
   killed: boolean;
   pid: number;
+  exitCode: number | null;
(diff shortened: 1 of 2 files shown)


lib/node.js
```diff
@@ -267,6 +267,7 @@ declare class child_process$ChildProcess extends events$EventEmitter {
   connected: boolean;
   killed: boolean;
   pid: number;
+  exitCode: number?;
```
mroch, line 270: i think the `?` is a syntax error. `exitCode: number | null` instead

isker:
This is what I get for typing things into the github editor and praying.  Will fix.

isker:
Should be good now @mroch

PR #8396, opened 2020-06-06
[easy 1 line change] Make exact the return type for Object.fromEntries

sebholl:
According to the ECMAScript spec, `Object.fromEntries` asserts:

> obj is an extensible ordinary object with no own properties.

As such, it seems appropriate that the object returned only contains key value pairs of this type, and no additional (unexpected) keys.

https://tc39.es/ecma262/#sec-object.fromentries

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -58,7 +58,7 @@ declare class Object {
         '0': K,
         '1': V,
         ...
-    }>): { [K]: V, ... };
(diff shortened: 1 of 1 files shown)


jbrown215:
This does not have a semantic difference. Objects with indexers aren't really treated as "exact" or "inexact" -- they are just objects with indexers.

PR #7602, opened 2019-03-31
BigInt typechecking

goodmind:
Fixes #6639 
Library definitions #7534 (BigInt builtin and typed arrays)

Probably signature verification should be rewritten too, but I don't know what it does.
There maybe missing cases in something like refinements, assignment operators (especially since they seem untyped)

Adds `sketchy-null-bigint` lint rule and supports `bigint` type in `sketchy-number` with different lint message

PR #7534, opened 2019-03-07
BigInt types

goodmind:
Follow up to #7471
(Currently these types are useless because `bigint` resolves to `any`)

#### Proposal
https://github.com/tc39/proposal-bigint
#### Docs
https://tc39.github.io/proposal-bigint/#sec-bigint-objects
https://tc39.github.io/proposal-bigint/#sec-typedarray-objects
https://tc39.github.io/proposal-bigint/#sec-dataview.prototype.getbigint64
https://tc39.github.io/proposal-bigint/#sec-dataview.prototype.getbituint64
https://tc39.github.io/proposal-bigint/#sec-dataview.prototype.setbigint64
https://tc39.github.io/proposal-bigint/#sec-dataview.prototype.setbiguint64

PR #8410, opened 2020-06-23
Make `offset` an optional argument for the sized Buffer read/writes

isker:
https://nodejs.org/docs/latest/api/buffer.html#buffer_buf_readbigint64be_offset

As far as I can tell this has always been optional.

Also @mroch, what is flow's policy for targeting versions of node?  `noAssert` was removed in node 10 (see the same linked docs).  If flow targets a particular version, maybe it should be removed.

Also!  `readBigInt64BE` and `*BigInt*` pals are missing here.  Ditto the same question as above: should they be added?


mroch:
@isker: our policy for targeting node is very... not good. I'd like the node definitions to move to flow-typed so they can be versioned. it'd keep them from interfering if you're working on a web project. and it'd also be great for making PRs like this not have to wait for a flow release.

the main problem then is that you don't get out-of-the-box support for node, but given how inaccurate it is now, I think it's right to trade off a bit of (arguable) usability for correctness.

it's similar but a bit better defined for React... we try to target the latest public release of React, but that can be a problem if you don't keep them both up to date together. moving react to flow-typed has similar benefits.

isker:
Got it.  Thanks for taking the time to spell that out.

For what it’s worth, the ergonomics on flow-typed aren’t great either.  It’s got the versioning problem addressed but the inability to specify dependencies among libdefs kills many nontrivial libs in the water.

That certainly wouldn’t be a problem for node or react, though 🙂.

I will open an additional MR that adds the missing APIs later.

isker:
> I will open an additional MR that adds the missing APIs later.

Nevermind.  These all return `BigInt`, support for which has been stuck in a series of doomed PRs.  🙃 

#7534 
#7602

PR #8428, opened 2020-07-04
[DOCS] remove suppress_comment option from docs

Naturalclar:
removed `suppress_comment` options from the option docs, as it is no longer supported from flow-bin v0.127.0


mroch:
Thanks for the PR! In https://github.com/facebook/flow/commit/42de9df5cf38d12b6fc51cb6a50f97b17e053828 we marked it as deprecated instead of removing it, for people still on old versions.

PR #8438, opened 2020-07-14
add aborted to http$IncomingMessage

petersolopov:
add aborted to http$IncomingMessage

https://nodejs.org/api/http.html#http_message_aborted


petersolopov:
@mroch sure. I've added some more missing definitions to `http$IncomingMessage` and update `.exp` file.

I took a definition for `destroy()` from the [LTS version](https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_message_destroy_error). Although [In current](https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_message_destroy_error) version method returns `this`.

PR #8453, opened 2020-08-03
Add support for `.webp` resources

TomasBarry:
This commit follows the work done in 6984935003906e52a72e406763748c97589780d6

The commit adds to the `module_resource_exts` an extension of `.webp`.
[`.webp`](https://developers.google.com/speed/webp) images are used in
some projects as they provide better compression in the Chromium based
browsers.

Relevant tests have also been added.


TomasBarry:
I have had to use `$FlowFixMe` comments quite extensively to allow for `.webp` assets to be used in my projects which has been a bit of a pain since I make use of ESLint rules to disallow `$FlowFixMe` comments meaning I think have to disable that ESLint rule each time I want to use a `.webp` asset which ends up looking like:

```es6
// eslint-disable-next-line flowtype/no-flow-fix-me-comments
// $FlowFixMe
import SomeWebPAsset from 'someAsset.webp'
```

Brianzchen:
I think this change is great as `webp` becomes more standard, but I did want to mention you could have used module name mapper to resolve importing `.webp` files.

https://flow.org/en/docs/config/options/#toc-module-name-mapper-extension-string-string
```
module.name_mapper.extension='webp' -> '<PROJECT_ROOT>/flow-mocks/empty-string.js'
```

Assuming you've setup your mock which I did in this sample repo
https://github.com/Brianzchen/webp-test/blob/master/flow-mocks/empty-string.js

TomasBarry:
> I think this change is great as `webp` becomes more standard, but I did want to mention you could have used module name mapper to resolve importing `.webp` files.
> 
> [flow.org/en/docs/config/options/#toc-module-name-mapper-extension-string-string](https://flow.org/en/docs/config/options/#toc-module-name-mapper-extension-string-string)
> 
> ```
> module.name_mapper.extension='webp' -> '<PROJECT_ROOT>/flow-mocks/empty-string.js'
> ```
> 
> Assuming you've setup your mock which I did in this sample repo
> [Brianzchen/webp-test@`master`/flow-mocks/empty-string.js](https://github.com/Brianzchen/webp-test/blob/master/flow-mocks/empty-string.js?rgh-link-date=2020-08-04T00%3A07%3A36Z)

That's definitely an option and something I'll probably go with if this feature is rejected. However, I would also like to do the same as I've done here for `.elm` files but I felt that was a separate discussion to `.webp` so I wanted to do them one at a time (particularly in case this `.webp` feature was accepted but the `.elm` one would be rejected for example).

TomasBarry:
@Brianzchen, any update on this one?

Brianzchen:
Feels like maybe this is parser related? These would be the people that can review this based on https://github.com/facebook/flow/blob/master/CONTRIBUTING.md
@mroch
@Hans-Halverson
@gkz

TomasBarry:
@nmote - thanks for merging this one. Before I add one for `.elm` files, is this something you would like? I am cognisant not to spam too many file extensions.

TomasBarry:
> I'm not familiar with it, but if it's a widespread file format I expect we'll merge it (eventually 😬).
> 
> Ideally it seems to me that this list of extensions should be a set of reasonable defaults, but that they should be extensible through `.flowconfig` options.

Fully agree that it would be nice to have it extendable through `.flowconfig`.

PR #8464, opened 2020-08-25
[DOCS] Barrel file FAQ

### LINKED ISSUE #8231: Flow and index.js imports
This is not a bug but something I suggest including in the documentation for flow as I can see others running into a similar problem.

If you do not add // @flow to index.js files you have that are re-exporting your components, the type checking will not work on any imports from those index.js files.

Of course this makes perfect sense once I knew what the cause was. it was losing the types since it importing from a file that was not set up for flow (even though the source file was). It is very easy to forget those index.js files even exist though, and it caused some serious frustration and almost made me give up on flow before I found out what was happening. 

I highly suggest mentioning this in the documentation to save others from that headache and wasted time.

Brianzchen:
Closes #8231 

@nmote

### DIFF
--- a/website/en/docs/faq/index.md
+++ b/website/en/docs/faq/index.md
@@ -321,3 +321,24 @@ const GenericButton = ({type, children}) => <button type={type} onClick={() => c
 
 const Button: React.ComponentType<{children: React.Node}> = setType(GenericButton);
 ```
+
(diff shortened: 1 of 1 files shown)


Brianzchen:
Hi @mroch, @nmote said you had review changes for me here?

gkz:
I feel like this isn't a common enough issue to put in the FAQ - in general if you haven't enabled Flow in a file it isn't going to type-check, I don't think we need to have this additional specific example

PR #8474, opened 2020-08-30
[PR] Return union void for some string functions

### LINKED ISSUE #8284: charAt, charCodeAt and codePointAt are weird
Flow version: 0.117.1

## Expected behavior

```js
const string = "ABC";

string.charAt();
string.charCodeAt();
string.codePointAt();
// ^ no errors

const maybeANumber = string.codePointAt(100);

Math.round(maybeANumber);
// ^ errors because maybeANumber could be undefined

if (typeof maybeANumber === "number") {
	Math.round(maybeANumber);
	// ^ should be fine now because we checked it this time
}

```
## Actual behavior
```js
const string = "ABC";

string.charAt();
string.charCodeAt();
string.codePointAt();
// ^ errors "Cannot call `string.[...]` because  function [1] requires another argument."

const maybeANumber = string.codePointAt(100);

Math.round(maybeANumber);
// ^ doesn't error but maybeANumber could be undefined

```

[here](https://flow.org/try/#0MYewdgzgLgBNBOBLMBzGBeGAiAggIQGEsBuAKFIWRQDpgALAQ3hygAoBKMy1Wx+AkABMApiw5coSHqBEAFEMihjOpAPSqYAPRjD48EPAjYCDMGBCxgDADbWYAA240A2tTcBdezABGwqwFcIYRgYADN-MGAoRHAYZwBGdxh4YQBHf0QUo1MLOl0YJhR-AFthMChqLHJQSFhihgBPXxwAORLfeAw4SSpaIWF5RTF4gAYRlVIAWQYoOmp9CMFWeqbRNuKOlXUtGEEQYQgwAHJYXX1O7386xub1jphQf2tBH2DF4VDkYUEgA)

## Explanation

In my opinion, the definitions of these methods are kind of wrong, because

1. they don't actually require an argument; passing no argument means `NaN` which means 0 in this context
2. `codePointAt` could return `undefined` when `pos` is greater than the string's length ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#Return_value))
3. there's no reason for the `charAt` parameter being called `pos` instead of `index` while the others are not ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charAt#Parameters))
	- MDN only calls the `codePointAt` parameter `pos` ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#Parameters))
	- TC39 calls all of them `pos`, but also "index" in their explanations ([ref](https://tc39.es/ecma262/#sec-properties-of-the-string-prototype-object))
	- we developers probably would all call them `index`

*Point 1* is debatable though, because ECMAScript actually specifies this:
**String.prototype.charCodeAt ( *pos* )**
and not
**String.prototype.charCodeAt ( [ *pos* ] )**
which they normally do for actually optional parameters, but in this case, `pos` will always get converted to `NaN` if it's undefined anyways. And passing `NaN` to those methods always returns the first index.
BUT, for example the flow definition of the global function `parseInt` is technically also not following ECMAScript. Flow says that `parseInt` accepts `mixed` as string, but it doesn't. And it says that `radix` is optional, but it actually isn't. Well, it won't throw, but almost every function of core JavaScript won't throw because of type-juggling and intelligent implementation. I get why Flow allows `mixed` here, otherwise we would have a pretty hard time checking our strings.
Same goes for `isNaN` for example. Would be weird if Flow errors on `isNaN("test")` just because we wanted to check that value, right?
In my opinion, `string.charCodeAt()` to get the first index should be allowed. It's a common pattern.

*Point 2* should be pretty clear.

*Point 3* is not **that** important but I think we should decide on one parameter name for all three methods here.

So my proposal is that this:
```js
	// ...
	charAt(pos: number): string;
    codePointAt(index: number): number;
	// ...
```
should be this:
```js
	// ...
	charAt(index?: number): string;
    charCodeAt(index?: number): number;
    codePointAt(index?: number): number | undefined;
	// ...
```
or maybe just this:
```js
	// ...
	charAt(index: number): string;
    codePointAt(index: number): number | undefined;
	// ...
```

Thoughts?

Brianzchen:
Closes #8284 

`codePointAt` and `''[123]` has the potential to return `undefined`. This change will enforce users to put guards around these use cases to provide safer string manipulation.


Brianzchen:
For my index test I had to write it as `(new String()[123]: string);` because `(''[123]: string)` wasn't throwing an error. Not sure if I'm missing something here because the previous test that uses string literal works fine.

nnmrts:
@Brianzchen 
Looks good, also thanks for the parameter name changes.

I don't really care about flow anymore these days though, sorry.

nnmrts (approved):

Brianzchen:
I've removed the index of check to make this PR a bit easier to get in. Flow team has already expressed that is an area of unsoundness that they are prepared to allow for an easier dev experience.

@nnmrts I understand, hope things get better and you feel interested again soon 😄

nnmrts (commented):
Reviewed the last commits but I don't know if this does anything or is even relevant anymore.

Brianzchen:
I think so, the definition if it hasn't been corrected is in fact wrong. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/codePointAt#return_value

It's documented as 
> If index is out of the range of 0 – str.length - 1, codePointAt() returns [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined).

SamChou19815:
https://github.com/facebook/flow/commit/ec8a209b46a4635b6a51bdf148a341980fdf3900 would resolve one of the issue. We don't plan to touch the other one, because TS is also being unsound there, and it would be too disruptive.

PR #8482, opened 2020-09-08
(Updated again) Flowlint no floating promises

SebastienGllmt:
This is yet another rebase of #8260 
There were a lot of nasty merge conflicts, so I just created a new PR instead since I don't think I updated everything correctly.

In practice, this PR seems to work well on our codebase (>100K of Javascript code)

TODO:
- [ ] solve `type_at_pos_type_destructors.diff `
- [ ] solve `async.diff `
- [ ] option to disable the lint for top-level async (?)
- [ ] add documentation

### DIFF
--- a/src/common/errors/error_codes.ml
+++ b/src/common/errors/error_codes.ml
@@ -131,6 +131,7 @@ type error_code =
   | UnclearAddition
   | UnclearType
   | UninitializedInstanceProperty
+  | NoFloatingPromises
   | UnnecessaryInvariant
   | UnnecessaryOptionalChain
   | UnreachableCode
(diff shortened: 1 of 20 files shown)


jeresig:
This is very exciting - we were talking about wanting this exact linter to use at Khan Academy. It would definitely help us to catch a lot of tricky bugs that would otherwise be nearly impossible to find. We likely won't be able to offer technical support but here is a huge 👍 to seeing this PR and hopefully it getting it merged in someday soon!

inactivist:
What can we do to help move this forward? I think it would be a helpful addition to Flow.

gkz:
Check out the Flow lint `unused-promise-in-async-scope`

https://beta.flow.org/en/docs/linting/rule-reference/#toc-unused-promise-in-async-scope

PR #8496, opened 2020-10-01
Fixed some vulnerability and upgrading the bootstrap to new version for fix a type of attack

MKokeshi:

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.0",
   "license": "CC-BY-4.0",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


Brianzchen:
Hi @MKokeshi, there was a previous attempt at this change [here](https://github.com/facebook/flow/pull/7285). Do you know if simply upgrading it actually works?

mroch:
yeah, there are definitely changes required to upgrade to 4.0.0. i'm not sure #7285 even gets them all but didn't have confidence that change was fully tested...

PR #8497, opened 2020-10-02
[DOCS] changelog: mark `untyped-type-import` as breaking

wchargin:
Summary:
This causes new errors for my project, which is understandable since
that’s the point of the change, but historically these things have been
called out at the top of the changelog to make upgrading easier.

wchargin-branch: changelog-untyped-type-import-breaking


wchargin:
Thanks!

PR #8504, opened 2020-10-08
[DOCS]: Fix broken urls caused by typo

saintmalik:
### Description 
(The link in the docs is leading to 404 pages, so I replaced it with the correct one.)

### DIFF
--- a/website/en/docs/config/declarations.md
+++ b/website/en/docs/config/declarations.md
@@ -16,7 +16,7 @@ etc are extracted and used by the typechecker when checking other code.
 
 Conceptually one can think of declaration mode as if Flow still typechecks the
 files but acts as if there is a comment that matches
-[`suppress_comment`](https://flow.org/en/docs/config/options/#toc-suppress-comment-regex) on every line.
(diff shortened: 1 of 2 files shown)


website/en/docs/config/declarations.md
```diff
@@ -16,7 +16,7 @@ etc are extracted and used by the typechecker when checking other code.
 
 Conceptually one can think of declaration mode as if Flow still typechecks the
 files but acts as if there is a comment that matches
-[`suppress_comment`](options#toc-suppress-comment-regex") on every line.
+[`suppress_comment`](https://flow.org/en/docs/config/options/#toc-suppress-comment-regex) on every line.
```
mroch, line 19: if we ever do translate flow.org, we don't want to hardcode `/en/`. I think `../options/` should work instead.

mroch (changes requested):
good catches, thanks! one minor fix first

website/en/docs/config/declarations.md
saintmalik: Okay on it

mroch (approved):

Issue #8478, opened 2020-09-02
The search functionality is mostly useless


Brianzchen:
I'm trying to jump to the flow options `exact_by_default` quickly but typing in the search gives me random results. I'd have to just type it in google or navigate manually through the docs section.

Attached screenshot of results I get.

<img width="1440" alt="Screen Shot 2020-09-03 at 9 51 31 am" src="https://user-images.githubusercontent.com/12436524/92048453-14b87b80-edcb-11ea-853c-43ce43fd2a98.png">

SamChou19815:
Since the new website is released, it should be usable now.

PR #8514, opened 2020-10-27
[DOCS] document react.runtime option

FezVrasta:

### DIFF
--- a/website/en/docs/react/index.md
+++ b/website/en/docs/react/index.md
@@ -22,3 +22,15 @@ follow [this guide](../tools/babel/).
 Babel also
 [works out of the box with Create React App](../tools/create-react-app/),
 just install Flow and create a `.flowconfig`.
+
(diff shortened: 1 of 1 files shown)


Brianzchen:
Should we add it here too?
https://flow.org/en/docs/config/options/

jbrown215:
@Brianzchen It's already documented there: https://flow.org/en/docs/config/options/#toc-react-runtime-automatic-classic

I'm ok with duplicating it on the React page for visibility.

FezVrasta:
Using the search functionality on the documentation and searching for `jsx` and `runtime` returns no results

Brianzchen:
> Using the search functionality on the documentation and searching for `jsx` or `runtime` returns no results

@FezVrasta I've also found it quite unusable https://github.com/facebook/flow/issues/8478

PR #8522, opened 2020-11-02
[lsp] consider Options.all for typeCoverage command

mochja:
Enables type coverage trough LSP in non pragma workspaces with `all=true` option. Previously the files were ignored and type coverage was only working on pragma files.

https://github.com/flowtype/flow-for-vscode/issues/138#issuecomment-720559892


mochja:
@mroch would this be something that could be potentially merged? I have read the original commit (https://github.com/facebook/flow/commit/d0b46ea63c9dd032f8b86b6540171ddc5a4f4232), but I think it should not be necessary to include \@flow pragma if `all` option is set to `true`.

With `force` set to true (in case of all option present), the type coverage works nicely as in the cli.

PR #8525, opened 2020-11-05
Method predicates

### LINKED ISSUE #7920: Supports %checks modifier on class methods
## Proposal

Allow the use of the `%checks` modifier on class methods.

## Use case
A big example is the lodash `flow-typed` file. Function such as `isNil` are defined like this:

```js
declare class Lodash {
    isNil(value: any): boolean
}
```

New feature would allow the following:

```js
declare class Lodash {
    isNil(value: any): boolean %checks (value != null);
}
```

Right now, the use of this method does not allows flow to know that the checked value is not null:

```js
const test: ?React$Element<any> = window.document.getElementById('test')
if (!isNil(test)) {
   return ReactDOM.createPortal(test, this.containerEl) // Throw flow error about test being undefined
}
```

mischnic:
Rebase of #7735

Closes #7735, Closes #7920

### DIFF
--- a/src/typing/errors/flow_error.ml
+++ b/src/typing/errors/flow_error.ml
@@ -1026,6 +1026,7 @@ let rec make_error_printable ?(speculation = false) (error : Loc.t t) : Loc.t Er
     | (Some loc, Error_message.Normal { features }) ->
       mk_error ~trace_infos ~kind loc (code_of_error error) features
     | (None, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
+    | (Some _, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
     | (None, PropMissing { loc; prop; reason_obj; use_op; suggestion }) ->
       mk_prop_missing_error loc prop reason_obj use_op suggestion
     | ( None,
--- a/src/typing/errors/flow_error.ml
+++ b/src/typing/errors/flow_error.ml
@@ -1026,7 +1026,6 @@ let rec make_error_printable ?(speculation = false) (error : Loc.t t) : Loc.t Er
     | (Some loc, Error_message.Normal { features }) ->
       mk_error ~trace_infos ~kind loc (code_of_error error) features
     | (None, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
-    | (Some _, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
     | (None, PropMissing { loc; prop; reason_obj; use_op; suggestion }) ->
       mk_prop_missing_error loc prop reason_obj use_op suggestion
     | ( None,
--- a/tests/predicates_declared/class-predicate.js
+++ b/tests/predicates_declared/class-predicate.js
@@ -1,0 +1,13 @@
+// @flow
+
+class Hello {
+  test(x: mixed): boolean %checks {
+    return typeof x === "string";
+  }
+}
+
+const m = new Hello();
+const foo: string | number = "";
+if (m.test(foo)) {
+  (foo: string);
+}
--- a/tests/predicates_declared/class-predicate.js
+++ b/tests/predicates_declared/class-predicate.js
@@ -9,5 +9,23 @@ class Hello {
 const m = new Hello();
 const foo: string | number = "";
 if (m.test(foo)) {
-  (foo: string);
+  (foo: string); // okay
+}
+
+class D1 extends Hello {
+  test(x: mixed): boolean %checks { // okay
+    return typeof x === "string";
+  }
+}
+
+class D2 extends Hello {
+  test(x: mixed): boolean %checks { // error incompatible predicates
+    return typeof x === "number";
+  }
+}
+
+class D3 extends Hello {
+  test(x: mixed): boolean { // error non-predicate method
+    return typeof x === "string";
+  }
 }
--- a/tests/predicates_declared/predicates_declared.exp
+++ b/tests/predicates_declared/predicates_declared.exp
@@ -12,6 +12,23 @@ References:
                ^^^^^^^^^^^ [2]
 
 
+Error ------------------------------------------------------------------------------------------ namespace-import.js:8:4
+
+Cannot cast `a` to number because null or undefined [1] is incompatible with number [2]. [incompatible-cast]
+
+   namespace-import.js:8:4
+   8|   (a: number); // error
+         ^
+
+References:
+   namespace-import.js:4:8
+   4| let a: ?number = 1;
+             ^^^^^^^ [1]
+   namespace-import.js:8:7
+   8|   (a: number); // error
+            ^^^^^^ [2]
+
+
 Error --------------------------------------------------------------------------------------- sanity-conditional.js:5:50
 
 Unexpected token `=`, expected the token `)`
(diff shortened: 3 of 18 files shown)


tests/predicates_declared/class-predicate.js
```diff
@@ -0,0 +1,13 @@
...
+  test(x: mixed): boolean %checks {
+    return typeof x === "string";
+  }
+}
+
+const m = new Hello();
+const foo: string | number = "";
+if (m.test(foo)) {
+  (foo: string);
+}
```
panagosg7, line 13: I'm trying out a couple more cases here
```js
class D1 extends Hello {
  test(x: mixed): boolean %checks { // okay
    return typeof x === "string";
  }
}

class D2 extends Hello {
  test(x: mixed): boolean %checks { // error incompatible predicates
    return typeof x === "number";
  }
}

class D3 extends Hello {
  test(x: mixed): boolean { // error non-predicate method
    return typeof x === "string";
  }
}
```
The errors look reasonable, but the second error message is not really informative:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ tests/predicates_method/class-predicate.js:23:5

Cannot extend C [1] with D2 because predicate encoded in return [2] is incompatible with predicate encoded in return [3]
in property test. [incompatible-extend]

 [3]  5│     return typeof x === "string";
       :
     20│
 [1] 21│ class D2 extends C {
     22│   test(x: mixed): boolean %checks { // error incompatible predicates
 [2] 23│     return typeof x === "number";
     24│   }
     25│ }
     26│


Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ tests/predicates_method/class-predicate.js:28:3

function [1]. Function is incompatible with function [2]. [function-predicate]

 [2]  4│   test(x: mixed): boolean %checks {
      5│     return typeof x === "string";
      6│   }
       :
     25│ }
     26│
     27│ class D3 extends C {
 [1] 28│   test(x: mixed): boolean { // error non-predicate method
     29│     return typeof x === "string";
     30│   }
     31│ }
     32│
```
Any chance we could get some better error messaging here? Perhaps by adding an use_op for the offending constraint.

panagosg7:
Thanks @mischnic ! This is very promising! 

One request for a better error message. Let me know if you find trouble with that.

mischnic:
That you for the quick answer, I've tried to do something in that direction, but there are a few problems still:

tests/predicates_declared/predicates_declared.exp
```diff
@@ -1,3 +1,76 @@
...
+References:
+   class-predicate.js:27:18
+   27| class D3 extends Hello {
+                        ^^^^^ [1]
+   class-predicate.js:4:3
+         v--------------------------------
+    4|   test(x: mixed): boolean %checks {
+    5|     return typeof x === "string";
+    6|   }
+         ^ [3]
```
mischnic, line 40: The new error message is still a bit odd (all of those periods)

src/typing/errors/flow_error.ml
```diff
@@ -1021,6 +1021,7 @@ let rec make_error_printable (error : Loc.t t) : Loc.t Errors.printable_error =
     | (Some loc, Error_message.Normal { features }) ->
       mk_error ~trace_infos ~kind loc (code_of_error error) features
     | (None, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
+    | (Some _, UseOp { loc; features; use_op }) -> mk_use_op_error loc use_op features
```
mischnic, line 1024: This is probably wrong.
I returned `UseOp` in error_message.ml for `EFunPredCustom`, but that other loc in the tuple is apparently still there...

tests/predicates_declared/predicates_declared.exp
panagosg7: The error is created in
https://github.com/facebook/flow/blob/7c7e60697a099209e4aff191b2ee8c68739104a0/src/typing/flow_js.ml#L4323-L4326
and is formatted in
https://github.com/facebook/flow/blob/7c7e60697a099209e4aff191b2ee8c68739104a0/src/typing/errors/error_message.ml#L2142-L2143

mischnic:
I think I've got it working properly now. There were also some existing test cases where the error message (and code: `function-predicate` to `incompatible-cast`) changed.

panagosg7:
There's one more thing missing here: support for exported classes with predicate methods. Eg. consider the case where class `Hello` is exported and you use `if (m.test(foo)) {}` in some other file.

In Flow's [types-first](https://medium.com/flow-type/types-first-a-scalable-new-architecture-for-flow-3d8c7ba1d4eb) mode, we need to be able to extract signatures from files that contain all important information to be used in an importing file. Currently, since methods can't have predicates this information is missing:
https://github.com/facebook/flow/blob/7c7e60697a099209e4aff191b2ee8c68739104a0/src/parser_utils/signature_builder_generate.ml#L1583

So, to add full support for this feature you'd have to include this information in the generated signature. There's actually an even newer version of these signature that would have to be updated too (`src/parser_utils/type_sig`), but we can take it one step at a time.

Feel free to give this a try, otherwise, I'll try to make some time to look into this.

mischnic:
-> 3456d1d

Should a test be added for that (if yes, not sure how and in which test folder)?

> There's actually an even newer version of these signature that would have to be updated too (src/parser_utils/type_sig), but we can take it one step at a time.

As far as I can tell, the class types use `obj_value_prop` which contain a `fun_sig` which already has a `predicate` value?

panagosg7:
> Should a test be added for that (if yes, not sure how to test this and in which folder that should go)?

You could add relevant tests in https://github.com/facebook/flow/blob/master/src/typing/__tests__/signature_generator_test.ml

> As far as I can tell, the class types use obj_value_prop which contain a fun_sig which already has a predicate value?

The type definition may have an entry, but it's not populated

https://github.com/facebook/flow/blob/5baf69342c995c5eaf38467c9b60fd0186fd483b/src/parser_utils/type_sig/type_sig_parse.ml#L1227

mischnic:
> but it's not populated

I had a look in there but I don't see where I can get the predicate from in that function. Looks like you'll have to implement this...

SamChou19815:
We now support type guards, which works on methods: https://flow.org/try/#1N4Igxg9gdgZglgcxALlAJwKYEMwBcD6aArlLnALYYrgA2WAzvXGCADQgYAeOBARgJ74AJhhhYiNXClzEM7DFCLl602e0hQhcMtCw18ufgAcqyMTXpyQ5LEZNCDx0+cvzOJtBQW49AOkMm+AhEWGhCKsgyRBgAvuoQItRgdIwABABiEBCpwAA6UKmplLgAFgkAFLyhyEVwnBhCAJQ1VWipcPSp9DJwUAg5+YWFmLhEaAUBGBAwqa2pALyLqQDk3Z59ywDcg6kx+Tu4GN2V1bX1TQMFQ3Az5aUdvsVlQidojY2XQ0OvNWu9CI1tldCnsrqCYmwQAA3DBoJjQahQgAMvgATABGFGokAxIA

Issue #6592, opened 2018-07-11
Can't run flow in cmd when VS Code is open in Windows


edele:
The only difference is `C` and `c` drive letters:
```
flow is running on a different directory.
server_root: c:\Users\edele\flowtest, client_root: C:\Users\edele\flowtest
```

But I think paths are the same because Windows paths are case-insensitive.

`server_root` was created by Flow Language Support for VS Code. And the error is thrown whenever I run `yarn run flow` in cmd.exe to see full list of errors in a project.

Looks like the issue originates in vscode.TextDocument.uri.fsPath which "normalizes" drive letter to lowercase according to documentation: https://code.visualstudio.com/docs/extensionAPI/vscode-api#Uri

Still thinking on a possible solution.

1. Should drive letter be forced to uppercase in Flow Language Support plugin because flow-bin uses uppercase?
2. Should flow-bin change the way it decides if `server_root` and `client_root` are the same just for Windows?
3. Am I missing something?

BTW. Just look at Microsoft nowadays. It's modern vibe feels so good. We have to consider Windows as a viable platform now.

mrkev:
After some quick skimming I didn't find where the uppercase `C:\` comes from, but I found where the comparison is done:

https://github.com/facebook/flow/blob/347281f31dd87719e48ea58bfb3f01c7051b44f1/src/server/commandHandler.ml#L29

I guess the paths could be normalized to lowercase before comparing on windows? Not 100% sure it wouldn't have negative effects, but feel free to PR. For reference this PR does a check to discern platform: https://github.com/facebook/flow/pull/6573

codespitzer:
Has a solution/workaround been discovered for this issue?

SergioMorchon:
Any news on this?

ArmorDarks:
> I guess the paths could be normalized to lowercase before comparing on windows? Not 100% sure it wouldn't have negative effects

that should be fine since on Windows paths aren't case sensitive, as it happens to be on Linux.

edele:
You can also run `flow check` instead as a workaround.

Diggsey:
Same issue here - windows 10.

ArmorDarks:
Any chances to give that issue priority? That would so much improve dev experience on Windows.

The fix should be quite straightforward, as Windows paths are case insensitive, which means that on win platform all paths can be simply lowercased before comparison.

It's just the OCaml thing that stops from making a PR... Too stupid, sorry 😥

edele:
[...]

PR #8530, opened 2020-11-11
don't normalize Windows drive letter in tool tests

mroch:
Summary: `tool test` uses `vscode-uri` to normalize paths, but that by default lowercases the drive letter, so we can't test what happens if you start Flow with an uppercase drive letter.

Differential Revision: D24893422


mroch:
this repro's #6592. with this change, the test runner starts the server with an uppercase `C:\`, and then makes LSP requests using `vscode-uri` which normalizes it to `c:\`.

closing this PR as it is squashed into the fix, #8531

PR #8532, opened 2020-11-13
FS: add 'withFileTypes' option to 'readdir' and 'readdirSync'

mrtnzlml:
- fs.readdir: https://nodejs.org/api/fs.html#fs_fs_readdir_path_options_callback
- fs.readdirSync: https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options
- fs.Dirent: https://nodejs.org/api/fs.html#fs_class_fs_dirent

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -925,6 +925,18 @@ declare module "fs" {
     bytesWritten: number;
   }
 
+  declare class Dirent {
+    name: string | Buffer;
+
+    isBlockDevice(): boolean;
+    isCharacterDevice(): boolean;
+    isDirectory(): boolean;
+    isFIFO(): boolean;
+    isFile(): boolean;
+    isSocket(): boolean;
+    isSymbolicLink(): boolean;
+  }
+
   declare function rename(oldPath: string, newPath: string, callback?: (err: ?ErrnoError) => void): void;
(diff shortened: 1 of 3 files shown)


lib/node.js
```diff
@@ -973,17 +985,26 @@ declare module "fs" {
   declare function mkdtempSync(prefix: string): string;
   declare function readdir(
     path: string,
-    options: string | { encoding: string, ... },
+    options: string | { encoding?: string, withFileTypes?: false, ... },
     callback: (err: ?ErrnoError, files: Array<string>) => void
   ): void;
+  declare function readdir(
+    path: string,
+    options: { encoding?: string, withFileTypes?: true, ... },
```
mroch, line 993: does `withFileTypes: true` need to be non-optional here? if i understand correctly, it's required to get an array of `Dirent`

mrtnzlml: Thanks for the review @mroch! It makes sense what you are saying - I updated the PR. Could you please check again?

mrtnzlml:
Hi @mroch! 👋 I noticed that in master landed the version _before_ your code review (`withFileTypes?: true` instead of your suggested `withFileTypes: true`). Was there something wrong with it? 🤔

PR #8499, opened 2020-10-03
[DOCS] Added Image to the README.md

san45:

PR #8446, opened 2020-07-23
feat(docs): Update README.md

iamdarshshah:
### Changes
- [x] Added Flow logo (image source: `https://opensource.facebook.com/projects`)
- [x] Aligned header elements to the center

> This will make it look a lot cleaner. It's good to include a logo in the docs, thus added it to the top.

Issue #8424, opened 2020-07-01
The Readme should provide a reason why I should use Flow


guettli:
I would like to know which benefits Flow has.

Please state this to potential users in the README.

Maybe add a comparison why Flow is better than TypeScript.

dylanirlbeck:
I'd see the documentation on [Flow's website](flow.org) about why and how they built Flow: https://flow.org/en/docs/lang/.

In general, I don't think READMEs for big projects need to make such an argument -- normally that is better done on a website, wiki, or external blog posts, for example.

Brianzchen:
Flow focuses on soundness
TypeScript focuses on completeness

Which one you prefer is up to you, but personally I use it because it helps me write more predictable and robust applications. Being 100% sound means that I'm very confident that I don't get runtime errors that can crash the application. Though it means I need to write a lot more checks to validate types. It's up to you whether this is a pro or con for you.

PR #8533, opened 2020-11-14
[DOCS] Improve readme introduction

Brianzchen:
There's been a few previous attempts to improve the readme though I don't think either hit the mark quite right (#8499 and #8446). This change adds the flow logo in the correct style, centers the logo and introductions.

I've also added a reference to soundness to give new comers the right impression when people want to evaluate what makes flow stand apart from other static typecheckers. Which I think also answers and should satisfy #8424.

## Before
Screen Shot 2020-11-14 at 12 12 54 pm

## After
Screen Shot 2020-11-14 at 12 10 22 pm


gkz:
I think I am satisfied with the current README.md, I think it's easier to read if you're opening it up as a text file. If you're on the web, you can check out the website.

PR #8540, opened 2020-11-20
libdefs: add typings for node's built-in Module module

gorakong:
Derived from the documentation and source code of Module:

- [Documentation](https://nodejs.org/api/module.html#module_modules_module_api)
- [Source Code](https://github.com/nodejs/node/blob/cbfa2d11d4e42ea7f6a847a6930fd1022773252e/lib/internal/modules/cjs/loader.js/)


Brianzchen:
Hi @gorakong. Can you please also update the test files. To do this, you need to setup and run the test suite and then update the `exp` with the `out` file that's generated.

lib/node.js
```diff
@@ -1654,6 +1654,68 @@ type net$connectOptions = {
   ...
 };
 
+declare module "module" {
+  declare type FilePath = string;
```
wbinnssmith, line 1658: I don't believe this is done anywhere else in this definition, e.g. the fs typings. Maybe just `string`?

gorakong:
hey @Brianzchen, i've updated the test files 🙂

Brianzchen:
@gorakong your changes have updated `enums.exp`, `get_def_enums.exp`, `node_tests.exp`. You'll need to update them locally then push new `exp` files

gorakong:
@Brianzchen gotcha. added!

SamChou19815:

PR #8561, opened 2020-12-28
Make runtests about 2x faster, by using parallelism continuously

gnprice:
The test suite tries to run up to N tests in parallel (with N=16 by
default).  But in practice it spends a lot of time with fewer tests
running, and as a result it gets much less of the benefit of
parallelism than it could.

The root of the issue is that we only start test T when all the tests
up to T-N finish -- so if e.g. test 100 is slow (and N=16), then
after some or all of the next few tests 101, 102, 103, ..., 115 have
finished, we'll still be waiting for test 100 to finish before
starting on test 116 or those after it.

This PR fixes the issue by using `xargs -P` to run N tests in parallel
continuously, starting a new task immediately after each one finishes.
The result is that on my desktop with 6 physical cores (12 logical),
the tests run in 127s instead of 285s, about 2.2x faster.

This PR preserves the nice features of the output: test results are
printed in order, and a line at the bottom shows the next test we're
currently waiting for.  I've tested that it works on macOS (with its
different `xargs` and very old `bash`) as well as on Linux.

The PR is split into two commits for easier review:
 * One commit is large but simple: it just moves a bunch of
   code into a new file, verbatim, plus makes minimal changes
   elsewhere to keep everything working.  This splits the "run one
   test" logic into its own script, to prepare for the next commit.
 * The second commit is much smaller, and makes the interesting
   changes to how we manage the parallelism.

The speedup is greatest with plenty of cores, but significant even on
a less-powerful machine.  On an old Mac Mini with a dual-core CPU (so
4 logical cores), with FLOW_RUNTESTS_PARALLELISM=6, the tests run
in 395s instead of 598s, about 1.5x faster.


gnprice:
> The PR is split into two commits for easier review:

I'd also be happy to split these into two PRs, if that'd be easier for your workflow. (If I were using Phabricator here, I'd probably make this a stack of two diffs.)

panagosg7:
I was wondering how the unit tests suddenly got so much faster. Thanks for the awesome perf win @gnprice !

PR #8567, opened 2021-01-13
[lib] Add missing composedPath function to Event

Brianzchen:
Docs: https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath


ebonow:
This seems like a big missing and hoping this can be merged soon.
[MDN Documentation](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath)

Perhaps also worth mentioning this has been in TypeScript for years.
https://github.com/microsoft/TypeScript/blob/19e7a89bbbbc25db1e0bf5d30420b509768bffe3/lib/lib.dom.d.ts#L5391-L5394

PR #8586, opened 2021-02-15
Add dns.promises typedef to node.js

STRML:
As of Node 10, [DNS has a `dns.promises` API](https://nodejs.org/api/dns.html#dns_dns_promises_api). 

Only tricky thing about this is that `dns.promises.lookup()` resolves an object `{address: string, family: number}`.

Also added missing methods (a few were added in v10) and removed `any` types where possible.

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -772,6 +772,54 @@ declare module "dns" {
     ...
   };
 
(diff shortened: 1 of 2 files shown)


SamChou19815:

Issue #8590, opened 2021-02-16
Missing Properties from PerformanceResourceTiming


MIGreenberg:
## Missing/Incorrect APIs
The following properties are missing from the Flow type definition for PerformanceResourceTiming:
* `decodedBodySize`
* `encodedBodySize`
* `nextHopProtocol`
* `transferSize`
* `workerStart`

https://github.com/facebook/flow/blob/v0.111.3/lib/bom.js#L411

## Relevant documentation
https://www.w3.org/TR/resource-timing-2/#sec-performanceresourcetiming

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/transferSize

MIGreenberg:
Fixed with https://github.com/facebook/flow/pull/8591

PR #8591, opened 2021-02-18
[lib] Adds Missing Attributes to PerformanceResourceTiming

MIGreenberg:
Summary:

Problem:

Flow's exported definition of PerformanceResourceTiming is missing some attributes. PerformanceResourceTiming's attributes are specified in the Resource Timing spec.
https://www.w3.org/TR/resource-timing-2/#sec-performanceresourcetiming
https://github.com/facebook/flow/issues/8590

Solution:

Add the missing attributes to the PerformanceResourceTiming class definition.

Test Plan:

`make test`


MIGreenberg:
I signed the individual Facebook CLA, but this change is motivated by some work I'm doing for Twitter Inc.

Issue #8538, opened 2020-11-19
Crash on Apple Silicon


STRML:
Flow version: 0.138

## Expected behavior

Flow should run as normal under Rosetta 2.

## Actual behavior

```
$ flow check
Out of shared memory:
Raised by primitive operation at file "src/heap/sharedMem.ml", line 55, characters 25-49
```

This may have to do with a quirk in Rosetta 2 / shared memory. I am not familiar enough with the codebase to know for sure, but would be happy to investigate if given some pointers.

Building a proper aarch64 version is stalled in https://github.com/facebook/flow/pull/8418/files which seems to simply need a review. If that can be done, we can get this into flow_bin and perhaps the problem will be solved.

*Edit:*

[Fix from @atulsmadhugiri](https://github.com/facebook/flow/issues/8538#issuecomment-755540541):

```
[options]
sharedmemory.heap_size=3221225472
```

alexgurr:
This is a **huge** blocker for our team, running new macs. Would be great to see this resolved.

DMXL:
Same here running Flow for VSCode, this is the log
```
usage: sudo dmesg
Dec 11 13:15:00.778 [error] Flow server (pid 45797) exited with code Out_of_shared_memory (15)
Dec 11 13:15:00.778 [info] Creating a new Flow server
Dec 11 13:15:00.782 [info] Spawned server #109 (pid=45801)
Dec 11 13:15:00.831 [error] Connection 'server #109' was closed from the other side
Unable to obtain kernel buffer: Operation not permitted
```

gkz:
We're looking into getting Macs with Apple silicon for the team (they are not yet standard issue for us).

STRML:
If anyone is committed to fixing this, I'll happily ship them one. Email me
- see my contact info at STRML.net.

Jasperrr91:
Same here, impossible to get this working
`Out of shared memory:
Raised by primitive operation at file "src/heap/sharedMem.ml", line 43, characters 6-40`

BotaniQue:
Same, really waiting for an update...
The project is unsupportable on M1 Macs

STRML:
Bounty is still claimable...

chapati23:
There's no way to get around this? e.g. For homebrew it's possible to run commands unser rosetta with `arch -x86_64 /bin/bash -c "yarn flow"` — which works for most of my CLIs that aren't yet natively M1-compatible

...except for flow 😞

chapati23:
[...]

PR #8616, opened 2021-03-15
Update ocaml to 4.10.2 for darwin-arm64 compatibility

radex:
These updates (inspired by a patch found posted by @maxsalven here: https://github.com/facebook/flow/issues/8538#issuecomment-756004548) are required to make flow compile on darwin-arm64. I tested this with Homebrew's formula - also works, allowing Homebrew users to easily build flow from source (and for Homebrew to host their own arm64 bottles) before Facebook is ready to ship arm64 binaries


radex:
@mroch ⬆️

samwgoldman:
Sorry for the delay and thank you for your patience so far. We are working to update our internal compiler toolchain, which is separate from the toolchain used to build Flow on Circle, to OCaml 4.12.0. We share a toolchain with a few other tools, including Hack, so this is a bit of a slow process. Once this is ready internally, I'll update you here.

mroch:
Fixed in https://github.com/facebook/flow/commit/be611f1e30796ed71c4d68b1dc73a839ce4234b3, thanks!

PR #8640, opened 2021-04-05
Fix the dune build (remove extraneous module)

anmonteiro:
The module `value` in `src/heap` doesn't exist. This makes the dune build work again for the parser.

### DIFF
--- a/src/heap/dune
+++ b/src/heap/dune
@@ -24,8 +24,7 @@
   (wrapped false)
   (modules
     prefix
-    sharedMem
-    value)
+    sharedMem)
   (preprocess (pps ppx_deriving.std visitors.ppx))
   (libraries
     heap_libc


anmonteiro:
@mroch while I have your attention, is there a reason why `flow_parser` isn't declared with a `public_name` field in the dune file?

Right now running `dune build -p flow_parser` doesn't do anything, even though there's a `flow_parser.opam` file.

anmonteiro:
> @mroch while I have your attention, is there a reason why `flow_parser` isn't declared with a `public_name` field in the dune file?
> 
> Right now running `dune build -p flow_parser` doesn't do anything, even though there's a `flow_parser.opam` file.

Looks like this was done in https://github.com/facebook/flow/commit/53c336e946ab77544f2af13d79049b6999298fe3

PR #8650, opened 2021-04-29
Add Utility types for `Pick` and `Omit`

nmn:
These are common utility function people reach for, but Flow doesn't have them. The $Omit function is a bit redundant as you achieve the same thing by using $Rest directly, but the implementation here takes care of some things people tend to forget (making the second object $ReadOnly so you don't need the exact value types)

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -2553,3 +2553,6 @@ declare var Atomics: {
   +[key: $SymbolToStringTag]: 'Atomics',
   ...
 };
+                             
+type $Pick<T: {...}, P: {...}> = $Exact<$ObjMapi<$Exact<P>, <K>(K) => $ElementType<T, K>>>;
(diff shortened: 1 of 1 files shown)


gkz:
Thanks for the idea. Right now we aren't looking to add any new utility types, while we improve the underlying type system, as the existing utility types (and these ones) have some underlying issues in their evaluation. "omit" and "pick" are utilities that are requested internally as well, so after we fix some of the underlying issues, we will probably look at some kind of solution to accomplishing this task.

FezVrasta:
@gkz do you think it could be possible to reconsider this now?

gkz:
We plan on implementing "mapped types" in general in 2023, which would allow for utilities types like `Pick` and `Omit` to be trivially defined using them. So yes, by the end of the year

PR #8654, opened 2021-05-06
Handle function type in useState

nmn:
This is not the most straightforward thing, but it does catch an edge case where the types allow incorrect behaviour the state value type is itself a function.

Context: https://flow.org/try/#0JYWwDg9gTgLgBAbwK4GcCmBlGBDGaC+cAZlBCHAORRrYDGMFA3AFDO0QB2K8A2t7mgA0cdDCwCAunAC8cVJhx4APAH4AFNyjAOAcwCU0gHwckAG1OG1J83pbNR4vGrUB9GAC4RMLboPGzprbMQA

This uses the overload type syntax. Hopefully that works correctly?

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -337,15 +337,10 @@ declare module react {
 
   declare export function useContext<T>(context: React$Context<T>): T;
     
-  type NonFunctions = number | boolean | Object | string | null | void | Array<NonFunctions>;
-
-  declare export function useState<S: NonFunctions>(
-    initialState: (() => S) | S,
-  ): [S, ((S => S) | S) => void];
-  
-  declare export function useState<S: Function>(
-    initialState: (() => S),
-  ): [S, ((S => S)) => void];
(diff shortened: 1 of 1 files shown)


lib/react.js
```diff
@@ -336,10 +336,16 @@ declare module react {
...
   declare export function useContext<T>(context: React$Context<T>): T;
+    
+  type NonFunctions = number | boolean | Object | string | null | void | Array<NonFunctions>;
 
-  declare export function useState<S>(
+  declare export function useState<S: NonFunctions>(
     initialState: (() => S) | S,
   ): [S, ((S => S) | S) => void];
+  
+  declare export function useState<S: Function>(
```
jamesisaac, line 346: `Function` is just an alias for `any`: https://flow.org/en/docs/linting/rule-reference/#toc-unclear-type

samwgoldman:
This change doesn't seem right. Aside from implementation details (using the `Function` type), `useState` is overloaded to call (lazily) a provided function. So you can't actually have a function-valued state in React, since React will instead try to call that function.

PR #8682, opened 2021-06-15
Bring dune build up to date

bobzhang:
This PR brings dune build up to date for all native targets, it is a safe PR that should not break any existing workflow, 
since we don't move any files around , it is purely additive.
 
It is split into small commits, each commit should be self-explained.

# How to test it?

```
flow$dune --version
2.8.5
flow$dune build src/flow.exe 
flow$dune build @all
```
It is tested on MacOS x64 and M1

### DIFF
--- a/dune
+++ b/dune
@@ -1,7 +1,7 @@
 (env
   (dev
-    (flags (:standard  -w -6-50 -error-style short ))
-    ; Use the default settings whicih is more strict
+    (flags (:standard -w @a-4-6-20-29-35-41-42-44-45-48-50 ))
+    ; TODO: Use the default settings which is more strict
     ; warning 6 should be turned on later
     ; this is needed to make the diff clean
     ; warning 50 is not a big deal
--- a/dune
+++ b/dune
@@ -1,5 +1,10 @@
 (env
-  (_
-    (flags (:standard -w @a-4-6-20-29-35-41-42-44-45-48-50 -error-style short ))
+  (dev
+    (flags (:standard  -w -6-50 -error-style short ))
+    ; Use the default settings whicih is more strict
+    ; warning 6 should be turned on later
+    ; this is needed to make the diff clean
+    ; warning 50 is not a big deal
+    
   )
 )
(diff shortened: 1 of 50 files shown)


src/state/heaps/context/dune
```diff
@@ -4,7 +4,7 @@
   (libraries
     flow_common_audit
     flow_parser
-    flow_shared_mem
+    ; flow_shared_mem
```
samwgoldman, line 7: Let's delete things instead of commenting

src/parser_utils/type_sig/dune
```diff
@@ -0,0 +1,17 @@
...
+ (name flow_type_sig)
+ (wrapped false)
+ (libraries
+  flow_parser_utils
+  cycle_hash
+  dtoa
+  flow_common
+  flow_common_utils
+  flow_common_utils_loc_utils
+  flow_common_tarjan
```
samwgoldman, line 11: I'm not looking closely at which dependencies you are adding, but I'm pretty familiar with this code and I'm surprised to see the dependency on tarjan here. I don't think we depend on this library.

dune
```diff
@@ -1,5 +1,10 @@
 (env
-  (_
-    (flags (:standard -w @a-4-6-20-29-35-41-42-44-45-48-50 -error-style short ))
+  (dev
+    (flags (:standard  -w -6-50 -error-style short ))
+    ; Use the default settings whicih is more strict
```
samwgoldman, line 4: s/whicih/which

dune
```diff
@@ -1,5 +1,10 @@
 (env
-  (_
-    (flags (:standard -w @a-4-6-20-29-35-41-42-44-45-48-50 -error-style short ))
+  (dev
+    (flags (:standard  -w -6-50 -error-style short ))
```
samwgoldman, line 3: I'd suggest keeping this in sync with _tags, otherwise we're going to keep breaking the dune build inadvertently.

src/parser_utils/type_sig/dune
bobzhang: indeed, flow_common_tarjan is not needed, fixed
There is an issue upstream https://github.com/ocaml/dune/issues/650, once the feature request is landed, we can find it automatically

dune
bobzhang: sounds good

Issue #7110, opened 2018-10-28
flow_parser can not be installed correctly using opam 2.0


vincentdchan:
It seems that opam 2.0 introduce sandbox feature to separate build and installation process, flow_parser can not be installed correctly. Here's the error:
```
# ocamlfind: Cannot mkdir /Users/duzhongchen/.opam/system/lib/flow_parser: Operation not permitted
# make: *** [ocamlfind-install] Error 2
```
No idea how to solve it.

jbrown215:
Does sudo work?

vincentdchan:
No. It seems that it's a restriction from opam, I solve it by execute `make ocamlfind-install` install by myself.

vicapow:
closed by mistake

giltho:
Hello, I managed to fix that and to have flow_parser install properly with opam 2.0.

However, the opam file in the master branch does not correspond to the one in the v0.96 release archive. I don't really know where to make the pull.

How should I do proceed ?

thanks !

Bannerets:
@giltho Could you post the fix?

giltho:
Sure :
The error comes from the sandboxing of opam 2. Basically, the build is not allowed to get "out" of its own folder. That is why the `build` and `install` operations are now separated.

The opam file should be 
```opam
opam-version: "2.0"
name: "flow_parser"
version: "X.Y.Z"
maintainer: "flow@fb.com"
authors: ["Flow Team <flow@fb.com>"]
homepage: "https://github.com/facebook/flow/tree/master/src/parser"
bug-reports: "https://github.com/facebook/flow/issues"
license: "MIT"

build: [ "sh" "-c" "cd src/parser && make ocamlfind-build" ]
install: [ "sh" "-c" "cd src/parser && make ocamlfind-install" ]
remove: ["ocamlfind" "remove" "flow_parser"]

depends: [
  "ocaml" {>= "4.03.0"}
  "ocamlfind" {build}
  "ocamlbuild" {build}
  "dtoa"
  "core_kernel"
  "ppx_deriving" {build}
  "ppx_gen_rec" {build}
  "sedlex" {< "2.0"}
  "ppx_tools_versioned" {= "5.2"}
  "wtf8"
]

dev-repo: "git+https://github.com/facebook/flow.git"
synopsis: "The Flow parser is a JavaScript parser written in OCaml"
description: """
[...]

PR #8683, opened 2021-06-15
Update lib/bom.js to have the getInstalledRelatedApps type

limaAniceto:
[W3C GetInstalledRelatedApps definition](https://wicg.github.io/get-installed-related-apps/spec/) 

[CanIUse for getInstalledRelatedApps](https://caniuse.com/?search=getInstalledRelatedApps)

[WebDEV tutorial on the getInstalledRelatedApps](https://web.dev/get-installed-related-apps/)

I wasn't able to build the flow binary due to some know issues apparently but hopefully this PR is simple enough to be read
- flow_parser can not be installed correctly using opam 2.0 #7110

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -256,10 +256,6 @@ declare class NavigatorConcurrentHardware {
     +hardwareConcurrency: number;
 }
 
-declare class NavigatorGetInstalledRelatedApps {
-  
-}
-
 declare class Navigator mixins
   NavigatorID,
   NavigatorLanguage,
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -256,6 +256,10 @@ declare class NavigatorConcurrentHardware {
     +hardwareConcurrency: number;
 }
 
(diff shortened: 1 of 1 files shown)


lib/bom.js
```diff
@@ -309,6 +309,13 @@ declare class Navigator mixins
     // Gecko
     taintEnabled?: () => false;
     oscpu: string;
+
+    // Chrome v0.70+
+    getInstalledRelatedApps?: $ReadOnlyArray<{
+      +platform: string,
+      +id: string
+      +url?: string
```
jbrown215, line 317: This object should have a version field AFAICT from https://github.com/WICG/get-installed-related-apps/blob/main/EXPLAINER.md

Is that right? Or are there other docs I should be looking at?

jbrown215 (changes requested):
Question inline

jbrown215:
Closing until the author responds to feedback

PR #8688, opened 2021-06-16
Represent shmem addrs as byte offsets into a char bigarray

samwgoldman:
Summary:
In the shared heap, addresses are stored as byte offsets from the heap base
pointer. Before this diff, when representing an address in OCaml, we would
convert this byte offset into a word offset. This diff removes the conversion,
so we also represent addresses as byte offsets in OCaml.

To make this work, I changed the word-based Bigarray to be char-based instead.
The parameterization of the Bigarray affects the get/set operations as well as
the size of the data returned, similar to changing a C array from int* to char*.

Because we still want to read/write words, just at char offsets, I changed from
using the get/set API to primitive functions, which do exactly what we want.

Note that we still represent sizes in word units, so we still need to take some
care when computing a new address from an existing address (byte offset) and a
size (in words).

The `(char, int8_unsigned_elt, c_layout) Array1.t` type is a more generally
useful type as well. It is sometimes called a "bigstring" and there are
libraries that provide useful functionality for it, notably Bin_prot.

Differential Revision: D28401827

### DIFF
--- a/src/heap/sharedMem.ml
+++ b/src/heap/sharedMem.ml
@@ -23,7 +23,7 @@ type table_stats = {
   slots: int;
 }
 
-type heap = (nativeint, Bigarray.nativeint_elt, Bigarray.c_layout) Bigarray.Array1.t
+type heap = (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
 
 (* Phantom type parameter provides type-safety to callers of this API.
  * Internally, these are all just ints, so be careful! *)
@@ -818,8 +818,6 @@ struct
 end
 
 module NewAPI = struct
-  open Bigarray
-
   type chunk = {
     heap: heap;
     mutable next_addr: int;
@@ -860,6 +858,10 @@ module NewAPI = struct
 
   type size = int
 
+  let bsize_wsize bsize = bsize * Sys.word_size / 8
+
+  let addr_offset addr size = addr + bsize_wsize size
+
   let get_heap () =
     match !heap_ref with
     | None -> failwith "get_heap: not connected"
@@ -873,7 +875,7 @@ module NewAPI = struct
     let x = f chunk in
     (* Ensure allocated space was initialized. *)
     assert (chunk.remaining_size = 0);
-    assert (chunk.next_addr = addr + size);
+    assert (chunk.next_addr = addr_offset addr size);
     x
 
   (* Addresses are relative to the hashtbl pointer, so the null address actually
@@ -1043,17 +1045,17 @@ module NewAPI = struct
 
   (* offsets *)
 
-  let module_addr file = file + 1
+  let module_addr file = addr_offset file 1
 
-  let module_refs_addr file = file + 2
+  let module_refs_addr file = addr_offset file 2
 
-  let local_defs_addr file = file + 3
+  let local_defs_addr file = addr_offset file 3
 
-  let remote_refs_addr file = file + 4
+  let remote_refs_addr file = addr_offset file 4
 
-  let pattern_defs_addr file = file + 5
+  let pattern_defs_addr file = addr_offset file 5
 
-  let patterns_addr file = file + 6
+  let patterns_addr file = addr_offset file 6
 
   (* read *)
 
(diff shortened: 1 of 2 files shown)


samwgoldman:
@bobzhang I can't tag you as a reviewer, but would appreciate your review on this change!

src/heap/sharedMem.ml
```diff
@@ -860,6 +858,10 @@ module NewAPI = struct
 
   type size = int
 
+  let bsize_wsize bsize = bsize * Sys.word_size / 8
+
+  let addr_offset addr size = addr + bsize_wsize size
```
bobzhang, line 863: `bsize_wsize` seems only used once, shall we inline it to avoid a new name to make people to understand the code easier?

src/heap/sharedMem.ml
```diff
@@ -1156,36 +1159,38 @@ module NewAPI = struct
...
 
-  let es_module_exports m = read_addr (get_heap ()) (m + 2)
+  let es_module_exports m = read_addr (get_heap ()) (addr_offset m 2)
 
-  let es_module_type_exports m = read_addr (get_heap ()) (m + 3)
+  let es_module_type_exports m = read_addr (get_heap ()) (addr_offset m 3)
 
   (* write *)
 
+  (* Read int64 from given byte offset in the heap. This is not bounds checked. *)
```
bobzhang, line 1176:
```suggestion
  (* Write int64 from given byte offset in the heap. This is not bounds checked. *)
```

src/heap/sharedMem.ml
```diff
@@ -1156,36 +1159,38 @@ module NewAPI = struct
...
+  external unsafe_write_int64 : heap -> int -> int64 -> unit = "%caml_bigstring_set64u"
+
```
bobzhang, line 1178: Shall we provide both `safe` and `unsafe` version? Is this used in some hot paths?

src/heap/sharedMem.ml
```diff
@@ -1156,36 +1159,38 @@ module NewAPI = struct
...
    * allocated. *)
-  let unsafe_write_addr_at heap dst addr =
-    Array1.unsafe_set heap dst Nativeint.(shift_left (of_int addr) 3)
+  let unsafe_write_addr_at heap dst addr = unsafe_write_int64 heap dst (Int64.of_int addr)
 
   (* Write a 64-bit hash at a specified address in the heap. This write is not
    * bounds checked; caller must ensure the given destination has already been
    * allocated. *)
-  let unsafe_write_hash_at heap dst hash = Array1.unsafe_set heap dst (Int64.to_nativeint hash)
+  let unsafe_write_hash_at heap dst hash = unsafe_write_int64 heap dst hash
```
bobzhang, line 1194: nit: seems could be simplified as an alias

bobzhang (approved):
Note I don't understand how flow shared memory works but this diff looks fine by me.

A couple of small questions: 
- Do we have CI running on 32 bit systems or is 32 bit system relevant? If not, somethings can be further simplified

- What  is the motivation of this diff? Is it mostly to reuse some third party libraries like bigstringaf

src/heap/sharedMem.ml
samwgoldman: Unless you feel strongly, I think I prefer giving a name to the wsize->bsize conversion. You're right that it's used only once, but my subjective opinion is that a separate function is easier to read/understand.

src/heap/sharedMem.ml
samwgoldman: Thanks! Will fix.

src/heap/sharedMem.ml
samwgoldman: You're right! Will change.

samwgoldman:
> Do we have CI running on 32 bit systems or is 32 bit system relevant? If not, somethings can be further simplified

Flow doesn't support 32 bit systems, so any simplification would be very welcome!

> What is the motivation of this diff? Is it mostly to reuse some third party libraries like bigstringaf

Main motivation is to to use Bin_prot for better control over serialization compared to Marshal. We serialize a lot of small values, and the 20 byte header emitted by Marshal's output ends up taking up more space than the value itself.

It is also nice to avoid the careful conversion between byte offsets stored in the heap and word offsets used to index into the word-based Bigarray. While writing this code I found it easy to get this wrong, so using byte offsets everywhere feels a bit simpler to me.

bobzhang:
> Flow doesn't support 32 bit systems, so any simplification would be very welcome!

If we only support 64 bit systems, some byte count maths can be simplified, for example, Sys.word_size is always 64.

>  and the 20 byte header emitted by Marshal's output ends up taking up more space than the value itself.

FYI, If we only care about 64 bit systems, the first 8 bytes are always the same which can be shaved off.

samwgoldman:
> Sys.word_size is always 64

I confirmed that ocamlc does enough constant folding that there's no difference in codegen between writing things in terms of Sys.word_size and hard-coding 64 bit. In light of that, I slightly prefer using `word_size`, but don't feel strongly.

> the first 8 bytes are always the same

Interesting! My read of extern.c was a little different; that we can have the small format or the large format on 64 bit arch. Will take another look! Thanks for the tip :)

src/heap/sharedMem.ml
samwgoldman: Oh, missed this question. I was using the unsafe variant before (Array1.unsafe_set) so this choice is mainly just behavior preserving. The code paths that lead to this call already have assertions that make eliminating the bounds check safe, which is why I used an unsafe set operation. In hindsight, I should have just used the bounds checked operation, since it's almost certainly not the bottleneck here.

PR #8696, opened 2021-06-28
Parser optimization

bobzhang:
- f7896b1 Added a end to end benchmark testsuite, the tested JS file is huge (12MB), glad 
   that Flow can handle it.
- 367403e 
   Refactoring to avoid expose too generic peek function, it is semantics preserving
- 165ed27
    Take advantage of the specialisation in previous commit, since we know the 
    lookahead buffer is at most 2, we pre-allocate it, so there is no need to implement it using
    a dynamic vector, also less code involved.

#  benchmarking
The improvement is impressive, it is around 10% improvement in the whole pipeline(including the IO time)
Screenshot 2021-06-28 at 11 23 59 AM

 Edit:
 
858d41c rename .js to .data to avoid tool issues
f9669fb guard the correctness for a missing case, previously lex_env could lex more in theory, fix that issue

a04192579a get rid of several runtime assertions, should also be slightly faster
7e09bad get rid of lex_until in junk function


bobzhang:
The windows build failure seems to be not relevant to my changes:
```
Failures
 - yarn - yarn not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
 NOTE: When you specify explicit sources, it overrides default sources.
If the package version is a prerelease and you didn't specify `--pre`,
 the package may not be found.
Please see https://chocolatey.or
```

mroch:
can you rename the test file to something other than .js? we have tools that search all .js files and don't want that to ever show up

bobzhang:
This PR is good for review,  note modulo the perf enhancement, it should also make the code generally more readable, there are some other opportunities, but I will stop here before we figure out a more typical workload for benchmarking

bobzhang:
b1d4106 simplify the logic of junk
c1d658e remove the indirection of la_results
The previous benchmark has lots of jitters, so I updated the benchmark that it is more scientific and reproducible:

Prior to this patch:
---- dune exec ./bench.exe
59.212305
59.155172

After this patch:
---- dune exec ./bench.exe
54.252545
54.472153

So the benefit of this PR is in two fold:
- It delivers around 10% boost for parsing a large file, the change is quite small, notably only touching one small module: Parser_env.Lookahead.
- It makes the lookahead code more readable, this is subjective, but would be happy to hear others' opinions

bobzhang:
rebased
- a5d1c29 swap the third party sedlexing runtime with a customzied version.
run twice the benchmark
```
42.909486
42.721931
```
This is a big jump from 54 -> 42

bobzhang:
The last commit added a c file, it may need some support from the build system.
If you have some trouble building it, we can skip the last commit temporarily.

Benchmark running on my machine compared with previous runs:

```
bench$dune exec ./bench.exe
31.823548bench$dune exec ./bench.exe
31.156909bench$dune exec ./bench.exe
```

It is a drop from 42 -> 31

655a5d7  This commit reduces the whole allocation by 20%, but it does not translate to too much CPU gain, probably because the allocation is really fast.

cb3699c This commit rewrite the utf8 encoding/decoding logic, it would be nice to have a careful review.
Other commits should be self explained

samwgoldman:
Ack -- looks like the JavaScript build of the Flow parser isn't happy with the C stub for initializing an array. Is your sense that this optimization is critical and we should make it work with the JS-compiled version? Easier to back it out, if not.

bobzhang:
@samwgoldman this optimisation should not have impact over JS.

> Easier to back it out, if not.

How to do that? Do we have some setup for conditional compilation?

samwgoldman:
I meant back out altogether, for native and JS. Was asking about whether
that particular optimization was significant in native.

bobzhang:
@samwgoldman After a number of trials, finally make it working for the CI! 
I also generalized the primitive a little bit so that it will be useful in other cases, e.g a large array of initilization of Nones. It has around 1-2% in my benchmark, these small things add up, so I tend to keep it.

src/parser/parser_env.ml
```diff
@@ -38,51 +38,32 @@ end
  * - Refactor this to memoize all requested lookahead, so we aren't lexing the
  *   same token multiple times.
  *)
-let maximum_lookahead = 2
 
 module Lookahead : sig
```
samwgoldman, line 42: Changes to this module look good! Aside from better perf, the simplicity makes this logic quite a bit easier to follow as well.

src/parser/parser_env.ml
```diff
@@ -357,12 +344,20 @@ let add_used_private env name loc =
...
-  Lookahead.peek !(env.lookahead) i
+let[@inline] lookahead_0 env = Lookahead.peek_0 !(env.lookahead)
+
+let[@inline] lookahead_1 env = Lookahead.peek_1 !(env.lookahead)
+
+let[@inline] lookahead ~i env =
+  match i with
+  | 0 -> lookahead_0 env
+  | 1 -> lookahead_1 env
+  | _ -> assert false
```
samwgoldman, line 355: Curious if you looked at generated code before adding the inline directives, since it seems pretty likely that ocamlopt would inline these definitions anyway.

src/parser/parser_env.ml
```diff
@@ -357,12 +344,20 @@ let add_used_private env name loc =
...
 
 (* functional operations: *)
-let with_strict in_strict_mode env = { env with in_strict_mode }
+let with_strict in_strict_mode env = 
+  if in_strict_mode = env.in_strict_mode then env
```
samwgoldman, line 359: If we're going to do this optimization, it might be nice to do it uniformly for all `with_foo` combinators. I imagine you made the changes based on what bubbled up from profiling, yeah? That seems like it risks overfitting your benchmark. If this is a useful optimization, let's apply it to the rest of these functions as well.

src/parser/parser_env.ml
```diff
@@ -616,12 +615,13 @@ module Peek = struct
 
   let ith_comments ~i env =
     let comments = Lex_result.comments (lookahead ~i env) in
+    match comments with 
+    | [] -> []
+    | _ ->
```
samwgoldman, line 620:
```suggestion
    (* perf: avoid closure allocations from List.filter for the common case of empty list *)
    match comments with 
    | [] -> []
    | _ -> 
```

We might also consider (a) using Base.List.filter here, which has the optimization we want built-in, and (b) proposing the eta-expansion in the stdlib definition. It's a small change, and clearly affected this code.

src/parser/token.ml
```diff
@@ -140,7 +140,7 @@ type t =
   | T_EOF
   (* JSX *)
   | T_JSX_IDENTIFIER of { raw: string }
-  | T_JSX_TEXT of (Loc.t * string * string) (* loc, value, raw *)
+  | T_JSX_TEXT of Loc.t * string * string (* loc, value, raw *)
```
samwgoldman, line 143: an inline record would be nice here, too

src/parser/token.ml
```diff
@@ -21,7 +21,7 @@ type t =
       value: string;
       raw: string;
     }
-  | T_REGEXP of (Loc.t * string * string) (* /pattern/flags *)
+  | T_REGEXP of Loc.t * string * string (* /pattern/flags *)
```
samwgoldman, line 24: nice catch -- an inline record would be nicer IMO

samwgoldman (changes requested):
This is great stuff! I confirmed that this leads to a nice speedup internally as well. Took a pass at much of this, but still need to spend time inside flow_sedlexing...

samwgoldman:
Oh, can you remove the sourcemap comment from the last line of flow_parser_dot_js.data? that comment is 3.8MB, and definitely not representative of a common parsing workload

src/parser/array_init.c
```diff
@@ -0,0 +1,39 @@
+#include <string.h>
+#include "caml/memory.h"
+#include "caml/alloc.h"
+#include "caml/mlvalues.h"
+
+/* The fast way to initialize an array filled with -1 */
+CAMLprim value caml_make_int_vect(value len, value init)
```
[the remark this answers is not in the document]
bobzhang: it is mainly to avoid a check called `caml_initialize`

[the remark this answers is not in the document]
samwgoldman: Where is `caml_initialize` called from `caml_make_vect`? This is the reason why OCaml promotes the initial value (if it is a young block).

src/parser/flow_lexer.ml
```diff
@@ -432,7 +448,7 @@ let string_escape env lexbuf =
   | eof
   | '\\' ->
     let str = lexeme lexbuf in
-    let codes = Sedlexing.lexeme lexbuf |> Array.map Uchar.to_int in
+    let codes = Sedlexing.lexeme lexbuf (*|> Array.map Uchar.to_int*) in
```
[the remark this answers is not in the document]
bobzhang: addressed in coming commits

src/parser/flow_lexer.ml
```diff
@@ -1766,21 +1786,33 @@ let jsx_child env =
...
+    {Lex_result.lex_token = child; lex_loc= loc; lex_comments = []; lex_errors = List.rev lex_errors_acc}
 
 let wrap f =
   let rec helper comments env =
     match f env env.lex_lb with
     | Token (env, t) ->
       let loc = loc_of_token env t in
-      let env = { env with lex_last_loc = loc } in
-      (env, t, loc, List.rev comments)
+      let lex_comments = if comments = [] then [] else List.rev comments in
```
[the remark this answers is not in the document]
bobzhang: Yeah, this is tiny to skip a function call, do you want to revert that?

src/parser/parser_env.ml
bobzhang: I guess this does not have impact on non-flambda mode, will remove the annotations to avoid confusions

src/parser/parser_env.ml
bobzhang: >  I imagine you made the changes based on what bubbled up from profiling, yeah?

Yes.
> If this is a useful optimization, let's apply it to the rest of these functions as well.

Done in coming commits

src/parser/parser_env.ml
bobzhang: addressed in coming commits. Note this would make flow_parser depends on base though

src/parser/flow_lexer.ml
```diff
@@ -163,6 +165,25 @@ let js_id_start = [%sedlex.regexp? '$' | '_' | id_start | unicode_escape | codep
...
+      | _ -> false
+    else true in 
+  let len = String.length x in 
+  if len = 0 then false 
+  else 
+    let first_char = String.unsafe_get x 0 in 
+    match first_char with 
+    | '$' | '_' | 'a' .. 'z' | 'A' .. 'Z'  (* approximation to js_id_start *)
+      -> aux x ~offset:1 len
+    | _ -> false
```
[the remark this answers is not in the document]
bobzhang: good suggestion. In theory it should be faster than my original patch. Addressed in coming commits

src/parser/array_init.c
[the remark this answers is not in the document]
bobzhang: https://github.com/ocaml/ocaml/pull/8657/files
It seems this minor enhancement was also done in upstream in 4.10.
I was doing this in 4.09, will remove that, it is a pity that I played lots of time with ocamlbuild on CI..

bobzhang:
> Oh, can you remove the sourcemap comment from the last line of flow_parser_dot_js.data? that comment is 3.8MB, and definitely not representative of a common parsing workload

I agree, can we leave it in a separate PR coming afterwards for future references, if we do it right now, it will leave all previous data points invalid, so we need do it again..

samwgoldman:
@bobzhang I'm having trouble importing due to a merge conflict. Would you mind rebasing onto a recent master and force pushing to get a clean history? I think that might help.

bobzhang:
@samwgoldman rebased, do you want to clean up the history? I guess that would make you review a bit more difficult

PR #8697, opened 2021-06-28
Support method unbound for fs/promises in flow 0.154

atabel:
This PR allows using fs/promises like this in flow 0.154:
```js
const {writeFile} = require('fs').promises;
```

Node fs/promises doc: https://nodejs.org/api/fs.html#fs_promises_api

### DIFF
--- a/lib/node.js
+++ b/lib/node.js
@@ -1321,7 +1321,7 @@ declare module "fs" {
   }
 
   declare type FSPromisePath = string | Buffer | URL;
-  declare class FSPromise {
+  declare type FSPromise = {
     access(path: FSPromisePath, mode?: number): Promise<void>,
     appendFile(path: FSPromisePath | FileHandle, data: string | Buffer, options: WriteOptions | string): Promise<void>,
     chmod(path: FSPromisePath, mode: number): Promise<void>,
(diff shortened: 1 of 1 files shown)


lib/node.js
```diff
@@ -1356,12 +1356,15 @@ declare module "fs" {
...
     }>,
     readdir(path: FSPromisePath, options?: string | EncodingOptions): Promise<string[]>,
-    readFile(path: FSPromisePath | FileHandle, options: string): Promise<string>,
-    readFile(path: FSPromisePath | FileHandle, options?: EncodingFlag): Promise<Buffer>,
-    readlink(path: FSPromisePath, options: BufferEncoding): Promise<Buffer>,
-    readlink(path: FSPromisePath, options?: string | EncodingOptions): Promise<string>,
-    realpath(path: FSPromisePath, options: BufferEncoding): Promise<Buffer>,
-    realpath(path: FSPromisePath, options?: string | EncodingOptions): Promise<string>,
+    readFile:
+      & (path: FSPromisePath | FileHandle, options: string) => Promise<string>
```
Brianzchen, line 1360: Have you tested this works? From my experience migrating lodash, this would mean

`(path: FSPromisePath | FileHandle, options: string)` return  `Promise<string> & ((path: FSPromisePath | FileHandle, options?: EncodingFlag) => Promise<Buffer>)`

You may want to change this to format like this
```
& ((path: FSPromisePath | FileHandle, options: string) => Promise<string>)
& ((path: FSPromisePath | FileHandle, options?: EncodingFlag) => Promise<Buffer>)
```

atabel: oh, absolutely. You're right

atabel: fixed, thanks ;)

PR #8698, opened 2021-07-01
Allow URL objects when starting workers

mischnic:
Passing URL object into worker construtors/`navigator.serviceWorker.register` should be allowed

For example described here: https://webpack.js.org/guides/web-workers/

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -735,7 +735,7 @@ type WorkerOptions = {
 }
 
 declare class Worker extends EventTarget {
-    constructor(stringUrl: string, workerOptions?: WorkerOptions): void;
+    constructor(url: string | URL, workerOptions?: WorkerOptions): void;
     onerror: null | (ev: any) => mixed;
     onmessage: null | (ev: MessageEvent) => mixed;
     onmessageerror: null | (ev: MessageEvent) => mixed;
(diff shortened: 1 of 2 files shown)


mvitousek:
The documentation you linked to looks good for `Worker` constructors; is there documentation for `SharedWorker` and `register`?

mischnic:
It's hard to find examples for this. The reason why passing an URL object works is that it is stringified and `url.toString()` just returns the "real" url. (Tthe benefit of the URL/import.meta.url thing is that the worker script is now actually relative to the JS file that creates the worker, just like an import - otherwise, it's relative to the current document. And since this `URL`->string conversion happens automatically, you can omit `toString()` in browsers.)

Webpack example for SharedWorker: https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/examples/worker/example.js#L25-L31

The arguments for `Worker`, `SharedWorker`, `register` are all `USVString` according to the spec, so `URL`s aren't specifically handled there (and all three should behave the same).

isker:
Worth considering:  https://github.com/microsoft/TypeScript/blob/a15030ff6fd63bc0c2a9ea1fe0ade45da2ecd698/lib/lib.webworker.d.ts#L2898

mischnic:
@mvitousek What do you think?

As far as I can tell, Flow already does what I'm proposing here for `fetch`: `RequestInfo` is the "url or fetch parameter" input type for `fetch`:
https://github.com/facebook/flow/blob/f9afdbc9c5971c2517ce96c64d996325cd6e6405/lib/bom.js#L1511

https://github.com/facebook/flow/blob/f9afdbc9c5971c2517ce96c64d996325cd6e6405/lib/bom.js#L1601

But the spec actually doesn't say that it can be a `URL` object (meaning Flow models this very common implicit `URL` to `string` conversion as done by `URL#toString`): https://fetch.spec.whatwg.org/#requestinfo
```js
typedef (Request or USVString) RequestInfo;
interface Request {
  constructor(RequestInfo input, optional RequestInit init = {});
  ...
}

...

fetch(RequestInfo input, optional RequestInit init = {});
```

SamChou19815:

PR #8706, opened 2021-07-17
Add support for the node 'console' module

isker:
- Define the 'console' module.  This module exports the Console
  constructor, which creates objects of the same type as the global
  `console`. https://nodejs.org/api/console.html
- Note that while the pre-existing type of the global `console` variable
  does not exactly conform to the type indicated in the node docs, it is
  functionally equivalent. For example, some methods on the global
  variable are typed `(...any): void` whereas the node docs indicate a
  named first parameter which is still any, as in `(any, ...any): void`.
- In order to ensure that the cross-environment `console` definition is
  correct for node, remove the `_exception` method.  I can find no
  evidence that a method named '_exception' ever existed on a console in
  any environment.  There is the deprecated `exception` method, which is
  an alias for `error` and MDN indicates is only supported in Firefox
  today:
  https://developer.mozilla.org/en-US/docs/Web/API/console#browser_compatibility.
  Because of that, it seems fine to eliminate this method without replacement.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -2439,7 +2439,6 @@ declare var console: {
     dir(...data: Array<any>): void,
     dirxml(...data: Array<any>): void,
     error(...data: Array<any>): void,
-    _exception(...data: Array<any>): void,
     group(...data: Array<any>): void,
     groupCollapsed(...data: Array<any>): void,
     groupEnd(): void,
--- a/lib/node.js
+++ b/lib/node.js
@@ -2802,6 +2802,23 @@ declare module 'repl' {
   }
 }
 
+declare module 'console' {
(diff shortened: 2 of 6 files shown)


isker:
Note that I didn't implement the functionality wherein:
```
> console === require('console');
true
```

Not sure how to do this in libdefs other than copying the entire `console` declaration into the module definition.  I can do that if you think it's worth it.  That being said I think approximately zero people are writing code that uses this exported value.

isker:
And of course the current libdef setup provides no way to implement `new console.Console(...)` which works just fine in node, because the global `console` is shared across environments.  I don't think flow provides any way at all to target a specific environment like that?

isker:
@mroch don't suppose you can take some time to look at this?

PR #8708, opened 2021-07-28
fix dune build src/flow.exe on Darwin

bobzhang:
I noticed in 0b6dece it was added, is this OS dependent thing? After removing it, it builds fine on MacOS

### DIFF
--- a/src/heap/dune
+++ b/src/heap/dune
@@ -7,7 +7,6 @@
   (names hh_shared)
   (flags
    (:standard -I../../src/third-party/lz4)))
- (c_library_flags (-latomic))
  (preprocess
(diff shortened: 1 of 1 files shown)


samwgoldman:
we need libatomic for the 16 byte compare-and-swap in hh_shared.c when compiling with gcc. Clang will just emit a `lock cmpxchg16b`, but gcc emits `__atomic_compare_exchange_16` which is provided by libatomic.

probably you are picking up clang on mac, but we use gcc internally, and gcc on darwin would also need libatomic. is there a better way to handle this in dune?

bobzhang:
wait for a better fix from upstream per discussions

PR #8710, opened 2021-07-29
fix suspicious equal

bobzhang:
IIUC, `==` is used as a short-cut optimisation to save allocations, it seems we should always use `==` instead of `=`


mroch:
indeed, this is critical for perf! shocked there are so many of them.

PR #8712, opened 2021-07-29
Use __sync intrinsic for hh_add instead of __atomic

samwgoldman:
Summary:
We can't convince gcc to emit `lock cmpxchg16b` for the `__atomic` intrinsic, but
it still does for `__sync`. We had some issues with an arm64 target during
D26950006 (https://github.com/facebook/flow/commit/7f66dc40797ec6055dd66b038f3da011f45356a3), but we should be able to fix this, since armv8-a has `ldaxp`/`stlxp`
at least.

Differential Revision: D29990640


samwgoldman:
Fixed in eb7ae8c04cfaba723e84f6a39d05893b7b1f24ea by avoiding need for DWCAS

Issue #8715, opened 2021-08-04
Eslint plugin: 0.0.2 released without exported rules


mrtnzlml:
Hello! 👋 This one is simple: the new package [`eslint-plugin-fb-flow`](https://www.npmjs.com/package/eslint-plugin-fb-flow) version 0.0.2 (latest) was released without any new rules available. Here is the package content after installation:

```text
.
├── CHANGELOG.md
├── LICENSE
├── README.md
├── index.js
├── package.json
└── rules
    ├── flow-enums-default-if-possible.js
    ├── no-flow-enums-object-mapping.js
    ├── use-exact-by-default-object-type.js
    ├── use-flow-enums.js
    └── use-indexed-access-type.js
```

And here is the `index.js` (notice the missing rules):

```js
/**
 * Copyright (c) Facebook, Inc. and its affiliates.
 *
 * This source code is licensed under the MIT license found in the
 * LICENSE file in the root directory of this source tree.
 *
 * @emails oncall+flow
 * @format
 */

'use strict';

module.exports = {
  rules: {
    'use-indexed-access-type': require('./rules/use-indexed-access-type'),
  },
};
```

Using a rule that is not specified in the `index.js` under `rules` would result in the following error:

```text
error: Definition for rule 'fb-flow/use-flow-enums' was not found (fb-flow/use-flow-enums) at …
```

Thanks for having a look! 😊 

cc @gkz

PR #8716, opened 2021-08-04
Eslint plugin: fix exported rules

mrtnzlml:
Closes: https://github.com/facebook/flow/issues/8715


gkz:
Thanks, after this I will bump and publish a new version, and also update how we test so it is not possible to forget to re-export in the future

PR #8729, opened 2021-08-16
Allow `mixed` args to `String.prototype.concat()`

### LINKED ISSUE #8728: `String.prototype.concat()` should accept `Array<mixed>`, not only `Array<string>`
## Missing/Incorrect APIs

`String.prototype.concat` should allow any type as its arguments.

`concat(...values: Array<string>): string` should be `concat(...values: Array<mixed>): string`

## Relevant documentation

According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat) for the `concat()` method of the `String` built-in type: 

> If the arguments are not of the type string, they are converted to string values before concatenating.

Passing numbers, objects, etc to `String.prototype.concat` is is valid JS behavior but causes an error in Flow.

This is important because the alternative to String#concat is the `+` operator, but that operator is problematic because using `+` to concatenate a string to an object will cause `valueOf` to be called on the object before it's coerced to a string. Some types (notably the new ECMAScript ["Temporal"](https://github.com/tc39/proposal-temporal) date/time API that's currently in Stage 3) will [throw when `valueOf` is called](https://tc39.es/proposal-temporal/docs/plaindate.html#valueOf) so that users don't inadvertently use `<` or `>` to compare objects that have specialized comparison methods.

BTW, I discovered this Flow problem while fixing React's use of the `+` operator on strings and objects, because React was crashing and/or displaying confusing error messages when Temporal instances were rendered or used as props. This issue is causing https://github.com/facebook/react/pull/22064 to fail its Flow check.

PR coming!

justingrant:
Fixes #8728 and unblocks https://github.com/facebook/react/pull/22064.

According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat) for the `concat()` method of the `String` built-in type: 

> If the arguments are not of the type string, they are converted to string values before concatenating.

Passing numbers, objects, etc to `String.prototype.concat` is is valid JS behavior but causes an error in Flow. It's also blocking https://github.com/facebook/react/pull/22064.

This PR changes the arg type from `Array<string>` to `Array<mixed>`.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -1103,9 +1103,11 @@ declare class String {
     codePointAt(index: number): number;
     /**
      * Returns a string that contains the concatenation of two or more strings.
-     * @param strings The strings to append to the end of the string.
+     * If the arguments are not of the type string, they are converted to string values before
+     * concatenating.
+     * @param values The values to append to the end of the string.
      */
-    concat(...strings: Array<string>): string;
+    concat(...values: Array<mixed>): string;
     constructor(value?: mixed): void;
     /**
      * Returns true if the sequence of elements of searchString converted to a String is the
--- a/newtests/autocomplete/test.js
+++ b/newtests/autocomplete/test.js
@@ -158,7 +158,7 @@ export default (suite(({addFile, flowCmd}) => [
              },
              {
                "name": "concat",
-               "type": "(...strings: Array<string>) => string"
+               "type": "(...values: Array<mixed>) => string"
              },
              {
                "name": "endsWith",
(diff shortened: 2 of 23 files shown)


justingrant:
BTW, I'm happy to fix the test failures and to add more tests, but before digging into that work I'd like to get some guidance from maintainers about whether this PR is welcome or not.

samwgoldman:
Hi, Justin. I left a comment on the linked React PR, since the context around that change serves as motivation for this one, and just to keep the dialog in one spot.

justingrant:
Looks like this PR isn't aligned with Flow's goals around avoiding implicit type casts, so I'll close this PR to keep the noise of everyone's plate.

Issue #7453, opened 2019-02-07
website: Font size too big especially on mobile devices


open-source-explorer:
Check out the [reponsinator view](http://www.responsinator.com/?url=https%3A%2F%2Fflow.org%2F) demonstrating overall apperance of the webite across various devices.

> It becomes weird especially in the landscape mode.

jamesgeorge007:
@jbrown215 I would love to work on this :clap: Do you have any sort of intentions on how it should be refactored :thinking:

jbrown215:
@jamesgeorge007: Nope! I have no design knowledge or preferences. I'll happily review, though

DHvyas:
> Check out the [reponsinator view](http://www.responsinator.com/?url=https%3A%2F%2Fflow.org%2F) demonstrating overall apperance of the webite across various devices.
> 
> > It becomes weird especially in the landscape mode.

I haven't checked the code but I am guessing the CSS for the font size must be in terms of vw (view-width). Changing it to REM should fix it.

nicollbryla:
Hi.👋🏻 I’d like to solve it.

Adityaintel:
Can I work on this if it's still open?

SamChou19815:
The new website should have fixed the issue.

PR #8734, opened 2021-08-23
Updated Code.

Adityaintel:
Updated styling as mentioned in #7453 . It was a bit rigid and had a very little scope of adaptation for different devices.


SamChou19815:
Sorry, but I think that using `vw` for anything other than `100vw` is not a good idea.

PR #8739, opened 2021-08-27
WIP: FlowMap performance enhancement

bobzhang:
Note this is not ready for merge, but it would be nice to confirm some performance improvement on the large code base.

Here are some numbers I got when testing 30 copies of ReactNative 
(compiler: 4.10.2+fp, supposed to be the same as the internal one).

```
Running `flow check` with works = 0 (single core)
user | system | cpu | total
-- | -- | -- | --
89.65 | 4.78 | 96 | 1:37.38
90.48 | 4.69 | 96 | 1:38.47
90.12 | 4.56 | 98 | 1:36.43
89.86 | 4.73 | 97 | 1:37.21
======================= after changes
85.88 | 4.74 | 97 | 1:33.23
85.38 | 4.59 | 96 | 1:33.16
85.57 | 4.25 | 99 | 1:29.86
85.93 | 4.26 | 99 | 1:30.22
```

@panagosg7 The added C stub function is in get_build_id.c

### DIFF
--- a/src/hack_forked/utils/collections/flow_map.ml
+++ b/src/hack_forked/utils/collections/flow_map.ml
@@ -4,6 +4,12 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *)
+module type OrderedType = sig
+  type t
+
+  val compare : t -> t -> int
+  (* val equal : t -> t -> bool *)
+end
 
 module type S = sig
   type key
@@ -20,8 +26,11 @@ module type S = sig
 
   val update : key -> ('a option -> 'a option) -> 'a t -> 'a t
 
+  val adjust : key -> ('a option -> 'a) -> 'a t -> 'a t
+
   val singleton : key -> 'a -> 'a t
 
+  (* when [remove k map] failed to remove [k], the original [map] is returned *)
   val remove : key -> 'a t -> 'a t
 
   val merge : (key -> 'a option -> 'b option -> 'c option) -> 'a t -> 'b t -> 'c t
@@ -48,9 +57,19 @@ module type S = sig
 
   val bindings : 'a t -> (key * 'a) list
 
+  val min_binding : 'a t -> key * 'a
+
+  val min_binding_opt : 'a t -> (key * 'a) option
+
+  val max_binding : 'a t -> key * 'a
+
+  val max_binding_opt : 'a t -> (key * 'a) option
+
   val choose : 'a t -> key * 'a
 
   val choose_opt : 'a t -> (key * 'a) option
+
+  val split : key -> 'a t -> 'a t * 'a option * 'a t
 
   val find : key -> 'a t -> 'a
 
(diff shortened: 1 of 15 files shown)


src/hack_forked/utils/core/get_build_id.c
```diff
@@ -18,10 +18,228 @@
...
+  item *stack;
+  item *limit;
+} stack;
+
+// Resize the stack and modify the
+// current point in place
+static void resize_stack(stack *stk,
+                         item **sp)
+{
+  asize_t newsize;
```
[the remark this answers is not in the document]
bobzhang: yes

src/hack_forked/utils/collections/flow_map.ml
```diff
@@ -0,0 +1,441 @@
...
+    | (Node _, (Empty | Leaf _)) -> assert false
+
+  let rec union f s1 s2 =
+    match (s1, s2) with
+    | (Empty, s)
+    | (s, Empty) ->
+      s
+    | (s, Leaf { v; d })
+    | (Leaf { v; d }, s) ->
+      add v d s
```
panagosg7, line 342: I believe the behavior in these cases diverges from the existing one. The fact that we're not using `f` at all here is a bad smell.

This repros in the following code 
```
declare var x: {y?: number};
if (x.y === undefined || x.y === null) {
    (x.y: number);
}
```
due to this call to `Key_map.union`: 

https://github.com/facebook/flow/blob/main/src/typing/statement.ml#L7171

Before, the resulting map would bind `x.y` to the entry of the first branch, but after `x.y` gets the entry of the second one.

src/hack_forked/utils/collections/flow_map.ml
```diff
@@ -0,0 +1,441 @@
...
+        let l = union f l1 l2 and r = union f r1 r2 in
+        match d2 with
+        | None -> join l v1 d1 r
+        | Some d2 -> concat_or_join l v1 (f v1 d1 d2) r
+      else
+        let (l1, d1, r1) = split v2 s1 in
+        let l = union f l1 l2 and r = union f r1 r2 in
+        (match d1 with
+        | None -> join l v2 d2 r
+        | Some d1 -> concat_or_join l v2 (f v2 d1 d2) r)
```
panagosg7, line 356: I'm trying out the following:
```suggestion
  let decompose = function
    | Empty -> failwith "impossible"
    | Leaf { v; d } -> (Empty, v, d, Empty, 1)
    | Node { l; v; d; r; h } -> (l, v, d, r, h)

  let rec union_aux f s1 s2 (l1, v1, d1, r1, h1) (l2, v2, d2, r2, h2) =
    if h1 >= h2 then
      let (l2, d2, r2) = split v1 s2 in
      let l = union f l1 l2 and r = union f r1 r2 in
      match d2 with
      | None -> join l v1 d1 r
      | Some d2 -> concat_or_join l v1 (f v1 d1 d2) r
    else
      let (l1, d1, r1) = split v2 s1 in
      let l = union f l1 l2 and r = union f r1 r2 in
      match d1 with
      | None -> join l v2 d2 r
      | Some d1 -> concat_or_join l v2 (f v2 d1 d2) r

  and union f s1 s2 =
    match (s1, s2) with
    | (Empty, s)
    | (s, Empty) ->
      s
    | (_, _) -> union_aux f s1 s2 (decompose s1) (decompose s2)
```

bobzhang:
@panagosg7  
6e1b0de fixes the Js build
d7cdd62 adopts a more efficient relativeLoc encoding, I noticed around 1-2% perf increase and lower memory usage
We can collect data points here and split this into small diffs, thanks!

src/typing/type.ml
```diff
@@ -2656,7 +2658,12 @@ module Constraint = struct
 
     type t = TypeTerm.use_t * (speculation_id * case_id) option
 
-    let compare = Stdlib.compare
+    let compare (x, assoc1) (y, assoc2) =
+      let v = [%derive.ord: (int * int) option] assoc1 assoc2 in
+      if v = 0 then
+        Stdlib.compare x y
+      else
+        v
```
panagosg7, line 2666:
```suggestion
    let compare (x, assoc1) (y, assoc2) =
        Stdlib.compare x y
```

src/hack_forked/utils/core/get_build_id.c
```diff
@@ -18,10 +18,227 @@
...
+        // Here the string comparison is different from the one used
+        // with lexical semantics
+        int len1 = caml_string_length(v1);
+        int len2 = caml_string_length(v2);
+        SIMPLE_COMPARE(len1, len2);
+        int res = memcmp(String_val(v1), String_val(v2), len1);
+        if (res < 0)
+          return LESS;
+        if (res > 0)
+          return GREATER;
```
panagosg7, line 178:
```suggestion
        int res = memcmp(String_val(v1), String_val(v2), (len1 < len2) ? len1 : len2);
        if (res < 0)
          return LESS;
        if (res > 0)
          return GREATER;
        if (len1 < len2) {
          return LESS;
        }
        else if (len1 > len2) {
          return GREATER;
        }
```

panagosg7 (commented):
Are you okay with the following changes to restore the error diff back to zero?

bobzhang:
> Are you okay with the above changes to restore the error diff back to zero?

Looks great, committed

bobzhang:
@panagosg7 The latest commits have a few minor tweaks (reverted the two commits as you proposed).
The C code is still pretty hot, I will have a look if there're some opportunities tomorrow and stop there.

PR #8746, opened 2021-09-15
more compact encoding of relativeLoc

bobzhang:
This is part of the work of #8739
It gives a more compact encoding when the Loc shares the same line which happens a lot

### DIFF
--- a/src/parser/relativeLoc.ml
+++ b/src/parser/relativeLoc.ml
@@ -5,44 +5,31 @@
  * LICENSE file in the root directory of this source tree.
  *)
 
-type relative_position = {
-  line_offset: int;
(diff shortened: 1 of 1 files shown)


panagosg7 (approved):

mroch:
@bobzhang this is great!

I'd love if the parser produced RelativeLocs directly. then we wouldn't have to convert them before storing ASTs in sharedmem and the parser wouldn't need to know about File_key.

bobzhang:
@mroch Yes, there are some more opportunities since the loc consumes lots of memory and it could be compressed a bit more since we are only targeting 64 bit plaforms (it is wasteful to use 64 bits to encode line number). Note the conversion itself is fast since those memory allocations are typically short-lived in the minor heap.

PR #8748, opened 2021-09-16
Flow type map perf improvement

bobzhang:
This  is part of the work of #8739

907f394 add a customized set/map
ac6d8ee add the fast C comparison function
46e535a 6cfd2cf 7fb4d90 8a73104 micro tweaks using the new API


panagosg7 (commented):
The changes in get_build_id.c could be placed in a more accurately named file.

bobzhang:
@panagosg7 comments addressed and rebased

PR #8765, opened 2021-10-16
[COMMENTS] Clarify ConcretizeTypeAppsT logic after Subtyping_kit refactor

gnprice:
These comments were originally written (in e4b2c0590 / D5648147)
to appear in sequence next to each other, and each one implicitly
refers to the one before it in order to be understood.

Now that one of these steps is in Subtyping_kit while the others are
in Flow_js, add a few more words to point at each other and stitch
them together.  This reduces the amount of code spelunking needed in
order to understand how this logic is meant to work.


gnprice:
Thanks!

PR #8767, opened 2021-10-18
Fix unsound subtyping for PolyT ~> PolyT

### LINKED ISSUE #8766: Unsound subtyping on polymorphic types
Flow version: v0.162.0; introduced in v0.71.0

## Expected behavior
The type `<T>() => mixed` should not be a subtype of `<S>() => S`.

That's because for many possible types `S` we could substitute on the right, there is no type `T` that we could substitute on the left to get a subtype.

For example, the upper type can be substituted as `() => empty`. Then no matter what we choose for `T`, the resulting lower type will be `() => mixed`, which is not a subtype of `() => empty`.

## Actual behavior
Flow does accept that subtyping.

This is unsound. For example, it can be used ([try-flow](https://flow.org/try/#0N4KA9GAEAqAWCWBnSjYHsCuAbAJpApgE6FqGQBGGALpDmvogHYDkVAXCABRpuQA80AHycAlJAC8gyAFt4AD3w4RvTnwDKwsZMhqtUtAG4Q4KAEkaAYwCGjCvkgZEiyFTSR40gA5Z80-I0t6Qgt4NFsAMxJpSBsATxdYz3tXGMZ4tCpYIg5wjEYLKlDbCyCLfD4AQQAaSAAhYTleCuU6yFBISBLGRBoAcwlIbl4BTQkpZpV1Ue1dMchDEA7CfCoMQlte1SFRObkRUSMAX2MISAB5MmlSexKvKwKsWI4uns7S-ABGAcqa+s5GyATVraThbaZSPbDDQ7Gb7ERGLglIhlTjMKzMFq+TxUWLwxHvD6o9GYrw4vEvGhWXiMDDSchEAZI4L4IkYoxWAB0rgAYvJFAdIKdCHlCn4CMRSCBjiBQKc4EgUOhsHgrFhEG4iCQyJQaHQGCx2FxwsNttA9DI+UpJtDZAolHNZtpwgjTgAJIj4ZjIJkhMIUKxOPB+zJWQ25fKFP0+8rVOoNJotWptRZvbp9AacY38U3m22KFqqG2W82OqTOlPLVbrSCbEacJ4wc17A5S4ycaOsknY3FGDtojFctC8u0CoUijz2TWkAwoQpYLCQRhublYNAAd3FWsQrZAQA)) to implement arbitrary type coercion:
```js
function coerce<A, B>(x: A): B {
  const g = (o: <T>() => A): (<S>() => S) => o;
  return g(<T>() => x)();
}

const a: number = coerce('a');
a.toFixed(); // runtime error, but no Flow error anywhere
```

An example with a slightly different flavor is `<T>(T) => mixed` ~> `<S>(mixed) => S`. This, too, can be used ([try-flow](https://flow.org/try/#0N4KA9GAEAqAWCWBnSjYHsCuAbAJpApgE6FqGQBGGALpDmvogHYDkVAXCABRpuQA80AHycAlJAC8gyAFt4AD3w4RvTnwDKwsZMhqtUtAG4Q4KAEkaAYwCGjCvkgZEiyFTSR40gA5Z80-I0t6Qgt4NFsAMxJpSBsATxdYz3tXGMZ4tCpYIg5wjEYLKlDbCyCLfD4AQQAaSAAhYTleCuU6yFBISBLGRBoAcwlIbl4BTQkpZpV1Ue1dMchDEA7CfCoMQlte1SFRObkRUSMAX2MISAB5MmlSexKvKwKsWI4uns7S-ABGAcqa+s5GyATVraThbaZSPbDDQ7Gb7ERGLglIhlTjMKzMFq+TxUWLwxHvD6o9GYrw4vEvGhWXiMDDSchEAZI4L4IkYoxWAB0rgAYvJFAdIKdCHlCn4CMRSCBjiBQKc4EgUOhsHgrFhEG4iCQyJQaHQGCx2FxwsNttA9DI+UpJtDZAolHNZtpwgjTgAJIj4ZjIJkhMIUKxOPB+zJWQ25fKFP0+8rVOoNJotWptRZvbp9AacY38U3m22KFqqG2W82OqTOlPLVbrSCbEacJ4wc17A5S4ycaOsknY3FGDtojFctC8u0CoUijz2TWkAwoQpYLCQRhublYNAAd3FWsQrZAQA)) to implement coercion:
```js
function coerce<A, B>(x: A): B {
  const g = (f: <T>(T) => mixed): (<S>(mixed) => S) => f;
  return g(<T>(y: T) => x)();
}

(coerce('a'): empty);
coerce('a').toFixed(); // runtime error; still no Flow errors
```

* Link to Try-Flow or Github repo: [try-flow](https://flow.org/try/#0N4KA9GAEAqAWCWBnSjYHsCuAbAJpApgE6FqGQBGGALpDmvogHYDkVAXCABRpuQA80AHycAlJAC8gyAFt4AD3w4RvTnwDKwsZMhqtUtAG4Q4KAEkaAYwCGjCvkgZEiyFTSR40gA5Z80-I0t6Qgt4NFsAMxJpSBsATxdYz3tXGMZ4tCpYIg5wjEYLKlDbCyCLfD4AQQAaSAAhYTleCuU6yFBISBLGRBoAcwlIbl4BTQkpZpV1Ue1dMchDEA7CfCoMQlte1SFRObkRUSMAX2MISAB5MmlSexKvKwKsWI4uns7S-ABGAcqa+s5GyATVraThbaZSPbDDQ7Gb7ERGLglIhlTjMKzMFq+TxUWLwxHvD6o9GYrw4vEvGhWXiMDDSchEAZI4L4IkYoxWAB0rgAYvJFAdIKdCHlCn4CMRSCBjiBQKc4EgUOhsHgrFhEG4iCQyJQaHQGCx2FxwsNttA9DI+UpJtDZAolHNZtpwgjTgAJIj4ZjIJkhMIUKxOPB+zJWQ25fKFP0+8rVOoNJotWptRZvbp9AacY38U3m22KFqqG2W82OqTOlPLVbrSCbEacJ4wc17A5S4ycaOsknY3FGDtojFctC8u0CoUijz2TWkAwoQpYLCQRhublYNAAd3FWsQrZAQA)

gnprice:
Fixes #8766.

When checking a flow between polymorphic types PolyT ~> PolyT, we had
an attempted shortcut that wasn't sound, and also rejected some valid
flows.

The shortcut was added in 1aa5c63ba / D7454529 as an optimization to
avoid reaching the general case for flows _ ~> PolyT, which at the
time was the exponential `generate_tests`.  Happily new-generics means
that the general case is now efficient as well as correct.  So, simply
delete the code for the shortcut.

The test suite takes about 161s to run on my machine before this
change, and about 161s after it too.

The logic in question exists in two versions, one in subtyping_kit.ml
and one in flow_js.ml, in the unify function.  Both have the same
unsoundness; in fact both would accept each pair of types given as
examples in #8766.  But the unify version seems to also be dead code:
I tried replacing that case's body with an assert and ran the test
suite, and the assert never fired.  In any event, delete both.
Also fix and update some comments on this code.

In addition to the unsoundness in #8766, the tests demonstrate some
valid cases we previously rejected, including where the PolyT's have
different numbers of parameters.  Plus for completeness a couple of
cases we already got right, where one end of the flow is polymorphic
but the other isn't.

### DIFF
--- a/src/typing/flow_js.ml
+++ b/src/typing/flow_js.ml
@@ -8233,52 +8233,6 @@ struct
           resolve_id cx trace ~use_op:(unify_flip use_op) id t
         | (DefT (_, _, PolyT { id = id1; _ }), DefT (_, _, PolyT { id = id2; _ })) when id1 = id2 ->
           ()
-        | ( DefT
-              (r1, _, PolyT { tparams_loc = tparams_loc1; tparams = params1; t_out = t1; id = id1 }),
-            DefT
-              (r2, _, PolyT { tparams_loc = tparams_loc2; tparams = params2; t_out = t2; id = id2 })
-          ) ->
-          let n1 = Nel.length params1 in
-          let n2 = Nel.length params2 in
-          if n2 > n1 then
-            add_output cx ~trace (Error_message.ETooManyTypeArgs (r2, r1, n1))
-          else if n2 < n1 then
-            add_output cx ~trace (Error_message.ETooFewTypeArgs (r2, r1, n1))
-          else
-            (* for equal-arity polymorphic types, unify param upper bounds
-               with each other, then instances parameterized by these *)
-            let args1 = instantiate_poly_param_upper_bounds cx params1 in
-            let args2 = instantiate_poly_param_upper_bounds cx params2 in
-            List.iter2 (rec_unify cx trace ~use_op) args1 args2;
-            let inst1 =
-              let r = reason_of_t t1 in
-              mk_typeapp_of_poly
-                cx
-                trace
-                ~use_op
-                ~reason_op:r
-                ~reason_tapp:r
-                id1
-                tparams_loc1
-                params1
-                t1
-                args1
-            in
-            let inst2 =
-              let r = reason_of_t t2 in
-              mk_typeapp_of_poly
-                cx
-                trace
-                ~use_op
-                ~reason_op:r
-                ~reason_tapp:r
-                id2
-                tparams_loc2
-                params2
-                t2
-                args2
-            in
-            rec_unify cx trace ~use_op inst1 inst2
         | (DefT (_, _, ArrT (ArrayAT (t1, ts1))), DefT (_, _, ArrT (ArrayAT (t2, ts2)))) ->
           let ts1 = Base.Option.value ~default:[] ts1 in
           let ts2 = Base.Option.value ~default:[] ts2 in
--- a/src/typing/subtyping_kit.ml
+++ b/src/typing/subtyping_kit.ml
@@ -978,66 +978,20 @@ module Make (Flow : INPUT) : OUTPUT = struct
      * we already deal with statements of this form when checking polymorphic
      * definitions! In particular, statements such as "there is some X:U...")
      * correspond to "create a type variable with that constraint and ...", and
-     * statements such as "show that for all X:U" correspond to "show that for
-     * both X = bottom and X = U, ...".
+     * statements such as "show that for all X:U" correspond to "introduce a
+     * GenericT with bound U".
      *
      * Thus, all we need to do when checking that any type flows to a
      * polymorphic type is to follow the same principles used when checking that
      * a polymorphic definition has a polymorphic type. This has the pleasant
      * side effect that the type we're checking does not itself need to be a
-     * polymorphic type at all! For example, we can let a non-generic method be
-     * overridden with a generic method, as long as the non-generic signature
-     * can be derived as a specialization of the generic signature.
+     * polymorphic type at all.
      *)
-    (* some shortcuts **)
+    (* shortcut: types are identical *)
     | (DefT (_, _, PolyT { id = id1; _ }), DefT (_, _, PolyT { id = id2; _ }))
       when Poly.equal_id id1 id2 ->
       if Context.is_verbose cx then prerr_endline "PolyT ~> PolyT fast path"
-    | ( DefT (r1, _, PolyT { tparams_loc = tparams_loc1; tparams = params1; t_out = t1; id = id1 }),
-        DefT (r2, _, PolyT { tparams_loc = tparams_loc2; tparams = params2; t_out = t2; id = id2 })
-      ) ->
-      let n1 = Nel.length params1 in
-      let n2 = Nel.length params2 in
-      if n2 > n1 then
-        add_output cx ~trace (Error_message.ETooManyTypeArgs (r2, r1, n1))
-      else if n2 < n1 then
-        add_output cx ~trace (Error_message.ETooFewTypeArgs (r2, r1, n1))
-      else
-        (* for equal-arity polymorphic types, flow param upper bounds, then
-         * instances parameterized by these *)
-        let args1 = instantiate_poly_param_upper_bounds cx params1 in
-        let args2 = instantiate_poly_param_upper_bounds cx params2 in
-        List.iter2 (fun arg1 arg2 -> rec_flow_t cx trace ~use_op (arg2, arg1)) args1 args2;
-        let inst1 =
-          let r = reason_of_t t1 in
-          mk_typeapp_of_poly
-            cx
-            trace
-            ~use_op
-            ~reason_op:r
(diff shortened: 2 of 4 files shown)


panagosg7:
Hi @gnprice !

@mvitousek and I took a closer look at this work, and it does indeed uncover an important soundness hole in the typechecker, so thanks again for raising this issue.

Mike, however, came up with a pattern of programs that now raise Flow errors (see [repro1](https://flow.org/try/#0FASwtgDg9gTgLgAgFQIIYGcECUCmqDGiAZjFGAgOQx6EUDcwjIAdnDjEQTggMpyw4APH1RsAfAgDewBAgDmOOIIDSALgQASZTgCe6YXFE4xYgBQBrXeuUBKdRoCiAGxxgcrACo6IQkWwA0CMpiDAC+jAAmOPhOqNQIOAAe0PAIRACuzIQgUMwI+LnM0Up+xqYyCOj81Op8AgZGYv7AdgjCtdW+huKBAAqkEOjqklUCtYEAdFOhZhUAwmTQRazquARwEwuQue5wXj6C-VCDTS0IALwSc7Ho+muEm4s7rIcD6CYMwHDe3KUXUgAfCqodRVGAsOTNWQAI3UzHSYGh7GaANCnyiMTi3AAbnFKp0OvVSmJGAVmFUEAB3EBwAAWdRqbR4hOoDR6CCOg2Go0ZPEm01msi2S12qxocA0wueex+r2O72aNguVxud3FkqeyyUnPeEnO+UKxTZZR5OBsDCAA) and [repro2](https://flow.org/try/#0C4TwDgpgBAysD2AnCAeOBDYEB8UC8UA3gFBRQDmEwKA0gFxQAkNEIAzmsJjtgBQDWrBjQCUDRgFEANhAC2EAHbAAKuFQYsAGig1sAbmIBfYsQAmEAMZT0yKBAAeYJMCgAzAK4KLwAJbwFUBb+CpbUGji8YlBoDHBI6lxY2NoACojwYGwMhGwIyLHaAHTFhnwi+LgAgoiI6CAoaRls+iagkFAAsiDh+EQAPqRQ6Ay5iD4K5JqDAEYMCu6y0xCIU32GBsRBCrlQAO4+wAAWcfnRMLF5qF3hyVCNmdm58QVQxYWlkRVQ1bX19829LYhbwoa6JCIiPRAA)). These two repros hit another bug in the checker -- unrelated to the current PolyT issue (see [standalone repro](https://flow.org/try/#0C4TwDgpgBAGlC8UDeUCGAuKA7ArgWwCMIAnAGigMwGdhiBLLAcygF8AoDgMxywGNg6AeyxoAPABVMMAHwAKYJnHkA7pgAkAaQggqE6QEp1AUQA2EPBCzBx4CBPKbtumdORsoHqMQjAcxEcAA2soAugDcbOxAA)). The path to these bugs get enabled when the PolyT ~> PolyT shortcut is bypassed.

Unfortunately, these patterns appear far too often in our internal repositories. (In one of our repos there were so many errors raised that the server ran out of memory.)

Also, even without these errors, it is possible that the fallback rules that would handle PolyT ~> PolyT constraints could cause a performance regression. It's hard to test this hypothesis, however, due to the sheer number of errors.

On the bright side, I think there is a path to address at least the soundness issue that you raised by properly fixing the PolyT ~> PolyT rule. The fix is to apply a sound rule for subtyping of the polymorphic case. See for example https://www.cis.upenn.edu/~bcpierce/papers/fsubnew.ps (page 3). According to rule ∀-local for example the change in Flow would be something like this: https://gist.github.com/panagosg7/29eb4a0b60a4e6103f7922b235905e49

This seems to address the coerce examples from https://github.com/facebook/flow/issues/8766, but it doesn't fix the arity-related tests added here. Arguably the latter are separate completeness, rather than soundness, issues. They could still be fixed as part of future work. The error diff from our repos with these changes is a lot more modest (non-blocking).

We really appreciate the effort you put into this and would like to include your extensive test suite. So, would you mind submitting those as a PR? Alternatively, we could also repurpose the current PR and change the wording a little bit.

PR #8773, opened 2021-10-28
[lib] Add copyFile overloaded definition

Brianzchen:
`fs.copyFile` doesn't require the `mode` param, it can have 3 args with the 3rd being the callback
ref: https://nodejs.org/api/fs.html#fscopyfilesrc-dest-mode-callback

Usage looks like
```
      fs.copyFile(srcPath, destPath, err => {
        if (err) {
          rej(err);
        } else {
          res();
        }
      });
```


tests/node_tests/node_tests.exp
```diff
@@ -1565,7 +1565,7 @@ Error --------------------------------------------------------------------------
 Unexpected end of input, expected a valid JSON value
 
    1|
-
```
mroch, line 1568: removing trailing whitespace is why the tests are failing 😞

try `runtest.sh -r` again, I guess? unless it's git that's stripping it?

mroch (changes requested):

tests/node_tests/node_tests.exp
Brianzchen: Nah it was me. Crazy didn't expect that but will keep it in mind next time

Issue #8251, opened 2020-01-07
Type unrefinement doesn't work when function is in conditional


dannyharding10:
Flow version: 0.115.0

## Expected behavior
Calling any function after a type refinement should undo that type refinement

## Actual behavior
When the function is called in the same conditional statement as the type refinement, the type refinement remains intact.

It is likely that flow is missing a lot of bugs because of this issue 😱 

* Link to Try-Flow or Github repo: https://flow.org/try/#0C4TwDgpgBAglC8UDeB7MwD8AuZBrCIOARiigDYQCGAdgL60BQDAxitQM7BTsoC2EAMQCu1ZgigAKSjhgBKBAD5kDAJAATCBWDRKAOjTAGUY1ABOEYENPUoAM0pl2EBowa2RzYAEs2UZgAsIZlx2KRl5JCMoL1tJAEI9AygAH2TuPkEPKVkIqONzS2soxmMogHoy9P5hUWyoCqgAVVE+fmouYH8vdigyL2poYBRuCEHAqAhTUxRTKE7KLnZ-FCEyNShKMEhKU3LKgGUACQB5RoAZABEoAFEAJVvj26gAIWuAYRhG-euqzNEoABSXwAKlALtcztdgdcrgY4lFWBxyBBdGQUABzKT6dC6fAgWRRFxMBowTxCBxkEBmETRQyIzhQXggY5EABWMnEqHQOCQeJwwFMQgg9CYASCIQkTJZrNkQA

Here's a code sandbox of the same code. You can see that it breaks
https://codesandbox.io/s/mystifying-agnesi-79k3k?fontsize=14&hidenavigation=1&theme=dark

Issue #8779, opened 2021-11-01
Havoc forgotten after for-loops and do-while loops


gnprice:
Flow version: v0.163.0

## Expected behavior
After a refinement is invalidated ("havoced"), in particular by a function call or a write to a property, it should stay invalidated wherever the control flow can reach from there (unless it gets refined again.)

## Actual behavior
When the invalidation happens in the body or update-expression of a for-loop, or the body or test of a do-while loop, or a block with a labelled break, it can get forgotten after the loop is complete.  This is unsound.

In particular we can write `coerce`.  For example, using a `for` body:
```js
const v: { f?: mixed } = {};

function store(x: mixed) {
  v.f = () => x;
};

function coerce<A, B>(x: A): B {
  v.f = () => { throw null };  // Refine v.f.
  for (let i = 0; i < 1; i++) {
    store(x);  // Invalidate the refinement.
  }
  return v.f(); // Should error!  Trying to still use the refinement.
}
```

[...]

PR #8781, opened 2021-11-01
Fix forgotten havoc, part 1: havoc_heap_refinements

gnprice:
When there's a branch in the control flow, we rely on some tricky
juggling of state in order to get the right environments (with the
right refinements, initialization states, etc.) at the right places.

A key workhorse of that juggling is Env.merge_env, which we invoke
where control-flow paths join up (at a number of callsites directly in
statement.ml, and several more indirectly through Env.in_refined_env)
in order to merge what happened in two different incoming paths to
produce an appropriate environment for the outgoing path.

The work that Env.merge_env does is all independent between different
bindings and refinements; it iterates through them.  Usually most of
them weren't touched, and are the same on all the paths; so as an
optimization, it relies on being passed a set of bindings and
refinement keys (a Changeset.t) which is supposed to identify all of
the binding-entries and refinements that may differ between the envs.

To make that work, everything in the Env module that mutates the
current environment needs to ensure that the current global changeset
has the appropriate keys.

We had a bug where Env.havoc_heap_refinements, which we use in
particular at function and method calls, wasn't adding the havoced
keys to the changeset.  As a result, when we later merge environments
at the end of some enclosing control-flow construct, we could forget
to propagate the invalidation to the surviving environment.

Fix that by adding the keys for any refis we find here to invalidate.

Along the way, simplify Scope.havoc_all_refis by dropping its unused
`?name` parameter.  There's never been a caller that actually passed
that parameter, and without it the function better matches its name
and becomes very simple, easy to reason about.  The simpler interface
also lets us write a clean, concise comment documenting what it's for.

Fixes: #8251

### DIFF
--- a/src/typing/env.ml
+++ b/src/typing/env.ml
@@ -1652,11 +1652,19 @@ module Env : Env_sig.S = struct
       Changeset.iter_type_updates havoc_entry havoc_refi
     )
 
-  (* Clear entries for heap refinement pseudovars in env.
-     If name is passed, clear only those refis that depend on it.
-     Real variables are left untouched.
-  *)
-  let havoc_heap_refinements () = iter_scopes Scope.havoc_all_refis
+  (* Clear all heap refinements (refis) in env.
+
+     Refinements on names bound directly in a scope are left
+     untouched; for those, see [havoc_local_refinements]. *)
+  let havoc_heap_refinements () =
+    iter_scopes (fun scope ->
+        if Changeset.Global.is_active () then
+          scope
+          |> Scope.iter_refis (fun key _ ->
+                 Changeset.Global.change_refi (scope.id, key, Changeset.Write)
+             );
+        Scope.havoc_all_refis scope
+    )
 
   let havoc_local_refinements ?(all = false) cx =
     iter_scopes (fun scope ->
--- a/src/typing/scope.ml
+++ b/src/typing/scope.ml
@@ -376,6 +376,9 @@ let remove_entry name scope = scope.entries <- NameUtils.Map.remove name scope.e
 (* get entry from scope, or None *)
 let get_entry name scope = NameUtils.Map.find_opt name scope.entries
 
+(* use passed f to iterate over all scope refis *)
+let iter_refis f scope = Key_map.iter f scope.refis
+
 (* use passed f to update all scope refis *)
 let update_refis f scope = scope.refis <- Key_map.mapi f scope.refis
 
--- a/src/typing/scope.ml
+++ b/src/typing/scope.ml
@@ -406,9 +406,8 @@ let havoc_refis ?name ~private_ scope =
     | Some name -> scope.refis |> filter_refis_using_propname ~private_ name
     | None -> Key_map.empty)
 
-let havoc_all_refis ?name scope =
-  havoc_refis ?name ~private_:false scope;
-  havoc_refis ?name ~private_:true scope
+(* havoc a scope's refinements: clear all its refis *)
+let havoc_all_refis scope = scope.refis <- Key_map.empty
 
 (* havoc a scope:
    - clear all refinements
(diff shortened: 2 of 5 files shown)


gnprice:
Marked the PR as draft only because it sits atop #8780 (which adds tests for this and related issues), and it probably makes sense to merge them as separate commits -- that way it's clear which test cases were already passing, and which were fixed here.

In Phabricator I'd represent this as a dependent diff stacked on another. Not sure if your importer plugin offers a way to get it to smoothly translate into that. Happy to tweak the PRs to make that work if so, or to adjust in some other way that'd be helpful for your workflow; let me know.

jbrown215:
Hi @gnprice -- thanks for the fix! I haven't taken a look yet but I wanted to let you know that we're currently working on replacing this env module entirely as part of the local type inference project. I'm hoping to ship the replacement in Q1 of 2022. Please don't work too hard on the Env module until then-- I'd hate to see your efforts go towards something that is going to be replaced so soon.

The new implementation currently lives in src/typing/new_env.ml, but that module is mostly just interpreting artifacts that come from src/analysis/env_builder/name_resolver.ml. The name_resolver tracks exactly which writes reach a read of a variable (almost ssa-style) along with the refinements that apply to those reads. I believe we havoc heap refinements correctly there. The new_env module consumes those write and read locations and turns them into types that can be queried by the type checker.

Since it is not fully integrated into Flow, we have a bunch of tests in `src/analysis/env_builder/__tests__/env_builder_refinement_tests.ml` which print the list of writes that reach a read + the refinements associated with them. My current focus is driving down the test case failures associated with using this new_env.ml implementation. If you pass `-e` to `runtests` you can see the results for yourself.

If you have any questions about any of that work please let me know and I'd be happy to explain.

I'm adding this PR to my TODO list and will hopefully get back to you by the end of the week.

gnprice:
Sounds good, thanks!

> I wanted to let you know that we're currently working on replacing this env module entirely as part of the local type inference project. I'm hoping to ship the replacement in Q1 of 2022.

Cool. I'd gathered from the commit logs that something like that was in the works, but wasn't sure how the pieces fit together; these details are good context, thanks. Will take a look at that code too.

I'll refrain from proposing any major changes to the env subsystem, since you're already in the middle of one. :-) I don't mind the time I've spent on it so far, even with it slated to be replaced soon; most of the time was on understanding the architecture, and those have a tendency to live on even when most of the code is rewritten.

This fix and the other two in my queue -- all fixing soundness bugs tested in #8780 -- are all pretty small code-wise, so I hope they won't be much work on your side to review, either. I'd guess the hardest part would be if one or more of them turns out to generate a lot of errors in your internal codebases; but if so, then dealing with those one way or another is something you'd need to do in any case as part of shipping a new, sounder env implementation.

The tests in #8780 should be useful tests for the new env module, too. Even after the fixes I have, #8779 will still be open, leaving TODO comments in that test file; with any luck, once you've implemented all the relevant pieces your new implementation will resolve them. And if not, that'll be a useful lead to debug ;-)

jbrown215:
I'm going to import and land #8780  first and then we'll get this one reviewed!

gnprice:
And I see #8780 is merged! Rebased this PR on top.

gnprice:
Ping :slightly_smiling_face: Is there anything I can add to help make this easy for you to review?

jbrown215:
Nope, sorry for the hold up. I actually just rebased your diff yesterday and took a look at the error diff. I found an error I couldn't quite figure out but also didn't manage to cut down the repro to something simpler to reproduce.

gnprice:
Cool, thanks for the update. Happy to take a look if you find a repro that you can share.

This is the kind of change where if it produces a mysterious error, then *probably* it's a bug somewhere else in the system (or else a subtle ill-typing in the code being checked) that was previously covered up by this bug. If so, then any time you spend debugging that will just be a head start on something you'd probably need to do anyway as part of rolling out local type inference. Hoping that's how this one turns out.

Issue #8782, opened 2021-11-05
replace-existentials codemod always errors


lyleunderwood:
Flow version: 0.162.1

## Expected behavior

`flow codemod replace-existentials` functions without error.

## Actual behavior

I've tried this codemod on multiple machines and had other developers try it, and always get the same error:

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.3 LTS"

$ ./node_modules/.bin/flow version
Flow, a static type checker for JavaScript, version 0.162.1

$ ./node_modules/.bin/flow codemod replace-existentials
Unhandled exception: (Failure hd)
Raised at file "stdlib.ml", line 29, characters 17-33
Called from file "list.ml", line 30, characters 10-23
Called from file "src/commands/commandUtils.ml", line 13, characters 4-32
```

gnprice:
Indeed, I reproduce.

I've just sent a fix as #8783. But the debugging that led to that fix also gives the following workaround.

When you see that traceback, pretend that it was instead an error message reading:
> Error: filenames or --input-file are required

(The filenames can be directories, so `flow codemod replace-existentials --write .` may be the whole command you need.)

When I try that in a small test repo, the actual codemod works great:
```diff
-const x: * = 3;
+const x: number = 3;
```

PR #8783, opened 2021-11-06
[codemod] Give proper error when run without arguments

gnprice:
When a codemod command is run with no arguments, it gives an error,
because it requires a list of filenames.

But rather than a proper error message, it was crashing with an
OCaml traceback:

  Unhandled exception: (Failure hd)
  Raised at file "stdlib.ml", line 29, characters 17-33
  Called from file "list.ml", line 30, characters 10-23
  Called from file "src/commands/commandUtils.ml", line 13, characters 4-32

There was already a conditional intended to provide a helpful error
message for the lack of filenames.  But earlier in the function, if
there's no --root, we were trying to take the first filename in order
to guess the root.  That caused the crash.

The fix is to move that check sooner.  (Nothing touches `filenames`
in between, so this doesn't have any other effect.)  Now the command
fails properly with:

  Error: filenames or --input-file are required

Also move a neighboring comment to the line it looks like it was
intended to apply to.

This fix would be good to backport to a v0.162.2, so that it can
apply to the replace-existentials codemod advertised in the v0.163.0
release notes.

Fixes: https://github.com/facebook/flow/issues/8782

### DIFF
--- a/src/commands/codemodCommand.ml
+++ b/src/commands/codemodCommand.ml
@@ -44,7 +44,12 @@ let main (module Runnable : Codemod_runner.RUNNABLE) codemod_flags () =
     codemod_flags
   in
   initialize_environment ();
+  (* Normalizes filepaths (symlinks and shortcuts) *)
   let filenames = CommandUtils.get_filenames_from_input input_file filenames in
+  if filenames = [] then (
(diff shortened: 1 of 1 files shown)


gkz:
Thanks for the fix @gnprice

PR #8814, opened 2021-12-02
Add 'navigate' as valid request ModeType

comp615:
Currently, in a SW, checking if the request is a full document reload might be done as request.mode === 'navigate'. However, this raises a flow error.

https://developer.mozilla.org/en-US/docs/Web/API/Request/mode

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -1497,7 +1497,7 @@ declare class URLSearchParams {
 
 type CacheType =  'default' | 'no-store' | 'reload' | 'no-cache' | 'force-cache' | 'only-if-cached';
 type CredentialsType = 'omit' | 'same-origin' | 'include';
-type ModeType = 'cors' | 'no-cors' | 'same-origin';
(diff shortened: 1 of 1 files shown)


lib/bom.js
```diff
@@ -1497,7 +1497,7 @@ declare class URLSearchParams {
 
 type CacheType =  'default' | 'no-store' | 'reload' | 'no-cache' | 'force-cache' | 'only-if-cached';
 type CredentialsType = 'omit' | 'same-origin' | 'include';
-type ModeType = 'cors' | 'no-cors' | 'same-origin';
+type ModeType = 'cors' | 'no-cors' | 'same-origin' | 'navigate;
```
mroch, line 1500: missing closing quote :)

mroch (changes requested):

lib/bom.js
comp615: Well that's embarrassing

mroch (approved):
thanks!

PR #8819, opened 2021-12-25
Update options.mdx

bobychaudhary:
Minor typo mistake

```diff
-signature-verfication-failure
+signature-verification-failure
```


gkz:
Can you update the same content currently in the `website/` directory? We're in the process of migrating our site and the current site at flow.org has content from `website/`, so both need to be updated. After you update I can import and merge

bobychaudhary:
Okay, sure 👍

PR #8821, opened 2021-12-28
[lib] globalise tls connect options

Brianzchen:
I'm trying to type 'pg' library right now but cannot access the type of tls options. Would be great to make it global just like the a lot of the node module.


SamChou19815:

PR #8880, opened 2022-05-05
An economic way of doing unicode lexing

bobzhang:
This would reduce the code size from 150k to 15k loc.
Note this is work in progress, but I noticed something exciting.

```
flow_parser.js 3.2M -> 1.1M
flow.js 5M -> 3.2M
```
I will do some more evaluations when I fixed the CI issues


samwgoldman:
I'm getting some parse errors when testing these changes internally. I was going to minimize the failing examples, but then I remembered that we have a test262 baseline, which is also affected by these changes.

The dune target parse_test262 defined in src/parser/test will produce a binary that runs the Flow parser over a test262 checkout. Internally the last import is https://github.com/tc39/test262/commit/19da3ca0757248f7595ee09d532bb83dd438f2b5.

Something like `parse_test262 -s path/to/test262/test -q > src/parser/test/test262_baseline.txt` will show you the diff from our current test262 status.

samwgoldman:
This looks good to me! I'll get this stamped and merged internally on during the day Friday US/Pacific. I'll run some longer perf experiments overnight to see if there is any observable effect from these changes on end-to-end times.

One thing I noticed from the updated test262 baselines is that Flow now accepts some programs with unicode escapes in keywords. I haven't paged in all of this code, but it kind of seems like we're being a bit loose with identifiers vs. keywords. But even with a few new false positives, the improvement in test262 overall is impressive, and awesome to see!

PR #8881, opened 2022-05-05
[lib] Add rm related funcs to `fs` libs

Brianzchen:
Add a bunch of missing functions from the lib definitions of `fs` mostly related to rm.

Implementation based off [fs.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts#L1419)


lib/node.js
```diff
@@ -986,8 +988,82 @@ declare module "fs" {
...
+  declare type RmDirOptions = {
+    /**
+     * If an `EBUSY`, `EMFILE`, `ENFILE`, `ENOTEMPTY`, or
+     * `EPERM` error is encountered, Node.js will retry the operation with a linear
+     * backoff wait of `retryDelay` ms longer on each try. This option represents the
+     * number of retries. This option is ignored if the `recursive` option is not
+     * `true`.
+     * @default 0
+     */
+    maxRetries?: number | undefined,
```
mroch, line 1001: use `void` instead of `undefined` (the tests are complaining)

PR #8883, opened 2022-05-11
Implement lock-free skip list

samwgoldman:
Summary:
This diff adds a concurrent lock-free skip list for Flow's shared memory
segment. The data structure is generic, but currently only exposed for modeling
sets of files, sorted by filename.

This skip list supports concurrent lock-free insert and delete and wait-free
search operations in O(logn) time and O(n) space.

In later work, we will use these skip lists to store the reverse dependencies of
modules.

Differential Revision: D36217992

### DIFF
--- a/src/heap/sharedMem.ml
+++ b/src/heap/sharedMem.ml
@@ -42,14 +42,15 @@ type tag =
   | Lib_file_tag
   | Haste_module_tag
   | File_module_tag
+  | Sklist_tag
   (* tags defined below this point are scanned for pointers *)
-  | String_tag (* 11 *)
+  | String_tag (* 12 *)
   | Int64_tag
   | Docblock_tag
   | ALoc_table_tag
   | Type_sig_tag
   (* tags defined above this point are serialized+compressed *)
-  | Serialized_tag (* 16 *)
+  | Serialized_tag (* 17 *)
   | Serialized_resolved_requires_tag
   | Serialized_ast_tag
   | Serialized_file_sig_tag
@@ -63,8 +64,8 @@ let tag_val : tag -> int = Obj.magic
 (* double-check integer values are consistent with hh_shared.c *)
 let () =
   assert (tag_val Entity_tag = 0);
-  assert (tag_val String_tag = 11);
-  assert (tag_val Serialized_tag = 16)
+  assert (tag_val String_tag = 12);
+  assert (tag_val Serialized_tag = 17)
 
 (* Addresses are relative to the hashtbl pointer, so the null address actually
  * points to the hash field of the first hashtbl entry, which is never a
@@ -784,6 +785,10 @@ module NewAPI = struct
 
   type 'a addr_tbl
 
+  type 'a sklist
+
+  type 'a sknode
+
   type ast
 
   type docblock
@@ -810,9 +815,11 @@ module NewAPI = struct
 
   type size = int
 
-  let bsize_wsize bsize = bsize * Sys.word_size / 8
+  let bsize_wsize wsize = wsize * Sys.word_size / 8
 
   let addr_offset addr size = addr + bsize_wsize size
+
+  let i64 = Int64.of_int
 
   let get_heap () =
     match !heap_ref with
@@ -857,8 +864,20 @@ module NewAPI = struct
   external unsafe_write_bytes_at : _ addr -> bytes -> pos:int -> len:int -> unit = "hh_write_bytes"
     [@@noalloc]
 
-  external compare_exchange_weak : _ addr -> int -> int -> bool = "hh_compare_exchange_weak"
+  external load_acquire : _ addr -> (int64[@unboxed]) = "hh_load_acquire_byte" "hh_load_acquire"
     [@@noalloc]
+
+  external store_release : _ addr -> (int64[@unboxed]) -> unit
+    = "hh_store_release_byte" "hh_store_release"
+    [@@noalloc]
+
+  external compare_exchange : weak:bool -> _ addr -> (int64[@unboxed]) -> (int64[@unboxed]) -> bool
+    = "hh_compare_exchange_byte" "hh_compare_exchange"
+    [@@noalloc]
+
+  let compare_exchange_weak = compare_exchange ~weak:true
+
+  let compare_exchange_strong = compare_exchange ~weak:false
 
   (** Addresses *)
 
@@ -866,7 +885,7 @@ module NewAPI = struct
 
   (* Write an address at a specified address in the heap. The caller must ensure
    * the given destination has already been allocated. *)
-  let unsafe_write_addr_at heap dst addr = buf_write_int64 heap dst (Int64.of_int addr)
+  let unsafe_write_addr_at heap dst addr = buf_write_int64 heap dst (i64 addr)
 
   (* Write an address into the given chunk and advance the chunk address. This
    * write is not bounds checked; caller must ensure the given destination has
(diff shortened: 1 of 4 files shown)


bobzhang:
Is this a port of runtime/If_skiplist.c from C to OCaml, would it make sense to keep the C version and expose it via bindings? I did not have a look at the implementation details, will have a closer look later this week.

samwgoldman:
It's not a straight port, the underlying algorithm is essentially the same, but there are a few differences which are enumerated in inline comments.

An important difference is that the skip list is stored in Flow's managed shared memory segment, whereas the OCaml runtime skip list is manually managed using malloc/free. Instead of a C struct we have something more like OCaml's blocks, word-sized data blocks containing either (relative) addresses or tagged integers, prefixed with a word-sized header.

Also different from the OCaml runtime implementation, and the textbook implementation referenced from there, is that we don't have sentinel head/tail nodes, and thus no need to define minimum/maximum values.

bobzhang (commented):
I did not finish the review, since reviewing concurrency code is hard and it takes time.
I knew sequential skiplist very well, had a working understanding of lock free skiplist. 
Some high level questions:
- The benefit of lockfree skip list will only be trickled down until OCaml 5 get merged internally, is that correct?
- Reviewing code of concurrency is difficult. The current implementation seems to mix lockfree skiplist with sharedMem encoding, would it make sense to factor it out as a stand alone module, a text book implementation of lockfree skip list in a stand alone module would be much easier to review than the current case.

src/heap/sharedMem.ml
```diff
@@ -1012,6 +1050,8 @@ module NewAPI = struct
 
   let read_string addr = read_string_generic String_tag addr 0
 
+  let compare_string a b = String.compare (read_string a) (read_string b)
+
```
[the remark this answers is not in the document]
samwgoldman: Yes. This could be further optimized to a memcmp over the shared segment directly. Let's leave this for follow-up work, though.

samwgoldman:
Thanks for looking! Appreciate having your eyes on this.

> The benefit of lockfree skip list will only be trickled down until OCaml 5 get merged internally, is that correct?

I'm not sure I understand. Once Multicore OCaml is a real thing, we should definitely investigate replacing Flow's shared memory stuff altogether with just regular shared OCaml values, and we might choose different data structures to represent the shared data. Is that what you're getting at?

> Reviewing code of concurrency is difficult. The current implementation seems to mix lockfree skiplist with sharedMem encoding, would it make sense to factor it out as a stand alone module, a text book implementation of lockfree skip list in a stand alone module would be much easier to review than the current case.

I think I might have failed to completely describe the context around these changes. This is a lock free skip list _for_ Flow's shared memory infrastructure. We need to allocate the objects that constitute the skip list in shared memory, which means that those objects need to be encoded correctly for that.

Maybe I'm not understanding your suggestion, and you have some proposal to separate the algorithms for the various operations and the underlying data representation?

src/heap/sharedMem.ml
```diff
@@ -1071,6 +1111,298 @@ module NewAPI = struct
...
+        (* We are at the end of the list at this level. *)
+        down level pred curr
+      else
+        let succ_addr = sknode_succ_addr level curr in
+        let succ64 = load_acquire succ_addr in
+        let succ = sklist_unmark succ64 in
+        if sklist_is_marked succ64 then
+          (* `curr` is logically deleted. Try to physically remove `curr` from
+           * this level by updating `pred` to point directly to `succ`. *)
+          if compare_exchange_strong pred (i64 curr) (i64 succ) then
```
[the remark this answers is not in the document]
bobzhang: it seems to depend on if compare_exchange is inlined or not

bobzhang:
I misread your intention, it seems you want to add a lockless skiplist data structure for shared memory given that we already have lockless hashtable. I will check with you during the meeting.

src/heap/sharedMem.ml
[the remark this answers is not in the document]
samwgoldman: I can run the tests under gdb, set a breakpoint on `hh_compare_exchange`, and observe that callers do not allocate Int64.t boxes here. I'm looking at x86, and the int values are just shifted right to decode from the tagged int representation and stored into the rsi/rdx registers before the call into C.

Once I publish my work that uses this code, we can run it through profilers to see what jumps out. Maybe there are surprise allocations, but it will be much easier to see later, and we can fix it then. Sound good?

src/heap/sharedMem.ml
```diff
@@ -1071,6 +1111,298 @@ module NewAPI = struct
...
+   * levels at 12 lets us store 16,777,216 (4^12) elements. *)
+  let sklist_num_levels = 12
+
+  (* Skip lists are probabilistic. Each node is created with a random height,
+   * given by this function. We take p=1/4, meaning that 3/4 of nodes will have
+   * height 1, 3/16 will have height 1, 3/64 will have height 2, and so on. *)
+  let sklist_random_level () =
+    let num_rolls = sklist_num_levels - 1 in
+    (* Random.int only gives 30 bits of randomness and we use 2 bits per roll *)
+    assert (num_rolls <= 15);
```
bobzhang, line 1158: could this assertion be lifted to toplevel

src/heap/sharedMem.ml
```diff
@@ -1071,6 +1111,298 @@ module NewAPI = struct
...
+    let mask = (1 lsl (2 * num_rolls)) - 1 in
+    let r = ref (Random.bits () land mask) in
+    let l = ref 0 in
+    while !r land 3 == 3 do
+      (* two random bits will be set with probability 1/4 *)
+      incr l;
```
bobzhang, line 1164: There is an interesting note on the ocaml implementation that " the least significant bits are
   "less random" than the most significant bits with a modulus of 2^m"

src/heap/sharedMem.ml
```diff
@@ -1071,6 +1111,298 @@ module NewAPI = struct
...
+        else if cmp (sknode_get_data curr) data < 0 then
+          (* Continue searching across a level until we find a node with a value
+           * that is greater than or equal to the target value. *)
+          search level succ_addr succ
+        else
+          down level pred curr
+    and down level pred curr =
+      (* Record predecessor and successor of the target at this level. *)
+      preds.(level) <- pred;
+      succs.(level) <- curr;
```
bobzhang, line 1261: Note there are several in-efficiencies if this is in a hot loop.

- unneededs bound checking
- hidden allocation for the mutual recursive function
- `cmp` passed as a function will trigger `caml_apply`
- `compare_file_by_name` seems to trigger allocation as mentioned above

bobzhang (approved):
As discussed offline, my preference for concurrent data structure would be encoded in low level C language so that you could have better control (no hidden allocation) and best perf delivered. 
But I trust your judgement assuming this piece of code is extensively tested.

src/heap/sharedMem.ml
samwgoldman: This is true for the LCG generator that they're using, but the OCaml Random module uses an algorithm that provides better mixing.

src/heap/sharedMem.ml
samwgoldman: It's optimized out by ocamlopt, probably just constant propagation, but sure I'll move this out.

src/heap/sharedMem.ml
samwgoldman: The loop is not particularly hot. It's a log(n) search with low n. We might iterate more if the list is under contention, but contention will be rare in practice.

The bounds checking I think is well worth the cost in this context. In general I'll take bounds checking unless profiling show that it's a meaningful bottleneck...

Regarding the comparison, we can optimize for sure, assuming it makes a difference in practice. I suspect the cost of caml_apply and allocations is likely overwhelmed by the actual string comparison. We could assign each file a unique integer ID and use that instead.

Issue #8886, opened 2022-05-29
Error thrown when parsing regexp with `d` flag on Node.js v10


fisker:
Flow version: 0.179.0

## Expected behavior

Not throw error

## Actual behavior

Error thrown

* Link to Try-Flow or Github repo:

```text
> require('flow-parser').parse('/./d')
Thrown:
SyntaxError: Invalid flags supplied to RegExp constructor 'd'
```

The value of RegExpLiteral should be `null` per ESTree spec https://github.com/estree/estree/blob/master/es5.md#regexpliteral

mroch:
this was fixed in 30e59baea97f8cdf84065d312df52d98fdb2e1d1 which is already in 0.179.0. are you sure you're using 0.179?

mroch:
or do you mean that Flow parses it but because of the `value` field, the runtime is throwing if it doesn't support `d`?

fisker:
> are you sure you're using 0.179?

Yes, I'm sure.

> or do you mean that Flow parses it but because of the value field, the runtime is throwing if it doesn't support d?

I think so.

PR #8887, opened 2022-05-29
fix js exception due to regex flags

mroch:
Summary:
in ESTree, the `value` property of a regex is an instantiated `RegExp` object. if it can't be represented for some reason, the `value` should be `null`. there is also a `regex` property with the pattern and flags as strings so you can still tell what it was.

JSON doesn't have regexes, so `flow ast` already returns null. `flow_parser.js` runs in a JS engine so it does try to create a `RegExp`. when creating a `RegExp` fatals, we previously tried to recover in a bogus way. it should return `null` instead.

this fixes a crash when the browser engine doesn't support the latest regex flags, like `y`, `u` or `d`.

Fixes https://github.com/facebook/flow/issues/8886

Changelog: [parser] Fix a runtime exception when parsing regexes via the JS version of the parser

Differential Revision: D36754295

### DIFF
--- a/src/parser/flow_parser_js.ml
+++ b/src/parser/flow_parser_js.ml
@@ -30,19 +30,17 @@ end = struct
 
   let null = Js.Unsafe.inject Js.null
 
-  let regexp loc pattern flags =
+  let regexp _loc pattern flags =
     let regexp =
       try
         Js.Unsafe.new_obj (Js.Unsafe.pure_js_expr "RegExp") [| string pattern; string flags |]
       with
       | _ ->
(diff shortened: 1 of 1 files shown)


fisker:
I'm not sure if we have the same problem with `BigInt`, it should work the same way, eg the BigInt is too big.

mroch:
looks like we don't even try to create a `value` for bigint :/

https://github.com/facebook/flow/blob/c0050829cdb3f383445962dd93ae490fbdd28cf6/src/parser/estree_translator.ml#L1324

PR #8898, opened 2022-06-22
fix parser crash on invalid UTF8 codepoints

mroch:
Summary:
UTF8 is a variable length encoding. there are some valid encodings that are invalid Unicode, like U+D800 (https://github.com/facebook/flow/commit/5f652a07562dbf042de4a3231cfb2491a7768ce2) - U+DFFF and above U+10FFFF. oddly, we check for U+D800 (https://github.com/facebook/flow/commit/5f652a07562dbf042de4a3231cfb2491a7768ce2) - U+DF00, but not U+DF01 - U+DFFF (I think this is just a typo) nor > U+10FFFF.

this allows invalid codepoints that don't match sedlex's `any` regex, causing it to raise `Failure "unreachable token"`.

the same bug exists upstream, but `Uchar.of_int` checks for these reserved ranges and catches it. I added the missing check for U+10FFFF here, but didn't touch the DF01-DFFF range yet (surrogates are confusing).

Differential Revision: D37323033

### DIFF
--- a/src/third-party/sedlex/flow_sedlexing.ml
+++ b/src/third-party/sedlex/flow_sedlexing.ml
@@ -152,33 +152,42 @@ let unsafe_utf8_of_string (s : string) slen (a : int array) : int =
     let spos_code = s.![!spos] in
     (match spos_code with
     | '\000' .. '\127' as c ->
+      (* U+0000 - U+007F: 0xxxxxxx *)
       a.!(!apos) <- Char.code c;
       incr spos
     | '\192' .. '\223' as c ->
+      (* U+0080 - U+07FF: 110xxxxx 10xxxxxx *)
       let n1 = Char.code c in
       let n2 = Char.code s.![!spos + 1] in
       if n2 lsr 6 != 0b10 then raise MalFormed;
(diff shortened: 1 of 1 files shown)


src/third-party/sedlex/flow_sedlexing.ml
```diff
@@ -152,33 +152,42 @@ let unsafe_utf8_of_string (s : string) slen (a : int array) : int =
...
         ((n1 land 0x07) lsl 18)
         lor ((n2 land 0x3f) lsl 12)
         lor ((n3 land 0x3f) lsl 6)
-        lor (n4 land 0x3f);
+        lor (n4 land 0x3f)
+      in
+      if p > 0x10ffff then raise MalFormed;
+      a.!(!apos) <- p;
       spos := !spos + 4
     | _ -> raise MalFormed);
```
bobzhang, line 192: This fix also applies to upstream https://github.com/ocaml-community/sedlex/blob/master/src/lib/sedlexing.ml#L287
Would you send a patch to upstream?

bobzhang (approved):
When tokenization, we turn the string into an array of unicodes, so doing sanity check on here would suffice.
The bug also appears in upstream, but its `__private_next_int` caught it using `Uchar.of_int` while our version does not do such check

PR #8909, opened 2022-07-26
fix: DOMRect has .x and .y properties

pmer:
In `lib/dom.js`, on `DOMRect` (named `ClientRect` in the types), we have the `top` / `right` / `bottom` / `left` orientation, but in the `x` / `y` / `width` / `height` orientation, the `x` and `y` are missing. This PR adds these properties. See https://developer.mozilla.org/en-US/docs/Web/API/DOMRect.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -4115,9 +4115,11 @@ declare class TextRange {
 
 declare class ClientRect { // extension
   left: number;
+  x: number;
   width: number;
   right: number;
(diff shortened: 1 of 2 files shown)


pmer:
Thanks @facebook-github-bot! Signed.

pmer:
Looks like this broke the CI tests Is there another edit I can make to get to passing?

PR #8910, opened 2022-07-26
fix: add DOMTokenList index signature

pmer:
In `lib/dom.js`, the `DOMTokenList` has the `item()` method but is missing the index signature way of accessing tokens.

Within https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList, we see:

> A DOMTokenList is indexed beginning with 0 as with JavaScript Array objects.

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1700,6 +1700,7 @@ declare class DOMTokenList {
   entries(): Iterator<[number, string]>;
   keys(): Iterator<number>;
   values(): Iterator<string>;
+  [index: number]: string;
 }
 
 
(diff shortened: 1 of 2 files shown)


pmer:
Looks like this broke the CI tests Is there another edit I can make to get to passing?

SamChou19815:
@pmer You will need to build the Flow binary, and then re-record the tests by

```
./runtests.sh -b path/to/compiled/flow-binary -r
./runtests.sh -b path/to/compiled/flow-binary -rw
./runtests.sh -b path/to/compiled/flow-binary -re
./runtests.sh -b path/to/compiled/flow-binary -rx
```

pmer:
Thanks, will do!

Issue #8914, opened 2022-08-07
Update `flow-remove-types/jest` transform to support `jest@28`


carloscuesta:
## System information

- `flow-bin`: `0.183.0`
- `flow-remove-types`: `2.184.0`
- `jest`: `28.1.3`

## Description

Hey! 👋🏼 

I was trying to use `flow-remove-types/jest` as a transform in a project to remove flow types. As suggested [in the documentation](https://www.npmjs.com/package/flow-remove-types) I added a custom transform to my package.json, like this:

> <img width="763" alt="Screenshot 2022-08-07 at 10 35 50" src="https://user-images.githubusercontent.com/7629661/183282481-646592f3-0dbc-455b-ba22-99534b8efe9f.png">

After running the tests, I found out that Jest was throwing an error:

```
STACK: Error: ● Invalid return value:
  `process()` or/and `processAsync()` method of code transformer found at
  "/node_modules/flow-remove-types/jest.js"
  should return an object or a Promise resolving to an object. The object
  must have `code` property with a string of processed code.
  This error may be caused by a breaking change in Jest 28:
  https://jestjs.io/docs/upgrading-to-jest28#transformer
  Code Transformation Documentation:
  https://jestjs.io/docs/code-transformation
```

In Jest 28, a [breaking change](https://jestjs.io/docs/upgrading-to-jest28#transformer) was introduced to transformer modules. They cannot return a string anymore, instead the should return an object:

https://github.com/facebook/flow/blob/19465a3e471bf2aaa3246a3f3e9af4a7a1dd0a35/packages/flow-remove-types/jest.js#L14

> <img width="554" alt="Screenshot 2022-08-07 at 10 40 04" src="https://user-images.githubusercontent.com/7629661/183282666-51e2e499-5013-4f3d-b54c-c457ed32bc9b.png">

## Expected behavior

I expect that `flow-remove-types/jest` works with the latest version of `jest` `v28.x`.

## Actual behavior

The transformer is returning an error as it's not compatible with `jest` `v28.x`.

PR #8915, opened 2022-08-07
Fix `flow-remove-types/jest` transformer

carloscuesta:
## Description

Hey! 👋🏼 

This PR updates the `flow-remove-types/jest` transformer to the new format defined by Jest in `v28`.

Related issue: https://github.com/facebook/flow/issues/8914

### DIFF
--- a/packages/flow-remove-types/jest.js
+++ b/packages/flow-remove-types/jest.js
@@ -11,6 +11,8 @@ var flowRemoveTypes = require('./index');
 
 module.exports = {
   process: function (src, filename) {
-    return flowRemoveTypes(src).toString();
+    return {
+      code: flowRemoveTypes(src).toString()
+    }
   },
 };


mroch:
thanks, I hadn't heard of this breaking change! do you know if this works with previous versions of jest? if it's not backwards compatible then we probably need to split this into a new transformer, e.g. `flow-remove-types/jest28`

mroch:
did some digging, it looks like this is backwards compatible to at least jest 20. 

https://github.com/facebook/jest/pull/12638 is the breaking change, which removed this legacy behavior:
```
      if (typeof processed === 'string') {
        transformed.code = processed;
```

this was added in https://github.com/facebook/jest/commit/f5eca01736965026b9901426a0c5995527ee5a32 in 2017.

i'm good with this!

carloscuesta:
Thanks for checking! 

I was not aware that those changes were made with vackwards compatibility in mind, so we should be good 🙏

PR #8936, opened 2022-11-04
Remove unnecessary calls to Obj.magic

nmote:
This module needs some calls to Obj.magic, but not as many as it currently has.


gkz:
> You can't import this pull request because Nat Mote (nmote) does not have a valid CLA on file.

lol

nmote:
@gkz, got the CLA sorted!

PR #8950, opened 2022-11-21
Treat *.ml{i} in Linguist as OCaml

kevinji:


kevinji (commented):
Linguist is marking some *.ml{i} files as SML instead of OCaml; this should fix the statistics.

PR #8968, opened 2022-12-12
Add Range. isPointInRange to dom.js

Dagur:
https://developer.mozilla.org/en-US/docs/Web/API/Range/isPointInRange


SamChou19815:
Some of the tests need re-recording. To do so, you need to compile flow, and then do `./runtests.sh -r -b path/to/compiled/flow-bin` and `./runtests.sh -rn -b path/to/compiled/flow-bin`

PR #8983, opened 2023-01-14
add Element.getAttributeNames(): Array<string>

gnoff:
according to MDN Element has a method `getAttributeNames` which returns and Array of strings.

https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttributeNames

It seems supported quite a ways back in every browser documented

### DIFF
--- a/lib/dom.js
+++ b/lib/dom.js
@@ -1735,6 +1735,7 @@ declare class Element extends Node implements Animatable {
   dispatchEvent(event: Event): boolean;
 
   getAttribute(name?: string): ?string;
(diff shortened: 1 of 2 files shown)


SamChou19815:
Some of the tests need re-recording. To do so, you need to compile flow, and then do `./runtests.sh -r -b path/to/compiled/flow-bin` and `./runtests.sh -rn -b path/to/compiled/flow-bin`

gnoff:
@SamChou19815 I updated the PR, seems a task is blocked still though :)

PR #9025, opened 2023-03-22
feat: add NodeJS stream promises

moroine:
Add NodeJS stream pipeline promise API https://nodejs.org/api/stream.html#streams-promises-api


SamChou19815:
To fix the tests, you need to compile flow and run `./runtests.sh -r -b path/to/flow`

PR #9045, opened 2023-07-03
[docs] Fix a few typos in Conditional Types, Type Guards, Mapped Types

pascalduez:


website/docs/types/conditional.md
```diff
@@ -35,7 +35,7 @@ type TypeOf<T> =
   T extends (...$ReadOnlyArray<empty>)=>mixed ? 'function' : 'object'
 
 type T1 = TypeOf<null>; // evaluates to 'null'
-type T2 = TypeOf<void>; // evaluates to 'void'
+type T2 = TypeOf<void>; // evaluates to 'undefined'
```
panagosg7, line 38: I believe this is intended. The terminology is 'undefined' for values, 'void' for types. "Evaluation" here means simplifying the type.

gkz: Yeah I thought about that as well at first, but I think in this specific example, the conditional type is returning a string literal attempting to mimic the expression `typeof <value>`: https://flow.org/en/docs/types/conditional/#toc-generic-conditional-types

PR #9062, opened 2023-07-21
fix generateMap "mappings" attribute is empty

liucan233:
if pretty option is false, generateMap will return empty string.

screenshot-20230721-160902

rollup [transform](https://rollupjs.org/plugin-development/#transform) need null or real sourcemap (not empty string), and react build script used generateMap by [this way](https://github.com/facebook/react/blob/main/scripts/rollup/build.js#L372).

### DIFF
--- a/packages/flow-remove-types/index.js
+++ b/packages/flow-remove-types/index.js
@@ -136,11 +136,17 @@ function resultPrinter(options, source, removedNodes) {
       return (result += source.slice(lastPos));
     },
     generateMap: function () {
+      if(!pretty){
+        console.info('flow-remove-types: "pretty" option is false, '
+          + 'and the current version will generate a sourcemap. previously, '
+          + 'the "mappings" attribute of sourcemap was an empty string.'
+        );
+      }
       return {
         version: 3,
         sources: ['source.js'],
         names: [],
-        mappings: pretty ? generateSourceMappings(removedNodes) : '',
+        mappings: generateSourceMappings(removedNodes),
       };
     },
   };


SamChou19815:
I'm not familiar with the setup. @mroch might have a better idea.

> rollup [transform](https://rollupjs.org/plugin-development/#transform) need null or real sourcemap (not empty string)

Then I think you should replace `''` with `null` instead. Also I would like see some test plan to ensure that this actually fixes the problem you are facing.

packages/flow-remove-types/index.js
```diff
@@ -136,11 +136,17 @@ function resultPrinter(options, source, removedNodes) {
       return (result += source.slice(lastPos));
     },
     generateMap: function () {
+      if(!pretty){
+        console.info('flow-remove-types: "pretty" option is false, '
+          + 'and the current version will generate a sourcemap. previously, '
+          + 'the "mappings" attribute of sourcemap was an empty string.'
+        );
+      }
```
SamChou19815, line 144: There shouldn't be any random printing

packages/flow-remove-types/index.js
```diff
@@ -136,11 +136,17 @@ function resultPrinter(options, source, removedNodes) {
...
       return {
         version: 3,
         sources: ['source.js'],
         names: [],
-        mappings: pretty ? generateSourceMappings(removedNodes) : '',
+        mappings: generateSourceMappings(removedNodes),
```
SamChou19815, line 149: Return null or undefined here?

SamChou19815 (changes requested):

PR #9066, opened 2023-08-08
Update React types to use better AbstractComponent<>

nmn:
Improve the type definitions for `React.forwardRef`, `React.memo` and `React.lazy` by leveraging the third type argument for `AbstractComponent`.

The "return" value of components, i.e. the `Render` type is now maintained.

### DIFF
--- a/lib/react.js
+++ b/lib/react.js
@@ -277,21 +277,21 @@ declare module react {
     ...
   };
 
-  declare export function forwardRef<Config, Instance>(
+  declare export function forwardRef<Config, Instance, Renders = React$Node>(
     render: (
       props: Config,
       ref: { current: null | Instance, ... } | ((null | Instance) => mixed),
-    ) => React$Node,
-  ): React$AbstractComponent<Config, Instance, React$Node>;
+    ) => Renders,
+  ): React$AbstractComponent<Config, Instance, Renders>;
 
-  declare export function memo<Config, Instance = mixed>(
-    component: React$AbstractComponent<Config, Instance, React$Node>,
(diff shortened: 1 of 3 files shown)


lib/react.js
```diff
@@ -277,21 +277,21 @@ declare module react {
     ...
   };
 
-  declare export function forwardRef<Config, Instance>(
+  declare export function forwardRef<Config, Instance, Renders = React$Node>(
```
SamChou19815, line 280: I don't think we need the default here. It can be easily inferred.

SamChou19815 (dismissed):
Please fix the test failures

lib/react.js
nmn: I kept the default value for all the existing usage of `forwardRef`. Without the default value, all the usages of `forwardRef<Props, Instance>` will break.

SamChou19815:
Added @jbrown215 as a reviewer

jbrown215:
These changes look good but I'm not quite ready to introduce behavior that depends on that type argument very widely yet-- I still have some upcoming changes that are semantically meaningful and don't want to introduce patterns that go out-of-date too quickly

nmn:
@jbrown215 Should I just leave the PR up for the future then?

jbrown215:
Yes! I think I’ll be able to merge this in the next month or two. --
Sent from Gmail Mobile

SamChou19815:
Done in https://github.com/facebook/flow/commit/22617771f568a260aec70104a86e8e8bd0fa8357 instead

PR #9074, opened 2023-09-02
[dom][node] Update URLSearchParams definitions

pascalduez:
Greetings, 

this add the following missing optional parameter `value` to `delete` and `has` methods.

1. `delete` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/delete
2. `has` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/has

Also add missing instance property `size`, as well as `toString` and `sort` methods.

1. `size` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/size
2. `toString` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/toString
3. `sort` https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/sort

Grab the occasion to harmonize the DOM and Node definitions.

See 
- https://nodejs.org/api/url.html#class-urlsearchparams
- https://url.spec.whatwg.org/#interface-urlsearchparams

### DIFF
--- a/lib/bom.js
+++ b/lib/bom.js
@@ -1537,17 +1537,22 @@ declare class Headers {
 
 declare class URLSearchParams {
     @@iterator(): Iterator<[string, string]>;
-    constructor(query?: string | URLSearchParams | Array<[string, string]> | { [string]: string, ... } ): void;
+
+    size: number;
+
+    constructor(init?: string | URLSearchParams | Array<[string, string]> | { [string]: string, ... } ): void;
     append(name: string, value: string): void;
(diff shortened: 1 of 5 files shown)


SamChou19815 (changes requested):
Need to fix some merge conflicts first

SamChou19815:
The changes do look good to me

pascalduez:
Conflicts resolved.

SamChou19815 (approved):

PR #9080, opened 2023-09-10
[flow][flow-parser-bin] replace nan with node-addon-api

diyorbek:
- Replaced `nan` library with `node-addon-api` which is more stable. 
 - Added missing `ocaml` library bindings
 - Added `arm64` to `package.json`  CPU list


packages/flow-parser-bin/package.json
```diff
@@ -31,6 +31,7 @@
     "linux"
   ],
   "cpu": [
-    "x64"
+    "x64",
+    "arm64"
```
SamChou19815, line 35: Please revert this. You can do it in another PR, where you also update the CI pipeline that actually include the arm64 artifact.

SamChou19815 (changes requested):
I'd like to understand your motivation first. If you'd like a flow parser with native performance, we recommend `hermes-parser` instead.

This package is currently not tested in any meaningful way except that it can build. It's depended on by nothing [according to npm](https://www.npmjs.com/package/flow-parser-bin?activeTab=dependents), so we might even stop publishing it one day. I have no idea what's going on here and whether it actually works. If you really want this change in, please write some automated tests.

PR #9112, opened 2024-01-08
Add annotations for the Storage Manager API

ZelJin:
This PR adds flow annotations for the [Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Storage_API), available via `window.navigator.storage`. This API is available in Web Workers as well.

I've only added methods that are a part of the [WHATWG standard](https://storage.spec.whatwg.org).

Storage API reference:
* MDN Docs for the Storage API: https://developer.mozilla.org/en-US/docs/Web/API/Storage_API
* MDN Docs for the `StorageManager` interface: https://developer.mozilla.org/en-US/docs/Web/API/StorageManager
* WHATWg living standard: https://storage.spec.whatwg.org


SamChou19815 (changes requested):
You need to compile Flow and re-record the tests using `./runtests.sh -b /path/to/compiled/flow -r`. If you don't want to do it, I can take over and do it for you.

ZelJin:
> You need to compile Flow and re-record the tests using `./runtests.sh -b /path/to/compiled/flow -r`. If you don't want to do it, I can take over and do it for you.

Thanks for the suggestion @SamChou19815! Fixed the tests and added a new one for the Storage API.

PR #9130, opened 2024-03-18
[ci] Properly test mac and linux in GitHub Actions CI

SamChou19815:
Currently, the mac and linux tests are passing, but if we look at the logs, they are failing for every single test. It turns out that we are not even downloading the artifacts from previous build jobs. We also need to set the executable bit to actually make the tests runnable (this is not needed in circle ci jobs and also not needed for windows builds, so it's unclear why, which needs further investigation)


SamChou19815:
> We also need to set the executable bit to actually make the tests runnable (this is not needed in circle ci jobs and also not needed for windows builds, so it's unclear why, which needs further investigation)

Note for myself:

https://github.com/actions/upload-artifact/issues/38

PR #9153, opened 2024-05-30
promise resolve generic

ayroblu:
Due to contravariant typing for function parameters, unions need to be generics otherwise the inputs need to satisfy all union variants. Change to resolve function to support only Promise or non promise values


ayroblu:
Looks like the signing for the CLA is broken

SamChou19815:
Can you add a flow.org/try example to demonstrate what this can unlock?

PR #9163, opened 2024-06-17
Update Array.prototype.{shift,pop} to `T | void`

zpao:
These currently return type `T`, which is incorrect by spec and inconsistent with `Array.prototype.at`. This also aligns with TypeScript, which consistently types these methods as `T | undefined`.

This does not modify regular indexed access. That remains `T` (which is consistent with default TypeScript behavior).

### DIFF
--- a/tests/node_tests/events/events.js
+++ b/tests/node_tests/events/events.js
@@ -16,7 +16,9 @@ emitter.emit({});                         // err: `event` must be a string
 emitter.eventNames().pop();               // ok: returns string[]
 emitter.eventNames('foo')                 // err: does not process args
 
-emitter.listeners('foo').pop()();         // ok: returns Function[]
+emitter.listeners('foo').pop();           // ok: returns Function[]
+emitter.listeners('foo').pop()();         // err: Function | void
+emitter.listeners('foo').pop()?.();       // ok: optional chaining of Function | void
 emitter.listeners();                      // err: requires `event`
 
 emitter.listenerCount('foo').toFixed();   // ok: returns a number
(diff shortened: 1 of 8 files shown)


tests/node_tests/events/events.js
```diff
@@ -16,7 +16,9 @@ emitter.emit({});                         // err: `event` must be a string
 emitter.eventNames().pop();               // ok: returns string[]
 emitter.eventNames('foo')                 // err: does not process args
 
-emitter.listeners('foo').pop()();         // ok: returns Function[]
+emitter.listeners('foo').pop();           // ok: returns Function[]
+emitter.listeners('foo').pop()();         // err: Function | void
+emitter.listeners('foo').pop()?.();       // ok: optional chaining of Function | void
```
zpao, line 21: Added a new call to just `pop` to show it's ok and that it's ok with chaining. The existing test case is now an error (but sandwiched between new tests).

Most of the changes in .exp are line numbers being off :(

tests/predicates_inferred/sanity-ordering.js
```diff
@@ -10,6 +10,9 @@ function dotAccess(head: { page: ?Object; }, create?: mixed) {
   const stack = path.split('.');
   do {
     const key = stack.shift();
+    if (!key) {
+      break;
+    }
```
zpao, line 15: I'll admit that this code is worse to write and in practice the falsey check isn't necessary for _this_ code (since `obj[undefined] === undefined`). But it does show a practical way to fix the issue that's surfaced.

PR #9177, opened 2024-07-01
[ci] Try bump xcode version

SamChou19815:


SamChou19815:
Doesn't seem to help fix the build error

PR #9185, opened 2024-07-08
[DOCS]

davidb4l:
From ESLint v9.0.0, the default configuration file is now eslint.config.js.
Update ESLint config to eslint.config.js as per v9.0.0 migration guidelines. see https://eslint.org/docs/latest/use/configure/migration-guide


SamChou19815 (changes requested):
Please keep both the old and the new filename, since not everyone who reads the docs will be on ESLint >9.

SamChou19815 (approved):

PR #9191, opened 2024-07-22
[ci] Fix windows build

SamChou19815:
Add `--accept-package-agreements` so that it doesn't get stuck


SamChou19815:
Doesn't seem to work. every once a while windows decides to be stupid

PR #9213, opened 2024-10-02
[flow-remove-types] fix interference when handling `as` expr

jbroma:
~~Fixed a case where `as` cast with generic was not removed when used with indexed access type~~

Fixed interference between `TypeParameterInstantion` and `as` expression removals.


SamChou19815 (changes requested):
What's the underlying issue here? We shouldn't keep adding more special cases

jbroma:
> What's the underlying issue here? We shouldn't keep adding more special cases

Hi 👋,

This is just a quick workaround to fix my previous implementation for handling generics. 

From my POV, it's the fact that `typeAnnotation` can have `IndexedAccessedType` and then inside of that we can have `objectType` with `type: 'GenericTypeAnnotation`. It looks like something that should be handled separately like `as const` tbh

SamChou19815:
The implementation should be removing the as and the type on the right completely. We should fix what's stopping that from happening, because otherwise after this PR you will keep finding things that got messed up

jbroma:
> The implementation should be removing the as and the type on the right completely. We should fix what's stopping that from happening, because otherwise after this PR you will keep finding things that got messed up

the most pragmatic way I've managed to find is to:
- limit `AsExpression` visitor to removing just the `as` part
- add `removeNode` as visitor to every `*TypeAnnotation` node like `StringTypeAnnotation` or `NumberTypeAnnotation`

let me know what you think of that, thanks!

jbroma:
I've updated the PR with the approach mentioned above.

After the changes, every kind of `as cast` gets removed correctly except for `HookTypeAnnotation` which needs a custom visitor since `removeNode` does not remove it correctly. I left it out for now since handling of it should be done in a separate PR.

SamChou19815 (changes requested):
Still, this is not a good fix. I did investigate and find the root cause that leads to https://github.com/facebook/flow/pull/9208 in the first place.

This rule: `TypeParameterInstantiation: removeNode` is causing interference. The intention of the rule is to remove type arguments from call expressions, but the name is also used as the type argument of type applications. e.g. the core removal engine sees instruction to remove `<string>` and `Foo<string>` and it starts to do weird things.

The fix should be to make them not interfere. This might not be easy given how the code is currently written, but listing all the annotation nodes will be very error prone and even worse.

SamChou19815:
One idea is that `TypeParameterInstantiation` should not be removed it's inside an annotation node

jbroma:
> One idea is that `TypeParameterInstantiation` should not be removed it's inside an annotation node

I'm having difficulty accessing the parent node (from my POV we would need to pass it to the visitor inside `visit`). We could also work this backwards from 'asExpression` with something like this:

```js
if (
  node.typeAnnotation.typeParameters &&
  node.typeAnnotation.typeParameters.type === 'TypeParameterInstantiation'
) {
  // handle 'TypeParameterInstantiation' case
} else {
  removeNode(context, node.typeAnnotation);
}
```

WDYT?

SamChou19815:
I don't have opinion on the strategy as long as it's attacking the core underlying issue

Another idea would be to fix `removeNode` or `resultPrinter` so that these overlapping removals are correctly handled

jbroma:
I addressed the interference issue through special handler for `TypeParameterInstantiation` and simplified the handler for `as` expression. Only 'GenericTypeAnnotation' seems to be the problem so that's when it bails out of removing the `TypeParameterInstantiation` node.

SamChou19815 (approved):

PR #9232, opened 2024-11-22
Type iterator helpers

### LINKED ISSUE #9230: Add iterator helpers proposal types
## Missing/Incorrect APIs
`Iterator.from()` and `Iterator.prototype.{*}` from the tc39 stage 4 [iterator-helpers-proposal](https://github.com/tc39/proposal-iterator-helpers).

## Relevant documentation
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator
- https://github.com/tc39/proposal-iterator-helpers
- https://tc39.es/proposal-iterator-helpers
- [TypeScript implementation](https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/lib/esnext.iterator.d.ts#L22)
- Safari is the last major browser to implement. Support [landed recently](https://bugs.webkit.org/show_bug.cgi?id=248650) but hasn't shipped yet.

hurrymaplelad:
Fixes #9230.

Includes tests: `$ make && bash runtests.sh bin/flow `.

## Changes

This PR is mostly additive, but there are some change. At least one is breaking.

### Iterators
- `Iterator` is now a class representing [the built-in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator), instead of a type representing an `IterableIterator` protocol.
   - This is the breaking change. Places that were using `Iterator` to represent the protocol need to do one of: 
      - Switch to `$IteratorProtocol` to continue accepting custom implementations.
      - Wrap value with `Iterator.from` to get `Iterator`s if they want the new helper methods.
   - The class now has three generic args, like `$Iterator`. It has defaults for the last two type arguments for backwards compatibility and simple uses.
- `$Iterator` is now a type alias for this class. It can likely be deprecated and removed.
- `$IteratorProtocol` is comparable to the old `$Iterator`, but does not require implementors to also implement 
`$Iterable` so that it imposes no extra constraints on `Iterator.from`.

### Iterables
- `Iterable` and `$Iterable` now only constrain `@@iterator` to `$IteratorProtocol`. The return value does not need to be iterable. Note that `Iterator.@@iterator` returns an `Iterator` which  implements both the iterable and iterator protocols.

### Generators
- `Generator` is now a type alias for a subclass of `Iterator`, rather than an interface. I [don't believe it's possible to implement a custom Generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator#constructor) so I believe this is correct.

## Consistency
I prioritized minimizing changes over internal consistency with this PR. One path back to consistency:
- Rename `$Iterable` to `$IterableProtocol` for consistency with `$IteratorProtocol` (optionally remove the $ prefix from both).
- Remove `$Iterator`.
These would have to be gradual migrations.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -1847,8 +1847,13 @@ declare class Iterator<+Yield,+Return=void,-Next=void> implements $IteratorProto
      * Iterates this iterator. Will hang if it's infinite and has no matches.
      * @param callbackfn  A function that accepts up to two arguments: the yielded value and its index.
      */
-     find<Refined: Yield>(predicate: (value: Yield, index: number) => implies value is Refined): Refined | void;
-     find(callbackfn: (value: Yield, index: number) => mixed): Yield | void;
+    find<Refined: Yield>(predicate: (value: Yield, index: number) => implies value is Refined): Refined | void;
+    find(callbackfn: (value: Yield, index: number) => mixed): Yield | void;
+    /**
+     * This method allows wrapping objects that implement the iterator protocol in an Iterator to get the built-in helpers.
+     * @returns The argument object if it's already an Iterator subclass, or a wrapping Iterator if the argument objects implements the iterable or iterator protocols.
+     */
+    static from<+Yield,+Return,-Next>(source: $IteratorProtocol<Yield,Return,Next> | $Iterable<Yield,Return,Next>): Iterator<Yield,Return,Next>;
 }
 
 type $Iterator<+Yield,+Return,-Next> = Iterator<Yield,Return,Next>;
--- a/lib/core.js
+++ b/lib/core.js
@@ -1750,6 +1750,21 @@ interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
      * @param callbackfn A function that accepts up to two arguments: the yielded value, and it's index. The map method calls the callbackfn function once per iteration.
      */
     map<U>(callbackfn: (value: Yield, index: number) => U): $Iterator<U, void, mixed>;
+    /**
+     * Returns a new iterator that yields only those elements of the iterator for which the provided predicate returns a truthy value.
+     * @param callbackfn A function that accepts up to two arguments: the yielded value and it's index. The filter method calls the predicate function once per iteration.
+     */
+    filter(callbackfn: typeof Boolean): $Iterator<$NonMaybeType<Yield>, void, mixed>;
+    /**
+     * Returns a new iterator that yields only those elements of the iterator for which the provided predicate returns a truthy value.
+     * @param callbackfn A function that accepts up to two arguments: the yielded value and it's index. The filter method calls the predicate function once per iteration.
+     */
+    filter<Refined: Yield>(callbackfn: (value: Yield, index: number) => implies value is Refined): $Iterator<Refined, void, mixed>;
+    /**
+     * Returns a new iterator that yields only those elements of the iterator for which the provided predicate returns a truthy value.
+     * @param callbackfn A function that accepts up to two arguments: the yielded value and it's index. The filter method calls the predicate function once per iteration.
+     */
+    filter(callbackfn: (value: Yield, index: number) => mixed): $Iterator<Yield, void, mixed>;
 }
 type Iterator<+T> = $Iterator<T,void,void>;
 
--- a/lib/core.js
+++ b/lib/core.js
@@ -1770,6 +1770,11 @@ interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
      * @param limit The maximum number of values to yield.
      */
     take(limit: number): $Iterator<Yield, void, mixed>;
+    /**
+     * Returns a new iterator that yields values from this iterator after skipping the provided count. If the this iterator has fewer than limit elements, the new iterator will be completed the first time next() is called.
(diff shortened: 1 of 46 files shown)


lib/core.js
```diff
@@ -1745,6 +1745,11 @@ type IteratorResult<+Yield,+Return> =
 interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
+    /**
+     * Calls a defined callback function on each element returned from this iterator.
+     * @param callbackfn A function that accepts up to three arguments. The map method calls the callbackfn function one time for each element in the array.
+     */
+    map<U>(callbackfn: (value: Yield, index: number) => U): $Iterator<U, void, Next>;
```
hurrymaplelad, line 1752: Note that [TypeScript's type](https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/lib/esnext.iterator.d.ts#L32) looks slightly wrong to me. It widens `Next` to `mixed`.

lib/core.js
```diff
@@ -1745,6 +1745,11 @@ type IteratorResult<+Yield,+Return> =
 interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
     @@iterator(): $Iterator<Yield,Return,Next>;
     next(value?: Next): IteratorResult<Yield,Return>;
+    /**
+     * Calls a defined callback function on each element returned from this iterator.
+     * @param callbackfn A function that accepts up to two arguments: the yielded value, and it's index. The map method calls the callbackfn function once per iteration.
+     */
+    map<U>(callbackfn: (value: Yield, index: number) => U): $Iterator<U, void, mixed>;
```
hurrymaplelad, line 1752: The `Next` type of the returned iterator is funny.

According to [the spec discussion](https://github.com/tc39/proposal-iterator-helpers/blob/main/DETAILS.md#passing-the-protocol) it should match the type of source iterator. In practice, Chrome discards any values passed to the `next` method of the returned iterator.

[TypeScript currently uses `mixed`](https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/lib/esnext.iterator.d.ts#L32), which existing behavior tolerates. It also tolerates working code if implementations eventually match the spec, albeit with reduced type safety.

We do the same here.

lib/core.js
```diff
@@ -1780,6 +1780,25 @@ interface $Iterator<+Yield,+Return,-Next> extends $Iterable<Yield,Return,Next> {
      * @param callbackfn A function that accepts up to two arguments: the yielded value, and it's index. The map method calls the callbackfn function once per iteration. The callback must return an iterable.
      */
     flatMap<U>(callbackfn: (value: Yield, index: number) => $Iterable<U, mixed, void>): $Iterator<U, void, mixed>;
+    /**
+     * Calls the argument "reducer" callback function on each value produced by this iterator, passing in the return value from the calculation on the preceding iteration. The final result of running the reducer across all elements is a single value.
+     * @param callbackfn A function that accepts up to three arguments: the accumulated value, the current value, and the current index.
+     * @param initialValue If initialValue is specified, it is used as the initial value to start
+     * the accumulation. If omitted, the first call of the callback gets the first two iterated values.
+     */
+    reduce<U>(
```
hurrymaplelad, line 1789: I believe this is a more accurate type for `reduce` without an initial value than  [the type we have for `Array.prototype.reduce`](https://github.com/facebook/flow/blob/efe35182d024b61d1c5a897eba77e9df23844f45/lib/core.js#L863) and the similar [type that TypeScript uses for `Iterator.prototype.reduce`](https://github.com/microsoft/TypeScript/blob/d85767abfd83880cea17cea70f9913e9c4496dcc/src/lib/esnext.iterator.d.ts#L69).

See motivating example in test. It likely rarely matters in practice. Happy to switch to the simpler type if I missed something, or if we just prefer simplicity.

tests/iterator_helpers/reduce.js
```diff
@@ -0,0 +1,15 @@
+/* @flow */
+declare const iterator: Iterator<number>;
+declare const numberReducer: (number,  number) => number;
+declare const stringReducer: (string,  number) => string;
+declare const flexibleReducer: (string | number,  number) => string;
+
+// Basic usage
+(iterator.reduce(numberReducer): number); // OK
+(iterator.reduce(flexibleReducer): string | number); // OK
```
hurrymaplelad, line 9: This is the motivating example for the more complex type signature for reduce with no initial value.

tests/babel_loose_array_spread/babel_loose_array_spread.exp
```diff
@@ -16,6 +16,24 @@ References:
                                     ^^^^^^^^^^^^^^^^^ [2]
 
 
+Error --------------------------------------------------------------------------------------------------- apply.js:11:15
```
hurrymaplelad, line 19: This is  a new error caused by splitting `$IteratorProtocol` from `Iterator`. It's in addition to the existing errors on this line, which were preserved.

SamChou19815 (approved):

Issue #9233, opened 2024-11-29
Set.prototype.has()


Dagur:
Set is defined like this

```
declare class Set<T> extends $ReadOnlySet<T> {
   /* ... */
    has(value: T): boolean;
   /* ... */
}
```

And I suspect that the type of `value` should be `mixed` because you can have situations like this

```
const values = new Set<number>([1,2,3]);
const maybeNumber: number | void = 3;
values.has(maybeNumber)
```

Which gives me this error

```
values.has(maybeNumber)
              ^ Cannot call `values.has` with `maybeNumber` bound to `value` because undefined [1] is incompatible with number [2]. [incompatible-call]
```

## Missing/Incorrect APIs

Set.prototype.has()

## Relevant documentation

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/has

## Test link

https://flow.org/try/#1N4Igxg9gdgZglgcxALlAIwIZoKYBsD6uEEAztvhgE6UYCe+JADpdhgCYowa5kA0I2KAFcAtiRQAXSkOz9sADwxgJ+NPTbYuQ3BMnTZA+Y2yU4IwRO4A6SFBIrGVDGM7c+h46fNRLuKxJIGWh8MeT0ZfhYlCStpHzNsFBAMIQkIEQwJODAQfiEyfBE4eWw2fDgofDBMsAALfAA3KjgsXGxxZC4eAw0G-GhcWn9aY3wWZldu-g1mbGqJUoBaCRHEzrcDEgBrbAk62kXhXFxJ923d-cPRHEpTgyEoMDaqZdW7vKgoOfaSKgOKpqmDA+d4gB5fMA-P6LCCMLLQbiLOoYCqgh6-GDYRYIXYLSgkRZkCR4jpddwPfJLZjpOBkO4AX34kA0SVs9gABE1cDISOyALzsr4Ad3ZAGVdgAeYQiG4APgAFABtACMvAATLwAMwAXQAlABuAA6j2gHIytBwADlriZkIKbZR2QAfTkQOBsfnsh4aeBfNhGqDGrk8qy1DAkeXmq0O3W5EANEwkODQJINAAMVjVAFYs1Y0yB6UA

gkz:
The current type is on purpose `T`. It's more useful to catch when people are passing in the wrong type. You can refine down to `number` first when doing this check. `maybeNumber != null && values.has(maybeNumber)`

Dagur:
[...]

PR #9234, opened 2024-11-29
Change prop type to mixed in Set.prototype.has()

Dagur:
See https://github.com/facebook/flow/issues/9233


gkz:
See issue comment.

PR #9265, opened 2025-03-13
Make string method index arguments optional

hurrymaplelad:
## What?

Updates `String.prototype.{at, charAt, charCodeAt, codePointAt}` to not require a position argument. These default to 0. 

- [Described by MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/charCodeAt#parameters)
  > Zero-based index of the character to be returned. [Converted to an integer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#integer_conversion) — undefined is converted to 0.
- [Defined in the spec](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.charcodeat)
  - Note that all 4 methods rely on [the `ToIntegerOrInfinity` definition](https://tc39.es/ecma262/multipage/abstract-operations.html#sec-tointegerorinfinity):
    > 1. Let number be ? [ToNumber]
    > 2. If number is one of NaN, +0𝔽, or -0𝔽, return 0.
  - `Number(undefined) -> NaN -> 0` 
- Implemented. Confirmed that MacOS latest Chrome, Safari, and Firefox all default to 0.
- Note that TypeScript library types [also have this bug](https://www.typescriptlang.org/play/?#code/OQQ2DoGMAsQJwMIHsAmBTAggFwBQEoBuIA). I may put up a fix there too. 

## Why?

Omitting the position argument is useful when we know we have a single character string. For example, converting a base64-encoded string to a `Uint8Array`:
```
Uint8Array.from(
  Iterator.from(
    atob(base64String)
  ).map(
    c => c.charCodeAt()
  )
)
```

## Is this a good idea?

I'm not sure. 

**Argument for**: I was surprised to see that common Stack Overflow suggestions ([example](https://stackoverflow.com/a/73147529/407845)) worked but did not typecheck. The default behavior is written in the spec and documented on MDN.

**Argument against**: Callsites are more readable with an argument. It's only one character. Even [this MDN example](https://developer.mozilla.org/en-US/docs/Web/API/Window/btoa#unicode_strings) adds the argument when it's not needed.

### DIFF
--- a/lib/core.js
+++ b/lib/core.js
@@ -1178,21 +1178,21 @@ declare class String {
      * @param index The index (position) of the string character to be returned.
      * Supports relative indexing from the end of the string when passed a negative index;
      * i.e. if a negative number is used, the character returned will be found by
-     * counting back from the end of the string.
+     * counting back from the end of the string. Defaults to 0.
      * @return A String consisting of the single UTF-16 code unit located at the specified position.
      * Returns undefined if the given index can not be found.
      */
-    at(index: number): string | void;
+    at(index?: number): string | void;
     /**
      * Returns the character at the specified index.
-     * @param pos The zero-based index of the desired character.
+     * @param pos The zero-based index of the desired character. Defaults to 0.
      */
-    charAt(pos: number): string;
+    charAt(pos?: number): string;
     /**
      * Returns the Unicode value of the character at the specified location.
-     * @param index The zero-based index of the desired character. If there is no character at the specified index, NaN is returned.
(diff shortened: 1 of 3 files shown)


gkz:
No, I don't think this is a good idea. The cases of people wanting it call it with no argument are much less than the logical bugs deriving from people forgetting to add an argument when they intended. Just do `c.charCodeAt(0)`

PR #9272, opened 2025-04-07
feat(flow-remove-types): remove type-only import declarations

### LINKED ISSUE #9271: [flow-remove-types] removal of type only import declarations
## Proposal

When dealing with `Identifier` style type only imports (e.g. `import {type x} from 'some-module'`, when all imports are type imports and after removing all type imports, it results in the following:

input:
```js
import { type x } from 'some-module';
```

output:
```js
import {} from 'some-module';
```

The statement above is a valid ES statement, equivalent to `import 'some-module'`, but this isn't the expected behaviour of this kind of type-stripping. 

When stripping flow types with `babel` (through `@babel/plugin-transform-strip-flow-types`), this import statement would be removed completely, and thus, would not cause inclusion of `some-module`;

I would like to propose adding this functionality to `flow-remove-types` since the crucial bit of information is knowing whether the import declaration was a `type-only` declaration. With current behaviour, the information is lost and removal of `import {} from 'some-module'` could result in erroneous behaviour. 

To make the change non-breaking, it's best that it's added behind a flag, similar to `uninitialized fields`

### Comparison of behaviour

For reference, I'm attaching the output of a comparison between `babel` and `flow-remove-types` in relevant test cases. 

Cases 1 & 4 should be explicitly handled by `flow-remove-types`

Case 3 in my opinion should not be handled, since both `import 'some-module'` and `import {} from 'some-module'` are valid ES statements resulting in the same behaviour.

```js
Case 1 - "removing type only imports":
Input                      [37]:  import { type x } from 'some-module';
Output - Babel             [37]:  import { type x } from 'some-module';
Output - Babel Strip Types [0]:
Output - Flow Strip Types  [37]:  import {        } from 'some-module';

Case 2 - "mixed type and non-type imports":
Input                      [40]:  import { type x, y } from 'some-module';
Output - Babel             [40]:  import { type x, y } from 'some-module';
Output - Babel Strip Types [32]:  import { y } from 'some-module';
Output - Flow Strip Types  [40]:  import {         y } from 'some-module';

Case 3 - "preserving empty import":
Input                      [29]:  import {} from 'some-module';
Output - Babel             [21]:  import 'some-module';
Output - Babel Strip Types [21]:  import 'some-module';
Output - Flow Strip Types  [29]:  import {} from 'some-module';

Case 4 - "mixed with DefaultImport":
Input                      [52]:  import DefaultImport, { type x } from 'some-module';
Output - Babel             [52]:  import DefaultImport, { type x } from 'some-module';
Output - Babel Strip Types [40]:  import DefaultImport from 'some-module';
Output - Flow Strip Types  [52]:  import DefaultImport, {        } from 'some-module';

Case 5 - "mixed with DefaultImport - type":
Input                      [57]:  import type DefaultImport, { type x } from 'some-module';
Output - Babel             [57]:  import type DefaultImport, { type x } from 'some-module';
Output - Babel Strip Types [0]:
Output - Flow Strip Types  [57]:

Case 6 - "mixed with DefaultImport - typeof":
Input                      [59]:  import typeof DefaultImport, { type x } from 'some-module';
Output - Babel             [59]:  import typeof DefaultImport, { type x } from 'some-module';
Output - Babel Strip Types [0]:
Output - Flow Strip Types  [59]:
```

repository with the comparisons: https://github.com/jbroma/flow-remove-imports

## Use case

In Re.Pack (https://github.com/callstack/repack/tree/main/packages/repack) we use `flow-remove-types` to remove flow types when transpiling with `swc. 

Recently, React Native 0.79 introduced some changes that cause a circular dependency problem because of the issue mentioned above.

jbroma:
Closes #9271 

- Added logic to handle removal of empty imports after removing type/typeof imports
- Added various test examples including:
  - Different types of imports (regular, default, named)
  - Type-only imports
  - Mixed default and named type imports
  - Various Flow type examples (functions, interfaces, classes, etc.)
- Updated test scripts and expected output files
- Updated CLI to include `--remove-empty-imports` flag


SamChou19815 (approved):

SamChou19815:
Since this is properly gated, I will accept, but please keep in mind that we do not intend to maintain this implementation of flow-remove-types in the long term. Eventually we will just replace it with a babel based implementation.

jbroma:
> Since this is properly gated, I will accept, but please keep in mind that we do not intend to maintain this implementation of flow-remove-types in the long term. Eventually we will just replace it with a babel based implementation.

Thank you for a quick response Sam! 

I'm well aware of the status of this library and we will be looking for alternatives as well, but for the moment this seems like a best fit. Happy to contribute tho 😄

PR #9274, opened 2025-04-12
fix(flow-remove-types): handle trailing comma in imports

jbroma:
This PR addresses an edge case when removing type-only imports (see https://github.com/facebook/flow/issues/9271)

Last import specifier can have a trailing comma and current implementation results in removal of trailing comma instead of a trailing brace, causing a syntax error. 

Input: 
```js
import DefaultImport, {
  type SomeType,
  typeof SomeOtherThing,
} from 'some-module';
```

Before (trailing brace present):
```js
import DefaultImport   
                
                        
} from 'some-module';
```

After (trailing brace correctly removed):
```js
import DefaultImport   
                
                        
  from 'some-module';
```


SamChou19815 (approved):

jbroma:
@SamChou19815, thank you for merging this so quickly! Would it be possible to publish a patch version that includes this PR?

SamChou19815:
The release of flow-remove-types is tied with the main Flow release in our release infra. We only do patch releases for severe issues. Since we do not intend to keep the current implementation of flow-remove-types long term, and the bug is behind a flag anyways, we will just wait until the next release.

PR #9275, opened 2025-04-12
prefer `Base.List.is_empty`

cometkim:
I'm having trouble building the parser with OCaml 4.14, and this seems to be the only dependency.

`List.is_empty` from the OCaml stdlib was added in 5.1.

The project already requires 5.2+, so you can ignore this. Alternatively, you can change other parts to prefer the OCaml stdlib for code consistency.


SamChou19815:
This has caused internal build failure for flow-parser.js. Since we are requiring OCaml 5.2, I will close this.

gkz:
We can't use `Base` in our parser code

PR #9293, opened 2025-06-21
feat: add --comment out flag

huhgawz:


SamChou19815:
Closing since CLA is not signed. In addition, we want to eventually stop maintaining this package in favor of hermes-parser based annotation removal transform

PR #9361, opened 2026-01-30
[flow] partial backout of D91720108

panagosg7:
Summary:
`make_path_relative_to_root` ended up only being used in one place and is a bit less efficient than before D91720108

Changelog: [internal]

Differential Revision: D91911938


SamChou19815:
already landed

PR #9395, opened 2026-04-08
[flow] Fix saved_state_reinit_libdef_change_mergebase test on macOS (#9395)

panagosg7:
Summary:

Add sleep between saved state creations so the lib file gets a different
mtime. On macOS, filesystem timestamps have second-level granularity, so
two writes in the same second look identical and Flow does not detect the
change. This matches the pattern used in saved_state_reinit_git and
saved_state_reinit_upstream_change_git.

Changelog: [internal]

Differential Revision: D100041960


SamChou19815:
already landed

PR #9402, opened 2026-04-24
[flow] Fix Windows path separator mismatches in File_key and flowconfig expansion (#9402)

panagosg7:
Summary:

Normalize directory separators to '/' in File_key suffixes at creation time (strip_project_root, lib_file_of_absolute) so that file keys are always '/'-based regardless of platform. This fixes the Graph.NodeNotFound crash when loading Linux-generated saved state on Windows, where '\'-based local file keys failed to match '/'-based saved state keys.

Also normalize separators in expand_project_root_token_as_relative so that flowconfig directives like module.system.node.root_relative_dirname with <PROJECT_ROOT> prefixes produce '/'-based paths, matching the now-normalized File_key suffixes. Without this, Files.is_prefix would append '\' on Windows while suffixes use '/', breaking root-relative module resolution.

Fix Files.is_prefix to append '/' instead of Filename.dir_sep when the prefix lacks a trailing separator. Previously, is_prefix would append '\' on Windows, which failed to match the now '/'-based File_key suffixes — breaking scoped root-relative module resolution (e.g., the `config_module_system_node_root_relative_dirnames` test).

Both normalizations are no-ops on Unix.

Changelog: [internal]

Differential Revision: D102285839

### DIFF
--- a/rust_port/crates/flow_common/src/files.rs
+++ b/rust_port/crates/flow_common/src/files.rs
@@ -908,11 +908,12 @@ pub fn expand_project_root_token(root: &Path, s: &str) -> String {
 
 pub fn expand_project_root_token_as_relative(s: &str) -> String {
     let s = s.replace(PROJECT_ROOT_TOKEN, "");
-    if s.starts_with('/') || s.starts_with('\\') {
+    let s = if s.starts_with('/') || s.starts_with('\\') {
         s[1..].to_string()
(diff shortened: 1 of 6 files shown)


SamChou19815:
landed in https://github.com/facebook/flow/commit/94108605643fcff630d72bd9562902329253c1fe

PR #9485, opened 2026-08-21
Process files sequentially in flow-upgrade runCodemods

### LINKED ISSUE #9407: `convertLegacyUtilityTypes` codemod replaces types with others from a random file
Flow version: 0.312.0

Not sure if this is an issue with codemods in general or just the `convertLegacyUtilityTypes` one, but it seems to be completely broken (see the linked repo).

It doesn't always take the types from A.js. It may randomly take the types from B.js or C.js instead.

* Link to Try-Flow or Github repo: https://github.com/mwiencek/flow-codemod-issue

rootkiller6788:
## Summary

Fixes #9407. When running a codemod over multiple files (e.g. `flow-codemod convertLegacyUtilityTypes`), every file was rewritten with the first file's transformed output.

### Root cause

`runCodemods` used `Promise.allSettled(filePaths.map(...))` so all files were transformed concurrently. `hermes-transform` is not safe to use this way:

- Prettier memoizes loaded parser plugins with `cacheKey: JSON.stringify` of the plugin options.
- `hermes-transform`'s printer passes a plugin whose only per-call difference is the `parse` closure returning the current file's AST. `JSON.stringify` drops functions, so the cache key is identical for every file.
- The second file's `prettier.format` therefore reuses the first file's parser, which returns the first file's AST, and every file gets printed with the first file's content.

Clearing `hermes-transform`/`prettier` from `require.cache` (the existing workaround) only helps when files are processed one at a time; under concurrent execution the reset races between files.

### Fix

Process files sequentially. The per-file failure tolerance of `Promise.allSettled` is preserved with a try/catch around each file.

Added two regression tests:
- `runCodemods-test.js` — verifies each file is transformed independently (the reported bug's exact scenario).
- `runCodemods-sequential-test.js` — mocks `hermes-transform` and asserts that transform invocations never overlap.

## Test plan

- `npx jest src/__tests__/runCodemods-test.js src/__tests__/runCodemods-sequential-test.js` — both pass.
- `npx jest src/` — all 14 suites / 91 tests pass.
- Reproduced the bug with the reporter's A/B/C fixture before the change (all files got A.js's output) and confirmed the fix yields each file's own correct output.


packages/flow-upgrade/src/__tests__/runCodemods-sequential-test.js
```diff
@@ -0,0 +1,70 @@
...
+    state.maxActive = Math.max(state.maxActive, state.active);
+    // Keep the transform pending long enough that a concurrent caller would
+    // observe overlapping invocations.
+    await new Promise(resolve => setTimeout(resolve, 25));
+    state.active -= 1;
+    return code;
+  },
+}));
+
+describe('runCodemods', () => {
```
SamChou19815: delete this test. it's only asserting that things are running sequentially which is an impl detail. it just adds blockers once the cache issue is properly fixed

rootkiller6788:
Thank you for the review. Fixed as suggested — I've deleted `runCodemods-sequential-test.js`.

You're right that the test only asserted files are processed one at a time, which is an implementation detail of the current workaround for the hermes-transform/prettier plugin cache collision (#9407). Once that cache issue is fixed properly at the root, files could again be transformed concurrently, and this test would only add blockers.

The behavior-level regression coverage stays in `runCodemods-test.js` ("transforms each file independently"), which verifies that every file is written with its own correct transformed output — the user-visible bug — without depending on how the work is scheduled.

Could you please take another look?

